PR #13, opened 2016-12-20
Create form/PasswordInput.vue

procoru:
Create form/PasswordInput.vue

### DIFF
--- a/src/components/forms/PasswordInput.vue
+++ b/src/components/forms/PasswordInput.vue
@@ -1,0 +1,68 @@
+<template lang="pug">
+  div(
+    class="input-group"
+    v-bind:class="classes"
+  )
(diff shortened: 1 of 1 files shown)


johnleider:
After seeing this I realize I completely left out the ability to select a type, which in the case of password or email for example, would fit nicely still within the text input. Possible name change to v-input and added ability to define type as a prop. Thoughts?

mfferreira:
@johnleider I like the name change.
I was just playing around and trying to set the input to password, and my instinct was just to add "type" to it. Which didn't work of course.

Congrats about the project by the way. Amazing work.

johnleider:
I'm going to add this functionality to the text-input component. Thank you for your contribution though.

PR #20, opened 2016-12-21
nextSibling whitespace (fix #17)

### LINKED ISSUE #17: CollapsibleHeader does not work with whitespace between v-collapsible-header and v-collapsible-body
When there is whitespace between the `v-collapsible-header` and `v-collapsible-body` the click event on the `v-collapsible-header` results in the error:

`Uncaught TypeError: this.$el.nextSibling.getAttribute is not a function `

This error occurs with the following example:
```
<v-collapsible>
  <v-collapsible-header>Item</v-collapsible-header>
  <v-collapsible-body>
    <v-card>
      <v-card-text class="grey lighten-3">...</v-card-text>
    </v-card>
  </v-collapsible-body>
</v-collapsible>
```
The problem is that the [nextSibling](https://developer.mozilla.org/en-US/docs/Web/API/Node/nextSibling) call in [CollapsibleHeader.vue](https://github.com/vuetifyjs/vuetify/blob/master/src/components/collapsible/CollapsibleHeader.vue#L17) returns a #text node instead of the intended `v-collapsible-body` element. 

Possible solutions include using [nextElementSibling](https://developer.mozilla.org/en-US/docs/Web/API/NonDocumentTypeChildNode/nextElementSibling) (with limited browser support) or the `.next()` function from jquery. A workaround is to collapse the whitespace in the html. 

I had this issue using Vue 2.1.6 and vuetify@51945ce6bfb8c20c478cdd8c705af1a5da18d85b

Pwntus:
The introduced method `getNextSibling` will look for the next **element**-sibling (nodeType 1) and thus ignoring any text nodes.

#17

### DIFF
--- a/src/components/collapsible/CollapsibleHeader.vue
+++ b/src/components/collapsible/CollapsibleHeader.vue
@@ -14,8 +14,18 @@
       click () {
         this.$vuetify.bus.pub(
           `collapse:toggle:${this.$parent._uid}`,
-          Number(this.$el.nextSibling.getAttribute('uid'))
+          Number(this.getNextSibling(this.$el).getAttribute('uid'))
         )
+      },
+      
+      getNextSibling (el) {
+        if (!(el = el.nextSibling)) return null
+        
+        while (el.nodeType != 1) {
+          if (!(el = el.nextSibling)) return null
+        }
+      
+        return el
       }
     }
   }


johnleider:
Thank you for your contribution. I will take a look at this later today.

johnleider:
Thank you for your contribution.

PR #117, opened 2017-02-05
added textarea component

dberube:
Created a new component, v-textarea in the forms collection.

v-textarea includes three boolean props: resizeable, resize-horizontal & resize-vertical (all default to false).  This component also accepts all of the same props as v-text-input with the exception of `type`.

Example:

```<v-textarea label="Enter Your Message" resize-vertical>```

This will produce a textarea with the label and same focus effects on the label and bottom border  as v-text-input that has a handle for the user to manually resize the element vertically only.

Focusing on v-textarea will also auto-expand the textarea from 3rem to 9rem.  If the input-group becomes 'dirty', it will remain at that height.


johnleider:
Thank you for your contribution. Can you please resubmit without the dist files included?

dberube:
Just created a new PR (#118) with dist files remaining untouched.

Thanks.

PR #118, opened 2017-02-06
added textarea component

dberube:
Created a new component, v-textarea in the forms collection.

v-textarea includes three boolean props: resizeable, resize-horizontal & resize-vertical (all default to false). This component also accepts all of the same props as v-text-input with the exception of `type`.

Example:

```<v-textarea label="Enter Your Message" resize-vertical>```

This will produce a textarea with the label and same focus effects on the label and bottom border as v-text-input that has a handle for the user to manually resize the element vertically only.

Focusing on v-textarea will also auto-expand the textarea from 3rem to 9rem. If the input-group becomes 'dirty', it will remain at that height.


johnleider:
Thank you. After next release I will merge it in.

dberube:
@johnleider Excellent, looking forward to it and very happy to help and contribute, you've created a great library here... :)

PR #144, opened 2017-02-16
Added focus highlight to select component

nekosaur:
Don't know if this was done by design, but the select component was missing the focus highlight transition that text input has. 

It made it hard to know which input field was active when using tab to switch between them, so I added the effect.


johnleider:
These are being addressed in the current form refactor that is releasing in 0.9.0. Thank you for your support though.

PR #153, opened 2017-02-19
CarouselItem: new boolean prop "img"

danikane:
Hi, this is my first ever commit and pull request to an official repo in GitHub.
I don't know what the etiquette and I don't consider myself THE developer so feel free to merge/modify/ignore/whatever this commit :)

I just noticed that on mobile devices the carousel item content doesn't scale well(if at all) when loaded as a background image. I fiddled with it locally when inside an  and it looked ok to me. The Vuetify CSS sets 500px height for the carousel slider div, but custom CSS rules are always viable and with  tag I thimk there are more possibilities.

Also for Carousel component **cycle** prop is _true_ by default, docs say _false_ so you may want to update the docs for that one as well. 

Thanks,
Daniel

### DIFF
--- a/src/components/carousel/CarouselItem.vue
+++ b/src/components/carousel/CarouselItem.vue
@@ -6,8 +6,13 @@
       class="slider__item"
       v-bind:class="{ 'reverse': reverse }"
       v-bind:style="styles"
+      v-bind:img="img"
       v-show="active"
     )
+      img(
+        v-if="img"
+        v-bind:src="src"
+        )
       slot
 </template>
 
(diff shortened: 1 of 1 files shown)


johnleider:
Thank you, I will review this.

johnleider:
Docs have been updated. Can you show an example of the differences it provides? I have not noticed this issue on mobile.

danikane:
Hey, I wasn't able to reproduce it on jsfiddle or using the startup webpack template.
Before I tried the img tag I was getting results similar to these:
https://vuetifyjs.com/components/carousel

image

On Chrome using the device emulator and on my phone.

image

Actually the jsfiddle snippet produced the opposite effect using the new prop so don't mind this :)

PR #177, opened 2017-03-03
Fixing max-width in grid

smith64fx:
Hey guys,
I noticed that the syntax here does not work with the current version of stylus.

```
for width in $container-max-widths
```

So I replaced it with 
```
for size, width in $container-max-widths
```

Also I added max-width: 100% for fluid containers, which is important, otherwise the "fluid" makes no sense :-)


crocodilered:
So how to use it? <v-container max-wdith="700px">?

PR #201, opened 2017-03-14
Pass value to rule callbacks

nekosaur:
This should enable users to write more generic rule functions which can be re-used in other parts of the app.

### DIFF
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -201,7 +201,7 @@ export default {
       this.errors = []
 
       this.rules.forEach(rule => {
-        const valid = rule()
+        const valid = rule(this.value)
 
         if (valid !== true) {
           this.errors.push(valid)


johnleider:
Could you provide an example of how the intended usage would look like.

nekosaur:
Something like:

```javascript
export default {
  password(value) {
    const pattern = new RegExp("^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#\$%\^&\*])(?=.{8,})")
    return pattern.test(value) || 'Incorrect password format'
  },
  onlyDigits(value) {
    return /^\d*$/.test(value) || 'String can only contain digits'
  }
}
```

```vue
<template>
  <v-container>
    <v-row>
      <v-col-spacer></v-col-spacer>
      <v-col xs4>
        <v-text-field
          label="password"
          v-model="password"
          :rules="[rules.password]"
        ></v-text-field>
        <v-text-field
          label="digits"
          v-model="digits"
          :rules="[rules.onlyDigits]"
        ></v-text-field>
      </v-col>
      <v-col-spacer></v-col-spacer>
    </v-row>
  </v-container>
</template>

<script>
  import rules from './rules.js'

  export default {
    data() {
      return {
        rules,
        password: '',
        digits: ''
      }
    }
  }
</script>
```

PR #236, opened 2017-03-19
Selects have long bottom margin/distance

nekosaur:
Selects have a longer bottom margin/distance compared to other inputs. Can be observed here http://codepen.io/anon/pen/gmoZEj

Left and right columns to compare, right shows this fix


johnleider:
After poking at this, the issue lies within the menu, not the select. Setting the menu to have a display of inline takes care of all the issues displayed, whereas the fix above only fixed one of them.

PR #269, opened 2017-03-25
Updating components elevations

mfferreira:


src/stylus/components/_pagination.styl
```diff
@@ -37,12 +37,12 @@
     text-decoration: none
     
     &--active
-      @extend .z-depth-2
+      elevation(4)
```
dohomi: @mfferreira I really like the elevation on active. shall we maybe introduce the same elevation(4) to hover?

src/stylus/components/_sidebar.styl
```diff
@@ -7,7 +7,7 @@
   width: $sidebar-width
   max-width: 100%
   z-index: 3
-  @extend .z-depth-2
+  elevation(16)
```
dohomi: Is elevetion 16 based on material-specs? I think its little bit too much. I think elevation(8) if is more suitable. I use a 4K monitor and it just looks bit off.

johnleider:
Need final approval from rest of members @dohomi @prograhammer

dohomi:
@johnleider I added 2 comments but they are both minor things, approved..

johnleider:
David is probably somewhere in the jungle, will give until the end of today for his approval.

PR #290, opened 2017-03-29
Completely reworked the tabs component

qw3n:
This started as a proof of concept, but I restructured the whole
component with the inital goal of adding support for tab changes without
the content elements  vuetifyjs#273.
The result still needs some work, but it is functioning. Here is a list
of the major changes.
1. Created a silder component that takes the current tab as a prop and
then adjusts accordingly.
2. Removed the need for slot defintions when creating the components
(render function takes care of spliiting based on tag-name)
3. Used the render function to also filter out tab-content if it didn't
have a matching id
4. Removed a bunch of code and imports that are no longer necessary
5. Hoefully created a leaner component

I've commented in where some parts may still need some tweaking or could
be improved. The only thing not working from the original is the forward
animation.

I realize that this diverges pretty drastically from the original, but I
thought I would commit this for at least your consideration.

Also, I apologize for the three extra commits I'm still new to creating pull requests and I'm not sure how to only submit the first commit.


johnleider:
Do you have an example of what markup would look with your implementation?

qw3n:
Exactly the same except no `slot="activators"` or `slot="content"` would be necessary.

qw3n:
Here is the test HTML I worked with

```html
<div id="app">
  <v-tabs grow>
    <v-tab-item
      v-for="i in 3" :key="i"
      v-bind:href="'#mobile-tabs-3-' + i"
    >
      Item {{ i }}
    </v-tab-item><v-tab-content
      v-for="i in 3" :key="i"
      v-bind:id="'mobile-tabs-3-' + i"
    >
      <v-card>
        <v-card-text>{{ text }} {{i}}</v-card-text>
      </v-card>
    </v-tab-content>
  </v-tabs>
  <v-tabs>
    <v-tab-item
      v-for="i in 3" :key="i"
      v-bind:href="'#mobile-tabs-3-test' + i"
    >
      Item {{ i }}
    </v-tab-item>
  </v-tabs>
</div>
<script type="text/javascript">
var app = new Vue({
  el: '#app',
  data: {
    text: 'Hello Vue!'
  }
})
</script>
```

johnleider:
Thank you, I'll review.

mfferreira:
@qw3n can you merge dev into your PR?

qw3n:
@mfferreira I merged, but I'm not sure what to do with all the conflicts..

johnleider:
I'll take a look at them this weekend.

johnleider:
I have not forgot about this. After next release, will be taking a dive in. Currently, does your implementation support all functionality that is available in Tabs now?

qw3n:
I believe so except for a glitch with the animation (it only works in reverse). The main thing I did was make the whole competent driven by one main variable `active`. So, an event is passed up when a tab is clicked and everything adjusts based on that tab click. I did use inject/provide so it requires Vue 2.2

qw3n:
One other problem is I'm not sure how the `v-tab-item` disabled option actually disables anything. If you could check that when you look this over.

johnleider:
Thank you for your patience. I've finally fully reviewed everything and really appreciate the PR to improve Vuetify. Unfortunately at this time, there are changes that I need to make to this component that would conflict with what you have done, and with time being precious, I don't have the resources to bug test a new implementation on top of my upcoming changes.

With that being said, there are a few good points that you have made with this request that I may implement, and will give the appropriate credit if I do.

staffordjh:
Thank you for your work on Vuetify. I was really happy to see the quality of your framework and its use of the Bootstrap layout in the demos. I was going to start playing around with it soon.

My question is in regards to the tabs component. I think adding a reverse slide direction would really improve the UI feel if going from a Right to Left tab. Both the Right to Left and Left to Right tab selections show the same transition. Can this be configured?

PR #325, opened 2017-04-07
Add border color classes

cawa-93:
As classes for color text, now there are classes for the color border. It can be used for the theming of the border. For example, for the block-quotes:
```html
<blockquote> Default blockquote.</blockquote>
<blockquote class="red--border border--lighten-4"> .red--border.border--lighten-4 </blockquote>
<blockquote class="red--border border--darken-4"> .red--border .border--darken-4 </blockquote>
```
image

### DIFF
--- a/src/stylus/tools/_colors.styl
+++ b/src/stylus/tools/_colors.styl
@@ -9,6 +9,9 @@ global-color(color_name, color_value)
   .{color_name}--after
     &:after
       background: color_value !important
+          
+  .{color_name}--border
+    border-color color_value !important
       
 global-color-accent(color_name, color_value)
   .{color_name}
@@ -23,3 +26,7 @@ global-color-accent(color_name, color_value)
   .{color_name}--text
     &.text--{color_type}
       color: color_value !important
+  
+  .{color_name}--border
+    &.border--{color_type}
+      border-color: color_value !important


johnleider:
We have had conversations about this. I'm not sure how useful this would be. It seems very niche. Will discuss with the team.

mfferreira:
I'm fine with being able to choose the blockquote color, but I don't think it makes sense to have that as classes, as the developer would have to set the class on every single blockquote that exists in the app.

Makes more sense to convert this:
```
blockquote
  border-left: 5px solid $theme.primary
  padding: $spacers.three.y 0 $spacers.three.y $spacers.four.x
  font-size: 1.5rem
  font-weight: 300
```

Into this:
```
blockquote
  border-left: 5px solid $blockquote.color
  padding: $spacers.three.y 0 $spacers.three.y $spacers.four.x
  font-size: 1.5rem
  font-weight: 300
```

And have the `$blockquote.color` set to `$theme.primary` as default.

johnleider:
I think he means the ability to just add a helper class to change border color, not so much setting default blockquote border color.

mfferreira:
In that case, I'll remove my review. I don't have anything against it, besides adding more classes.

johnleider:
What is the increased build size of the vuetify.min.css with and without these classes @cawa-93 ?

cawa-93:
+16736 B

johnleider:
After careful thought, I have no problem including this, but as an optional (bool) that the dev could enable when manually building the style file. Similar to how ```$color-pack``` works, https://github.com/vuetifyjs/vuetify/blob/master/src/stylus/generic/_bootstrap.styl#L7

johnleider:
Any plans to still update this?

johnleider:
Change the bootstrapping to be conditionally based upon a stylus variable that can be defined by the dev and I'll re-open and merge.

PR #331, opened 2017-04-08
Sort data tables on first sortable column by default

nekosaur:
In my opinion data tables should be sorted on their first sortable column by default. Not sure if what I've done is the best way to do it though


johnleider:
Great idea. I'd be interested in talking about what some of the other data table components/libraries do so we can put in the best possible implementation.

johnleider:
There have been changes to datatable since this PR. I will merge this with master and then take a look.

PR #341, opened 2017-04-10
Added validate option to EditDialog

nekosaur:
Added validate option to EditDialog. Setting it to true will make the dialog validate the input field inside of it, and not allow saving if the field has any validation errors.

You can see it in action here.
http://i.imgur.com/ENWFUIg.gifv

### DIFF
--- a/src/components/tables/EditDialog.js
+++ b/src/components/tables/EditDialog.js
@@ -41,7 +41,7 @@ export default {
 
   computed: {
     input () {
-      return this.$slots.input[0].componentInstance
+      return this.$slots.input.map(e => e.componentInstance).find(c => 'errors' in c)
     }
   },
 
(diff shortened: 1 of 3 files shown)


johnleider:
The only issue I have with this is it assumes input will always be at the 0 index of the slot.

nekosaur:
Good point. I changed it to look for the first component with an errors property

src/components/tables/EditDialog.js
```diff
@@ -34,6 +39,12 @@ export default {
     }
   },
 
+  computed: {
+    input () {
+      return this.$slots.input.map(e => e.componentInstance).find(c => 'errors' in c)
```
johnleider: Doing it this way is n2, could shorting to only the find.

PR #366, opened 2017-04-14
Feature/dialogs

mfferreira:


src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+      return {
+        'dialog': true,
+        'dialog--active': this.isActive,
+        'dialog--persistent': this.persistent,
+        'dialog--fullscreen': this.fullscreen,
+        'dialog--stacked-actions': this.stackedActions && !this.fullscreen,
+      }
+    },
+
+    computedOrigin () {
```
johnleider: If we are not actually going to compute anything, these can be removed.

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+      return this.origin
+    },
+
+    computedTransition () {
```
johnleider: Same for this fn

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+    },
+
+    closeConditional (e) {
+      return this.persistent ? false : this.wasClickInside()
+    },
+
+    resize () {
+      if (!this.isActive) return
+
+      let actions = this.$children.filter(c => c.$options.propsData.actions )
```
johnleider: Extra space after actions

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+
+      // make sure we have the actions card row
+      if (actions.length) {
+        let btns = actions[0].$slots.default
+
+        let maxButtonWidth = (this.$slots.default[0].elm.offsetWidth - 8 - (8 * btns.length) ) / btns.length
+        let shouldStack = false
+
+        for (let i=btns.length; i--;) {
```
johnleider: improper spacing

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+    resize () {
+      if (!this.isActive) return
+
+      let actions = this.$children.filter(c => c.$options.propsData.actions )
+
+      // make sure we have the actions card row
+      if (actions.length) {
+        let btns = actions[0].$slots.default
+
+        let maxButtonWidth = (this.$slots.default[0].elm.offsetWidth - 8 - (8 * btns.length) ) / btns.length
```
johnleider: I'm a fan of using refs over slots as the instance you receive from slots as I've had mixed behavior.

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
...
+
+    overlayClasses () {
+      return {
+        'overlay--modal-bottom': false  // this.bottom
+      }
+    }
+  },
+
+  methods: {
+    wasClickInside (target) {
```
johnleider: Method not needed, put logic into closeConditional

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,127 @@
+import Toggleable from '../../mixins/toggleable'
```
johnleider: We talked about the mixin for card-actions. If possible, I'd prefer to use that as we can ensure the behavior is the same across different components. https://github.com/vuetifyjs/vuetify/blob/feature/pickers/src/mixins/card-actions.js

src/components/dialogs/Modal.vue
```diff
@@ -0,0 +1,97 @@
+<template lang="pug">
```
johnleider: This should be deprecated per our conversation, if I'm remembering correctly.

src/stylus/components/_dialogs.styl
```diff
@@ -0,0 +1,64 @@
...
+      height 56px           // TODO: height with two lines of text: 80dp
+      padding 0 !important
+
+      &__title
+        font-size 20px      // TODO: convert from 20sp
+        padding: 20px 0     // TODO: convert from 20dp
+
+      .btn
+        &:first-child      // TODO: full width 72dp
+          width: 24px !important
```
johnleider: We should try to avoid important when possible. These can be changed by reducing the min-width which is on all buttons. min-width: 24px width: 24px or even just doing max-width: 24px.

src/stylus/components/_dialogs.styl
```diff
@@ -0,0 +1,64 @@
...
+          height: 24px !important
+          margin: 0 32px 0 16px !important
+        &:last-child
+          margin 0 !important
+          height 100%
+          font-size 14px    // TODO: convert from 14sp
+
+    .card
+      height 100% !important
```
johnleider: Card height is set by a prop, we could mention this within the documentation, but I don't want to fight with the inline-style and an important style.

src/stylus/components/_dialogs.styl
```diff
@@ -0,0 +1,64 @@
...
+      margin: 0
+      padding: 0 0 8px
+
+      .btn
+        width: 100%
+        height: 48px
+        margin: 0
+        padding: 0 16px
+        display: flex
+        flex-direction: row-reverse
```
johnleider: new line for next selector

src/stylus/components/_overlay.styl
```diff
@@ -12,10 +13,10 @@
   filter: blur(10%)
   pointer-events: none
   z-index: 1000
-  
+
   &--active
     opacity: 1
     pointer-events: auto
-    
+
```
johnleider: this shouldn't be needed with modal deprecation.

mfferreira: which part?

src/stylus/components/_dialogs.styl
mfferreira: This is meant for the fullscreen variation

src/components/dialogs/Modal.vue
mfferreira: Yes, this wasn't supposed to be pushed. My mistake

src/components/dialogs/Dialog.js
mfferreira: where?

src/components/dialogs/Dialog.js
mfferreira: And the method name was actually misleading :)

src/stylus/components/_overlay.styl
johnleider: the &--modal part

mfferreira: ah yes!

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,113 @@
...
+
+      // make sure we have the actions card row
+      if (actions.length) {
+        let btns = actions[0].$slots.default
+
+        let maxButtonWidth = (this.$refs.dialog.offsetWidth - 8 - (8 * btns.length)) / btns.length
+        let shouldStack = false
+
+        for (let i=btns.length; i--;) {
+          let span = btns[i].child._vnode.children[0].elm
```
nekosaur: $slots array can include empty vnodes (see https://github.com/vuejs/vue/issues/5329). Perhaps include a check if child is defined?

src/components/dialogs/Dialog.js
```diff
@@ -0,0 +1,113 @@
...
+    closeConditional (e) {
+      console.log(e.target)
+      // close dialog if !persistent and clicked outside
+      return this.persistent ? false : true
+    },
+
+    resize () {
+      if (!this.isActive) return
+
+      let actions = this.$children.filter(c => c.$options.propsData.actions)
```
johnleider: neko is correct, it should be ```c => c.actions```.

PR #399, opened 2017-04-20
Correcting snackbars multi-line

abotsi:
If found it resizes better with these two lines, hope it helped!


johnleider:
this has already been fixed, ty for the contrib though.

PR #420, opened 2017-04-23
Pass v-text-field name to domsProps

leonardoarroyo:
Necessary for using plugins such as vee-validate with vuetify.


johnleider:
This is already present https://github.com/vuetifyjs/vuetify/blob/master/src/components/forms/TextField.js#L169

leonardoarroyo:
Seems like I had an outdated version.

Issue #384, opened 2017-04-17
Sidebar (drawer) should have a backdrop (scrim)


sindrepm:
As per the [specs](https://material.io/guidelines/patterns/navigation-drawer.html), a navigation drawer when open should have everything behind it darkened by a backdrop (scrim), e.g. with an opacity value of `rgba(0, 0, 0, .54)`. This should only apply when it was triggered by a user action (i.e. when the initial state of the drawer is "hidden"). In addition, the drawer should span the height of the screen and not be clipped under the top bar.

I think these changes are important to stay more consistent with the Material Design spec.

johnleider:
As far as the sidebar not being clipped under the toolbar, that's at the discretion of the developer, we have options to do that or not do it. 

The overlay, however, is something that we need to do.

Issue #370, opened 2017-04-15
Light and dark theme for sidebar?


sindrepm:
Is there a simple way to change the color theme of the sidebar? The background and text color seems hard coded to dark/white. I could change the background color by applying a `class="blue"` for example, but the text color does not update. I think it would be useful to have some more options to customize the theme for the sidebar, e.g. a dark and light option, and also make it inherit the `<color>--text` class.

johnleider:
Great idea. Will do

PR #429, opened 2017-04-24
Enhancement/sidebars

johnleider:
This PR is for #384 and #370

Sidebar update to support overlay and mini variants. Also updated some parts of dialog since the overlay changes affect it. Docs are available [here](https://github.com/vuetifyjs/docs/tree/enhancement/sidebars).


src/components/grid/index.js
```diff
@@ -4,6 +4,7 @@ const Col = {
   functional: true,
 
   render: (h, { data, children }) => {
+    console.log(data)
```
nekosaur: rogue console.log

Issue #391, opened 2017-04-18
Use breakpoints from the spec


sindrepm:
> For optimal user experience, material design user interfaces should adapt layouts for the following breakpoint widths: 480, 600, 840, 960, 1280, 1440, and 1600dp.

https://material.io/guidelines/layout/responsive-ui.html#responsive-ui-breakpoints

Maybe the breakpoints in Vuetify should follow these?

https://github.com/vuetifyjs/vuetify/blob/master/src/stylus/settings/_variables.styl#L37

capttrousers:
Sorry if I don't understand GH well, but is this released yet? I see it says merged but that looks like only to the dev branch. 

The current breakpoint values have a huge gap between 600 and 1024, so having a few bp's in there as well as the <600 bp at 480 would be really nice to make more responsive layouts.

johnleider:
The breakpoints have been released and as within spec.

capttrousers:
Hi John, thanks for the response. 

The spec linked above lists the following breakpoints along with the number of grid columns and gutters between them. 

It seems the vuetify breakpoints at 600, 1024, 1440, and 1920 are missing a number of useful breakpoints, especially in the range from 360 - 600, and 600 - 1024. Am I missing something?


Breakpoint (dp) | Columns | Gutter
-- | -- | --
0 | 4 | 16
360 | 4 | 16
400 | 4 | 16
480 | 4 | 16
600 | 8 | 16/24*
720 | 8 | 16/24*
840 | 12 | 16/24*
960 | 12 | 24
1024** |  12 | 24
1280** |  12 | 24
1440** |  12 | 24
1600** |  12 | 24
1920** |  12 | 24

johnleider:
You are correct, we do not have the incremental breakpoints and it's something that we will not add. We strictly support the window break points as to cover all would be unnecessary bloating of the framework.

Huixiong91:
Vuetify breakpoints at 600, 1024, 1440, and 1920 for the respective devices xs, sm, md, lg, xl

This seems to differ slightly from the material design breakpoints of 600, 960, 1280, 1920. 
That is for the max-width, 960 vs 1024 for sm & 1024 vs 1280 for md.

May I know the reason for the slight deviation and if in the future will this be aligned?

johnleider:
Because more than 5 breakpoints is excessive and doesn't offer a real benefit for the increased size.  If you truly want this, you can easily expand on the grid hash and simply add them for you specific implementation.

[...]

PR #430, opened 2017-04-24
Grid enhancement

johnleider:
Enhancements for #391

### DIFF
--- a/src/stylus/settings/_variables.styl
+++ b/src/stylus/settings/_variables.styl
@@ -35,10 +35,10 @@ $material-twelve-percent-dark := rgba(#000, 0.12)
 // Grid
 // ============================================================
 $grid-breakpoints := {
-  xs: 480px
-  sm: 960px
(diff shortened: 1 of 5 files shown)


src/stylus/settings/_variables.styl
```diff
@@ -36,10 +36,10 @@ $material-twelve-percent-dark := rgba(#000, 0.12)
...
 $grid-breakpoints := {
   xs: 0
-  sm: 576px
-  md: 768px
-  lg: 992px
-  xl: 1200px
+  sm: 480px
+  md: 1024px
+  lg: 1440px
+  xl: 1920px
```
nekosaur: According to the breakpoint table, small should start at 600.

Also the way I interpret their second note ("Desktop breakpoints are 16dp below the listed values to accommodate variations in browser chrome") is that values >= 1024 should be -16px, so 1008, 1264, etc

Issue #345, opened 2017-04-12
enhancement for async data-tables  


dohomi:
To provide better handling with async data we need following fields to improve on data-tables:

- [ ] emit/watch sort change
- [ ] emit/watch row-per-page 
- [ ] bind total-number-of-items to a property
- [ ] bind loading to a property

Sarpedion:
I'd like to offer a couple of suggestions. 

1. Emitting paging event, with rows-per-page and current page parameters, would enable to fully customize async loading (use case: fetching a certain page from backend, while using Elasticsearch, based on current page and rows-per-page prop)

2. Visual indicator that async loading is in progress. As far as I know, this feature is not defined by Material Design standards, but would nevertheless improve user experience while using async loading. 
Example of such feature would be placing indeterminate linear progress bar between header row and first row.
For example:
![image](https://cloud.githubusercontent.com/assets/16826299/25384995/b58813e8-29c2-11e7-9a94-dc5ed3b7afaa.png)

nekosaur:
Addressed in #446

PR #446, opened 2017-04-26
Data table update

nekosaur:
Update to support server side pagination and sorting (vuetifyjs/vuetify#345)

### DIFF
--- a/src/components/tables/DataTable.js
+++ b/src/components/tables/DataTable.js
@@ -88,8 +88,8 @@ export default {
   },
 
   computed: {
-    length () {
-      return this.totalItems ? this.totalItems : this.value.length
+    itemsLength () {
+      return this.totalItems || this.value.length
     },
     indeterminate () {
       return this.selectAll && this.someItems && !this.everyItem
@@ -110,7 +110,7 @@ export default {
       const page = this.rowsPerPage === Object(this.rowsPerPage)
         ? this.rowsPerPage.value
         : this.rowsPerPage
-      return page === -1 ? this.length : this.page * page
+      return page === -1 ? this.itemsLength : this.page * page
     },
     filteredItems () {
       if (this.totalItems)
@@ -146,9 +146,7 @@ export default {
 
   watch: {
     rowsPerPage () {
-      if (this.page == 1) {
-        this.update()
-      }
+      this.page === 1 && this.update()
 
       this.page = 1
     },
--- a/src/components/tables/DataTable.js
+++ b/src/components/tables/DataTable.js
@@ -1,11 +1,12 @@
 import Head from './mixins/head'
 import Body from './mixins/body'
 import Foot from './mixins/foot'
+import Progress from './mixins/progress'
 
 export default {
   name: 'datatable',
 
-  mixins: [Head, Body, Foot],
+  mixins: [Head, Body, Foot, Progress],
 
   data () {
     return {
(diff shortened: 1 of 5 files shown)


src/components/tables/DataTable.js
```diff
@@ -75,10 +76,21 @@ export default {
...
+    },
+    loading: {
+      type: Boolean,
+      default: false
     }
   },
 
   computed: {
+    length () {
+      return this.totalItems ? this.totalItems : this.value.length
```
johnleider: ```return this.totalItems || this.value.length```

src/components/tables/DataTable.js
```diff
@@ -98,9 +110,12 @@ export default {
       const page = this.rowsPerPage === Object(this.rowsPerPage)
         ? this.rowsPerPage.value
         : this.rowsPerPage
-      return page === -1 ? this.value.length : this.page * page
+      return page === -1 ? this.length : this.page * page
     },
     filteredItems () {
+      if (this.totalItems)
```
johnleider: put this on one line

src/components/tables/DataTable.js
```diff
@@ -131,6 +146,10 @@ export default {
 
   watch: {
     rowsPerPage () {
+      if (this.page == 1) {
```
johnleider: absolute comparison, also move to 1 line
```this.page === 1 && this.update()```

File: src/components/tables/DataTable.js [hunk 4]
```diff
@@ -141,10 +160,21 @@ export default {
     },
     search () {
       this.page = 1
+    },
+    page () {
+      this.update()
```
johnleider: update is getting called multiple times, not sure if necessary. when ```rowsPerPage``` changes, it can call update, and will change ```this.page```, which also triggers reactivity and calls ```this.update()```

src/components/tables/DataTable.js
```diff
@@ -75,10 +76,21 @@ export default {
...
+      default: null
+    },
+    loading: {
+      type: Boolean,
+      default: false
     }
   },
 
   computed: {
+    length () {
```
johnleider: can we give this a more descriptive name? seeing ```this.value.length``` and ```this.length``` throughout was a bit confusing.

src/stylus/components/_datatables.styl
```diff
@@ -41,30 +41,47 @@
...
+  tr
+    height: auto !important
+
+  th
+    padding: 0 !important
+
+    .progress-linear
+      top: -3px
+      margin: 0 0 -3px
+      opacity: 0
```
johnleider: progress linear can be hidden and displayed by using the ```active``` [prop](https://github.com/vuetifyjs/vuetify/blob/master/src/components/progress/ProgressLinear.vue#L23). Can we use that instead of writing specific styles?

nekosaur: I forgot about active :)

src/components/tables/DataTable.js
nekosaur: Are you saying it gets called multiple times in the code, or when using it? As far as I can tell it only triggers once when paging/changing rows per page. The update call in the rowsPerPage watch is so that it emits update if it's already on page 1, since the page watcher won't active then.

src/components/tables/DataTable.js
nekosaur: I can't put this on one line, if you mean `this.totalItems && return this.value`. That results in unexpected token for the `return`

nekosaur: Nevermind, I was being stupid

PR #450, opened 2017-04-27
adding placeholder prop to textfield

### LINKED ISSUE #436: Ability to specify placeholder for text inputs and selects
Some times it is nice to be able to specify a placeholder text for input control that are different from the label (e.g. a short explanatory text). I know we already have the `hint` prop, but a placeholder fills a slightly different purpose. It is also part of the HTML spec.

Vue-material already have this. Have a look at the second example to see what I mean:
https://vuematerial.github.io/#/components/input

What do you think?

mfferreira:
closes #436

### DIFF
--- a/src/components/forms/TextField.js
+++ b/src/components/forms/TextField.js
@@ -166,6 +166,7 @@ export default {
         ref: 'input'
       }
 
+      if (this.placeholder) inputData.domProps.placeholder = this.placeholder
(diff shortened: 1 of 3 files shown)


mfferreira:
Focused
screen shot 2017-04-27 at 1 26 32 pm
Blured
screen shot 2017-04-27 at 1 26 21 pm

PR #474, opened 2017-05-01
add support for sorting nested properties

mfferreira:
Right now, when setting up the headers and their corresponding values, a key of the v-model is specified. That key is only valid for plain objects.

For example, in an object like `{a: 1, b: 2, c: {d:3, e: e}}` I can only access `a` and `b`.
This PR will allow the dev to setup headers like this:
```
addressHeaders: [
  {text: 'Header label', value: 'obj.nested.value'},
]
```

### DIFF
--- a/src/components/tables/DataTable.js
+++ b/src/components/tables/DataTable.js
@@ -2,6 +2,7 @@ import Head from './mixins/head'
 import Body from './mixins/body'
 import Foot from './mixins/foot'
 import Progress from './mixins/progress'
+import { getObjectValueByPath } from '../../util/helpers'
 
 export default {
   name: 'datatable',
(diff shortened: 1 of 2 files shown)


johnleider:
Could you provide an example of what this would allow the developer to do?

mfferreira:
Right now, when setting up the headers and their corresponding values, a key of the v-model is specified. That key is only valid for plain objects.

For example, in an object like `{a: 1, b: 2, c: {d:3, e: e}}` I can only access `a` and `b`.
This PR will allow the dev to setup headers like this:
```
addressHeaders: [
  {text: 'Header label', value: 'obj.nested.value'},
]
```

Issue #480, opened 2017-05-02
DataTable allow override sort / filter functions


avengerweb:
- What will it allow you to do that you can't do today?

Example:
I have a "created_at" column in unix timestamp, before I used my own value getter, which allow me for sort use int timestamp and for showing / filtering - formated date, like 05/02/2017 3:55 PM

Example:
I have move faster search / sort algorithm, specified for my data.

------------------------
That easy for implement and very useful, it`s help solve: (small example)
Pull request #474 - that guy can wrote own filter / sort with nested props for example.
issue #316 p.3 - u can modify filter and add his own conditions (in this case search by specified field)

PR #482, opened 2017-05-02
Allow override default search and filter functions

avengerweb:
Issue #480 
My code sample:
```
            <v-data-table
                    v-bind:headers="headers"
                    :search="filterKey"
                    v-model="data"
                    rows-per-page="50"
                    class="elevation-1"
                    :customFilter="customFilter"
                    :customSort="customSort"
            >
                    <td class="text-xs-right" v-for="entry in headers" :key="entry.value" :class="entry.style" v-html="getValue(entry.value, props.item)"></td>
                </template>
            </v-data-table>
```
```
methods: 
{
            customFilter(items, filterKey)
            {
                if (filterKey)
                {
                    let search = new RegExp(filterKey.replace(/ [\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&"), 'i');

                    return items.filter(row =>
                        this.headers.some(column =>
                        {
                            return search.test(this.getValue(column.value, row));
                        })
                    );
                }

                return items;
            },
            customSort(items, index, desc)
            {
                let modifier = (desc ? 1 : -1);
                return items.sort((var1, var2) =>
                {
                    if (!index)
                        return 0;

                    let getter = this.callbacks[index];

                    let val1 = var1[index];
                    let val2 = var2[index];
                    if (getter)
                    {
                        val1 = getter(index, var1, true);
                        val2 = getter(index, var2, true);
                    }

                    if (isFinite(val1) && isFinite(val2))
                        return modifier * (val1 - val2);

                    return modifier * val1.toString().localeCompare(val2);
                });
            }
}
```

### DIFF
--- a/src/components/tables/DataTable.js
+++ b/src/components/tables/DataTable.js
@@ -72,6 +72,34 @@ export default {
           val.toString().toLowerCase().indexOf(search) !== -1
       }
     },
+    customFilter:
+    {
+      type: Function,
+      default: (items, search, filter) => {
+          search = search.toString().toLowerCase()
+          return items.filter(i => Object.keys(i).some(j => filter(i[j], search)))
+      }
+    },
+    customSort:
+    {
+      type: Function,
+      default: (items, index, desc) => {
+        return items.sort((a, b) => {
+              const sortA = a[index]
+              const sortB = b[index]
+
+              if (desc) {
+                  if (sortA < sortB) return 1
+                  if (sortA > sortB) return -1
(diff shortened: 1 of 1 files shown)


johnleider:
Could you please provide an example that you would use in documentation to explain what this change allows?

avengerweb:
I will. Tomorrow

avengerweb:
@johnleider, so, I show how I use this.

Issue #485, opened 2017-05-03
Provide alternative namespace-like naming


nickmessing:
In XML we have namespaces which look like `<namespace:tag/>`. I suggest providing `v` pseudo-namespace as alternative to `v-` prefix because it looks more friendly to people familiar with XML.

- **What will it allow you to do that you can't do today?**
  Group the entire application into namespaces for code like:
  ```
  <v:content>
    <v:container fluid>
      <user:authentication />
    </v:container>
  </v:content>
  ```
  IMO this looks cleaner and easier to follow than prefixing with `{something}-`
- **How will it make current work-arounds straightforward?**
  People often use dash-separated component names and it is no visually different from prefixes.
- **What potential bugs and edge cases does it help to avoid?**
  It just makes it easier to follow the code IMO

I suggest adding namespace-like components as aliases to existing ones.

P.S. It can be achieved using `componentPrefix`, this is a suggestion to default behaviour

mfferreira:
@nickmessing Personally, I don't see any particular issues supporting other tag styles.

My only argument is that kebab-style, although not a rule, it's a convention within html. I'm not sure about the real reasons behind this, but I can imagine it could be to keep it consistent with #Id and .class names that don't use colons in their names.

johnleider:
It's actually built in, or was built in, to be able to change that. Problem is some components use other components which would not work as they are coded with the v- prefix.

PR #486, opened 2017-05-03
Support both v- and v: prefixes by default for namespace-like look

nickmessing:
Needs to be discussed in #485


johnleider:
Unfortunately this will not work as some components use others, which have their prefix hard-coded as v-. It would require significant changes throughout the framework to support this.

PR #514, opened 2017-05-10
[dialog] Scroll improvements 

avengerweb:
Fixed scroll when dialog opened
Add global modifier for indicate when dialog opened (for applications with custom scroll area)

### DIFF
--- a/src/components/dialogs/Dialog.js
+++ b/src/components/dialogs/Dialog.js
@@ -46,6 +46,7 @@ export default {
 
   watch: {
     isActive (val) {
+      document.querySelector('body').classList.toggle('application--dialog-opened', val);
       if (val) {
         !this.fullscreen && !this.hideOverlay && this.genOverlay()
       } else {
--- a/src/stylus/elements/_global.styl
+++ b/src/stylus/elements/_global.styl
@@ -9,10 +9,12 @@ html
   -webkit-font-smoothing: antialiased
   -moz-osx-font-smoothing: grayscale
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0)
+  overflow-y: auto
 
 body
   font-family: $body-font-family
   line-height: $line-height-root
+  overflow-y: scroll
 
 header
   transition: .3s padding $transition.swing


johnleider:
Could you show/describe how this would change a developers current implementation? I'm fuzzy on what this accomplishes. Thank you.

avengerweb:
You about js part? I thought about it, I think in this cases this part is useless...

johnleider:
Could you rephrase that response? I don't understand.

avengerweb:
I will make some examples tomorrow

avengerweb:
Update first message

src/stylus/components/_dialogs.styl
```diff
@@ -75,6 +74,8 @@
 
     >
       .card
+        height: 100% !important // TODO: remove important (card has 'height' property with default 'auto')
```
johnleider: this can/should be defined on the card.

src/components/dialogs/Dialog.js
```diff
@@ -46,6 +46,7 @@ export default {
 
   watch: {
     isActive (val) {
+      document.querySelector('body').classList.toggle('application--dialog-opened', val);
```
johnleider: This doesn't seem to be used, is it needed?

src/stylus/components/_dialogs.styl
avengerweb: Yes, but it need specify in all dialogs, not user friendly

avengerweb: May be good question, why card have default value 'auto' everywhere?

src/components/dialogs/Dialog.js
avengerweb: Optional, last part of first message. I can give some screenshots if needed.

src/stylus/components/_dialogs.styl
johnleider: Because it will auto height based upon content but user can define something specifically if they choose.

src/components/dialogs/Dialog.js
johnleider: I'm not very fond of this as something built in.

src/stylus/components/_dialogs.styl
avengerweb: two ways:
 * Specify 100% height for card in fullscreen dialogs (but in every fullscreen dialog we need specify it, coz dialog has 100% height)
* My way, with specify height by css

johnleider:
Are you still working on this?

avengerweb:
@johnleider, what wrong with this pull request?

johnleider:
You made no modification from my review points.

jamesYapKM:
capture
@mfferreira The scroll bar still there, please help to fix this bug. Thanks

smares:
The current implementation hiding the scrollbar isn't that great, it makes the content of the page jump around.

johnleider:
You can choose to disable that functionality if it doesn't fit you. If I try to add padding to compensate for the change, people don't like it, if I have a scrollbar, people don't like it, if the scrollbar hides, people don't like it.  I'm not sure what else I can do.

smares:
The biggest problem I see is that a fixed value of 17px is used instead of calculating the actual width of the scrollbar like http://jsfiddle.net/UU9kg/17/ for example. With Chrome under Ubuntu, the scrollbar width is 15px, with Firefox it's 5px.
This will still not fix the problem with toolbar elements placed on right jumping around. Dunno, I was trying to get this working in a codepen after having some success in my own app by setting the overflow to hidden / auto on the body, but then in the codepen I suddenly had two scrollbars and a horizontal one to make of the used space by the second vertical one. This sucks indeed. The other frameworks I've been looking at circumvent this problem by simply not having the scroll on the html / body, but on the main element that shows the scrollable content.

johnleider:
The problem with that is you lose the ability to hook into window.addEventListener('scroll'). Then you have to have all developers know to add that listener to the content area instead. I experimented with it before and I was not happen with it.

smares:
Why not make the body not overflow? https://codepen.io/anon/pen/mwpwOo At least that fixes the problem with content jumping around for me.

johnleider:
I'll try that

johnleider:
So I thought I had tried that approach before, and I did. The problem with removing the body overflow is it jumps you to the top of the screen. Now, there are ways to save the scroll position and put yourself back after the scroll is re-enabled, but it was all very hacky.

johnleider:
I think I may have a fix.

johnleider:
Fix is in dev, this should be glorious now.

TechWACO:
Hi,

Is this case still open? I encountered a similar behavior using v-navigation-drawer when the drawer is opened, the  scroll is disabled but still can scroll with the scrollbar.

Any idea how to disable the scrollbar when the drawer is open?

Thanks

Issue #498, opened 2017-05-05
DataTable: changing rows per page mutates a prop, resulting in a warning


grk:
### Steps to reproduce
When the dropdown at the bottom of the table is used to change rows per page, a warning is shown by Vue:

```
"[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: 'rowsPerPage' 
(found in <Datatable>)"
```

### Versions
Vuetify 0.11.1, current master
Vue 2.3.0

### Reproduction Link
https://codepen.io/anon/pen/GmMvOp

nekosaur:
Fixed in #516

Issue #400, opened 2017-04-20
Manage selected state in the data-table component instead of in the data itself


jakobrosenberg:
Currently the data-table stores its state within the data itself, which I find highly problematic.

**Storing view state in data models**

**Pros**
Not sure

**Cons**
- You have to map selected state to every data item.
- That gets complicated if your data is reactive.
- And even more complicated if your data is immutable, e.g. from a data store like Vuex.
- And even more complicated if the same data is used in a table more than once as the selected state is remembered in the store.
- And even more complicated if your server enforces schema or doesn't sanitize your submitted data
- Lastly, you risk data corruption in case your data entries already have a value property and you're unaware that the selected state by default overwrites the value.

Many of the cons listed can be tackled through various boilerplate, methods, computed methods, watchers, server sanitation etc., but the crux of the matter is, view state shouldn't leak into the data in the first place.

The current data-table uses the following scope 
```
  props: {
    item: {
      value: false,
      name: 'Ice cream sandwich',
      calories: 237,
      fat: 9.0,
      carbs: 37,
      protein: 4.3,
      sodium: 129,
      calcium: '8%',
      iron: '1%'
    }
  }
```

I propose this
```
  props: {
    checked: false,
    item: {
      name: 'Ice cream sandwich',
      calories: 237,
      fat: 9.0,
      carbs: 37,
      protein: 4.3,
      sodium: 129,
      calcium: '8%',
      iron: '1%'
    }
  }
```

Move the checked state out of the item and into props and keep the logic inside the data table component. The selected state can be be queried through $ref as well as emitted from the data table component.

johnleider:
Could you provide some pseudo code in what a data table example would look like with this implementation?

jakobrosenberg:
The data table example would remain the same as that in the current documentation except for the omission of ``value: false`` in each row.

```
{
   value: false, //omitted
   name: 'Ice cream sandwich',
   calories: 237,
   fat: 9.0,
   carbs: 37,
   protein: 4.3,
   sodium: 129,
   calcium: '8%',
   iron: '1%'
 }
```

An additional ``key`` property, like the one used in ``v-for`` could prove helpful.


The current method of getting checked items looks like this
``checkedItems = this.myItems.filter(row => row.value)``

With a ref it would look like this
``checkedItems = this.$ref.myTable.rows.filter(row => row.checked)``

[...]

PR #516, opened 2017-05-10
Datatable sync

nekosaur:
Made the following changes:

- added synced `pagination` object prop, and moved `page`, `rowsPerPage`, `sortBy` and `descending` into it (solves #498)
- v-model is now optional and used for selected array (together with regular prop `selectedKey`) (solves #400)
- incoming data moved to regular `items` prop
- items slot now exposes a boolean `selected` which can be used in `v-checkbox`
- removed `defaultSort` as pagination prop can be used to same effect

This change allows for external control of pagination, sorting and selection.

Selected example (selectedKey defaults to 'id')
```vue
<v-data-table
  v-model="selected"
  :items="items"
  :headers="headers"
  select-all
  >
    <td><v-checkbox primary v-model="props.selected"></v-checkbox></td>
    <td>#{{ props.item.id }}</td>
    <td>{{ props.item.first_name }}</td>
    <td>{{ props.item.email }}</td>
    <td>{{ props.item.gender }}</td>
  </template>
</v-data-table>
...
<script>
  export default {
    data () {
      return {
        selected: []
      }
    }
  }
</script>
```

Pagination example (change default sort)
```vue
<v-data-table
  :pagination.sync="pagination"
  :items="items"
  :headers="headers"
  select-all
  >
    <td><v-checkbox primary v-model="props.selected"></v-checkbox></td>
    <td>#{{ props.item.id }}</td>
    <td>{{ props.item.first_name }}</td>
    <td>{{ props.item.email }}</td>
    <td>{{ props.item.gender }}</td>
  </template>
</v-data-table>
...
<script>
  export default {
    data () {
      return {
        pagination: {
          sortBy: 'first_name'
        }
      }
    }
  }
</script>
```


src/components/tables/DataTable.js
```diff
@@ -177,56 +177,59 @@ export default {
...
-      this.$emit('update', {
-        page: this.page,
-        rowsPerPage: this.rowsPerPage,
-        sorting: this.sorting,
-        desc: this.desc
-      })
+    isSelected (item) {
+      return this.selected && this.selected.find(i => i[this.selectedKey] === item[this.selectedKey])
     },
     sort (index) {
```
johnleider: I think that this method could be cleaned up a bit. The same call is made for every scenario, the only thing that changes is the object that is emitted. Assign the value of that object and only have one this.$emit.

src/components/tables/DataTable.js
```diff
@@ -115,36 +105,51 @@ export default {
       type: Boolean,
       default: false
     },
-    defaultSort: {
+    selected: {
```
johnleider: I think we should let the selected be controlled under the model and the items be under a prop items, and just sync that. Would be curious to hear others thoughts.

```v-model="yourSelectedRows"```
```:items.sync="tableItems"```

File: src/components/tables/DataTable.js [hunk 3]
```diff
@@ -177,56 +177,59 @@ export default {
...
     }
   },
 
   mounted () {
-    const header = this.headers.find(h => !('sortable' in h) || h.sortable)
-    this.desc = this.defaultSort ? this.defaultSort.desc : this.desc
-    this.sorting = this.defaultSort ? this.defaultSort.field : header.value
+    const firstSortable = this.headers.find(h => !('sortable' in h) || h.sortable)
+
+    let defaultPagination = {
```
johnleider: should be const as it doesn't change value, could also just put it into the Object.assign arguments.

avengerweb:
Why default sort is part of pagination? Let's locate it in headers maybe?

nekosaur:
@avengerweb It's part of the synced prop so that sorting can be controlled outside of the data table. The end result is that it can also be used to set the default sorted column. Placing an option in the headers as well duplicates functionality which already exists.

src/components/tables/DataTable.js
```diff
@@ -3,7 +3,7 @@ import Body from './mixins/body'
 import Foot from './mixins/foot'
 import Progress from './mixins/progress'
 import { getObjectValueByPath } from '../../util/helpers'
-
+import Vue from 'vue'
```
johnleider: What's this guy for?

johnleider:
All reviewers have until Sunday to provide feedback.

PR #559, opened 2017-05-19
Remove 'execute' permission of all JS/Vue files and one .yml

stephane:
Only scripts such as `build/release.sh` must have the execution bit set on Unix system.


stephane:
Rebase against dev..

PR #560, opened 2017-05-19
Remove 'execute' permission of all JS/Vue files and few others

stephane:
Only directories and few script files must have the execution bit on Unix system.


johnleider:
Would it be useful to just perform these changes in build/release.sh? I fear over time permissions could get accidently moved to the incorrect ones.

stephane:
It's possible to add a command such as`find -name '*.js' -exec chmod -x {} \;` in `release.sh` but these commands (find and chmod) aren't available on all systems (or the syntax is different, find MacOS vs find on Linux for example) and if by mistake, someone changes the permissions of a file, git will detect the change.

[1] `node_modules` must be excluded

johnleider:
So that is the issue I have with this. We can update the file permissions from this PR. But in the future I will probably forget and one might get changed to something improperly.

stephane:
You can set the core.filemode setting of your got to avoid that.

johnleider:
Wonderful, ty for that. When I get home from work today I'll go ahead and get this in.

stephane:
Updated.

Issue #576, opened 2017-05-22
Min / Max / Step Support for Number Input Types


StratusBase:
### Versions
master

### What is expected ?
When specifying the available "min", or "max" values, the numeric input types abide by them.

Also, support for the native "step" property would be awesome.

### What is actually happening ?
Nothing, when specified the numeric input does not receive the value and it allows the input to go beyond the specified min / max values and it completely ignores "step".

Fiddle:
https://jsfiddle.net/b00jL1mt/1/

johnleider:
Please follow the <a href="https://github.com/vuetifyjs/vuetify/blob/master/.github/CONTRIBUTING.md">guidelines</a> on how to report an issue. In particular, provide an example on <a href="https://jsfiddle.net/">www.jsfiddle.net</a> (or a similar service) that reproduces the problem. If necessary, create a repository for us to clone with a minimal reproduction. repositories of actual projects will generally not be accepted .

Thank you.

StratusBase:
@johnleider - Added a Fiddle to the request

johnleider:
A few things from review.

- maxlength does not work on number inputs as far as I'm aware
- min/max are primarily used for the counter and are not actually applied to the element

If min/max are explicitly applied, counter would be useless as you would never be able to go outside of them and would only serve as a display. I'm contemplating on how I'd like to address this.

StratusBase:
Sorry for the confusion, the maxlength property can be ignored as it does not belong on a number input, I think that was a copy/paste error on my part.

What about a new "v-number-field" component?

Then down the road, inputs like, tel, email, date, etc can just be their own self-contained components with their own properties, etc...

creativecoder:
+1 for this enhancement. Right now when you set `type="number"`, some browsers (like Chrome) enforce a default `step` value of 1, which means you can't enter a decimal value like `1.1` when submitting a form--the browser's implementation of form validation will block form submission.

I don't like the idea of a separate component for each input type, I think this would be hard to maintain and possibly more complicated to use.

Note that min, max, and step are also valid attributes for `date-time` in addition to `number` input types.

What about setting min and max attributes only when counter is false? It seems like you would use min/max attributes to validate numeric input values rather than a counter anyway, so no functionality is lost.

The step field should be able to come through as is, since it's not currently used in the vuetify API

johnleider:
I agree with applying min/max if counter is false, seems like the best implementation.

NicKAnderson038:
[...]

PR #589, opened 2017-05-23
[#576] Pass through min, max, and step attributes for v-text-field

creativecoder:
Allows setting these attributes on number and date-time type inputs.

Min and max are passed through as attributes unless using a counter,
and so are used as min and max lengths for the counter instead.

### DIFF
--- a/src/components/forms/TextField.js
+++ b/src/components/forms/TextField.js
@@ -20,14 +20,9 @@ export default {
     id: String,
     name: String,
     maxlength: [Number, String],
-    max: {
-      type: [Number, String],
-      default: 25
-    },
-    min: {
-      type: [Number, String],
-      default: 0
-    },
+    max: [Number, String],
+    min: [Number, String],
+    step: [Number, String],
     multiLine: Boolean,
     prefix: String,
     readonly: Boolean,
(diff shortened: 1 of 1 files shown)


src/components/forms/TextField.js
```diff
@@ -20,14 +20,9 @@ export default {
...
-    max: {
-      type: [Number, String],
-      default: 25
-    },
-    min: {
-      type: [Number, String],
-      default: 0
-    },
+    max: [Number, String],
+    min: [Number, String],
```
johnleider: Allowing a string for min and max will cause logic issues down below.

src/components/forms/TextField.js
```diff
@@ -60,11 +55,17 @@ export default {
       const inputLength = (this.inputValue && this.inputValue.toString() || '').length
       let min = inputLength
 
-      if (this.min !== 0 && inputLength < this.min) {
-        min = this.min
+      if (this.counterMin !== 0 && inputLength < this.counterMin) {
```
johnleider: You compare this to an absolute equal of 0, but since a string can be passed, it could be '0'.

creativecoder: @johnleider Good catch--looks like this is a bug in the current version as well. Take a look at the revised computed properties for `counterMin` and `counterMax`. Or let me know if you have another way you prefer to handle this issue.

johnleider: I don't see any changes.

src/components/forms/TextField.js
```diff
@@ -60,11 +55,19 @@ export default {
...
-      return `${min} / ${this.max}`
+      return `${min} / ${this.counterMax}`
+    },
+    counterMin () {
+      const parsedMin = Number.parseInt(this.min, 10)
+      return Number.isNaN(parsedMin) ? 0 : parsedMin
+    },
+    counterMax () {
+      const parsedMax = Number.parseInt(this.max, 10)
+      return Number.isNaN(parsedMax) ? 25 : parsedMax
```
creativecoder: @johnleider See above. `counterMin` and `counterMax` will always return integers.

johnleider: looks good.

PR #610, opened 2017-05-26
Fix Icon Component inner-text parsing when using FA icon with multiple words

beauvankirk:
For cases where you want to use a font-awesome icon specified by multiple font-awesome classes in a Vuetify Icon component, the inner-text parsing/prefixing logic wasn't working. For example, when the desired rendering is a rotated or flipped icon, with plain font-awesome html, one would use
`<i class="fa fa-shield fa-rotate-90"></i> fa-rotate-90<br>`
The current master branch splits inner text and joins with 'fa- ' which in no case was giving me the desired result.  The following seemed to ensure that regardless of inner text leading/trailing whitespace, any multi-word inner text content would have each word correctly prefixed with 'fa-' and adding the necessary leading space to the overall string:
`` ` fa-${text.trim().split(' ').join(' fa-')}` ``


mfferreira:
@beauvankirk Please submit the PR to the `dev` branch for review.

PR #643, opened 2017-05-31
Provide item's index in DataTable scoped slot

mtpc:


johnleider:
I had to revert this. It was my fault for not testing it as it seemed harmless on the surface, but it completely breaks the data tables. Will have to revisit later.

mtpc:
> but it completely breaks the data tables.
Does it? I found no issue using the patch whatsoever 🤔

PR #646, opened 2017-05-31
637 nav drawer list per guidelines

basejump:


johnleider:
Is this in addition to #629 or does this super-cede that?

basejump:
@johnleider I'll merge these into one simple pull reqest

Issue #644, opened 2017-05-31
Second `v-navigation-drawer` with `persistent` and `right` expected to push content to left when shown.


duebbert:
### Steps to reproduce

1. Use layout 6 as basis.
2. Add `v-navigation-drawer` with attributes `right` and `persistent`

### Versions

0.12.2

### What is expected ?

When showing the right navigation drawer the content would be pushed to the left just like it is pushed to the right when showing the left navigation drawer.

### What is actually happening ?

Right navigation drawer is opening as overlay over the `main` content.

### Reproduction Link

https://codepen.io/anon/pen/OmKxar

maurop123:
Just dropping my two cents. Having the option to push the content or overlay it seems like a good option to have either the left hand drawer or the right hand one.

maurop123:
Interestingly, in this version of your fiddle, the right-hand drawer is placed above the content, and when it's opened it pushes the content like the left-hand drawer. Odd, huh?

https://codepen.io/maurop123/pen/XRvqao

maurop123:
I've got a fix for this issue. Submitting a pull request

duebbert:
Thanks Mauro, this is a great start and useful if we have a navigation drawer on the right with `persistent`. 

My goal would be to have 2 navigation drawers: one on the left and one of the right. Both should push the contents accordingly to the right or left. In your improved version this still doesn't work because the navigation drawer expects a toolbar to come as direct sibling.

Or should I open a new issue to cover having 2 navigation drawers before a toolbar?

maurop123:
I think I see what you're saying. I'll take another look.

jdriesen:
If I can ask... any news about the 2 navigation drawers ?
Maybe an example ? Thanks in advance for any reply.

johnleider:
No news, already works.

jdriesen:
I'm searching now since a rather long time for a template example with 2 navigation drawers...
@johnleider  ... can you maybe point me in the right direction ?

johnleider:
https://vuetifyjs.com/layout/pre-defined example #3

jdriesen:
First of all, thank you for replying on a Sunday evening.
I guess there is a small misunderstanding in the definition of "double navigation drawers" (100% my fault, I'm new to vuetify ...)

I'm trying to rebuild [this layout ](http://develop.ireporter.be/) with Vuetify (it's currently made with BS)
See the double navigation bar on top...

[...]

PR #650, opened 2017-05-31
fix for #644

maurop123:
Can test against this code pen https://codepen.io/maurop123/pen/XRvqao


maurop123:
If there's any other styles than the one I changed that may be involved in the drawers opening an closing, with a point in the right direction I can figure it out.

johnleider:
I had a few changes, and the PR should go to dev. re-opening.

PR #656, opened 2017-06-01
Extend allowed dates to time pickers

### LINKED ISSUE #594: Extend allowed dates to time pickers
It would be cool if the allowed dates functionality from the date pickers could be implemented to time pickers as well. Use case can be a site where you can create apointments only during the opening times and only every half an hour.

nekosaur:
Fixes #594


johnleider:
Extensive changes! At a glance looks good, will dive deeper later.

src/components/pickers/TimePicker.js
```diff
@@ -145,6 +160,56 @@ export default {
...
+      const allowed = this[`allowed${type.charAt(0).toUpperCase() + type.slice(1)}s`]
+
+      if (!allowed) return true
+
+      if (Array.isArray(allowed)) {
+        return !!allowed.find(v => v === value)
+      } else if (allowed instanceof Function) {
+        return allowed(value)
+      } else if (allowed instanceof Object) {
+        const range = type === 'minute' && this.ranges.minutes || !this.is24hr && this.ranges.hours.slice(1, 13) || this.ranges.hours
```
johnleider: Can we move the conditionals to new lines.
```
type === 'minute' &&
  this.range.minutes ||
  ...
```

src/components/pickers/TimePicker.js
```diff
@@ -145,6 +160,56 @@ export default {
...
+        const mod = type === 'minute' && 60 || !this.is24hr && 12 || 24
+        const offset = type === 'hour' && !this.is24hr ? 1 : 0
+
+        let steps = allowed.max - allowed.min
+        steps < 0 && (steps = steps + mod)
+
+        let j = allowed.min
```
johnleider: j is never reassigned, could be const.

src/components/pickers/mixins/time-body.js
```diff
@@ -106,24 +113,19 @@ export default {
...
-          ? 12 : time > 0 && this.hour === 12
-          ? 1 : this.hour + time
-      } else {
-        this.hour = time < 0 && this.hour === 0
-          ? 23 : time > 0 && this.hour === 23
-          ? 0 : this.hour + time
-      }
+      let range = this.generateRange('hour', this.hour)
+
+      time < 0 && (range = range.reverse().slice(1))
```
johnleider: I think this should be triggering lint, I could be wrong.

src/components/pickers/TimePicker.js
```diff
@@ -145,6 +160,56 @@ export default {
...
+        return !!allowed.find(v => v === value)
+      } else if (allowed instanceof Function) {
+        return allowed(value)
+      } else if (allowed instanceof Object) {
+        const range = type === 'minute' && this.ranges.minutes || !this.is24hr && this.ranges.hours.slice(1, 13) || this.ranges.hours
+        const mod = type === 'minute' && 60 || !this.is24hr && 12 || 24
+        const offset = type === 'hour' && !this.is24hr ? 1 : 0
+
+        let steps = allowed.max - allowed.min
+        steps < 0 && (steps = steps + mod)
```
johnleider: I think this should be triggering lint, I could be wrong.

src/components/pickers/mixins/time-body.js
```diff
@@ -106,24 +113,19 @@ export default {
...
       return true
     },
     changeMinute (time) {
       const current = Number(this.minute)
+      let range = this.generateRange('minute', current)
 
-      const minute = time < 0 && current === 0
-        ? 59 : time > 0 && current === 59
-        ? 0 : current + time
+      time < 0 && (range = range.reverse().slice(1))
```
johnleider: I think this should be triggering lint, I could be wrong.

src/components/pickers/TimePicker.js
```diff
@@ -171,16 +171,18 @@ export default {
       } else if (allowed instanceof Function) {
         return allowed(value)
       } else if (allowed instanceof Object) {
-        const range = type === 'minute' && this.ranges.minutes || !this.is24hr && this.ranges.hours.slice(1, 13) || this.ranges.hours
+        const range = type === 'minute' && this.ranges.minutes ||
+          !this.is24hr && this.ranges.hours.slice(1, 13) ||
```
johnleider: .slice returns a copy and doesn't actually modify the array. So I'm not sure what this line does.

nekosaur: It limits the range of allowable hours to 1-12 when not in 24hr mode. Although now that I think about it that gives me no way of distinguishing between AM/PM hours. 12hr format is the worst.

Issue #622, opened 2017-05-29
Feature request: Being able to specify first day of the week in date picker


Brimstedt:
It would be great if it was possible to specify first day of week in date picker. 

It is an i18n issue, lots of places consider Monday to be first day of the week and it is confusing to have the "wrong" day first.

> What will it allow you to do that you can't do today?

View the calendar according to users locale.

> How will it make current work-arounds straightforward?

Not aware of any workarounds for this.

> What potential bugs and edge cases does it help to avoid?

It will make it easier for users who does not look at the day titles to select the correct day (i.e it's easy to assume that monday is in the leftmost column)

mornir:
An attribute similar to the firstDayOfWeek attribute in Element CSS Framework would be awesome:
http://element.eleme.io/#/en-US/component/date-picker#picker-options

PR #705, opened 2017-06-06
Added firstDayOfWeek prop

nekosaur:
Solves #622


mornir:
Awesome! 
If I understand your pull request correctly, Monday will be the default first day of the week? As a European citizen, I can only approve. However, our American friends might not be too happy about it ^_^

nekosaur:
Oh, no I was planning on having it be Sunday. Must have forgotten to change it back after testing.

Issue #708, opened 2017-06-06
list-group has no reactive height


cawa-93:
### Steps to reproduce
1. Create a group that is open by default. This will set a fixed value for the height of the list. (In my example, this is 224 pixels.)
2. The elements of the list must be reactive.
3. Add a new item to the list. Because of the fixed height, the new elements of the list go beyond it.
### Versions
v0.12.7

### What is expected ?
The height of the list should change according to the dynamic content
### What is actually happening ?
The first time the page loads, a fixed list height is set. It stays that way until you hide it and open the list again. After that, the height is indicated as "auto".

### Reproduction Link
https://codepen.io/cawa-93/pen/vZOKPd

PR #709, opened 2017-06-06
Fix for #708

cawa-93:
Fix for #708
Set the default height value as "auto"

### DIFF
--- a/src/components/lists/ListGroup.js
+++ b/src/components/lists/ListGroup.js
@@ -33,7 +33,7 @@ export default {
     },
     styles () {
       return {
-        height: `${this.height}px`
+        height: `auto`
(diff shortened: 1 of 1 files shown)


johnleider:
There are a few issues with this. Your implementation is fine but the expand-transition class already sets the height to auto after transition. The problem is that we are bothering to set the height at all within this component.

I've also tested your current implementation and it has some visual issues when opening and closing lists rapidly. With that being said, I have a fix that I'm going to push up that will reference this. I'll still make sure you get credit for this in the release notes.

cawa-93:
@johnleider Of course, you are better at this. For me, it's important that the bug is fixed. Thank you )

PR #751, opened 2017-06-12
Add prefix to component names

KaelWD:
Having the component's name the same as its tag allows JetBrains IDEs to autocomplete components and their properties while typing. This has the slight side-effect of adding a `V` to the front of the names in vue-devtools (`TabsBar` => `VTabsBar`, `Btn` => `VBtn`).

Most functional components are generated with the `createSimpleFunctional` helper, and could not have names added, all others are listed below. 

Functional components with existing names: 
 - Card
 - ListTileAction

Functional components with added names: 
 - App
 - BottomNav
 - CardRow
 - Divider
 - Footer
 - Flex
 - Layout
 - Container
 - Icon
 - Subheader
 - TabsItems

### DIFF
--- a/src/components/alerts/Alert.js
+++ b/src/components/alerts/Alert.js
@@ -2,7 +2,7 @@ import Toggleable from '../../mixins/toggleable'
 import Contextualable from '../../mixins/contextualable'
 
 export default {
-  name: 'alert',
+  name: 'v-alert',
 
   mixins: [Contextualable, Toggleable],
 
--- a/src/components/app/App.js
+++ b/src/components/app/App.js
@@ -1,6 +1,8 @@
 export default {
   functional: true,
 
+  name: 'v-app',
+
   props: {
     light: {
       type: Boolean,
--- a/src/components/app/AppBar.js
+++ b/src/components/app/AppBar.js
@@ -1,5 +1,5 @@
 export default {
-  name: 'app-bar',
+  name: 'v-app-bar',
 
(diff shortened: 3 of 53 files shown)


johnleider:
Thank you for the hard work on this. Some comments though:

I believe doing it this way (and technically the current way) leaves room for error. I'd rather add this to the build process where the name is automatically assigned.

Particularly on this line: https://github.com/vuetifyjs/vuetify/blob/master/src/index.js#L8
That way we could remove the name from all components and have it generated from the build.

Thoughts?

KaelWD:
I thought about doing it that way, but the problem I'm having is precisely *because* part of the naming is done dynamically.  
I'm not sure exactly how the IDE is looking for the source location, but it seems to need a static string (that's why `createSimpleFunctional` was a problem for me), and it doesn't care about the dist files - so adding the text with webpack wouldn't work either.

Ideally we'd be able to have a separate generated file with IDE hints (like TypeScript definitions I guess), but that's something for Jetbrains to deal with. 

If you look at [the way buefy handles it](https://github.com/rafaelpimpa/buefy/blob/dev/src/index.js#L57), they're setting the component name from `component.name`, which I think is pretty elegant. The problem then is adding a name to everything generated by `createSimple(Functional|Transition)`.

johnleider:
I'm not sure. I really feel ya with the ide issues. But there has to be something you can configure that doesn't require such a massive change to the system.

PR #759, opened 2017-06-13
Allow to disable min-width on content element.

jursenbo:
There are cases when the width of `activator` element of Menu component is wider than the needed width of `content` element. For example, Date picker `content` is 290px wide, but if the `activator` is wider, then by setting `min-width` style on `content`, unneeded white space appears right to the Date picker.

Here is a CodePen to show the issue - https://codepen.io/anon/pen/bReKmX. Click on the input field and see the extra white space on the right of the Date picker.

I have added a new prop for Menu element - `disableContentMinWidth`, which sets the `min-width` style of the `content` to `0` if set to `true`.

### DIFF
--- a/src/components/menus/Menu.js
+++ b/src/components/menus/Menu.js
@@ -127,20 +127,22 @@ export default {
       type: Boolean,
       default: false
     },
-    disableContentMinWidth: {
-      type: Boolean,
-      default: false
+    minWidth: {
+      type: Number,
+      default: null
     }
   },
 
   computed: {
-    minWidth () {
-      return this.dimensions.activator.width + this.nudgeWidth + (this.auto ? 16 : 0)
+    calculatedMinWidth () {
+      return this.minWidth !== null
+        ? this.minWidth
+        : this.dimensions.activator.width + this.nudgeWidth + (this.auto ? 16 : 0)
     },
     styles () {
       return {
         maxHeight: this.auto ? '200px' : isNaN(this.maxHeight) ? this.maxHeight : `${this.maxHeight}px`,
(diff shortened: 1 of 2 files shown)


johnleider:
In this particular situation, do you feel it may be better to instead just add a prop for minWidth, rename the current computed property, and let the dev simply overwrite the minimum width. Thoughts?

jursenbo:
That was my other idea, I guess it would make more sense to do it that way, as it would be more flexible and in the future might become handy for devs.

src/components/menus/Menu.js
```diff
@@ -126,6 +126,10 @@ export default {
     positionAbsolutely: {
       type: Boolean,
       default: false
+    },
```
johnleider: rename this to minWidth, should accept number or string

src/components/menus/Menu.js
```diff
@@ -136,7 +140,7 @@ export default {
     styles () {
       return {
         maxHeight: this.auto ? '200px' : isNaN(this.maxHeight) ? this.maxHeight : `${this.maxHeight}px`,
-        minWidth: `${this.minWidth}px`,
+        minWidth: this.disableContentMinWidth ? 0 : `${this.minWidth}px`,
```
johnleider: this should only reference ```this.calculatedMinWidth``` which will be the new name for ```this.minWidth```

jursenbo: Good, will do.

jursenbo:
PR has been updated with agreed changes. I had to update also `position.js` file, which used `minWidth`.

PR #768, opened 2017-06-14
Use variable $card-background

jleveugle:
Hello,

Actually, we can't override the background-color of cards.

In _variables file, there is a variable for that : $card-background which is never used.

```
$card-actions-border-top := 1px solid rgba($material-theme.fg-color, .1)
$card-background := $material-theme.bg-color
$card-padding := 16px
...
```

I just adding the line to use this one.

Have a good day,

### DIFF
--- a/src/stylus/components/_cards.styl
+++ b/src/stylus/components/_cards.styl
@@ -1,12 +1,13 @@
 .card
   elevation(2)
   position: relative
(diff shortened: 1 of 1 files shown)


johnleider:
Card background is not determined that way anymore. Please check dev for the latest implementation.

Issue #762, opened 2017-06-14
Tabs not rendering properly after deletion


eclipseAce:
### Steps to reproduce
Use tabs like this
```xml
<div id="app">
  <v-tabs scroll-bars light v-model="current">
    <v-tabs-bar slot="activators">
      <v-tabs-item v-for="tab in tabs" :key="tab.id" :href="'#' + tab.id" @dblclick.native="closeTab(tab)">
        {{ tab.text }}
      </v-tabs-item>
      <v-tabs-slider></v-tabs-slider>
    </v-tabs-bar>
    <v-tabs-content v-for="tab in tabs" :key="tab.id" :id="tab.id">
      <v-card flat>
        <v-card-text>This is {{ tab.text }}</v-card-text>
      </v-card>
    </v-tabs-content>
  </v-tabs>
</div>
```
and the scripts
```javascript
export default {
  name: 'app',
  data() {
    return {
      tabs: [
        { id: 'item1', text: 'Item 1' },
        { id: 'item2', text: 'Item 2 blablablabla' },
        { id: 'item3', text: 'Item 3' },
        { id: 'item4', text: 'Item 4' }
      ],
      current: null
    }
  },
  methods: {
    closeTab(tab) {
      const index = this.tabs.indexOf(tab);
      this.tabs.splice(index, 1);
      const nextIndex = Math.min(index, this.tabs.length - 1);
      this.current = this.tabs[nextIndex].id;
    }
  }
}
```
Double click on tab 'ITEM 2 blablablabla...'
### Versions
Vue 2.3.4
Vuetify 0.12.7

### What is expected ?
Tab 'ITEM 3' should be focused

### What is actually happening ?
Tab 'ITEM 2 blablabla...' disappeared, but the length of slider bar was not changed and 'ITEM 3' was not focused

I've read the Tabs.js in src, I think the problem is the watcher of 'activeIndex' wasn't triggered when reducing tabs

### Reproduction Link
https://codepen.io/eclipseAce/pen/wezKNz

johnleider:
Tabs doesn't (and cant) watch the length of its children.

johnleider:
I see that you are working on a fork with potential changes, just make sure that you are basing these changes off of what exists in DEV as the structure is quite different.

PR #769, opened 2017-06-15
Fix tabs issue #762

eclipseAce:
Now tabs works fine, in case of tab count get changed but active index doesn't


johnleider:
So I guess I didn't vet this as good as I should have. The more I look at this, the implementation seems very specific towards reacting to a click event, which may not always be the reason that the tab-item is removed. Also, this implementation does not work in current dev. I'm going to work on something that should accommodate your issue, but with a different setup.

PR #799, opened 2017-06-20
Prevent possible error "Cannot read property 'staticClass' of undefined"

smith64fx:
Hi,
Sometimes I get this error.

I updated the check for this variable, so it's more safe than now.


johnleider:
Please re-open the PR towards **dev** branch with only your specific changes, no dist files please.

Issue #796, opened 2017-06-20
Badge directive should have an option to hide it


keystorm:
Sometimes you need to reactively show or hide a badge. It would be great to be able to do something like this:
``` vue
<span v-badge="{value: pending, hidden: pending == 0}">
```
Is this even possible out of the box? I've tried to `v-badge="null"` and others only to have it spit an `undefined` into the badge.

This should be a really easy fix, but I've never done a PR before.

PR #803, opened 2017-06-20
Added functionality to hide/show badges

nekosaur:
Related to https://github.com/vuetifyjs/vuetify/issues/796

Example at https://github.com/vuetifyjs/docs/pull/85

### DIFF
--- a/src/directives/badge.js
+++ b/src/directives/badge.js
@@ -15,8 +15,7 @@ function directive (el, binding) {
   )
 
   if (!config.visible || (binding.expression && !binding.value)) {
-    el.classList.remove('badge', 'badge--left', 'badge--icon', 'badge--bottom', 'badge--overlap')
-    el.removeAttribute('data-badge')
+    el.classList.remove('badge')
 
     return
   }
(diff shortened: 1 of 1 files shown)


src/directives/badge.js
```diff
@@ -9,10 +9,18 @@ function directive (el, binding) {
       icon: false,
       left: false,
       bottom: false,
-      overlap: false
+      overlap: false,
+      visible: true
```
johnleider: I like this, but whatever property name is used, we must ensure that if we add functionality similar to other directives, or if there is already functionality like this, that the names match up. Similar API is important.

src/directives/badge.js
```diff
@@ -9,10 +9,18 @@ function directive (el, binding) {
...
     }
   )
 
+  if (!config.visible || (binding.expression && !binding.value)) {
```
johnleider: I like the way you did this, instead of creating a new class that hides it, you remove all applicable classes.

Question, what do you think about just adding one of the display classes instead? Would be more future proof as the names badge--left, badge--icon could change or be added to in the future.

PR #837, opened 2017-06-28
Allow a route-link to have a download attribute

catskull:
Pretty self-explanatory. I needed to use a `<v-list-title>` to link to a file and download it instead of opening it in a new tab or the current window.

This prop can be used like this:
```
<v-list-tile href="file.pdf" download='true'>
    <v-list-tile-title>Download</v-list-tile-title>
</v-list-tile>
```
The above will generate something like:
`<a href="file.pdf" download>Download</a>`

You can also pass in a string you want the file download's file name to be (only works on same-origin urls per HTML5 spec, if the domain is not same-origin, it will just ignore your string and use the file name the file was originally called). You would use it like this:
```
<v-list-tile href="file.pdf" download='fileName'>
    <v-list-tile-title>Download</v-list-tile-title>
</v-list-tile>
```
And it would generate something like this:
`<a href="file.pdf" download="fileName">Download</a>`

### DIFF
--- a/src/mixins/route-link.js
+++ b/src/mixins/route-link.js
@@ -10,7 +10,8 @@ export default {
     ripple: Boolean,
     router: Boolean,
     tag: String,
-    target: String
+    target: String,
+    download: String
   },
 
   methods: {
@@ -51,6 +52,12 @@ export default {
         if (tag === 'a') {
           data.attrs.href = options || 'javascript:;'
           if (this.target) data.attrs.target = this.target
+
+          if (this.download === 'true') {
+            data.attrs.download = ''
+          } else if (this.download) {
+            data.attrs.download = this.download
+          }
         }
 
         data.on = { click: this.click }


johnleider:
While this will work, and is traditionally how I did it in the past. I feel there is a never-ender supply of attributes that could technically be added to this. I'd like to simple just tell Vue to put those on a specific element as opposed to what it's doing now.

You **can** do this with functional components, I assume there must be a way for regular.

johnleider:
$vnode.data.attrs is where they are located.

catskull:
If we're talking about just <a> tags, then there's only a handful of attributes possible and even less that make sense to use. Here's the full list: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

Not sure what you mean by $vnode.data.attrs

johnleider:
My comment above, where instead of adding each attribute as a prop, we could simply parse them from $vnode (this.$vnode), and then place them wherever and instead of having to have prop declarations this can just handle automatically. Does that make sense?

catskull:
Ah, so $vnode.data.attrs would kind of be a dynamic list of optional params. I'll look into it and submit a PR if I get it working.

johnleider:
In light of that, I'm going to close this and wait for a different implementation.

PR #849, opened 2017-06-29
Allow download attribute on any component that uses route-link mixin

catskull:
Updated version of #837

Instead of using a prop on route-link mixin, this will just look at the attributes on the tag that's wrapping route-link. So basically it's an optional prop.

It works on any component that uses route-link. For example, on a button, you would do this:
```
      <v-btn
        dark
        flat
        tag="a"
        href="/file.pdf"
        download
      >button text
      </v-btn>
```


catskull:
Closing for now

PR #856, opened 2017-06-30
Vuetify is an application dependency

stephane:


mornir:
https://github.com/vuetifyjs/webpack/pull/1

jamesxv7:
This is something that I'm curious about but I dont clearly understand the difference or benefits moving it to an application dependency since any how is going to be bundle in the final package? Maybe I'm missing something :/

stephane:
I think the rule of dumb here is to put dependencies imported by your app in _'dependencies'_ and the others (build chain or test tools) in _'devDependencies'_. Not sure this convention has other side effects.

nekosaur:
My two cents is that since vuetify is not actually required to run your project (i.e. when doing `npm run start` or equivalent), it should remain in dev dependencies. As @jamesxv7 said, it's going to be bundled when built.

stephane:
@nekosaur I don't understand what you mean, I can't serve my app if vuetify isn't installed (it's required as Vue or vue-router are)! The source code of vuetify will end in the final production build (but not the source code of the other devDependencies).

stephane:
BTW it's just  a matter of convention because once npm/yarn install them, they just are stored the same way in `node_modules`.
The meaning of each sections is better explained here:
https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

nekosaur:
I can see two scenarios

- The thing you're building is a self-contained product of some kind, say a website. When you want to use it you clone the repo, do `npm install` (which will install dev dependencies), then as an example `npm run start` (which will build your app and serve it). In this case vuetify does not have to be a regular dependency since it is bundled in the compiled js files

- The thing you're building is not self-contained. Perhaps it is a vuetify plugin, or some small part of a larger framework. In this case you would do `npm install --save/--save-dev your-thing` from whatever project. If the npm package includes already compiled files then you won't need vuetify in dev dependencies. However, if the package doesn't have compiled files then you do need vuetify in regular dependencies.

Like you said in the end it's a matter of convention (and personal preference), but I don't see how `--save-dev` is somehow incorrect in this case

stephane:
> In this case vuetify does not have to be a regular dependency since it is bundled in the compiled js files

Vuetify is bundled in the compiled JS file because it's required to run so that's the definition of a regular dependency as explained in the Stackoverflow link:

> dependencies are required to run, devDependencies only to develop, e.g.: unit tests, Coffeescript to Javascript transpilation, minification, ...

Yes, the use of `--save-dev` isn't incorrect, it just disturbing for me (and for others vuetifyjs/webpack#1)

jamesxv7:
Curiously, the templates aim to use it as a dependency. I think it would be neat to align this to it.

nekosaur:
> Vuetify is bundled in the compiled JS file because it's required to run so that's the definition of a regular dependency as explained in the Stackoverflow link

I'm not sure I agree. By that logic webpack is also a regular dependency if you're using chunking or code splitting since the compiled js files include a bootstrapper. Or babel if you're using es6 since it's required to transpile for your app to run.

But like I said earlier, this is just my two cents.

stephane:
Yes I understand that (and you can also include stylus loader if you push the idea this way).
When I look at the `package.json` of a project, I find it convenient to look at `dependencies`section to know the required work to upgrade. For example, to upgrade from v0.12 to v0.13, let's read the new devguide, code, code, tests and maybe user changes in the app. Upgrading a dev dependencies? may be it's just a `yarn add` or/and a few tweaks in `build/`directory).

johnleider:
Well, good points on all sides. Throwing it to a vote, upvote this for changing it to a dependency, downvote for leaving it as it is!

PR #868, opened 2017-07-03
Allow component as row in datatable

stephane:
With this change, it's possible to use component in datatable body such as:

```
<v-data-table
  v-bind:headers="headers"
  v-bind:items="items"
>
    <foo-row :bar="props.item"></foo-row>
  </template>
</v-data-table>
```

BTW I think it would be better if `<tr>` were required (explicit is better than implicit).

### DIFF
--- a/src/components/tables/mixins/body.js
+++ b/src/components/tables/mixins/body.js
@@ -22,10 +22,10 @@ export default {
 
           const row = this.$scopedSlots.items(props)
 
-          if (row.length && row[0].tag === 'tr') {
-            return row
-          } else {
+          if (row.length && row[0].tag === 'td') {
             return this.genTR(row, { attrs: { active: this.isSelected(item) } })
+          } else {
+            return row
           }
         })
       }


stephane:
What I can't do to help to review this patch (better explanation, unit tests, ...)?
@nekosaur you know very well this part of the code.

stephane:
I'm on vacation I'm not sure I'll able to merge it in the next days. Feel free to merge it in the meantime. Than you

S-YOU:
jeez, this break too much. I am selecting which columns to render inside template.
Now, It rendering every columns, I have to rewrite the the whole template. Do it in 2.0. Damn!

nekosaur:
Can you show us your markup in a codepen?

S-YOU:
I am not wasting the time to create damn codepen.

nekosaur:
There's not much we can do to help without seeing your markup.

S-YOU:
Template tag supposed to have many child tags.
Checking only first tag is td and drop all other children in template, 
returning the whole row is bullshit.

```
 if (row.length && row[0].tag === 'td') {
              return this.genTR(row, { attrs: { active: this.isSelected(item) } })
 ```

nekosaur:
Please keep it civil. Language like that is not tolerated.

Like I've mentioned twice now, unless you can show us your markup for the data-table, there's not much we can do to try to help.

johnleider:
If you can't be bothered to create a codepen to let us help you, then we can't be bothered to provide support for you.

smtlk:
@stephane you saved my ass on the project im developing. for that i humbly thank you :) 
I hope someday i'll be able to contribute something to this amazing project :) Thank you.

Issue #890, opened 2017-07-05
v-data-table: binding select-all to true throws error 


duebbert:
### Steps to reproduce
1. Create `v-data-table` with `select-all` set to `true`: `<v-data-table :select-all="true" ... >`

### Versions

0.13

### What is expected ?

Shouldn't result in any errors.

### What is actually happening ?

Throws an error about the color attribute wrongly set.

### Reproduction Link

https://codepen.io/anon/pen/GEGdbV?editors=1010

PR #893, opened 2017-07-05
Fixed bug when explicitly setting select-all prop to true

nekosaur:
Solves #890

### DIFF
--- a/src/components/tables/mixins/head.js
+++ b/src/components/tables/mixins/head.js
@@ -17,7 +17,7 @@ export default {
           props: {
             dark: this.dark,
             light: this.light,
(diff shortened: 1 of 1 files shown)


src/components/tables/mixins/head.js
```diff
@@ -17,7 +17,7 @@ export default {
           props: {
             dark: this.dark,
             light: this.light,
-            color: this.hasSelectAll ? this.selectAll : '',
+            color: this.selectAll === true && '' || this.selectAll,
```
johnleider: There is a computed prop for this, can we move the logic there?

PR #902, opened 2017-07-05
Display fixed toolbars in front of "default" toolbars

jrast:
"Default" toolbars are displayed in front of fixed toolbars if the toolbar is scrolled.
This PR fixes this and sets the `z-index` only for fixed toolbars.

As I had no time to test this, this PR is more like a issue with a proposed solution.


jrast:
Sorry, used the wrong base branch...

PR #910, opened 2017-07-06
Comply with the specs

arvanoid:
According to the [specs](https://material.io/guidelines/components/data-tables.html#data-tables-structure)

Table content (table.table tbody td) should be 13sp Roboto Regular
Column headers (rule table.table thead th) should be 12sp Roboto Medium

### DIFF
--- a/src/stylus/components/_tables.styl
+++ b/src/stylus/components/_tables.styl
@@ -55,7 +55,7 @@ table.table
       height: 48px
 
     td
-      font-weight: 300
+      font-weight: 400
       font-size: 13px
 
   .input-group--selection-controls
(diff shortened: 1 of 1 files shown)


johnleider:
Please re-submit to dev branch and I'll approve.

arvanoid:
I also noticed that Hover background is **Grey 200 (#EEEEEE)** instead of `rgba($material-theme.fg-color, $material-theme.divider-percent)`. Right now grey is darker than expected.

Do you think that it should also be changed from

```stylus
&:hover
  background: rgba($material-theme.fg-color, $material-theme.divider-percent)
```

 to

```stylus
&:hover
  background: $grey.lighten-3
```

PR #917, opened 2017-07-06
Added disabled items option

qw3n:
The function genTile now has an optional boolean variable for disabled items. I changed the noDataText item to be disabled. This prevents someone from clicking "No Data Available" and it returning that as the selected item. Now it merely returns an empty string. 

This would also allow in the future a public API addition that allowed disabled divider items in lists.

### DIFF
--- a/src/components/selects/mixins/generators.js
+++ b/src/components/selects/mixins/generators.js
@@ -91,7 +91,7 @@ export default {
       })
 
       if (!children.length) {
-        children.push(this.genTile(this.noDataText))
+        children.push(this.genTile(this.noDataText,true))
       }
 
       return this.$createElement('v-card', [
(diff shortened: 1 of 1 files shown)


src/components/selects/mixins/generators.js
```diff
@@ -91,7 +91,7 @@ export default {
       })
 
       if (!children.length) {
-        children.push(this.genTile(this.noDataText))
+        children.push(this.genTile(this.noDataText,true))
```
johnleider: fix spacing after , for lint

src/components/selects/mixins/generators.js
```diff
@@ -110,10 +110,10 @@ export default {
         props: item
       })
     },
-    genTile (item) {
+    genTile (item,disabled) {
```
johnleider: fix spacing after , for lint

src/components/selects/mixins/generators.js
```diff
@@ -110,10 +110,10 @@ export default {
...
       const active = this.selectedItems.indexOf(item) !== -1
       const data = {
-        nativeOn: { click: () => this.selectItem(item) },
+        nativeOn: { click: () => this.selectItem(disabled?'':item) },
```
johnleider: add spacing for lint between ternary operators

qw3n:
Changes made.

Issue #239, opened 2017-03-20
Icons not Displaying


tenjojeremy:
Hello,

Icons are not displaying  such as <v-icon medium class="grey--text text--darken-2">lock</v-icon>.

I just get the lock text.

This is my setup:

![example](https://cloud.githubusercontent.com/assets/25231946/24111516/0d7f2586-0d6d-11e7-96c4-c05bc62fff15.PNG)

Thank you in advance.

johnleider:
Please follow the <a href="https://github.com/vuetifyjs/vuetify/blob/master/.github/CONTRIBUTING.md">guidelines</a> on how to report an issue. In particular, provide an example on <a href="https://jsfiddle.net/">www.jsfiddle.net</a> (or a similar service) that reproduces the problem. If necessary, create a repository for us to clone with a minimal reproduction. repositories of actual projects will generally not be accepted .

Thank you.

iamike:
reproduction? almost all of the android didn't show icon well. don't you notice it?
![980929239](https://user-images.githubusercontent.com/526854/27983889-9a6adef2-63fa-11e7-9e73-5ccd5325706d.jpg)

PR #927, opened 2017-07-08
Fixed icon button's flex shrinking problem

patoezequiel:
Set the flex-shrink property of the .btn--icon class to 0, to preserve its round shape even when the screen shrinks too low. Currently the button takes oval shape.

image

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -69,6 +69,7 @@
     height: $button-height
     width: $button-height
     min-width: 0
+    flex-shrink: 0
 
     .btn__content
       height: $button-height


iamike:
Hi sorry for interrupted.
is there anyone can help this guy? because I met the same problem. I don't think this is a small issues? here is the  [link](https://github.com/vuetifyjs/vuetify/issues/239)

and also attached a screen shot. it's not on minority android. almost on every android 5 browser



if this is the problem that we can not fixed, then I will try to ask our team to using other html tag like <a> or <span> with btn class for prevent using v-btn...

please help us out. thanks.

johnleider:
@iamike Please follow the guidelines and post an issue with the requested information. The issue you are referencing did not do so and never did.

johnleider:
Re-submit to dev branch and I'll review.

Issue #734, opened 2017-06-09
Can swipe feature be added to the sidebar nav on mobile


zapping:
For this component https://vuetifyjs.com/layout/pre-defined

![swipefeature](https://user-images.githubusercontent.com/4123026/26970754-1fabf5b8-4d28-11e7-9609-a85c4f6dc249.PNG)

Can swipe feature be added to this on so that it can be closed on a mobile. 

- What will it allow you to do that you can't do today?
You need to have an explicit close button to close the sidenav bar or you have to leave enough space around the component so that it can be closed by touching outside it.

- How will it make current work-arounds straightforward?
If a swipe feature can be added to it then instead of the above or rather in addition to the about features a user can conveniently close it by swiping on the sidenav.

johnleider:
Fixed with #931

Issue #556, opened 2017-05-18
Support touch devices in v-carousel


Darkside73:
It would be nice, if we can slide carousel items with swipe

smares:
I think swiping should be added to all controls where it makes sense, like pickers and tabs for example, too.

evromalarkey:
better support would be great, it's the only thing that's missing to make a perfect hybrid mobile app

ram-you:
Make it MOBILE FIRST.

johnleider:
Fixed with #931

PR #931, opened 2017-07-08
enhancement/touch

nekosaur:
Added touch support for carousel, tabs, date-picker, navigation-drawer

Related to https://github.com/vuetifyjs/vuetify/issues/556, https://github.com/vuetifyjs/vuetify/issues/734


src/components/tabs/TabsBar.js
```diff
@@ -1,7 +1,9 @@
+import touch from '~util/touch'
+
 export default {
   name: 'tabs-bar',
 
-  inject: ['isScrollable'],
+  inject: ['isScrollable', 'isMobile'],
```
johnleider: This will not be reactive since it is an inject. I'm not sure if it matters to support the resizing of the window.

src/components/tabs/Tabs.js
```diff
@@ -6,11 +6,14 @@ export default {
   mixins: [Themeable],
 
   provide () {
-    return {
+    const obj = {
```
johnleider: is there a reason to assign and return instead of just returning an object?

src/util/touch.js
```diff
@@ -0,0 +1,126 @@
...
+
+  el.addEventListener('touchstart', e => touchstart(e, wrapper), { passive: true })
+  el.addEventListener('touchend', e => touchend(e, wrapper), { passive: true })
+  if (move) el.addEventListener('touchmove', e => touchmove(e, wrapper), { passive: true })
+
+  return wrapper
+}
+
+export default {
+  install,
```
johnleider: This may be grasping straws, but I'm not a fan of (install|uninstall). Even though this will be mainly used for developing components, this may be something we expose in the framework for the public. With that being said, maybe something _more_ semantic, add/remove.

I'm not sure, what are your thoughts?

johnleider: Maybe bind|unbind, so it's similar to a vue directive

src/components/tabs/TabsBar.js
nekosaur: Yeah I know. In this particular instance it doesn't matter, because these functions are only used inside a computed prop which also depends on a variable that changes with resize (isOverflowing).

src/components/tabs/Tabs.js
nekosaur: Nope, forgot to revert it after trying something out

src/util/touch.js
nekosaur: bind|unbind is better 👍

adallow:
How do I use this? I tried the example at https://vuetifyjs.com/examples/layouts/baseline and if I trigger the navigation draw by clicking on the top left menu option on my mobile device then eh draw pops out from the left as expected, however if I then try to swipe to close it it doesn't do anything. I have to click off it to close the draw. Any information is much appreciated.

johnleider:
The issue board is strictly for bugs. Please direct all questions to the official [Discord](https://chat.vuetifyjs.com) server.

Issue #795, opened 2017-06-20
"v-menu" appears as fixed if parent component is inside a scrolling container


matt-e-king:
I did my best to see if this has already been reported, but could find anything. I wouldn't really call this an issue, I am more curious what your thoughts are on adding a prop to `v-menu` to avoid mounting to the root of the app, that way a menu can exist inside of a scrolling container also.

Could possible check for a prop in the `mount` method:
(https://github.com/vuetifyjs/vuetify/blob/a79b018c557789b2cd4cf89ac3badbca40abd5fe/src/components/menus/Menu.js#L138)

### Steps to reproduce
Put a `<v-menu>` inside of a component that has a `overflow: auto` or `overflow-y: scroll`

### Versions
0.12.7

### What is expected ?
The menu should "stick" to the activator that opens the menu, e.g. Dropdown, Select.

### What is actually happening ?
If you place any component that has a `<v-menu>` inside of a parent component that is scrollable, the menu appears to be fixed and becomes "disconnected" from the activator

### Reproduction Link
Click the "Dropdown" button and then start scrolling up or down.
https://codepen.io/anon/pen/BZWOry

Thanks!!!

johnleider:
Maybe we could provide an ** absolute** prop for the detachable mixin that would append it to its parent instead of root. Thoughts?

matt-e-king:
Is there a "detachable" mixin? At first glance I thought it was all happening in the `mount()` method. 

Either way, yeah, a boolean prop seems to be the simplest. Could even be called `detach` and it is set to true by default. 

Anything else seems like overkill, unless it's on the roadmap to create some type of  "detachable" or "portal" mixin that allows you to control where in the DOM something mounts.

johnleider:
I know that neko had a change for this, but at this point, I don't know what the best course of action is.

matt-e-king:
Gotcha. Thanks for the follow up. So you're thinking that even this commit alone:
https://github.com/vuetifyjs/vuetify/pull/936/commits/57db3e5de48efeae527384ca01aa8073369f21f5
would still create some unwanted side-effects?

Related issue: I'm sure this has already been pointed out but, I also noticed if you've applied hover styles to a component that is the direct parent of a menu, when you hover over the menu, the hover styles are removed from the parent.

https://codepen.io/anon/pen/BZWOry

Thanks for all your work!

johnleider:
[...]

PR #936, opened 2017-07-09
enhancement/relative menu

nekosaur:
Menu now moves with scrolling parent even when using overflow. Also fixed issue where menu displayed on top of toolbar when scrolling past it.

Related to https://github.com/vuetifyjs/vuetify/issues/795


johnleider:
Neko, I don't think this is something we really need to pursue anymore. It will still have certain issues depending upon the setup and I'd rather solve the issue in another way.

PR #938, opened 2017-07-09
bug/router link

nekosaur:
- fixed router-link mixin so that href prop produces <a> tag
- added button tests, added vue-router dep for tests, bumped avoriaz dep


nekosaur:
hold off on merge, found a problem

PR #967, opened 2017-07-12
fix undefined overlay

greabock:


johnleider:
You included this in #969

PR #969, opened 2017-07-12
Improved toolbar behavior

greabock:
Better `<main>` section responsibility depends on toolbar state.

now css rule `padding-top` of  folowing `<main>`  take into consideration toolbar modifiers:

* toolbar-fixed
* toolbar-extended
* toolbar-dense
* toolbar-prominent

in all combinations

[old behavior](https://codepen.io/anon/pen/RgEeBy)
[new behavior](https://codepen.io/anon/pen/BZvqwO)

### DIFF
--- a/src/components/toolbar/Toolbar.js
+++ b/src/components/toolbar/Toolbar.js
@@ -25,14 +25,11 @@ export default {
         'toolbar--fixed': this.fixed,
         'toolbar--floating': this.floating,
         'toolbar--prominent': this.prominent,
-        'toolbar--extended': this.isExtended(),
+        'toolbar--extended': this.isExtended,
         'theme--dark': this.dark,
         'theme--light': this.light
       }
-    }
-  },
-
-  methods: {
+    },
     isExtended () {
       return this.$slots.extension || this.extended
     }
@@ -48,7 +45,7 @@ export default {
       'class': 'toolbar__content'
     }, this.$slots.default))
 
-    if (this.isExtended()) {
+    if (this.isExtended) {
       children.push(h('div', {
         'class': 'toolbar__extension'
       }, this.$slots.extension))
--- a/src/components/toolbar/Toolbar.js
+++ b/src/components/toolbar/Toolbar.js
@@ -25,9 +25,16 @@ export default {
         'toolbar--fixed': this.fixed,
         'toolbar--floating': this.floating,
         'toolbar--prominent': this.prominent,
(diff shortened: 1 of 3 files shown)


src/components/toolbar/Toolbar.js
```diff
@@ -25,12 +25,19 @@ export default {
...
         'toolbar--floating': this.floating,
         'toolbar--prominent': this.prominent,
+        'toolbar--extended': this.isExtended(),
         'theme--dark': this.dark,
         'theme--light': this.light
       }
     }
   },
 
+  methods: {
```
johnleider: For similarity of how we do this in other places, move this to a computed property

src/components/toolbar/Toolbar.js
```diff
@@ -41,9 +48,7 @@ export default {
       'class': 'toolbar__content'
     }, this.$slots.default))
 
-    if (this.$slots.extension ||
-      this.extended
-    ) {
+    if (this.isExtended()) {
```
johnleider: With the above change, this would be ```this.isExtended```

src/mixins/overlayable.js
```diff
@@ -35,11 +35,11 @@ export default {
 
       if (this.absolute) {
         // Required for IE11
-        this.$el.parentNode.insertBefore(overlay, this.$el.parentNode.firstChild)
+        this.$el.parentNode.insertBefore(this.overlay, this.$el.parentNode.firstChild)
```
johnleider: good catch, thank you!

greabock:
If you accept these changes next step will be to improve navigation-drawer--mini-variant behavior that also must depend on toolbar state. But it maybe more difficult cuz of css doesn't have 
"previous element" selector and we can't to define this rules based on toolbar state in css =(.

johnleider:
For expand-ability, I think a different approach will be needed. While this solves the issue now, it doesn't cover all potential issues or changes in the future. I'll have to talk with the team.

akaspick:
Sorry if this is the wrong place to comment, but I'm using Vuetify 0.16.6 and my toolbar is still covering my content as is shown in the OPs 'old behavior' example.  The 'new behavior' is what I'd like to see in my app.

I've used the same example with my own codepen using vuetify 0.16.6 https://codepen.io/anon/pen/RLYoOE

I'd like to see the content pushed down initially below the toolbar.

jacekkarczmarczyk:
@akaspick Can you please create a new issue? Thank you

PR #978, opened 2017-07-13
Better spacing (support grid-breakpoints)

Fl0pZz:
[Here](https://repl.it/J0HM/1) i try to show, how it works:
The big spaces between blocks will be only screen size more then 1024px

Sorry, my English is too bad to write documentation


johnleider:
Please make all PR's to dev

PR #983, opened 2017-07-13
Better spacing (support grid-breakpoints)

Fl0pZz:
[Here](https://repl.it/J0HM/1) i try to show, how it works:
The big spaces between blocks will be only screen size more then 1024px as example

Sorry, my English is too bad to write documentation

### DIFF
--- a/src/components/selects/Select.js
+++ b/src/components/selects/Select.js
@@ -164,11 +164,7 @@ export default {
           this.inputValue !== null &&
           typeof this.inputValue !== 'undefined'
         ) {
-        this.$nextTick(() => {
-          this.$refs.input.value = this.returnObject
-            ? this.getText(this.inputValue)
-            : this.getValue(this.inputValue)
-        })
+        this.$nextTick(() => (this.$refs.input.value = this.getValue(this.inputValue)))
       }
     },
     genLabel () {
--- a/src/components/selects/Select.js
+++ b/src/components/selects/Select.js
@@ -164,7 +164,11 @@ export default {
           this.inputValue !== null &&
           typeof this.inputValue !== 'undefined'
         ) {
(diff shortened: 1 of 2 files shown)


johnleider:
I like this, I only worry about the size it adds to the package, will check out later tonight.

johnleider:
Looking for some input from other dev members.

hrayr-artunyan:
This is very useful, I'm curious why it hasn't been reviewed and merged yet.

johnleider:
It has been reviewed, I forgot to remove the tag. It hasn't been implemented yet as there is still consideration for the increase to the output css size.

johnleider:
One of my current issues is that you remove the existing spacing declarations, making a pretty large breaking change. Is there a specific reason for this?

jacekkarczmarczyk:
You can always leave the old classes (.ma-4 etc) and just add new ones. But of course it will make the CSS size even bigger...

HendrikGrobler:
Any update on this? Would really like to use it ASAP.

johnleider:
I'm considering suggesting a different way of handling this but I'm not sure. Recently a global variable was added for determining the viewport size which would allow something like:

```
:class="{ 'pa-3': $vuetify.breakpoint.sm, 'pa-5': $vuetify.breakpoint.md-and-up }"
```

While this is a bit more to type, I'm considering this to be the solution, willing to hear feedback.

dohomi:
@johnleider I think you are right on this. With the variables dev has full flexibility and the bit of typing does not harm 👍

hrayr-artunyan:
@johnleider this is an excellent solution. The global variables allows for a lot more flexibility in other areas as well.

johnleider:
Closing this in favor of using global variable.

Issue #979, opened 2017-07-13
v-select item-text feature request


TheGinJiCode:
In v-select, can you add a way to make item-text and item-value source from an nested object.
Ex:
Inventories = [
   {
      id: 1,
      item: {
         id: 1
         name: 'Test'
      }
   }
]
item-text="item.name"
item-value="item.id"

Thanks in Advance :)

johnleider:
Fixed with #985

PR #985, opened 2017-07-13
Fix for #979

nekosaur:


johnleider:
Less of a bug and more of it wasn't a feature.

Issue #986, opened 2017-07-14
Pagination - ALL Bug


TheGinJiCode:
I have a bug in Pagination, using ALL in the rows per page dropdown. When I select ALL, the next page is still clickable even though the return next_page_url is null.

![image](https://user-images.githubusercontent.com/29909936/28197939-ce6308f6-688c-11e7-9a32-db5f3d629ff2.png)

![image](https://user-images.githubusercontent.com/29909936/28197969-046b06ba-688d-11e7-86c3-a0f67d927ec6.png)

johnleider:
Fixed with #987

PR #987, opened 2017-07-14
Fix for #986

nekosaur:
#986


johnleider:
What are chances of a test being written for this case? =]

PR #990, opened 2017-07-14
Fix editable select

Fl0pZz:
[Here](https://jsfiddle.net/fhLwq3m9/) is example how it works now (with bug)

On the picture - how it should works
screenshot_20170714_165947

### DIFF
--- a/src/components/selects/Select.js
+++ b/src/components/selects/Select.js
@@ -164,7 +164,11 @@ export default {
           this.inputValue !== null &&
           typeof this.inputValue !== 'undefined'
         ) {
-        this.$nextTick(() => (this.$refs.input.value = this.getValue(this.inputValue)))
+        this.$nextTick(() => {
+          this.$refs.input.value = this.returnObject
+            ? this.getText(this.inputValue)
+            : this.getValue(this.inputValue)
+        })
       }
     },
     genLabel () {


johnleider:
I'm not entirely sure what this fixes or how the returnObject property would define which field is used for the value?

Fl0pZz:
Nope, use `item-text` and `item-value` props to define field value

johnleider:
I'm going to need some more information about what this addresses, I still don't understand.

Fl0pZz:
Ok, let's look at these lines of [code](https://github.com/Fl0pZz/vuetify/blob/fix/select/src/components/selects/Select.js):
```js
this.$refs.input.value = this.returnObject
            ? this.getText(this.inputValue)
            : this.getValue(this.inputValue)
```
It does not change `this.inputValue`, we only show `item-text` instead of `item-value`

johnleider:
What I'm asking is _why_. I see the lines of code, I see what you are doing, I don't understand _why_ it matters. Maybe someone else can explain or help explain to me?

Fl0pZz:
For me, this is important, because I believe that the user should see the text, and not some identifier, which is then sent to the server.
What happens if there is a choice of language, but the identifiers will be UUID?
[Example](https://jsfiddle.net/fhLwq3m9/1/)

johnleider:
I think the best thing we can do here is you provide 2 examples of what it currently does and what it will do with your change. That's the part I don't understand, the difference it makes under what situations.

Fl0pZz:
I hope 2 of these pictures will be enough

How it looks now:
screenshot_20170714_211252

Language (And this is how it should look like):
screenshot_20170714_211401

Fl0pZz:
`> I'm not entirely sure what this fixes or how the returnObject property would define which field is used for the value?`
Talking about the `returnObject` property, in your code it use like [here](https://github.com/vuetifyjs/vuetify/blob/dev/src/components/selects/Select.js#L203)
and this is still an undocumented property

VasiliyIsaichkin:
It's gif show what happend

johnleider:
I honestly just needed the steps shown above, I go through lots of code and I wasn't using it like that. I see the issue now, reviewing.

Fl0pZz:
@johnleider hey! I also want to be in the release notes :)

PR #991, opened 2017-07-14
Keep data input type equal for boolean values

loduis:


johnleider:
Can you expand on what this fixes or enables?

loduis:
selection_001

selection_003
selection_004

johnleider:
Thank you! I'll write a test for this, then get it merged in.

johnleider:
After looking at this, and writing a test. It doesn't actually make sense. Can you post some markup? The only reason it would ever return null is if you had a value assigned, and if you are using the checkbox for booleans, there is no reason to assign a value.

loduis:
As you can see this is connected to a [store](https://vuex.vuejs.org/en/state.html)
If I convert this store to json (JSON.stringify), to save to a server, this data is sent with null rather than false

johnleider:
You're missing what I'm saying. ```v-checkbox v-model="model">``` will always alternate between true and false.

loduis:
this is is my markup
```html
    <v-card class="mt-4">
      <v-card-title class="title" primary-title>{{ $t('settings_.email_notification') }}</v-card-title>
      <v-card-text>
        <v-checkbox :label="'settings_.notifications.' + label"
          :key="label"
          :value="alert"
          v-model="dailyNotifications[label]"
          hide-details v-for="(alert, label) in dailyNotifications" />
      </v-card-text>
    </v-card>
```
this is part of that api json
```json
"daily_notifications": {
    "transactions": false,
    "payments": false,
    "due": false,
    "overdue": false
  }
```
after apply https://vuex.vuejs.org/en/forms.html Two-way Computed Property

I get object dailyNotifications, from store

Something happens in this configuration that is not true and false always

for this config vuetify run https://github.com/vuetifyjs/vuetify/blob/dev/src/mixins/selectable.js#L62

What am I doing wrong?


I am running vuetify the dev branch
selection_006

And within the project I'm importing the components I need as you can see

johnleider:
The problem is you are assigning :value when you don't need to though.

loduis:
Yes, If you have all the reason I had not noticed that simple fact since I was using the v-model

Thanks.

Issue #973, opened 2017-07-12
v-toolbar fixed with extended prop is overflowing content


dohomi:
An extended `toolbar fixed` is overflowing the main content of a page. Here a codepen as a showcase:
https://codepen.io/anon/pen/WOPQMr

dohomi:
A second showcase for showing the content overflow is with the v-tab example included and fixed toolbar prop.
https://codepen.io/anon/pen/yXZYKg

johnleider:
Fixed with #969

Denfie:
Solve this fix also the problem with [#5 Flexible toolbar and card toolbar](https://vuetifyjs.com/components/toolbars) when the toolbar is fixed and have card toolbar as overlay?

PR #994, opened 2017-07-14
Fix for #973

Denfie:
To give the main section the real top padding then the toolbar height changed.


johnleider:
This was fixed with #973

PR #1006, opened 2017-07-16
Fix new bug

loduis:
This can be something webpack
screenshot from 2017-07-16 09-51-29


johnleider:
What are the steps to produce this particular issue?

loduis:
This is fixed now

PR #1023, opened 2017-07-17
Refactored touch util to directive

nekosaur:


src/components/navigation-drawer/NavigationDrawer.js
```diff
@@ -29,7 +28,10 @@ export default {
     persistent: Boolean,
     right: Boolean,
     temporary: Boolean,
-    touchless: Boolean,
+    touchless: {
+      type: Boolean,
+      default: false
```
johnleider: This is redundant, Boolean type props default to undefined which evaluates to false.

johnleider:
Can we address some of the issues introduced as outlined by codebeat?

PR #1033, opened 2017-07-18
add to ignore __snapshots__

loduis:
This is generate when you run and test


nekosaur:
No, snapshouts _should_ be included. Otherwise there would be nothing to compare with.

PR #1044, opened 2017-07-18
enhancement/icon

nekosaur:
- Removed explicit mdi and fa props
- Instead if icon name starts with `something-` prefix, we use that instead of material icons. This is compatible with at least FontAwesome and MDI icons.


src/components/icons/Icon.spec.js
```diff
@@ -0,0 +1,53 @@
+import Icon from '~components/icons/Icon'
+import { test } from '~util/testing'
+import { compileToFunctions } from 'vue-template-compiler'
```
johnleider: Unused import

src/components/buttons/ButtonToggle.js
```diff
@@ -0,0 +1,86 @@
...
+
+      item.text && children.push(h('span', item.text))
+      item.icon && children.push(h('v-icon', item.icon))
+
+      return h('v-btn', {
+        key: index,
+        props: {
+          flat: true
+        },
+        nativeOn: {
```
johnleider: can be changed to ```on``` instead of ```nativeOn```

src/components/buttons/ButtonToggle.js
```diff
@@ -0,0 +1,86 @@
...
+    inputValue: {
+      required: false
+    }
+  },
+
+  computed: {
+    classes () {
+      return {
+        'btn-toggle': true,
+        'btn-toggle--selected': this.inputValue && !this.multiple || this.inputValue && this.inputValue.length > 0
```
johnleider: Can drop this.inputValue.length > 0 and just use this.inputValue.length

src/components/buttons/ButtonToggle.js
```diff
@@ -0,0 +1,86 @@
...
+  },
+
+  props: {
+    items: {
+      type: Array,
+      default: () => []
+    },
+    multiple: Boolean,
+    mandatory: Boolean,
+    inputValue: {
```
johnleider: alphabetize props

PR #1050, opened 2017-07-19
Allows to import components independently.

routegroup:
Right now if you wanted to extend any component you had to have some webpack settings from this package in your own and many more. It allows to load component or helper independently. 

Ex.
```js
import {Components} from 'vuetify'

let TextField = Components.TextField
``` 

Ex. 2 (of helper)
```js
import {Helpers} from 'vuetify'

let createSimpleFunctional = Helpers.createSimpleFunctional

const CardIcon = createSimpleFunctional('card__icon')
``` 

It gives you direct access to component object.


johnleider:
Thank you for your contribution, but the team is already working on a solution for a la carte components. You can follow the project here https://github.com/vuetifyjs/vuetify/projects/8

Issue #937, opened 2017-07-09
Please add support for stepper dark mode


shongololo:
It would be helpful to have a dark mode for the steppers so that we can use dark colour backgrounds. For now the text can be manually switched to white, but the icons and dividers are still black.

Codepen: https://codepen.io/shongololo/pen/pwxZKL

PR #1056, opened 2017-07-19
Fix for #937

nekosaur:
Added theme support to stepper

### DIFF
--- a/src/components/steppers/Stepper.js
+++ b/src/components/steppers/Stepper.js
@@ -1,5 +1,9 @@
+import Themeable from '~mixins/themeable'
+
 export default {
   name: 'stepper',
+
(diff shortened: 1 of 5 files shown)


src/components/steppers/Stepper.js
```diff
@@ -31,7 +35,9 @@ export default {
         'stepper--is-booted': this.isBooted,
         'stepper--vertical': this.vertical,
         'stepper--alt-labels': this.altLabels,
-        'stepper--non-linear': this.nonLinear
+        'stepper--non-linear': this.nonLinear,
+        'theme--dark dark--bg': this.dark,
```
johnleider: can remove dark--bg light--bg, not used

Issue #340, opened 2017-04-10
DataTable initial row per page value


MitchelLabonte:
### Steps to reproduce
- Add a datatable
- Set rows-per-page-items with an array that doesn't contain 5

### Versions
Firefox 45.8.0
Vuetify 0.10.4

### What is expected ?
The selected value of the "rows per page" dropdown should be the first option.

### What is actually happening ?
The "rows per page" dropdown is initially set to an empty value while the table displays 5 items per page.

### Reproduction Link
https://codepen.io/ElMitchL/pen/yMmZYz

silvioiannone:
The issue is still present on for 0.12.7: https://codepen.io/anon/pen/WOjbRb

maesklaas:
Still not working, any clue when this will be fixed?

xgrounds:
This issue has been fixed now.

PR #1083, opened 2017-07-21
Fix for #340

nekosaur:


src/components/tables/VDataTable.js
```diff
@@ -223,6 +223,7 @@ export default {
   created () {
     const firstSortable = this.headers.find(h => !('sortable' in h) || h.sortable)
     this.defaultPagination.sortBy = firstSortable ? firstSortable.value : null
+    this.defaultPagination.rowsPerPage = this.rowsPerPageItems[0]
```
johnleider: ```this.rowsPerPageItems[0]```, is it possible that this is empty and would throw an error?

nekosaur: Only if you provide an empty array to rowsPerPage prop. I'd call that user error :)

johnleider: I know that it seems redundant, but could we provide a warning if it is empty?

Issue #1082, opened 2017-07-21
Make <v-btn-toggle> a linking component for <v-btn>


qw3n:
## Idea
Right now `<v-btn-toggle>` accepts an items prop. What I propose is instead have `<v-btn-toggle>` use slots instead and accept `<v-btn>` components as children. In the backend the component would provide the same behavior it currently does by linking the provided buttons into a set of togglable buttons.

## API
The `<v-btn-toggle>` would have the same API except the items prop would no longer be necessary and it would have a slot for `<v-btn>`.

## Advantages
This would make the component much more flexible and allow for styling of the individualized buttons. Also, events could be attached to the button or to the button group depending on the needs of the developer. Also things like tooltips could be added to individual buttons something that is now currently not possible. Another advantage would be now that buttons are all the same if you want to change or reuse a button it can be copied and pasted instead of completely changing the way it is created.

nekosaur:
Has been merged into dev. Will be available in next release

PR #1087, opened 2017-07-21
enhancement/button toggle

nekosaur:
- updated btn-toggle according to #1082
- in the process, removed toggleable mixin from btn
- as a result, had to refactor bottom-nav, which now uses the same button-group mixin as btn-toggle


src/components/bottom-nav/VBottomNav.js
```diff
@@ -1,19 +1,51 @@
...
-      if (props.shift) data.staticClass += ' bottom-nav--shift'
-      if (props.value) data.staticClass += ' bottom-nav--active'
+    methods: {
+      isSelected (i) {
+        const item = this.getValue(i)
+        return this.active === item
+      },
+      updateValue (i) {
+        const item = this.getValue(i)
+        return this.$emit('update:active', item)
```
johnleider: Is the return necessary?

nekosaur: nope

johnleider:
This will be merged for 0.15 release.

PR #1090, opened 2017-07-21
fix warning on import directives.

loduis:
selection_001


johnleider:
Please only commit the changed files, or is yarn.lock needed for this?

PR #1093, opened 2017-07-22
Close drawer in mobil after change route

loduis:
1. In mobil drawer is alway close
selection_005

2. Open drawer for select one option

selection_003

3. Option is selected
selection_004

But the drawer no close automatic after select option

### DIFF
--- a/src/components/navigation-drawer/VNavigationDrawer.js
+++ b/src/components/navigation-drawer/VNavigationDrawer.js
@@ -66,7 +66,7 @@ export default {
 
   watch: {
     $route () {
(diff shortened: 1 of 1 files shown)


loduis:
this is my markup
```html
  <v-navigation-drawer
    dark
    persistent
    enable-resize-watcher
    disable-route-watcher
    :mini-variant.sync="mini"
    v-model="drawer">
  </v-navigation-drawer
```

johnleider:
I don't particularly like this, you are using disable route watcher. If we add a conditional, where it could be mobile or disabled-route-watcher, then I believe you can get un-intended consequences.

PR #1103, opened 2017-07-23
Fix import ripple directive

loduis:


johnleider:
This isn't really a fix, and more of a preference. The current implementation is how we plan to move forward with directive imports.

loduis:
What I personally see in the new implementation you have in mind is whether a future is thought to be able to have individual components, is that you need to import all the directives just to use one or two. For what I think personally is not optimal

johnleider:
Valid point, but webpack tree shaking takes care of that.

PR #1104, opened 2017-07-23
Fix import directive clickOutside

loduis:


johnleider:
The space is intentional, ty for the PR though.

PR #1114, opened 2017-07-24
add lint to test.

loduis:
Not more pull request with lint errors.
One command npm run test

see example: https://github.com/vuejs/vue/blob/dev/package.json#L26

This must be merged after the current lint errors are adjusted, that is why it fails


johnleider:
I would rather just add this to the CI task as opposed to changing the script honestly.

johnleider:
Added lint to travis CI

Issue #1117, opened 2017-07-24
Icons in floating Buttons and Fabs are not centered in MSIE11


julkuh:
### Steps to reproduce
Go to https://vuetifyjs.com/components/buttons and scroll down to **#6 Floating** Example.

### Versions
Microsoft Internet Explorer 11,
Vuetify JS 0.14.5

### What is expected ?
The icons inside floating Buttons are nicely centered.

### What is actually happening ?
The icons are right aligned due to unnecessary margin definition from default .btn__content

### Reproduction Link
https://vuetifyjs.com/components/buttons

Christilut:
I'm seeing them left aligned with vuetify 0.14.7 on iOS 10 (simulator) and iOS 9 (device)

johnleider:
The fix hasn't been released yet.

Christilut:
Oh, but it said in the release log:
`#1117 Fixed a bug where v-btn with the fab prop was not centering the icon on IE11`

johnleider:
Referring to this, 16116ecf0d8bfef31ab1a3dcda33a9e8366d7344. This bug fix caused the safari issue.

johnleider:
Referring to 80e0d2f25d09aff7c21ca795eb7267e702e68a75. This IE11 fix caused the safari issue, was fixed in the before mentioned commit. That particular commit hasn't been released.

PR #1118, opened 2017-07-24
fix for #1117

julkuh:
Overwrite the buttons default margin definition which is only needed for non-circular buttons.


johnleider:
Didn't see this until after, already fixed with https://github.com/vuetifyjs/vuetify/commit/ce918e7e7911ea6209e009df42f7383968721bc3. Thank you though!

PR #1138, opened 2017-07-25
Add directive clickOutside, fixed lint error.

loduis:
This is used in the line: https://github.com/vuetifyjs/vuetify/blob/dev/src/components/menus/VMenu.js#L220

### DIFF
--- a/src/components/menus/VMenu.js
+++ b/src/components/menus/VMenu.js
@@ -21,6 +21,10 @@ export default {
     Toggleable
   ],
 
+  directives: {
+    clickOutside
+  },
+
   data () {
     return {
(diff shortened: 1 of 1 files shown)


johnleider:
I actually removed this temporarily because I need to look into why when I add the directive that way, menu will not close on outside click.

loduis:
I would comment on the liine: https://github.com/loduis/vuetify/blob/ab4ae12f9db7f66fe5d7b8d0b5b30db557594f1e/src/components/menus/VMenu.js#L9

johnleider:
This is actually a separate issue that I'm fixing right now, ty.

PR #1143, opened 2017-07-26
support typescript class constructor

OniVe:
For objects with a constructor like:
`var MyClass = (function () {
	function MyClass(foo, bar) {
		this.foo = foo;
		this.bar = bar;
	}
	return MyClass;
}());`
Condition (`obj.constructor === Object`) fails.


johnleider:
Curious, why does that cause it to fail?

OniVe:
Here's a test: [codepen.io](https://codepen.io/OniVe/pen/yoYMQK)

PR #1153, opened 2017-07-27
Add TypeScript definitions for Vuetify plugin and createSimpleTransition

isaaclyman:
If you're writing a Vue app with TypeScript and you write `Vue.use(Vuetify)`, TypeScript throws an error because the Vuetify module hasn't been declared.

This pull request includes a tiny TypeScript definition file for the Vuetify plugin and one for the `createSimpleTransition` method as documented here: https://vuetifyjs.com/motion/transitions

I believe this covers the entire documented API surface area of Vuetify imports in a TypeScript application.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@ import './src/util/helpers'
 import { PluginFunction } from 'vue'
 
 declare class Vuetify {
-    static install: PluginFunction<never>
+  static install: PluginFunction<never>
 }
 
 export = Vuetify
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,8 @@
+import './src/util/helpers'
+import { PluginFunction } from 'vue'
+
+declare class Vuetify {
+    static install: PluginFunction<never>
+}
+
+export = Vuetify
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "license": "MIT",
   "homepage": "http://vuetifyjs.com",
   "main": "dist/vuetify.js",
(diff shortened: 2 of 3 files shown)


johnleider:
Why only the createSimpleTransition from the utils?

isaaclyman:
`createSimpleTranslation(string)` is the only function from the utils that's described the documentation (unless I missed something). My goal was to define types for documented methods, although if you can make an argument for documenting more methods (or less) I could definitely update the pull request.

johnleider:
Well I would think that there are probably a considerable amount of changes that we can make for typescript compatibility. I'm not versed in it at all, so I'll have to defer to other's opinions. There are, however, many more components in src/util/helpers.

isaaclyman:
I understand. To be honest, I'm scratching my own itch here. To meet the needs of 99% of users, all you need is the definition for the plugin function (so you can do `Vue.use(Vuetify)`). That's the biggest hangup for me, trying to include Vuetify in a TS project.

You shouldn't need to change anything to be compatible with TypeScript; the definition files can handle just about anything. But the only important things to put in the definition files are the classes and methods that you expect people to use. I took documentation as a proxy for actual usage data.

johnleider:
I just want to make sure I'm not opening up a rabbit hole for additional changes that need to take place.

isaaclyman:
@johnleider Sure, I've added a commit. Just to be clear, I don't know what those functions are or how to use them (because they aren't documented) but I've gathered as much information as I can from the code.

Issue #1166, opened 2017-07-28
Add class to expansion panel container


jacekkarczmarczyk:
Expansion panel provides css classes for header and body but not for the their container. Adding class to the container makes custom styling easier (actually it makes it possible at all, especially for the active panel)

My proposal is to add "expansion-panel__container" class for the container and "expansion-panel__container--active" for the active container.

I'm working now on the patch

johnleider:
Couldn't you just use the selector ```.expansion-panel > li```?

jacekkarczmarczyk:
How do i target expanded panel?

johnleider:
Can you show me a codepen of what you are trying to do please.

jacekkarczmarczyk:
https://codepen.io/anon/pen/QMjJZw

This is an example of custom styling the expanded panel

johnleider:
So what is wrong with what you have provided?

jacekkarczmarczyk:
Ok, now I got it, it should be something like this:

https://codepen.io/anon/pen/qXOLqM

So now the only thing that's wrong this solution is that I just personally found it more elegant if these classes are provided by the framework

johnleider:
Now this I like. Would you like to PR this or do you want me to integrate it?

jacekkarczmarczyk:
I've already made w PR (https://github.com/vuetifyjs/vuetify/pull/1167), not sure what to do next so maybe it's better if you integrate it
Sorry, I'm new in github...

PR #1167, opened 2017-07-28
fix for #1166

jacekkarczmarczyk:
This is my first PR for github project, hopefully everything is fine

### DIFF
--- a/src/components/expansion-panel/VExpansionPanelContent.js
+++ b/src/components/expansion-panel/VExpansionPanelContent.js
@@ -22,6 +22,12 @@ export default {
         'expansion-panel__header': true,
         'expansion-panel__header--active': this.isActive
       }
+    },
+    containerClasses() {
+      return {
+        'expansion-panel__container': true,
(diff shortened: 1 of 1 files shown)


johnleider:
Can you update this PR to reflect the styles you showed in the codepen please.

jacekkarczmarczyk:
@johnleider - not sure what you mean, the styles in this codepen are not in the scope of this issue, it's only example how one can style the expansion panel using new classes (without necessity of setting them manually). BTW - it was inspired by http://materializecss.com/collapsible.html#popout

However I've updated the classes in the codepen to reflect my commit, and added "popout" class to the panel just for testing purposes

johnleider:
Okay, maybe I'm misunderstanding. So the animation/style I like, I thought that's what you had meant. As far as needing the classes, I think it is perfectly acceptable to target the way you did in the codepen. Adding the classes just allows you to type one less selector.

Thank you for the PR and your contribution.

PR #1182, opened 2017-07-30
Added 'ripple' prop to v-checkbox

jrast:
As discussed in the chat, this adds a ripple prop to the checkbox component. Ripple can therefore be disabled with:

````
<v-checkbox :ripple="false"></v-checkbox>
````

This prop would also be nice for Radios and Switches. However, this would require a bigger code change.

### DIFF
--- a/src/components/selection-controls/VCheckbox.js
+++ b/src/components/selection-controls/VCheckbox.js
@@ -24,7 +24,11 @@ export default {
   },
 
   props: {
-    indeterminate: Boolean
+    indeterminate: Boolean,
+    ripple: {
+      type: [Boolean, Object],
+      default: true
+    }
   },
 
   computed: {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/components/selection-controls/VCheckbox.js [hunk 1]
```diff
@@ -24,7 +24,11 @@ export default {
   },
 
   props: {
-    indeterminate: Boolean
+    indeterminate: Boolean,
+    ripple: {
+      type: [Boolean, Object],
+      default: true
```
johnleider: make the default the default action we expect:

```{ center: true }```

```default: () => ({ center: true })```

File: src/components/selection-controls/VCheckbox.js [hunk 2]
```diff
@@ -56,17 +60,17 @@ export default {
...
-    const ripple = h('div', {
+    const rippleEl = h('div', {
       'class': 'input-group--selection-controls__ripple',
       on: Object.assign({}, {
         click: this.toggle
       }, this.$listeners),
       directives: [{
         name: 'ripple',
-        value: { center: true }
+        value: this.ripple ? { center: true } : false
```
johnleider: This should just be this.ripple

src/components/selection-controls/VCheckbox.js
```diff
@@ -56,17 +60,17 @@ export default {
       }, this.icon)
     ])
 
-    const ripple = h('div', {
+    const rippleEl = h('div', {
```
johnleider: Is there a reason to change the name?

jrast: No longer. Was fiddeling around with this and didn't change it back...

jrast:
Maybe this PR should be closed in favour of https://github.com/vuetifyjs/vuetify/pull/1184

PR #1184, opened 2017-07-30
Added ripple prop to Checkbox, Radio, and Switch components. 

waxter013:
Having the ripple prop makes it optional for these components to have a ripple effect.

### DIFF
--- a/src/components/selection-controls/Checkbox.js
+++ b/src/components/selection-controls/Checkbox.js
@@ -12,7 +12,11 @@ export default {
   },
 
   props: {
-    indeterminate: Boolean
+    indeterminate: Boolean,
+    ripple: {
+      type: [Boolean, Object],
+      default: true
+    }
   },
 
   computed: {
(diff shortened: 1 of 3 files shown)


src/components/selection-controls/Checkbox.js
```diff
@@ -12,7 +12,11 @@ export default {
   },
 
   props: {
-    indeterminate: Boolean
+    indeterminate: Boolean,
+    ripple: {
+      type: [Boolean, Object],
+      default: true
```
jrast: This should be `default: () => ({ center: true })`

In the other files you used `default: { center: true }`, I don't know if this works too.

jrast: `default: () => ({ center: true })` recommended @johnleider  in https://github.com/vuetifyjs/vuetify/pull/1182

johnleider:
You made changes on an older branch where the naming convention has changed. Please update your vuetify version and resubmit this request.

PR #1187, opened 2017-07-31
Fix #1041

### LINKED ISSUE #1041: DataTable Edit Dialog does not position itself correctly near edge of viewport
### Steps to reproduce
1. Create a datatable with an editable column. 
2. Size the window so that the editable column is at the edge of the screen. 
3. Open the edit dialog popup. It will not position itself according to the available screen space.

### Versions
Bug started happening when i updated to Vuetify 0.14.2 from 0.13

### What is expected ? 
The dialog popup positions itself to be fully visible.

The dialog positions its left side on the cell value, leaving it's right side cut off. 

### Reproduction Link
Click on any <td> value under the last column "Iron". The edit dialog will appear cutoff on the right side.
https://jsfiddle.net/s32pfnvb/3/

Kylart:
I did it mom!

Tested with `vuetifyjs/webpack` and my own app.

### DIFF
--- a/src/components/menus/mixins/position.js
+++ b/src/components/menus/mixins/position.js
@@ -82,13 +82,13 @@ export default {
     calcXOverflow (left) {
       const hasWindow = typeof window !== 'undefined'
       const innerWidth = hasWindow ? window.innerWidth : 0
-      const totalWidth = left + maxWidth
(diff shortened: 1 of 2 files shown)


Kylart:
Oups, seems my older commits are in too, should I remake?

Issue #1172, opened 2017-07-28
tfoot in data tables


robodude:
- What will it allow you to do that you can't do today?
I'd like to have a 'summary' column at the bottom of my table. Maybe displaying the sum or mean of a column.

On some tables, I have a 'add row' footer to quickly add rows to the table.

  - How will it make current work-arounds straightforward?
Current work around 'faking' the structure of the table with another table. Being able to use 1 data table to replicate the function of an html table would be ideal.

  - What potential bugs and edge cases does it help to avoid?

guanzo:
+1 to this.

LanFeusT23:
+1

PR #1190, opened 2017-07-31
enhancement/data-table-footer

nekosaur:
See https://github.com/vuetifyjs/vuetify/issues/1172

Added `footer` slot. Works the same way as `items` slot in that you can either supply one or more `<tr>` tags, or one or more `<td>` tags which will be wrapped internally in a `<tr>` tag.

### DIFF
--- a/src/components/tables/mixins/foot.js
+++ b/src/components/tables/mixins/foot.js
@@ -99,7 +99,8 @@ export default {
         ]))
       }
 
-      return this.$createElement('tfoot', [children])
+      if (!children.length) return null     
+      return this.$createElement('tfoot', children)
     }
   }
 }
--- a/src/components/tables/mixins/foot.js
+++ b/src/components/tables/mixins/foot.js
@@ -81,13 +81,25 @@ export default {
       ])]
     },
     genTFoot () {
-      return this.$createElement('tfoot', [
-        this.genTR([
+      const children = []
+
+      if (this.$slots.footer) {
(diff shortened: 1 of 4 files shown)


src/components/tables/mixins/foot.js
```diff
@@ -81,13 +81,25 @@ export default {
...
+      if (!this.hideActions) {
+        children.push(this.genTR([
           this.$createElement('td', {
             attrs: { colspan: '100%' }
           }, this.genActions())
-        ])
-      ])
+        ]))
+      }
+
```
johnleider: If !children.length return null

src/components/tables/mixins/foot.js
```diff
@@ -81,13 +81,25 @@ export default {
...
+      return this.$createElement('tfoot', [children])
```
johnleider: children is already an array, no need to wrap inside another

johnleider:
CI error needs to be fixed, otherwise good 2 go.

nekosaur:
Damn you whitespace

src/components/tables/mixins/foot.js
```diff
@@ -81,13 +81,26 @@ export default {
       ])]
     },
     genTFoot () {
-      return this.$createElement('tfoot', [
-        this.genTR([
+      const children = []
+
+      if (this.$slots.footer) {
```
robodude: I think this needs to be changed to this.$scopedSlots.footer

Issue #1197, opened 2017-07-31
Remove overflow-y from html element


DedDorito:
### Steps to reproduce
Use Vuetify.

### Versions
Latest of everything.

### What is expected ?
`html` element to have no `overflow-y` set or `overflow-y: auto`

### What is actually happening ?
`html` element has `overflow-y: scroll`, which forces a scrollbar to always be present.

### Reproduction Link
https://codepen.io/anon/pen/JyXrNz

Ragura:
For me the biggest problem with this is the shifting of the page content that occurs when opening a dialog, because the dialog removes the scrollbar entirely (or pushes it aside?).
I worked around it by manually setting the overflow-y to auto, but I've read in previous issue reports on this matter that it messes up the menus (don't have any in my app atm).

DedDorito:
Yeah, I experienced the same thing with dialogs as well (navigation drawer too, etc.).  I was assuming there is a reason for this, but perhaps there's a better way to achieve the same outcome than always persisting a scrollbar (main reason why I created this).

johnleider:
The problem with this is there are just as many people who are already used to this. On top of that, a few points:

- The scroll-bar appearing and disappearing can be jarring when you have an SPA that switches between pages
- It's part of the reset we use and I specifically wanted

This has been brought up in the past, I just don't think it's difficult for a developer to just turn it off if they don't want it.

Ragura:
> The scroll-bar appearing and disappearing can be jarring when you have an SPA that switches between pages
I agree, but this is exactly what's happening with the current behaviour for components that do disable the scrollbar, like the dialog components. Content on the page jumps around because of the extra width. Unless there is an easy way to disable this as well?

johnleider:
I don't understand that last part.

DedDorito:
> The scroll-bar appearing and disappearing can be jarring when you have an SPA that switches between pages

Isn't this the default behavior of nearly every page on the web?  Generally speaking, the default for a scrollbar is almost always hidden unless the content requires it to be there.

I definitely agree it's trivial for a dev to override, but it seems to be an odd default in my opinion, which is why I initially opened this issue.  If this is indeed intended and the way you'd like to keep it, feel free to close this issue.

Thanks :)

DedDorito:
> I don't understand that last part.

[...]

PR #1200, opened 2017-08-01
Fix for #1197

cawa-93:


johnleider:
After speaking with the backers and devs, the consensus is to leave it as it is and allow the user to change it.

PR #1223, opened 2017-08-02
fixes #1209

### LINKED ISSUE #1209: Fix footer cut element in view
https://jsfiddle.net/Ewocker/auefnyd4/

Part of the main component is being cut by a fixed footer.
Also, there is no fixed prop in footer doc, but I have not yet find another way to fixed the footer without fixed attr. Am I doing the wrong thing, or should the doc be updated?

Discord Chat:
Ewocker - Today at 8:05 PM
Hi, for th footer issue
https://jsfiddle.net/Ewocker/auefnyd4/
Notice that the very last 'zero' is cut

johnleider - Today at 8:07 PM
that's a bug
submit an issue

ChavaSobreyra:
Adds bottom padding to main when application has fixed footer

### DIFF
--- a/src/stylus/generic/_application.styl
+++ b/src/stylus/generic/_application.styl
@@ -20,6 +20,8 @@
   &--footer-fixed
     > aside.navigation-drawer
       max-height: 'calc(100vh - %s)' % $footer-height
+    > main
+      padding-bottom: $footer-height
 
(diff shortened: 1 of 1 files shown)


johnleider:
Just noticed this, thank you. Ended up using margin instead of padding.

PR #1232, opened 2017-08-03
Fix/vue version validation

jacekkarczmarczyk:
Sorry, accidentally changed filemode from 644 to 755


johnleider:
You committed a revert, is that intentional?

jacekkarczmarczyk:
In the first commit 48cd74b there's a change in the source (intended) and change of file mode 100644 → 100755 (not intended), in the second commit 3222da9 I've reverted only file mode change, so both commits together result only in the changes in the source code, file mode remains 100644

johnleider:
image
This is what I see, I also confirmed by pulling down your branch.

jacekkarczmarczyk:
And this is what you should see when looking at BOTH commits merged together, if you look at only one you'll see the file mode change

johnleider:
Yea, I realized that last night. Sorry for the confusion.

PR #1237, opened 2017-08-04
Issue1235

ChavaSobreyra:
`activeClass` was being received as prop, but was not being used in computed classes function. 

Question: I noticed vBreadCrumbsItem also receives active class, but does nothing with it. However I think with breadcrumbs,  when a breadcrumb is 'active' it is actually 'disabled'. What should happen there? I took a shot at what I think the intended logic is there. Let me know if it makes sense.


johnleider:
Lol I was literally fixing CI when you pushed your commit :)

PR #1238, opened 2017-08-04
Added a callback for when a snackbar times out

qw3n:


src/components/snackbars/VSnackbar.js
```diff
@@ -23,6 +23,7 @@ export default {
       type: Number,
       default: 6000
     },
+    timeoutFn: Function,
```
johnleider:
```
timeoutFn: {
  type: Function,
  default: () => {}
}
```

src/components/snackbars/VSnackbar.js
```diff
@@ -58,6 +59,9 @@ export default {
       if (this.isActive && this.timeout) {
         this.activeTimeout = setTimeout(() => {
           this.isActive = false
+          if(this.timeoutFn) {
```
johnleider: ```this.timeoutFn()```

johnleider:
Update the documentation page with the new prop please https://github.com/vuetifyjs/docs/pulls

jacekkarczmarczyk:
Isn't it better to emit timeout event instead of passing callback as a prop?

nekosaur:
Yes, I think I would prefer event as well. It's more consistent with the rest of the library

johnleider:
Not necessarily. Text-field append and prepend icon cb as an example.

nekosaur:
True. But you could make an argument that those are active tasks (clicking an icon), compared to the snackbar disappearing which is passive. Or that those should be changed also ;)

johnleider:
Valid point, this is something we should decide and document for future
use.

qw3n:
And I was just trying to add what I thought was going to be a simple feature.  Here is my two cents. 

I think both methods are equally valid. Neither is more complicated for the end user `callback="fn"` or `@event="fn"`. I think in general callbacks are more familiar since in most scenarios there isn't an event system in place to use. But, since Vue supplies one it seems more "Vue-like" to use an emitted event. Also, since Vue encourages "data down events up" it may make more since to use that approach to stay more inline with general philosophy of Vue.

Whatever is the call I can rework this once you've made a decision.

johnleider:
Active tasks will be through prop, Passive events will be through event. Can you please update this to emit an event, ty.

johnleider:
As I was re-reviewing this, I thought to myself, why not just watch the changing of the model? When the snackbar closes, it fires a model event. You can easily watch this model and perform your action whenever the model changes.

qw3n:
But the v-model changing could be from a the user clicking a close button in the snackbar. This would mean your watcher would also have to have checks for when the v-model changed, but was not from a timeout.  Do you think that is something the developer should handle or not?

johnleider:
I think it is something the developer should handle.

qw3n:
Would you still be interested in adding the test I wrote. It tests the timeout functionality not just an added callback. I don't have a great handle on what does and doesn't need tested, but I can update it and resubmit it if you think it would be useful.

PR #1245, opened 2017-08-04
Fixed autofocus typo

### LINKED ISSUE #1178: focus event not propagating from inside stepper
### Steps to reproduce
Place a v-text-field with autofocus inside v-stepper-step.  Reload page.  Notice the text field highlights blue, but does not get an active cursor.

### Versions

Tested in Chrome and FF current versions.  Using MacOS.  

"vue": "^2.3.4",
"vuetify": "^0.14.7",

### What is expected ?

I would expect the active cursor to take the position of the text-field with the autofocus in it.

### Reproduction Link

Example with failing behavior: https://codepen.io/shanemgrey/pen/BdjpzO
Example with correct behavior: https://codepen.io/shanemgrey/pen/mMVREX

If using codepen, you may have to Click "Change View" button and change to full screen for the working version to show correct behavior.

I suspect this line of code is causing the problem, but haven't been able to work out the flow of the source to be confident.  
https://github.com/vuetifyjs/vuetify/blob/dev/src/components/steppers/VStepperStep.js#L53

hrayr-artunyan:
This fixes #1178

### DIFF
--- a/src/components/text-fields/VTextField.js
+++ b/src/components/text-fields/VTextField.js
@@ -166,7 +166,7 @@ export default {
           'height': this.inputHeight && `${this.inputHeight}px`
         },
         domProps: {
-          autofucus: this.autofocus,
+          autofocus: this.autofocus,
           disabled: this.disabled,
           required: this.required,
           value: this.lazyValue


johnleider:
Funny typo, ty!

andyburke:
has this made it into a release? I am still seeing the behavior described in #1178

jacekkarczmarczyk:
Yes
https://github.com/vuetifyjs/vuetify/blob/22d66f51244fb5ec9627b552b3ad99c5f0ce4ef1/src/components/VTextField/VTextField.vue

Create a separate issue if you can reproduce it in codepen

PR #1251, opened 2017-08-04
Fixes #1218 - change triggered on blur

### LINKED ISSUE #1217: Text Field component not behaving correctly with validation (VeeValidate)
### Steps to reproduce

1. Create a `v-text-field` component
2. Apply validation through VeeValidate

### Versions

0.14.7

### What is expected ?

The component should turn red only if there are errors.

The component is red from the very beginning and there is also an error in the console:

 > "[Vue warn]: \<transition-group\> children must be keyed: \<div\>"

### Reproduction Link

[CodePen](https://codepen.io/anon/pen/KvgPYy?editors=1111)

### LINKED ISSUE #1218: Text fields change event called twice
### Steps to reproduce
Use change event on v-text-field

### Versions
Vue @latest
Vuetify @v0.14.7
Chrome 60.0.3112.78

### What is expected ?
Change event fired once.

### Reproduction Link
https://codepen.io/anon/pen/EvgxMj

jacekkarczmarczyk:
First commit removes emitting 'change' event on blur. Now the 'change' event is emitted only when input.value changes. However it emits the event object instead of input value (as it should do according to documentation). It also emits event when value after applying modifiers has not changed - i.e. when trailing spaces were added to the value of the text field with "trim" modifier. Input element fires the 'change' event, but the effective value (input.value.trim()) doesn't change

Second commit fixes the last issue, and the 3rd commit fixes the emitted value (input value instead of event object)

Playground.vue for testing:

```
<template>
    <v-container>
        <v-text-field v-model.trim="value" label="Input" @change="change($event)"></v-text-field>
        <p>Type 'abc' then blur, 'change' event should be emitted. Then add some spaces at the end and blur - change event should not be emitted</p>
    </v-container>
</template>


<script>
    export default {
        data: () => ({value: ''}),
        methods: {
            change: e => console.log('Event emitted', e)
        }
    }
</script>
```

### DIFF
--- a/src/components/text-fields/VTextField.js
+++ b/src/components/text-fields/VTextField.js
@@ -142,9 +142,9 @@ export default {
       this.inputValue = e.target.value
       this.multiLine && this.autoGrow && this.calculateInputHeight()
     },
-    onChange (e) {
+    onChange () {
       if (this.originalValue !== this.inputValue) {
-        this.$emit('change', e)
+        this.$emit('change', this.inputValue)
       }
     },
     blur (e) {
--- a/src/components/text-fields/VTextField.js
+++ b/src/components/text-fields/VTextField.js
@@ -106,10 +106,8 @@ export default {
   },
 
   watch: {
-    focused (val) {
+    focused () {
       this.hasFocused = true
-
-      !val && this.$emit('change', this.lazyValue)
     },
     value () {
       this.lazyValue = this.value
--- a/src/components/text-fields/VTextField.js
+++ b/src/components/text-fields/VTextField.js
@@ -10,6 +10,7 @@ export default {
   data () {
     return {
       hasFocused: false,
+      originalValue: null,
       inputHeight: null
     }
   },
@@ -106,8 +107,11 @@ export default {
   },
 
   watch: {
-    focused () {
+    focused (val) {
       this.hasFocused = true
+      if (val) {
+        this.originalValue = this.inputValue
+      }
     },
     value () {
       this.lazyValue = this.value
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
Currently docs say that the "change" event emits String value, after merging this PR "change" event will emit String or Number (Number if v-model has "number" modifier), so I guess the docs should be updated as well

johnleider:
I had already fixed this issue, forgot to push. Thank you for your contribution however.

https://github.com/vuetifyjs/vuetify/commit/ba176f520c75c4cbb6896f6a439224d2d2d94942

jacekkarczmarczyk:
@johnleider - it doesn't seem to be fixed for me. Try to use this Playground.vue I've provided, it still sometimes emits 2 events:
- type "abc   " (with spaces at the end)
- blur - 2 events are emitted (one with event object, one with input value)
- focus again and blur - event is still emitted although value hasn't changed

johnleider:
Well I haven't released it yet

johnleider:
Also blur and change are expected, no?

jacekkarczmarczyk:
To be precise - when I wrote "2 events are emitted" I meant "2 'change' events are emmited", I guest the 'blue' event also was emitted but I didn't check it

And in the last step I've also checked only 'change' event. It was emitted, but I'd expect only blur event if the value has not changed

My change also takes into account the "trim" or "number" modifier as I wrote before, so "a   " and "a" values are equals (after trimming) so "change" event imho shouldn't be emitted

johnleider:
Can you go on chat tonight? I'd like to discuss more there.
Https://chat.vuetifyjs.com

PR #1260, opened 2017-08-05
fixes #1258

### LINKED ISSUE #1258: label covers invalid value of number text field
### Steps to reproduce
https://codepen.io/anon/pen/vJybJV

### Versions
0.14.8

### What is expected ?
Label should not cover value

Label covers value

jacekkarczmarczyk:


johnleider:
Create a test for this and I will merge.

nekosaur:
@jacekkarczmarczyk Is this still valid in current dev? If so I think you'd better create a new PR

PR #1264, opened 2017-08-05
allow click on label of selection control

vanhoavn:


johnleider:
Isn't pointer-events auto by default?

vanhoavn:
Yes, it just added by other's recent commits.

PR #1265, opened 2017-08-05
allow customize label by slot

vanhoavn:
This allow us to customize labels by using slot:

```
                <v-checkbox>
                    <span slot="label">this is <b>important</b> field</span>
                </v-checkbox>
```

### DIFF
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -107,7 +107,7 @@ export default {
 
       if (this.id) data.attrs = { for: this.id }
 
-      return this.$createElement('label', data, this.label)
+      return this.$createElement('label', data, this.$slots.label || this.label)
     },
     genMessages () {
       let messages = []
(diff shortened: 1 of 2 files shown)


src/mixins/input.js
```diff
@@ -107,7 +107,7 @@ export default {
 
       if (this.id) data.attrs = { for: this.id }
 
-      return this.$createElement('label', data, this.label)
+      return this.$createElement('label', data, this.$slots.label || this.label)
```
johnleider: This should pass label through as a prop.

vanhoavn: You mean functional slots?

vanhoavn: `this.label` is the props provided by user, which is used for simple string input, while label slot `this.$slots.label` is used for rendering more complex labels.

johnleider: Valid point, I just realized my mistake.

PR #1266, opened 2017-08-05
allow accent to be used as contextual value

vanhoavn:


johnleider:
We are actually evaluating removing contextualable as a mixin in favor of colorable. I'll leave this up for the other dev team members to review.

nekosaur:
Yes, I think we should be working towards converting components to use colorable instead.

PR #1267, opened 2017-08-05
fixes #1263 - emit event on drawer toggle

### LINKED ISSUE #1263: [Feature Request]: Start/stop events when toggling navigation drawer
I use Google Maps in my main view, and the map needs to be resized every time its container is resized.

When the navigation drawer is toggled, I use a timeout in order to wait for the animation to finish, otherwise the map will be resized too early.

This is the method called by my toolbar-icon click event:

    toggleSidebar: function() {
      this.sidenav = !this.sidenav;
      setTimeout(function() {
        eventBus.$emit('toggleSidebar');
      }, 300);

It would be nice to have events triggered by the navigation drawer itself when the toggle starts and stops.

jacekkarczmarczyk:
Playground.vue for testing - switching on/off left drawer (persistent) should write 'toggle' in console after end of transition, same for changing mini variant, for right (temporary) drawer no event should be emitted:

```
<template>
    <v-app>
        <v-navigation-drawer @toggle="toggle()" persistent :mini-variant="mini" v-model="drawer">
            <v-btn @click="mini = !mini" icon>
                <v-icon v-if="mini">chevron_right</v-icon>
                <v-icon v-else>chevron_left</v-icon>
            </v-btn>
        </v-navigation-drawer>

        <v-toolbar>
            <v-toolbar-side-icon @click.native.stop="drawer = !drawer"></v-toolbar-side-icon>
            <v-spacer></v-spacer>
            <v-toolbar-side-icon @click.native.stop="temporary = !temporary"></v-toolbar-side-icon>
        </v-toolbar>

        <main>
            <v-navigation-drawer @toggle="toggle()" temporary v-model="temporary" right></v-navigation-drawer>
        </main>
    </v-app>
</template>

<script>
    export default {
        data() {
            return {
                drawer: true,
                temporary: false,
                mini: false,
            };
        },
        methods: {
            toggle() {
                console.log('toggle');
            },
        },
    };
</script>
```

### DIFF
--- a/src/components/navigation-drawer/VNavigationDrawer.js
+++ b/src/components/navigation-drawer/VNavigationDrawer.js
@@ -94,11 +94,12 @@ export default {
 
   mounted () {
     this.$vuetify.load(this.init)
-    this.$el.addEventListener('transitionend', this.onTransitionend)
+    this.$el.addEventListener('transitionend', this.onTransitionend, false)
   },
 
-  destroyed () {
-    this.$el.removeEventListener('transitionend', this.onTransitionend)
+  beforeDestroy () {
+    this.$el &&
+      this.$el.removeEventListener('transitionend', this.onTransitionend, false)
   },
 
   methods: {
--- a/src/components/navigation-drawer/VNavigationDrawer.js
+++ b/src/components/navigation-drawer/VNavigationDrawer.js
@@ -94,6 +94,11 @@ export default {
 
   mounted () {
     this.$vuetify.load(this.init)
+    this.$el.addEventListener('transitionend', this.onTransitionend)
+  },
+
+  destroyed () {
+    this.$el.removeEventListener('transitionend', this.onTransitionend)
   },
 
   methods: {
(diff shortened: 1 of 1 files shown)


johnleider:
What are people's thoughts on this? There is nothing that ever stated that a toggle event would emit when the drawer closes or opens, and I'm not too sure why this should be in by default.

jacekkarczmarczyk:
Do you mean that it should emit event only when some prop is set to true?

johnleider:
I don't think it should emit an event at all, You can watch the model and make updates accordingly.

jacekkarczmarczyk:
But afaic the model changes immediately after clicking the triggering button and I want to generate event when then transition ends, so that some heavy layout recalculations can be done just once (see the issue description). Of cource setTimeout(300) somehow solves the problem, but it's not very DRY

johnleider:
It solves the problem because the drawer has a 300ms animation time. I'm just trying to decide whether animation start/stop are valid events to track for the dev because there are a LOT of components that could potentially use it.

johnleider:
Closed parent issue.

everhardt:
I'd like to use this event to redraw a d3 graph. Now that this hasn't been merged in, what is the suggested way to get notified when the toggle is completed? setTimeout(300)?

jacekkarczmarczyk:
You may try to add `@transisionend="..."` to the `<v-content>` and check whether the `propertyName` is `padding-left`/`padding-right`. Theoretically it should work, let us know if it does on discord chat https://chat.vuetifyjs.com

PR #1273, opened 2017-08-06
allow usage of customized select filter function

vanhoavn:


src/components/selects/mixins/autocomplete.js
```diff
@@ -7,12 +13,8 @@ export default {
 
   methods: {
     filterSearch () {
-      return this.items.filter(i => {
-        const text = this.getText(i)
-        if (typeof text !== 'string') return false
-
-        return text.toLowerCase().indexOf(this.searchValue.toLowerCase()) !== -1
-      })
+      const query = this.searchValue.toLowerCase()
```
jacekkarczmarczyk: imho query.toLowerCase() should be evaluated in the default filter function instead of filterSearch(), otherwise it's not possible to pass the case sensitive filter function

vanhoavn: Yes right!

src/components/selects/mixins/autocomplete.js
```diff
@@ -1,4 +1,13 @@
 export default {
+  props: {
+    filter: {
+      type: Function,
+      default: () => (item, query, itemText) => (
+        typeof itemText === 'string' &&
+        itemText.toLowerCase().indexOf(query.toLowerCase()) !== -1
```
johnleider: add indentation and change evaluation to > -1 instead of !== -1

PR #1274, opened 2017-08-06
allow customize menu's z-index

vanhoavn:
Allow zIndex of menu to be configurable to support complex layering applications.


src/components/menus/VMenu.js
```diff
@@ -135,7 +135,11 @@ export default {
       default: false
     },
     maxWidth: [Number, String],
-    minWidth: [Number, String]
+    minWidth: [Number, String],
+    zIndex: {
+      type: [Number, String],
+      default: null
```
johnleider: Default should be 6, see https://github.com/vuetifyjs/vuetify/blob/dev/src/stylus/components/_menus.styl#L25.

Also, update the stylus file and remove the defined z-index.

src/components/selects/VSelect.js
```diff
@@ -88,6 +88,10 @@ export default {
       type: [Boolean, Number, String],
       default: false
     },
+    menuZIndex: {
+      type: [Number, String],
+      default: null
```
johnleider: This should simply match what default is, 6

src/components/menus/VMenu.js
```diff
@@ -150,13 +154,19 @@ export default {
       return maxWidth < minWidth ? maxWidth : minWidth
     },
     styles () {
-      return {
+      const styles = {
```
johnleider: This can just return, zIndex will always be defined, add zIndex explicitly

src/components/menus/VMenu.js
```diff
@@ -135,7 +135,11 @@ export default {
       default: false
     },
     maxWidth: [Number, String],
-    minWidth: [Number, String]
+    minWidth: [Number, String],
+    zIndex: {
+      type: [Number, String],
+      default: 6
```
johnleider: Just in case, can we add required: true please

vanhoavn: What's the reason for both `default` and `required` ?

johnleider: Yea, you're right :P

johnleider:
Can you please update for the conflicts.

vanhoavn:
ok i'll do it soon today

PR #1281, opened 2017-08-08
Accessibility and testing changes

Phlow2001:
Accessibility changes to VTextField and VCheckbox.  Removed console.warn mock in favor of spy's and new expect() matchers, toHaveBeenWarned and toHaveBeenTipped, with failure on unexpected console.error and console.warn messages.  Compensating changes where console.warn was being asserted.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,4 @@
-build/*.js
+﻿build/*.js
 src/stylus/*.styl
+src/util/testing.js
+src/util/to-have-been-warned.js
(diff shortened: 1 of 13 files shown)


nekosaur:
Haven't looked at this in detail yet, but I feel the hasBeenWarned being console.error and hasBeenTipped being console.warn is slightly misleading.

Phlow2001:
Yeah, I thought the same thing.  Vue does [Vue warn] via console.error and [Vue tip] via console.warn.  Since it's mostly intact from Vue source (I added a default export, and reporting for unexpected console.warns), I didn't change the name, but I'm not against making it more Vuetify centric.

PR #1285, opened 2017-08-08
Fixed menu position

cawa-93:
Now the menu can be fixed near the activator and goes beyond the screen

### DIFF
--- a/src/components/menus/VMenu.js
+++ b/src/components/menus/VMenu.js
@@ -70,6 +70,8 @@ export default {
     auto: Boolean,
     offsetX: Boolean,
     offsetY: Boolean,
+    fixedOffsetX: Boolean,
+    fixedOffsetY: Boolean,
     disabled: Boolean,
     maxHeight: {
       default: 'auto'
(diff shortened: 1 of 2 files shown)


johnleider:
Could you please post an example implementation so I can load it in the playground.

cawa-93:
Old:
```vue
<v-menu></v-menu>
```
New
```vue
<v-menu fixed-offset-y ></v-menu>
<v-menu fixed-offset-x ></v-menu>
```

johnleider:
What is the usecase for this?

cawa-93:
https://codepen.io/cawa-93/pen/LjWmEo/?editors=1010
The current behavior is as follows:
If you try to open a menu whose activator is at the bottom of the visible area, the menu will move up and to the left so that it fits into the visible area.
The presence of the `fixed-offset-*` attribute prevents such behavior, forcing the menus always appear in the same place, and go outside the visible zone.

johnleider:
The behavior of menu has since changed and this PR is no longer needed. Thank you for your contributions.

cawa-93:
How exactly did the behavior change?

PR #1289, opened 2017-08-08
Add prop to allow configuration of which keys execute a toggle

ckaczor:
Currently checkboxes (and other toggles) will toggle on either space or enter.  I needed my checkbox to only toggle on space like a standard checkbox.

This add a prop (defaulting to the current of space and enter) to control which keys will toggle.  For my case I'll use an array with just 32 - other uses might be to disable all keys or add other keys.

Thanks!


Phlow2001:
Actually, you're right, checkbox accessibility spec says it's spacebar only.  Perhaps this should be an array overridden at the checkbox level.

ckaczor:
Yeah - I wasn't quite sure where to put the property.  I put it in the input mixin because that's where the keypresses were being handled but it seems like all of it should go in toggleable instead?

johnleider:
Add prop to docs and I'll merge

ckaczor:
Doc pull request created: https://github.com/vuetifyjs/docs/pull/151

Thanks!

johnleider:
Okay, I forgot, need to write a test for this, or I can.

ckaczor:
I've been trying to add some tests that trigger keydown events and look for changes (basically the normal click tests but with keydown instead) but I haven't found the right way to trigger the events.  

I'm happy to add the tests if someone has pointers on how to do the keyboard testing - it doesn't look like there's any keyboard testing right now to use as an example.

ckaczor:
Looks like I figured it out - tests coming shortly.

johnleider:
You might want to wait a bit for your changes. We merged in a la carte branch and that will break a lot of PR's. Also, please do not commit dist files, thank you!

PR #1292, opened 2017-08-09
Dynamic classes for Grid

cawa-93:
This allows you to dynamically change classes
Implementation by analogy with icons: https://github.com/vuetifyjs/vuetify/blob/97a852eb1672b5fef37e3ba89f16543c42767e70/src/components/icons/VIcon.js#L60


johnleider:
While this does fix the issue, it introduces an extra function when it's not really needed.

PR #1296, opened 2017-08-10
Adding component snapshot tests

mfferreira:
The goal is to cover all components, and their individual props.


johnleider:
Fix CI errors and I'll merge

PR #1302, opened 2017-08-10
align-content support in v-layout

jacekkarczmarczyk:
I was hesitating between `<v-layout align-content-start />` and `<v-layout content-start />`, chosen the first one which is longer but more descriptive

### DIFF
--- a/src/stylus/generic/_grid.styl
+++ b/src/stylus/generic/_grid.styl
@@ -104,6 +104,25 @@
   &-baseline
     align-items: baseline
     
+.align-content
+  &-stretch
+    align-content: stretch
+
+  &-start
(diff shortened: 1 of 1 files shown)


src/stylus/generic/_grid.styl
```diff
@@ -104,6 +104,25 @@
   &-baseline
     align-items: baseline
     
+.align-content
+  &-stretch
```
johnleider: I'm not sure if this one is needed as stretch is the default.

johnleider:
Update docs with new prop information and I'll merge.

jacekkarczmarczyk:
https://github.com/vuetifyjs/docs/pull/165

PR #1317, opened 2017-08-11
Improving margin/gutters

ChavaSobreyra:
This PR includes what should be a non-controversial non-breaking improvement to padding. 

That said, in the docs examples, there are some examples where columns don't have column classes. Some issues could arise there. Maybe I don't understand the intention for the system there.

When using the negative margin trick for the grid, it is important that all direct children of the layout rows have padding to compensate. Some of the docs examples don't do they, they just add gutters/margin manually using `ma-1` for example. 

For example, the way bootstrap deals with this is to say "Content should be placed within columns, and only columns may be immediate children of rows." Not that their word is gospel -- just an example. 

I also added some notes on some MD spec breakpoints we are missing. I can tell you the missing sm and md breakpoints are sorely needed. If you look at the new homepage, there's a wiiiide gap between 1024 and 1440. At 1420 window width the page content looks way too small.

### DIFF
--- a/src/stylus/generic/_grid.styl
+++ b/src/stylus/generic/_grid.styl
@@ -2,46 +2,68 @@
   margin-right: auto
   margin-left: auto
   flex-basis: 100%
-  padding: $grid-gutters.lg
+
+  @media $display-breakpoints.xs-only
+    padding: $grid-gutters.xs
+  @media $display-breakpoints.sm-only
+    padding: $grid-gutters.sm
+  @media $display-breakpoints.md-only
+    padding: $grid-gutters.md
+  @media $display-breakpoints.lg-only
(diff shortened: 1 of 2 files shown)


johnleider:
The only thing that I believe this could change (not break) would be the grid-lists implementation. I'd have to look at it when I get home though.

johnleider:
It appears this would have an issue with explicitly defining the grid-list size on the wrapping container.

johnleider:
After reviewing this again, it is out of spec. The breakpoints defined are not intended to happen automatically due to screen size. On desktop it is 16px padding for containers and 24px for mobile. As far as the gutters between the actual flex items, that is controlled by grid lists. Due to this we cannot merge this in at this time. Thank you for your contribution.

PR #1342, opened 2017-08-13
Remove CSS Blur

c4software:
Hi,

The overlay menu, is a bit sluggish on low power device, I notice it's a bit better without (the barely noticeable) blur effect.

Since the 10% of blur is slightly noticeable, removing it, seems not a big loss, but on mobile (like a Nexus 5) make the animation more smooth.

I hope this little deletion will be merge.

### DIFF
--- a/src/stylus/components/_overlay.styl
+++ b/src/stylus/components/_overlay.styl
@@ -14,7 +14,6 @@
     background-color: rgba(33, 33, 33, 1)
     bottom: 0
     content: ''
-    filter: blur(10%)
     height: 100%
     left: 0
     opacity: 0


Phlow2001:
Why not override this in your own use of the library?

johnleider:
I'm not necessarily opposed to this, let me re-check the spec.

c4software:
@Phlow2001 Just because its an enhancement for all user, the current overall overlay effect its not smooth, this little suppression make it a bit better.

But to be honest, I think the overall implementation could be better (finger following, better performance, etc).

For the performance part, a didn’t see why its « so sluggish » (even on my Nexus 6P, its not really really smooth, but its okay). I think its also because of the div add and removal, but i’m not sure. I have used other « toolkit » like the material-ui (or more simple material look a like lib) and the overlay is always smooth.

c4software:
@johnleider I didn’t see any blur specification in overlay on the material guideline if you find it let me know.

johnleider:
Neither can I.

PR #1384, opened 2017-08-17
fixes #1326 - directive for closing active components with esc key

### LINKED ISSUE #1326: [feature request] dialogs should close after pressing "esc" key
Dialogs are not closing when user presses the "esc" key. Closing with "esc" key is so common behaviour that imho it should be the default behavior of Vuetify dialogs, even if it could be easy implemented by the dev

jacekkarczmarczyk:
PR created again as the previous one messed up after merging with the dev branch

This directive allows to close the active component (which is by default defined as a component with isActive set to true) by pressing the "esc" key. At the moment it applies to the dialog (and probably bottom nav, but I haven't checked it yet), select and menu. Components are closed in the reversed order, if the dialog contains menu then first "esc" closes the menu and next one closes the dialog.

If the value of the directive is a function that returns `false` then the component is not deactivated (same mechanizm as in click-outside)

Playground.vue:
```
<template>
    <v-app>
        <v-checkbox v-model="disableEsc" label="disable esc for Dialog 1"></v-checkbox>
        <v-btn @click.stop="dialog = true">Open dialog</v-btn>
        <v-dialog v-model="dialog" :disable-esc="disableEsc" persistent width="800">
            <v-card>
                <v-card-title>Dialog 1</v-card-title>
                <v-card-text>
                    <v-select label="Select" :items="['one', 'two']"></v-select>
                    <v-btn @click.stop="inner = true">Inner dialog</v-btn>
                    <v-dialog v-model = "inner" width="600">
                        <v-card>
                            <v-card-title>Dialog 2</v-card-title>
                            <v-card-text>
                                <p>Inner dialog</p>
                                <v-select label="Select" :items="['one', 'two']"></v-select>
                            </v-card-text>
                        </v-card>
                    </v-dialog>
                    <v-menu offset-y>
                        <v-btn slot="activator">Menu</v-btn>
                        <v-list>
                            <v-list-tile>
                                <v-list-tile-title>Item 1</v-list-tile-title>
                            </v-list-tile>
                            <v-list-tile>
                                <v-list-tile-title>Item 2</v-list-tile-title>
                            </v-list-tile>
                        </v-list>
                    </v-menu>
                </v-card-text>
            </v-card>
        </v-dialog>
    </v-app>
</template>

<script>
    export default {
        data: () => ({
            dialog: false,
            inner: false,
            disableEsc: false
        }),
    }
</script>
```

### DIFF
--- a/src/components/VDataTable/VEditDialog.vue
+++ b/src/components/VDataTable/VEditDialog.vue
@@ -66,7 +66,7 @@
           'class': 'small-dialog__actions',
           directives: [
             { name: 'show', value: this.large },
-            { name: 'esc', value: () => this.cancel() }
+            { name: 'esc' }
           ]
         }, [
           this.genButton(this.cancel, this.cancelText),
--- a/src/components/VDataTable/VEditDialog.vue
+++ b/src/components/VDataTable/VEditDialog.vue
@@ -64,10 +64,10 @@
       genActions () {
         return this.$createElement('div', {
           'class': 'small-dialog__actions',
-          directives: [{
-            name: 'show',
-            value: this.large
-          }]
+          directives: [
+            { name: 'show', value: this.large },
+            { name: 'esc', value: () => this.cancel() }
+          ]
         }, [
           this.genButton(this.cancel, this.cancelText),
           this.genButton(this.save, this.saveText)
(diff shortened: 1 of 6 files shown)


nekosaur:
I haven't looked too closely at this yet, but I think we need a better name for the directive than just `esc`.

jacekkarczmarczyk:
Yes, let me know if you have a better name ;)
I'm also thinking about possibility to change the "active" prop - now it's `isActive` so after pressing "esc" the `isActive` property is set to `false`, but i've seen changes on dev that allow to set other property for this. But I think I'll do this after this PR will be reviewed

PR #1385, opened 2017-08-17
Enhancement/carousel v-model

### LINKED ISSUE #1358: Carousel improvements
There should be an event that fires on item change with payload that tells which item is active. 

What If I want to add caption or another components that corresponds to the active item?

jacekkarczmarczyk:
This PR adds model to carousel component and fixes #1358 

Playground:

```
<template>
    <v-app>
        <main>
            <v-container>
                <div class="headline">With v-model</div>
                <v-carousel v-model="model" style="height: 300px">
                    <v-carousel-item v-for="(item,i) in items" v-bind:src="item.src" :key="i"></v-carousel-item>
                </v-carousel>
                <div class="headline">Without v-model</div>
                <v-carousel style="height: 300px">
                    <v-carousel-item v-for="(item,i) in items" v-bind:src="item.src" :key="i"></v-carousel-item>
                </v-carousel>
            </v-container>
        </main>
    </v-app>
</template>

<script>
    export default {
        data () {
            return {
                model: 0,
                items: [1,2,3,4].map(i => ({
                    src: `http://lorempixel.com/400/200/?${i}`
                }))
            }
        },

        watch: {
            model () {
                console.log(`Current element: ${this.model}`)
            }
        }
    }
</script>
```

### DIFF
--- a/src/components/VCarousel/VCarousel.vue
+++ b/src/components/VCarousel/VCarousel.vue
@@ -24,6 +24,7 @@
     },
 
     props: {
+      value: Number,
       cycle: {
         type: Boolean,
         default: true
@@ -121,7 +122,7 @@
         this.$nextTick(this.startInterval)
       },
       init () {
-        this.current = this.value >>> 0
+        this.current = this.value || 0
       },
       next () {
         this.reverse = false
--- a/src/components/VCarousel/VCarousel.vue
+++ b/src/components/VCarousel/VCarousel.vue
@@ -16,7 +16,7 @@
 
     data () {
       return {
-        current: null,
+        inputValue: null,
         items: [],
         slideInterval: {},
         reverse: false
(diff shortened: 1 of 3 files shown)


src/components/VCarousel/VCarousel.vue
```diff
@@ -125,7 +121,7 @@
         this.$nextTick(this.startInterval)
       },
       init () {
-        this.current = 0
+        this.current = this.value >>> 0
```
nekosaur: `this.value || 0` is more readable, and what we usually use

src/components/VCarousel/VCarousel.vue
```diff
@@ -71,6 +62,11 @@
 
         !this.isBooted && this.cycle && this.restartInterval()
         this.isBooted = true
+
+        this.$emit('input', this.current)
+      },
+      value () {
```
nekosaur: You haven't defined `value` as a prop. I would probably try to remove `current` entirely in favor of `value` and emitting `input`.

johnleider: I'm in favor of this.

jacekkarczmarczyk:
@nekosaur tests ready

src/components/VCarousel/VCarousel.vue
nekosaur: After talking to @jacekkarczmarczyk, I think I'm fine with keeping current, as it allows one to use the carousel without declaring a v-model. What do you think @johnleider?

johnleider: I'm okay with it, can we name it inputValue though?

PR #1421, opened 2017-08-19
label covers invalid value of number text field

### LINKED ISSUE #1258: label covers invalid value of number text field

jacekkarczmarczyk:
Updated PR https://github.com/vuetifyjs/vuetify/pull/1260

### DIFF
--- a/src/components/VTextField/VTextField.vue
+++ b/src/components/VTextField/VTextField.vue
@@ -10,7 +10,8 @@
 
     data () {
       return {
-        inputHeight: null
+        inputHeight: null,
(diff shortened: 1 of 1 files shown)


nekosaur:
Please create a test for this

jacekkarczmarczyk:
As discussed before on discord this could be tested only in browser environment (i.e. nightwatch), it's can't be tested with avoriaz

PR #1425, opened 2017-08-20
Handle up/down arrow key events on VSelect

kevmo314:
This change allows the up and down arrow keys to select the next and previous items respectively when the v-select input box is focused but the menu is not open. If the value is not set, then the down arrow key selects the first item and the up arrow does nothing.

### DIFF
--- a/src/components/VSelect/VSelect.vue
+++ b/src/components/VSelect/VSelect.vue
@@ -329,7 +329,21 @@
           click: () => {
             if (!this.isActive) this.isActive = true
           },
-          keydown: this.onKeyDown // Located in mixins/autocomplete.js
+          keydown: (e) => {
+            if (!this.isActive &&
+                !this.multiple &&
+                !this.autocomplete &&
+                [38, 40].includes(e.keyCode)) {
+              const i = this.items.findIndex((i) => {
+                return this.getValue(i) === this.getValue(this.inputValue)
+              }) + e.keyCode - 39
+              if (i >= 0 && i < this.items.length) {
+                this.$emit('change', this.inputValue = this.items[i])
+              }
+            } else {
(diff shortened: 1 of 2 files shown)


src/components/VSelect/VSelect.vue
```diff
@@ -329,7 +329,21 @@
           click: () => {
             if (!this.isActive) this.isActive = true
           },
-          keydown: this.onKeyDown // Located in mixins/autocomplete.js
+          keydown: (e) => {
+            if (!this.isActive &&
+                !this.multiple &&
```
johnleider: This is the only part that confuses me. Is there a reason this shouldn't be supported on multiple?

kevmo314: What would be the expected behavior for multiple?

johnleider: The same I would think

kevmo314: I'm not clear on what that means though unless I misunderstood? If a multiple select has A, B, C selected, which one does it increment when the up/down key is pressed?

johnleider: Potentially from the one selected furthest down in the DOM. I'm thinking of some other things I can do to improve this, and may take it a different route.

kevmo314: I don't follow, why does the DOM position matter? The menu is not showing, moving the furthest one down seems like a somewhat arbitrary choice. If you mean the last item, then it also seems somewhat arbitrary because the items are sorted alphabetically.

johnleider:
I have handled this in a different manner due to the massive ```v-select``` updates this release. I have made sure to give the proper credit.

PR #1427, opened 2017-08-20
Disallow focusing of VSelect when disabled attribute is present.

kevmo314:
This change removes the ability to tab to a `v-select` if it's disabled and blurs the field when it becomes disabled.

### DIFF
--- a/src/components/VSelect/VSelect.vue
+++ b/src/components/VSelect/VSelect.vue
@@ -69,6 +69,7 @@
         type: Number,
         default: 200
       },
+      disabled: Boolean,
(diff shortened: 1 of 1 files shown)


johnleider:
Thank you for your contributions but this is something I've already addressed on dev.

kevmo314:
This doesn't seem to be fixed in the 0.15 beta, at least not on dev.vuetifyjs.com: http://i.imgur.com/K9bZMQe.png

johnleider:
@kevmo314 I'm pushing a fix now for this regression

PR #1456, opened 2017-08-24
defaultPagination.sortBy isn't initialized

stephane:
customSort() receives an index set to undefined (and checks against null) then it doesn't
return items but try to sort them.

https://github.com/stephane/vuetify/blob/0912ccdd41e1efc106560a31286ecb12660796a4/src/components/VDataTable/VDataTable.vue#L100

BTW Do you think it would an improvement to rename `index` argument to `sortBy`?

### DIFF
--- a/src/components/VDataTable/VDataTable.vue
+++ b/src/components/VDataTable/VDataTable.vue
@@ -33,9 +33,10 @@
         all: false,
         searchLength: 0,
         defaultPagination: {
+          descending: false,
           page: 1,
           rowsPerPage: 5,
-          descending: false,
+          sortBy: null,
           totalItems: 0
         }
       }


nekosaur:
Can you show some example code where this would lead to an error?

nekosaur:
ping @stephane

stephane:
I will try to provide an example today!

stephane:
In this codepen https://codepen.io/sraimbault/pen/brQeNa with an empty list of items, the customSort() function is called here:

https://github.com/stephane/vuetify/blob/0912ccdd41e1efc106560a31286ecb12660796a4/src/components/VDataTable/VDataTable.vue#L219

with `items = []` and `index = undefined` because `sortBy` isn't initialised so the customSort function try to sort the list without a valid `index` (getObjectValueByPath returns null).

vdatatable-index

stephane:
@johnleider could you merge it, please?

PR #1457, opened 2017-08-24
enhancement/dialog-scroll

nekosaur:
on desktop (lg and bigger), we disable scroll by intercepting mousewheel and keyboard events and doing `.preventDefault()`

on mobile we simply set `overflow: hidden`


canhongpoxiao:
you only set `document.documentElement.style.overflow = 'hidden'` its not work for ios, you should set `document.body`

PR #1473, opened 2017-08-26
remove hard coded footer height

dzift:


johnleider:
Right now this is necessary for proper container height.

Issue #1304, opened 2017-08-10
v-select noDataText tile is disabled by default after #982. It should be controlled via prop.


chancet1982:
After fixing #982 it seems that now the "no data" tile is disabled by default. Before this fix this was not the case. This is caused by **generators.js, line 100** where:
`children.push(this.genTile(this.noDataText, true))
`
This prevents any action being taken if a user is searching for an item which is currently not found (for example if item is not found then noDataText can be "add new" and by listening to the input/change you could trigger an method to add this missing item.

I think a better approach here instead of assuming that the "no data" tile should be non-clickable is to provide a prop (noDataTileDisabled -type: Boolean, default: true) and use that instead of the hardcoded 'true'

Basically:
`children.push(this.genTile(this.noDataText, this.noDataTileDisabled ))
`
This will open the door to "add new" if searchValue isnt in the items

### Steps to reproduce
1. Add a v-select to your project with autocomplete
2. Start typing (searching for an item that is not in the list)
3. Click on the "no data found" tile.

### Versions
All versions. change happened upgrading from 0.14.4 to 0.14.8
 
### What is expected ?
No data found item shouldnt be disabled by default, instead it should be controlled via prop.

### What is actually happening ?
item is now disabled, i believe it should be controlled via prop

### Reproduction Link
https://jsfiddle.net/chancet1982/knp56m9w/

chancet1982:
After fixing #982 it seems that now the "no data" tile is disabled by default. Before this fix this was not the case. This is caused by generators.js, line 100 where:
`children.push(this.genTile(this.noDataText, true))`

This prevents any action being taken if a user is searching for an item which is currently not found (for example if item is not found then noDataText can be "add new" and by listening to the input/change you could trigger an method to add this missing item.

I think a better approach here instead of assuming that the "no data" tile should be non-clickable is to provide a prop (noDataTileDisabled -type: Boolean, default: true) and use that instead of the hardcoded 'true'

Basically:
`children.push(this.genTile(this.noDataText, this.noDataTileDisabled ))`

This will open the door to "add new" if searchValue isnt in the items

PR #1505, opened 2017-08-28
fix for #1304

chancet1982:
This fix is intended to solve 2 issues:
1. The "noDataText" tile isnt clickable, and that is hardcoded. Instead it should look for a prop "no-data-clickable" which is by default false for its "disabled" state.
2. Instead of hardcoded value for "noDataText" (which is language specific) when generating the noDataText tile the value can be set to any string, using the prop no-data-value

### DIFF
--- a/src/components/VSelect/mixins/generators.js
+++ b/src/components/VSelect/mixins/generators.js
@@ -115,7 +115,12 @@ export default {
       })
 
       if (!children.length) {
-        children.push(this.genTile(this.noDataText, true))
+        const noDataItem = {
+          text: this.noDataText,
+          value: this.noDataValue
+        }
+
+        children.push(this.genTile(noDataItem, !this.noDataClickable))
       }
 
       return this.$createElement('v-card', [
--- a/src/mixins/filterable.js
+++ b/src/mixins/filterable.js
@@ -3,6 +3,14 @@ export default {
     noDataText: {
       type: String,
       default: 'No data available'
+    },
+    noDataValue: {
+      type: String,
+      default: 'NO_DATA'
+    },
+    noDataClickable: {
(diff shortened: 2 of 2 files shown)


johnleider:
I'm okay with these changes but there are a few notes:

- Can you please add tests that verify this functionality
- What is the intended behavior for ```v-data-table``` which also implements the **filterable** mixin

chancet1982:
Regarding your question. I think current behavior of v-data-table is fine. I think the enhancement is only relevant for the v-select. Regarding the tests - I cant seem to run the tests locally using "npm test" is there anything im missing?

johnleider:
You need to have jest installed globally I believe, but without seeing the error, I'm not sure.

chancet1982:
Tests dont seem to currently work on combination of windows 10, Git Bash, and Jest. **Adding -i to the configuration seems to solve it.**

Link to thread:
https://github.com/facebook/jest/issues/3079

johnleider:
I think you can remove the need for no-data-value as it already accepts no-data-text.

chancet1982:
Hi, sorry but I am quite a crappy test writer as it seems... I dont think ill be able to right now. ill explain the logic I had with adding a value to the "no data" tile and you can decide if you wish to take these changes and write the tests or dump the idea altogether...
I basically made 2 changes:
1. no datatext clickable as a prop instead of hardcoded.
2. the nodatatext as it is today is language specific, comparing a value string which is not language dependent is simpler -if the prop is not used then no harm done :)

Once more apologize but will not be able to write the tests.

johnleider:
No worries, I'll take it from here.

johnleider:
I have resolved this differently, ty for the contribution.

PR #1507, opened 2017-08-28
Bug 0.15 - Fix labels on selection-controls to be clickable

ckaczor:
The selector on selection controls label wasn't right so pointer-events were disabled and the z-index was too low.  Fixed selector so it applies the right styles.


johnleider:
Can you please only provide the commit that changes the one file instead of the above 10 please.

ckaczor:
All set.

PR #1526, opened 2017-08-29
fixes cursor on readonly input in menu activator

jacekkarczmarczyk:
If menu activator contains readonly input then cursor is set to default instead of pointer, it's visible for example in "#3 Date pickers - In dialog and menu" - https://codepen.io/anon/pen/WEYZvx

Playground.vue

```
<template>
  <v-app>
    <v-menu>
      <v-text-field slot="activator" label="Input" readonly></v-text-field>
      <v-card><v-card-text>Hello world!</v-card-text></v-card>
    </v-menu>
  </v-app>
</template>

<script>
export default {
  data: () => ({date:null})
}
</script>

```

### DIFF
--- a/src/stylus/components/_menus.styl
+++ b/src/stylus/components/_menus.styl
@@ -18,6 +18,9 @@
     cursor: pointer
     position: relative
 
+    input[readonly]
+      cursor: pointer
+
     .toolbar__side-icon
       margin: 0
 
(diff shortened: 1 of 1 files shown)


johnleider:
I tried to see what this did, confused, please help.

jacekkarczmarczyk:
https://codepen.io/anon/pen/EvrpJB

Hover cursor on the text area (on "Picker in menu" text)
With "fix off" - cursor is set to default
With "fix on" - cursor is set to pointer, which I think is correct behaviour for menu activator

PR #1572, opened 2017-09-02
Fix #1571 - Delete VForm $listeners.input before render

### LINKED ISSUE #1571: v-form passes back InputEvent instead of a boolean
### Steps to reproduce
1. Create a `v-input` inside a `v-form`, binding the form's model
2. Input something into the text field

### Versions
 - Vuetify 0.15.1
 - Chrome 60

### What is expected ?
The `v-model` of the form should be a boolean corresponding to its validity

The InputEvent from the `v-input` is being returned instead. 

### Reproduction Link
https://codepen.io/anon/pen/dzaBvB?editors=1011

KaelWD:
This fix will prevent any `input` event listeners from being passed to VForm's internal form element. I'll write a test case for this too if I can figure out how jest works.

### DIFF
--- a/src/components/VForm/VForm.js
+++ b/src/components/VForm/VForm.js
@@ -73,6 +73,7 @@ export default {
   },
 
   render (h) {
+    this.$listeners && delete this.$listeners.input
     return h('form', {
       attrs: this.$attrs,
       on: this.$listeners


johnleider:
Thank you for your PR, however I've opted to take a different path. The issue is resolved though.

PR #1581, opened 2017-09-03
Make v-form validation great again

KaelWD:
Form validation (`<v-form v-model="valid">`) is kinda useless right now: 
 - It only works on immediate children, so nested components such as date pickers don't count
 - `valid` only changes when every single form field has been updated, whether they require validation or not. This makes optional fields impossible without having to call `form.validate()`
 - If a field is dynamically added to the form (with `v-if` or similar), it doesn't *ever* count towards validation, even if `form.validate()` is called
 - If a field is dynamically removed from the form and is invalid, the form will remain in an invalid state, even if `form.reset()` or `form.validate()` is called

This PR fixes all these issues, making validation of [forms like this](https://codepen.io/anon/pen/VzRKEW?editors=1010) or [this](https://codepen.io/nelxaz/pen/xLemZB?editors=1010#0) possible. 
 - Deeply nested children will be included in the validation
 - `valid` will change after every update. `form.validate()` can still be called to check overall form validity and force validation errors to display on untouched invalid fields
 - A new boolean prop `lazy-validation` is added that only takes into account touched fields when determining overall validity (this is the same as it is currently) - so `form.reset()` will result in `valid` being `true`
 - Dynamically added fields will work the same way as static ones
 - Dynamically removed fields will be removed from consideration, so if an invalid field is removed from an otherwise valid form, `valid` should now be `true`

### DIFF
--- a/src/components/VForm/VForm.js
+++ b/src/components/VForm/VForm.js
@@ -80,15 +80,11 @@ export default {
       }
     },
     validate () {
-      const errors = this.getInputs().reduce((errors, child) => {
-        const error = !child.validate(true)
-        return errors || error
-      }, false)
-
-      return !errors
+      return this.getInputs().every(i => i.validate(true))
     },
     reset () {
       this.getInputs().forEach((input) => input.reset())
+      this.errorBag = {}
     }
   },
 
--- a/src/components/VForm/VForm.js
+++ b/src/components/VForm/VForm.js
@@ -11,7 +11,8 @@ export default {
   },
 
   props: {
-    value: Boolean
+    value: Boolean,
+    lazyValidation: Boolean
   },
 
   watch: {
@@ -44,18 +45,8 @@ export default {
 
       return search(this.$children)
     },
-    watchInputs () {
-      const inputs = this.getInputs()
-
-      if (inputs.length < this.inputs.length) {
-        // Something was removed, we don't want it in the errorBag any more
-        const removed = this.inputs.filter(i => !inputs.includes(i))
-
-        for (const input of removed) {
-          this.$delete(this.errorBag, input._uid)
-          this.$delete(this.inputs, this.inputs.indexOf(input))
-        }
-      }
+    watchInputs (inputs) {
+      inputs === undefined && (inputs = this.getInputs())
 
       for (const child of inputs) {
         if (this.inputs.includes(child)) {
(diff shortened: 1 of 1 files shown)


johnleider:
Create a test corresponding to the changes and I'll finish my review and merge, thank you.

src/components/VForm/VForm.js
```diff
@@ -35,41 +30,80 @@ export default {
...
       })
     },
     validate () {
-      const errors = this.getInputs().reduce((errors, child) => {
-        const error = !child.validate(true)
-        return errors || error
-      }, false)
-
-      return !errors
+      return !this.inputs.filter(input => input.validate(true)).length
```
johnleider: This could/should be ```!this.inputs.every(input => input.validate(true))```

src/components/VForm/VForm.js
```diff
@@ -35,41 +30,80 @@ export default {
...
+    watchInputs (inputs) {
+      inputs === undefined && (inputs = this.getInputs())
+
+      for (const child of inputs) {
+        if (this.inputs.includes(child)) {
+          continue // We already know about this input
+        }
+
+        this.inputs.push(child)
+
```
johnleider: remove space

src/components/VForm/VForm.js
```diff
@@ -35,41 +30,80 @@ export default {
...
+      this.inputs.forEach((input) => input.reset())
+      Object.keys(this.errorBag).forEach(key => this.$delete(this.errorBag, key))
     }
   },
 
   mounted () {
-    this.$vuetify.load(() => {
-      this.getInputs().forEach((child) => {
-        this.inputs += 1
+    this.$vuetify.load(() => this.watchInputs())
```
johnleider: This should be ```this.$vuetify.load(this.watchInputs)```

azaars:
Brilliant!

Now I begin to think that external validators are second class. I believe this enhancement should work on `v-stepper`, that is by wrapping it with `v-form`, or perhaps even better if we have `v-stepper` built in with v-form, which can optionally be enabled with some props like `use-form`.

Looking forward for a quick merge!

src/components/VForm/VForm.js
KaelWD: I had that in 9a5063bf, but it stops at the first invalid field. I can change it to `...length === 0` to make it more clear what is happening.

src/components/VForm/VForm.js
KaelWD: That will pass the global `Window` object to `watchInputs`, instead of `undefined`. Is `.load()` even needed? Nothing seems to change if I remove it.

src/components/VForm/VForm.js
nekosaur: Could also possibly use `this.inputs.some(input => !input.validate(true))`

KaelWD: `.some` has the same problem as `.every`. 

> The every method executes the provided callback function once for each element present in the array until it finds one where callback returns a falsy value. If such an element is found, the every method immediately returns false.

> some() executes the callback function once for each element present in the array until it finds one where callback returns a truthy value (a value that becomes true when converted to a Boolean). If such an element is found, some() immediately returns true. 

`input.validate()` must be called on every element so that validation errors are forced to display.

src/components/VForm/VForm.js
johnleider: It's for SSR protection lol

johnleider: Also valid point, no need to change this.

src/components/VForm/VForm.js
```diff
@@ -78,11 +75,13 @@ export default {
       })
     },
     validate () {
-      return !this.inputs.filter(input => input.validate(true)).length
+      return !this.inputs.filter(input => input.validate(true))
```
johnleider: What is the reasoning for this change?

KaelWD: Oops, thanks for pointing that out. That wasn't meant to be pushed.

PR #1583, opened 2017-09-04
Add keyboard capability to VList (and consequently VNavigationDrawer), VIcon aria-hidden

Phlow2001:
Add keyboard capability to VList (and consequently VNavigationDrawer)
Make all VIcon's aria-hidden=true if not otherwise set.
Updated snapshots.

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -9,10 +9,8 @@ button($material)
     color: $material.text.primary
 
   &.btn--disabled
-    box-shadow: none !important
     color: $material.buttons.disabled
-    pointer-events: none
-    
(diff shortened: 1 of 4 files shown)


johnleider:
So I'm testing this now and I'm not honestly sure what I'm supposed to see. Which I click on list-tiles they now highlight, but that's it. Could you provide some example markup?

Phlow2001:
Here's an example of keyboard nav that isn't functional, and the PR fixes.
https://codepen.io/jayblanchard/pen/LjKGNq?editors=1010

johnleider:
Everything looks good except the list items are getting the outline even when just being clicked.

PR #1598, opened 2017-09-05
Allow any data type in button values

### LINKED ISSUE #1596: v-btn-toggle can't toggle a boolean
Since 0.15 and the deprecated prop items in v-btn-toggle, when using the default slot, the component can't toggle a boolean because the value attribute of v-btn return a string for "true" and a number for "false"

### Versions
0.15+

### What is expected ?
Toggling a boolean

### Reproduction Link
https://codepen.io/anon/pen/GvLyRZ

Could be related: https://github.com/vuetifyjs/vuetify/issues/1369

KaelWD:
This PR adds a `value` prop to `v-btn` that lets you store any type of data on it. This allows `v-btn-toggle` to toggle anything instead of just strings, it also fixes the bug where the button's index is returned if its value is `false`

Sample codepen: https://codepen.io/anon/pen/RZmPgx?editors=1010

Fixes #1596

### DIFF
--- a/src/components/VBtn/VBtn.vue
+++ b/src/components/VBtn/VBtn.vue
@@ -41,7 +41,8 @@
       type: {
         type: String,
         default: 'button'
-      }
+      },
+      value: null
     },
 
     computed: {
@@ -124,6 +125,10 @@
 
       tag === 'button' && (data.attrs.type = this.type)
       this.loading && children.push(this.genLoader())
+
+      if (typeof this.value !== 'undefined') {
+        data.attrs.value = (typeof this.value === 'object') ? JSON.stringify(this.value) : this.value
+      }
 
       return h(tag, data, children)
     }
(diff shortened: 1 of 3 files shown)


johnleider:
Please do not submit PR's with **--no-verify**

KaelWD:
Sorry about that, I got stuck and needed someone else to see the errors and I thought this was the easiest way to do so.

johnleider:
This won't work with how route-links work.

KaelWD:
Can you give me an example? `route-link` doesn't have a `value` prop that I can see, and `toggleable` uses `inputValue` as its model.  
I added a `<v-btn to="/some/path">` to my Playground and it works as expected.

src/components/VBtn/VBtn.vue
```diff
@@ -125,6 +126,10 @@
       tag === 'button' && (data.attrs.type = this.type)
       this.loading && children.push(this.genLoader())
 
+      if (typeof this.value !== 'undefined') {
+        data.attrs.value = (typeof this.value === 'object') ? JSON.stringify(this.value) : this.value
```
KaelWD: @johnleider we decided in discord that the serialisation is a bad idea and can just be
```js
data.attrs.value = this.value
```

johnleider: updated, ty

PR #1643, opened 2017-09-08
Force click-outside to respond to clicks on body

jeffberry:
This fixes issues with the `click-outside` directive that are caused when `[data-app]` is not the outermost container of the page for those that are using Vuetify in an isolated portion of their site.


johnleider:
You cannot bind click events to the body on iOS devices, which is why I have it binding to the main v-app. Thank you for your contribution, unless something new has come up that I'm not aware of, we cannot use this.

PR #1659, opened 2017-09-10
Time picker filter does not style "00" minutes as selectable

jacekkarczmarczyk:


nekosaur:
Could you possibly create a test for this? Make it a unit test of the function

Issue #1667, opened 2017-09-11
Radio values are double quoted in the DOM


KaelWD:
### Steps to reproduce
 - Create a `v-radio` with a string `value` (`<v-radio value="some-string">`)
 - Inspect the HTML element generated

### Versions
0.15.3

### What is expected ?
A hidden `<input type="radio" value="some-string">`

### What is actually happening ?
The generated `<input>` element has an extra set of quotes in its value

### Reproduction Link
https://codepen.io/anon/pen/oGvNEe

Issue #1661, opened 2017-09-10
0.15.3 breaks horizontal v-radios in v-radio-group


smares:
### Steps to reproduce
Create a `v-radio-group` with several `v-radios` and set it to display the radios horizontally (`row`). This worked in 0.15.2.

### Versions
Vue 2.4.2, Vuetify 0.15.3, Edge 15

### What is expected ?
Radio buttons should be displayed in a row

### What is actually happening ?
Radio buttons are displayed as column

### Reproduction Link
https://codepen.io/smares/pen/EvqrMx

smares:
The issue seems to be caused by the `display: block` rule of `.radio-group--column .input-group__input`

PR #1669, opened 2017-09-11
Fix several radio issues

### LINKED ISSUE #1654: v-radios all have an input with name test
### Steps to reproduce
Insert a v-radio and inspect the element using the browser dev tools. You will see an input with the name test as part of the node.

### Versions
Vue 2.4.2, Vuetify 0.15.2, Edge 15

### What is expected ?
Either an input with individual name or no input (?)

All v-radios have an input with name test

### Reproduction Link
https://codepen.io/smares/pen/RZXZjz

KaelWD:
Fixes #1654, #1661, and #1667

### DIFF
--- a/src/components/VBtn/VBtn.vue
+++ b/src/components/VBtn/VBtn.vue
@@ -126,9 +126,9 @@
       tag === 'button' && (data.attrs.type = this.type)
       this.loading && children.push(this.genLoader())
 
-      if (typeof this.value !== 'undefined') {
-        data.attrs.value = this.value
-      }
+      data.attrs.value = ['string', 'number'].includes(typeof this.value)
+        ? this.value
+        : JSON.stringify(this.value)
 
       return h(tag, data, children)
     }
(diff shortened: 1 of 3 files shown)


src/components/VBtn/VBtn.vue
```diff
@@ -126,9 +126,9 @@
       tag === 'button' && (data.attrs.type = this.type)
       this.loading && children.push(this.genLoader())
 
-      if (typeof this.value !== 'undefined') {
-        data.attrs.value = this.value
-      }
+      data.attrs.value = ['string', 'number'].includes(typeof this.value)
+        ? this.value
+        : JSON.stringify(this.value)
```
johnleider: Didn't we talk about this and decided to not stringify the value?

KaelWD: Yeah, but `value = false` would result in no value at all, and I figured it'd make sense to have similar behaviour for both radios and button toggles. I did try using `toString` in one of the other commits in this PR, but it was messy and still didn't work properly.

PR #1674, opened 2017-09-11
Remove ripple and change cursor on disabled selection controls

### LINKED ISSUE #1665: disabled radio button has ripple and hover
### Steps to reproduce
Create disabled radio button

### Versions
0.15.3

### What is expected ?
Disabled radio button show use default mouse cursor and should not have ripple effect

Disabled radio button has pointer mouse cursor on hover and has ripple effect

### Reproduction Link
https://codepen.io/anon/pen/ZXzzKM

EDIT: same for chekbox/switch

KaelWD:
Unsure whether the cursor should be `default`, or `not-allowed`

Fixes #1665

### DIFF
--- a/src/components/VCheckbox/VCheckbox.vue
+++ b/src/components/VCheckbox/VCheckbox.vue
@@ -69,7 +69,7 @@
         }, this.icon)
       ])
 
-      const ripple = h('div', {
+      const ripple = !this.disabled && h('div', {
         'class': 'input-group--selection-controls__ripple',
         on: Object.assign({}, {
           click: this.toggle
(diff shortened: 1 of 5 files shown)


johnleider:
Looks good, but 3 duplications of the same code **should** be triggering codebeat. Is there a way we could DRY this up?

KaelWD:
It's not triggering codebeat because I basically just moved some existing stuff behind an `if`. I don't have time now, but I'll have a look again tomorrow and see what I can do.

johnleider:
Something like that, will have to look into it further.

johnleider:
I've already corrected this, thank you.

PR #1702, opened 2017-09-12
Fix issue where text field without a model loses its value on blur

KaelWD:
Partially reverts fc774a26

@johnleider Might want to have a look at what was actually wrong here, I'm too tired to see it right now


KaelWD:
Looks like you've fixed this manually.

PR #1703, opened 2017-09-12
VTextField.spec.js cleanup

jacekkarczmarczyk:


src/components/VTextField/VTextField.spec.js
```diff
@@ -166,7 +166,6 @@ test('VTextField.js', ({ mount }) => {
   })
 
   it('should not clear input if not clearble and has appended icon (without callback)', async () => {
```
KaelWD: Should be "clearable", same on line 151

src/components/VTextField/VTextField.spec.js
```diff
@@ -178,7 +177,6 @@ test('VTextField.js', ({ mount }) => {
     icon.trigger('click')
     await wrapper.vm.$nextTick()
     expect(wrapper.vm.inputValue).toBe('foo')
-    expect(called).toBe(0)
```
KaelWD: You can use `jest.fn()` to mock this (in the previous test)

```js
const callback = jest.fn()
...
appendIconCb: callback
...
expect(callback.mock.calls.length).toBe(1)

jacekkarczmarczyk: done

PR #1713, opened 2017-09-13
Cannot add property blur, object is not extensible

jacekkarczmarczyk:
I can't reproduce it in Playground.vue or jsfiddle, error shows only in codepen, but this change should fix it (didn't check in codepen, just run tests and they all passed)


jacekkarczmarczyk:
Preparing new patch

PR #1720, opened 2017-09-13
[vue 2.4.3] fixes #1719

### LINKED ISSUE #1719: Stepper component causing error
### Steps to reproduce
Create a stepper, see e.g. the default vuetify stepper codepen or forked copy with link pasted below

### Versions
Seems to happen to all versions of vuetify that I've tried this for since 0.15.2 -> 0.15.6. Chrome version 61.0

### What is expected ?
For the stepper content to work

Stepper raises an error:
```
Error in render function: "TypeError: Cannot add property beforeEnter, object is not extensible"

found in

---> <VStepperContent>
       <VStepper>
         <VApp>
           <Root>
```
Followed by:
```
TypeError: Cannot add property beforeEnter, object is not extensible
    at render (vuetify.min.js:6)
    at createFunctionalComponent (vue.js:3519)
    at createComponent (vue.js:3678)
    at _createElement (vue.js:3857)
    at createElement (vue.js:3796)
    at vm.$createElement (vue.js:4139)
    at Proxy.render (vuetify.min.js:6)
    at VueComponent.Vue._render (vue.js:4190)
    at VueComponent.updateComponent (vue.js:2578)
    at Watcher.get (vue.js:2919)
```
In Safari a different error occurs:
```
[Vue warn]: Error in render function: "TypeError: Attempted to assign to readonly property."

found in

---> <VStepperContent>
       <VStepper>
         <VApp>
           <Root>
```
### Reproduction Link
https://codepen.io/shongololo/pen/RLwzPR

jacekkarczmarczyk:
### disclaimer

Probably it would be better to find out why `data.on` is not extensible and change it to extensible, but i'm not sure how to approach it

**Edit:**
Changing `on: this.$listeners` to `on: { ...this.$listeners }` could do the thing, but actually maybe it's better to create extensible object only when needed, so this PR should be enough

image

### DIFF
--- a/src/util/helpers.js
+++ b/src/util/helpers.js
@@ -28,7 +28,7 @@ export function createSimpleTransition (name, origin = 'top center 0', mode) {
     render (h, context) {
       context.data = context.data || {}
       context.data.props = { name }
-      context.data.on = context.data.on || {}
+      context.data.on = context.data.on ? { ...context.data.on } : {}
 
       if (mode) context.data.props.mode = mode
 
--- a/src/util/helpers.js
+++ b/src/util/helpers.js
@@ -28,7 +28,10 @@ export function createSimpleTransition (name, origin = 'top center 0', mode) {
     render (h, context) {
       context.data = context.data || {}
       context.data.props = { name }
-      context.data.on = context.data.on ? { ...context.data.on } : {}
+      context.data.on = context.data.on || {}
+      if (!Object.isExtensible(context.data.on)) {
+        context.data.on = { ...context.data.on }
+      }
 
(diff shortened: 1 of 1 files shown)


KaelWD:
> it would be better to find out why `data.on` is not extensible

Vue 2.4.3 includes [this change](https://github.com/vuejs/vue/pull/6441/files#diff-37f4deddf9f54795e5be4c5540ceb906R236). `emptyObject` is a [frozen empty object](https://github.com/vuejs/vue/blob/d6999c19c4c18d5eedfd3a7741e17b0e34e8a8df/src/core/util/lang.js#L3), I guess to help performance.

PR #1747, opened 2017-09-15
Fixes rotated clearable icon

### LINKED ISSUE #1741: clearble icon is rotated when select is focused
Actually the icon is rotated only in css, now it is the "X" sign so after rotating it looks same. If (for any reason) another icon is applied it could be a problem though

### Steps to reproduce
Create v-select, set the value, focus

### Versions
0.15.7

### What is expected ?
Clearable icon should not be rotated

Clearable icon is rotated

### Reproduction Link
It doesn't load icons font to show the rotation
https://codepen.io/anon/pen/LzVjdy

jacekkarczmarczyk:
Fixes #1741

### DIFF
--- a/src/components/VTextField/VTextField.spec.js
+++ b/src/components/VTextField/VTextField.spec.js
@@ -246,4 +246,19 @@ test('VTextField.js', ({ mount }) => {
     expect(change).toBeCalledWith('fgh')
     expect(change.mock.calls).toHaveLength(1)
   })
+
+  it('should not make prepend icon clearable', () => {
+    const wrapper = mount(VTextField, {
+      propsData: {
+        prependIcon: 'check',
+        appendIcon: 'check',
+        value: 'test',
+        clearable: true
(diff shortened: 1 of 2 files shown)


johnleider:
I actually handled this differently in dev yesterday but haven't pushed. Will review when I get home.

jacekkarczmarczyk:
Even if you decide not to merge it you may consider to merge at least changes in _input-groups.styl, there is no need for theming 'padding' or 'transition' props and it also increases size of compiled css

johnleider:
This user-select: none was very intentional. Why would that need to be removed?

jacekkarczmarczyk:
It's not removed, it's just moved outside the theme(), so instead of
```css
.application--light
  .input-group__append-icon
    user-select: none
.application--light
  .input-group__append-icon  
     user-select: none
```

we have

```css
.input-group__append-icon 
  user-select: none
```

PR #1750, opened 2017-09-15
Button active prop applies no styling

jacekkarczmarczyk:


johnleider:
This is already possible, just not discussed properly it seems. This is however, not a bug.

jacekkarczmarczyk:
So the `active` prop should be removed from docs?

johnleider:
Yes

PR #1751, opened 2017-09-15
added accent class for buttons

jacekkarczmarczyk:


johnleider:
I have a task to remove "contextualable" from all components. I will be converting this component in the process. You'll be able to just do **color="accent"**.

PR #1771, opened 2017-09-17
Make VIcon aria-hidden by default, remove explicit aria-hidden in consumers, update snapshots.

Phlow2001:


johnleider:
Is this still being pursued?

Phlow2001:
I'll have to redo when I have time, it's way out of sync at this point.

Issue #1637, opened 2017-09-08
[feature request/enh] Add possibility to display list tile with @click handler or to/href prop as a non-link tile


jacekkarczmarczyk:
Usecase:
https://codepen.io/anon/pen/RZzYeP

### What will it allow you to do that you can't do today?

It will add more control to the list item. Now `list__tile--link` class is added depending on the existence of `@click` / `to=""` / `href=""`

### How will it make current work-arounds straightforward?

Proposed functionality can be achieved by adding a class to the list tile and some custom css as it's show in the codepen. It would be easier and more intuitive to control it just by some bool property (similar to `disabled`)

### What potential bugs and edge cases does it help to avoid?

Workaround might cause a problems if the Vuetify internals (class names, html structure, styles etc) change in the future

PR #1786, opened 2017-09-17
Adds possibility to display list tile with @click handler or to/href prop as a non-link tile

jacekkarczmarczyk:
Adds bool `active` prop to `v-list-tile`, default `true`

If list tile is not active (`:active="false"`)
- tile doesn't get `list__tile list__tile--link` class so it doesn't look like a link
- href/to props are ignored

Edit: needs docs update if merged


johnleider:
I'm not particularly a fan of this. If the tag isn't **a**, and there is no **href** and there is no **to** and there is no **click**, that's VERY specific. You might as well simply add the class yourself no? I dunno what others thoughts are.

jacekkarczmarczyk:
First of all I don't agree that non-interactive list (without click/to/href) is VERY specific. I wouldn't call it even "specific" (without "very"). It's just a list of items that doesn't provide any action - this actually could be done by adding `disabled` prop, but it adds the 'disabled' style which is not always desirable

It can be solved by adding a class and overriding vuetify styles (as I did in codepen in #1637), it just doesn't seem to me proper solution for not unusual case. Actually I wish that this construction worked:
```html
<v-list-tile @click="condition ? handler : null">
```
but it doesn't, when `condition` is `false` `$listeners['click']` is not null as some would expect

johnleider:
I definitely misunderstood a few things, my apologies. This is perfectly fine.

PR #1788, opened 2017-09-17
VSelect -  Re-evaluate selection when items changes

### LINKED ISSUE #1743: v-select default v-model not show up correctly when item get from async
### Steps to reproduce
get v-select items from async/await
### Versions

v0.15.7
### What is expected ?

v-select will show up the default value

the default value not display correctly
### Reproduction Link

https://codepen.io/guanxudong/pen/dVoZYB?editors=1010

KaelWD:
Fixes #1743 

`selectedItems` should probably be a computed property instead of data, but I don't feel like wading through that much spaghetti today.

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -250,6 +250,8 @@ export default {
       this.searchValue && this.$nextTick(() => {
         this.$refs.menu.listIndex = 0
       })
+
+      this.genSelectedItems()
     },
     menuIsActive (val) {
       if (!val) return
@@ -430,9 +432,7 @@ export default {
 
       return typeof value === 'undefined' ? item : value
     },
-    genSelectedItems (val) {
-      val = val || this.inputValue
-
+    genSelectedItems (val = this.inputValue) {
       // If we are using tags, don't filter results
       if (this.tags) return (this.selectedItems = val)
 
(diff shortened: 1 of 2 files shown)


src/components/VSelect/VSelect.js
```diff
@@ -378,11 +380,12 @@ export default {
       this.isActive = true
       this.isFocused = true
 
-      if (this.$refs.input && this.isAutocomplete) {
-        this.$nextTick(() => {
-          // this.$refs.input.focus()
```
KaelWD: This wasn't doing anything so I marked it with a TODO so it's visible

src/components/VSelect/VSelect.js
```diff
@@ -378,11 +380,12 @@ export default {
...
-        })
-      }
+      // TODO
```
johnleider: I think you can just go ahead and remove this.

KaelWD: Moved it to my other PR instead

gygoo:
When will these changes be published? In version 1.0.17, v-select is not reactive. Does not respond to the change in the status of the selected item.

KaelWD:
If you check the merge commit you'll notice it was released with v0.16. If you're having a different problem please open a new issue, although it may have been fixed already in the 1.1 alpha so check with that first.

PR #1790, opened 2017-09-17
Add tags with tab and arrow keys as well as enter

### LINKED ISSUE #1746: [enhancement] ability to add new tags with "tab" key
Currently in `<v-select chips tags>` new tags can be added only by pressing "enter" key, presonally I thinks it's more intuitive to add them by pressing "tab" key. There might be situations that other keys (comma, space) could be used so I think that it should possible to choose the allowed keys as a prop, similar to `toggle-keys` prop. I would suggest to use `[9, 13]` as a default value as it is very likely that after typing user want's to add new tag and move to another field by pressing tab

Codepen just shows the `<v-select chips tags>` component
https://codepen.io/anon/pen/aLOKRG

KaelWD:
Update: I've decided against custom keys, it was way too messy and confusing to configure. Tab and enter are now hardcoded instead. I will attempt to add configurable delimiters also (eg. split a string into tags on commas)

### Tasks
 - [x] Basic functionality
 - [x] Tests
 - [x] More tests
 - [x] Tab should add the currently highlighted search result as a tag
 - [x] Pressing the right arrow will deselect the search results and allow you to add a new value instead
 - [x] Pressing the left arrow at the left hand side of the search text will add the current text and select the previous tag
 - [x] Refactor everything to be not disgusting

### Related ideas that have come up
 - Allow items to be added when space or comma is entered
 - Currently highlighted search result should be inserted into the textfield: image image

Fixes #1746


src/components/VSelect/VSelect.js
```diff
@@ -346,7 +345,8 @@ export default {
         return
       }
 
-      if (![8, 46].includes(keyCode)) return
+      // TODO: We already returned up there though?
+      if (![8, 46].includes(keyCode)) return // backspace/delete
```
johnleider: can be removed

src/components/VSelect/mixins/select-autocomplete.js
```diff
@@ -30,32 +34,41 @@ export default {
         i, this.searchValue, this.getText(i))
       )
     },
-    onKeyDown (e) {
-      // If enter or space is pressed, open menu
-      if (!this.menuIsActive &&
-        [13, 32, 38, 40].includes(e.keyCode)
-      ) {
+    async onKeyDown (e) {
```
johnleider: no component level async calls please, something we can explore in the future

src/components/VSelect/mixins/select-autocomplete.js
```diff
@@ -30,32 +34,41 @@ export default {
...
+        this.searchValue &&
         !this.filteredItems.length
       ) {
         this.selectedItems.push(this.searchValue)
 
-        this.$nextTick(() => {
-          this.searchValue = null
-          this.$emit('change', this.selectedItems)
-        })
+        await this.$nextTick()
```
johnleider: revert 49->52

src/components/VSelect/mixins/select-autocomplete.js
KaelWD: Is this because of a-la-carte, or do you just not like having random promises all over the place?

src/components/VMenu/mixins/menu-keyable.js
```diff
@@ -60,20 +60,9 @@ export default {
       } else if (e.keyCode === 13 && this.listIndex !== -1) {
         this.tiles[this.listIndex].click()
       }
-
-      if (this.listIndex === -1) this.setActiveListIndex()
```
KaelWD: Can you tell me what this was for? It doesn't seem to do anything useful as far as I can tell.

johnleider:
This will be a post 0.16 merge.

PR #1791, opened 2017-09-18
Avoid operate documentElement by inline style, it will conflict with some rem solu…

kimixu717:
Avoid operate documentElement(html) by inline style, it will conflict with some rem solution. And fix rolling through the overlay by add touch-action

### DIFF
--- a/src/mixins/overlayable.js
+++ b/src/mixins/overlayable.js
@@ -158,7 +158,7 @@ export default {
     },
     hideScroll () {
       if (this.$vuetify.breakpoint.mdAndDown) {
-        document.documentElement.style.overflow = 'hidden'
+        document.documentElement.classList.add('ovf-hidden')
       } else {
         window.addEventListener('mousewheel', this.scrollListener)
         window.addEventListener('keydown', this.scrollListener)
@@ -166,7 +166,7 @@ export default {
     },
     showScroll () {
       if (this.$vuetify.breakpoint.mdAndDown) {
-        document.documentElement.removeAttribute('style')
+        document.documentElement.classList.remove('ovf-hidden')
       } else {
         window.removeEventListener('mousewheel', this.scrollListener)
         window.removeEventListener('keydown', this.scrollListener)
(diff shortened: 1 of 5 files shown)


johnleider:
I understand the touch change, but I do not understand the documentElement issue. Also, modify the css within the .styl files, not the dist file. That gets overwritten every release.

kimixu717:
@johnleider I have revert the change of css files. About the documentElement issue: in some project, the 
whole page is written in rem unit rather than px, dynamic modification the documentElement's font-size to adaptive the page(show the same content in diffrent screens) , such as the m.tmall.com in iphone 6 it shows '<html data-dpr="1" style="font-size: 37.5px;">', in iphone 6 plus it shows '<html data-dpr="1" style="font-size: 41.4px;">' . So in this projects, Vuetify modification then documentElement is conflict with it.

src/stylus/elements/_global.styl
```diff
@@ -11,6 +11,8 @@ html
   -moz-osx-font-smoothing: grayscale
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0)
   overflow-x: hidden
+  &.ovf-hidden
```
johnleider: add this to src/stylus/trumps/_display.styl as helper classes for **.overflow-hidden**, **.overflow-x-hidden** and **overflow-y-hidden**.

kimixu717:
@johnleider done

src/mixins/overlayable.js
```diff
@@ -158,15 +158,15 @@ export default {
...
+        document.documentElement.classList.add('overflow-y-hidden')
       } else {
         window.addEventListener('mousewheel', this.scrollListener)
         window.addEventListener('keydown', this.scrollListener)
       }
     },
     showScroll () {
       if (this.$vuetify.breakpoint.mdAndDown) {
-        document.documentElement.classList.remove('ovf-hidden')
+        document.documentElement.classList.remove('overflow-y-hidden')
```
johnleider: Before you had .ovf-hidden hide all overflow, is there a reason for only y here now?

kimixu717: because the documentElement already has the overflow-x:hidden; attribute, in src/stylus/elements/_global.styl

Issue #1324, opened 2017-08-12
[Bug] Hover broken on menu


Kylart:
### Steps to reproduce
Hover and leave multiple times on that button

### Versions
vuetify 0.14.9

### What is expected ?
Disappear on `mouseleave` and appear on `mouseenter`

### What is actually happening ?
What happens on the codepen. If you hover and leave multiple times, you won't be able to make the menu appear again.

In my app I have actually another bug where the menu visibly disappear but I can still click on the button and it executes the `@click` method.

### Reproduction Link
https://codepen.io/Kylart/pen/JyyZGV

PR #1796, opened 2017-09-18
fixes #1794

### LINKED ISSUE #1794: menu with hover option disappear delay
### Reproduction Link
https://codepen.io/cayaner/pen/RLrbxB

### Steps to reproduce
open [codepen](https://codepen.io/cayaner/pen/RLrbxB) link, then hover the drop down btn, then move away and the drop done will stay visible for couple of seconds.
### Versions
vuetify 15.6

### What is expected ?
menu with hover option should disappear immediately after moving mouse away.

The menu stay visible for a while after moving mouse away.

jacekkarczmarczyk:
why just control close delay if we can control open delay as well

`<v-menu open-delay="500" close-delay="0">`

Needs docs update if merged

### DIFF
--- a/src/components/VMenu/mixins/menu-activator.js
+++ b/src/components/VMenu/mixins/menu-activator.js
@@ -1,12 +1,23 @@
 /**
  * Menu activator
- * 
+ *
  * @mixin
  *
  * Handles the click and hover activation
  * Supports slotted and detached activators
  */
 export default {
+  props: {
+    openDelay: {
+      type: [Number, String],
+      default: 0
+    },
+    closeDelay: {
+      type: [Number, String],
+      default: 500
+    }
+  },
+
   methods: {
     activatorClickHandler (e) {
       if (this.disabled) return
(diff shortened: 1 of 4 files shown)


johnleider:
The delay wasn't actually arbitrary, it was to protect against a potential type of bug from happening. I'll have to get my hands dirty with this one.

johnleider:
So this bug was fixed by adding the timeout, https://github.com/vuetifyjs/vuetify/issues/1324. However, through experimenting, you can go all the way down to 100 ms and it was still fine. So my change request is to add a validator to the close-delay prop for the value to always be >= than 100

src/components/VMenu/mixins/menu-activator.js
```diff
@@ -1,12 +1,23 @@
...
  */
 export default {
+  props: {
+    openDelay: {
+      type: [Number, String],
+      default: 0
+    },
+    closeDelay: {
+      type: [Number, String],
+      default: 500
```
johnleider: add validator for values >= 100

jacekkarczmarczyk:
I'm not fan of "magic values" like mentioned "100", so I started to dig deeper, and here's what i've found:

- there are 3 vars controlling the visibility of menu - `hasJustFocused`, `isActive`, `isContentActive` - I still don't fully understand how it works, but the problem was caused by `isContentActive`
- `isContentActive` is set to `true` in `menuable.js/startTransition()`, there are 2 `requestAnimationFrame`s before it gets the `true` value (1 in `startTransition` and 1 in `VMenu.js/activate()`), so I suspected that the minimum close delay is somehow related to raf
- indeed, I was able to replicate the issue with delay 31, but not 32 (or 30/31 respectively, don't remember) which is about 2 raf intervals
- meanwhile i've found that adding `this.callDeactivate()` in `menu-activator.js/mouseLeaveHandler` allows to close the menu if it stuck, but
- setting `isActive = true` in the `requestAnimationFrame` solved the problem - https://github.com/vuetifyjs/vuetify/pull/1796/commits/a5d690ebaf4a183ecf2e17d0b44889816fe764b5#diff-19e53fa58c8437f379cfe169ce66b36bL49 - I mean I can't reproduce it anymore with `closeDelay="0"`, but I've left there `callDeactivate()` just in case

There's still some magic not very clear to me, but I prefer going this way than adding some arbitrary restrictions to delays. Do you this there is a way to simplify the menu? These 3 variables seem to me overcomplicated, but maybe there is a reason for this, I'm open for any suggestions

PR #1821, opened 2017-09-20
Focus v-select's internal input when focus() is called

### LINKED ISSUE #1819: v-select function focus() does not work
### Steps to reproduce
1) create v-select with autocomplete
2) call focus()

### Versions
0.15.7

### What is expected ?
keyboard focus into select

select highlighted, but no focused

### Reproduction Link
https://codepen.io/anon/pen/gGrezV?editors=1010#0

Not work after commit e96ca54317995eaf56eef53cd6724a8bad57b2e2
src/components/VSelect/VSelect.vue
line 369

KaelWD:
Fixes #1819

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -392,12 +392,11 @@ export default {
     focus () {
       this.isActive = true
       this.isFocused = true
-      this.$el.focus()
 
       if (this.$refs.input && this.isAutocomplete) {
-        this.$nextTick(() => {
-          this.$refs.input.focus()
-        })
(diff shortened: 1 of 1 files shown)


src/components/VSelect/VSelect.js
```diff
@@ -248,7 +248,7 @@ export default {
     },
     isBooted () {
       this.$nextTick(() => {
-        if (this.content) {
+        if (this.content instanceof EventTarget) {
```
KaelWD: `this.content` is always an object, so this was basically doing `if (true) {`

PR #1822, opened 2017-09-20
Focus menu activator when clicked

### LINKED ISSUE #1817: Menu popup doesn't close on tab if activating with button
### Steps to reproduce
1. Click Menu Button 1
2. Hit Tab and notice the menu is still open

Behavior if you activate the menu with space and then tab is correct.

### Versions
Latest

### What is expected ?
Tabbing away from the menu should close it

Tab after click doesn't cause the menu to close

### Reproduction Link
https://codepen.io/jayblanchard/pen/mBPpdd?editors=1010

KaelWD:
Fixes #1817

### DIFF
--- a/src/components/VMenu/mixins/menu-activator.js
+++ b/src/components/VMenu/mixins/menu-activator.js
@@ -21,11 +21,15 @@ export default {
   methods: {
     activatorClickHandler (e) {
       if (this.disabled) return
-      else if (this.openOnClick && !this.isActive) {
+      if (this.openOnClick && !this.isActive) {
+        this.$slots.activator[0].elm.focus()
         this.isActive = true
         this.absoluteX = e.clientX
         this.absoluteY = e.clientY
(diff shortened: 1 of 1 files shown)


KaelWD:
wtf codebeat

src/components/VMenu/mixins/menu-activator.js
```diff
@@ -21,11 +21,15 @@ export default {
   methods: {
     activatorClickHandler (e) {
       if (this.disabled) return
-      else if (this.openOnClick && !this.isActive) {
+      if (this.openOnClick && !this.isActive) {
+        this.$slots.activator[0].elm.focus()
```
johnleider: What happens if this.$slots.activator is empty?

KaelWD: Fixed, I didn't realise that there was an `activator` prop. We really need to figure out how to get mixins into the docs.

PR #1827, opened 2017-09-21
Enhancement/masks

### LINKED ISSUE #861: Masking Inputs
Hi guys. Great library. Are there any plans for built-in support for masking text inputs? I think most serious applications employ masking for phone number inputs and the like.  Given that this will continue to be a need for developers using vuetify, even if built-in support isn't on the roadmap, I think an example on how to integrate with a 3rd party solution would be useful. 

> What will it allow you to do that you can't do today?

Mask inputs

> How will it make current work-arounds straightforward?

All I see right now is a closed issue directing users to get help using 3rd party libraries on gitter.

> What potential bugs and edge cases does it help to avoid?

Helps users to input data correctly the first time without reyling on hints and form errors.

johnleider:
resolves #861


KaelWD:
I don't know if there's any way to fix this, but if you edit somewhere in the middle of the value it deletes some characters. Also the cursor goes to the end of the textbox so you can't change more than one character without having to click back to the correct place. 

eg. with mask `(##) #### ####`:
 - Enter my phone number
image
 - I want to change the area code to `(02)` instead, so I click behind the `1` and press backspace
image
 - The cursor has moved back to the end, and the `3` and `6` have vanished
 - I click back to after `0` and enter `2`
image
 - Now my area code is correct, but I'm two numbers short - it should read `(02) 2345 6789`

johnleider:
Yea, that's a pickle.

johnleider:
So this I might need some help with thinking up an idea how to solve the caret position. Everything works except ensuring that the input caret is at the proper position when editing the field.

**SEND HELP!**

KaelWD:
Now when I press `3` it deletes most of the other characters 
image

src/mixins/maskable.js
```diff
@@ -0,0 +1,130 @@
...
+    masked () {
+      const preDefined = this.preDefined[this.mask]
+      const mask = preDefined || this.mask || ''
+
+      return mask.split('')
+    }
+  },
+
+  methods: {
+    maskText (text, newText = []) {
```
jacekkarczmarczyk: Last but not least - as discussed before I'd suggest to extract the mask logic to separate file (src/util/*), it will follow the single responsibility rule and make the code analysis or testing easier

azaars:
I made a PR. It's about the caret problem. #1879

evaldobratti:
@johnleider is it going to have currency mask support?

johnleider:
@evaldobratti I would say yes, could you provide an example?

src/util/mask.js
```diff
@@ -13,22 +18,50 @@ const allowedMasks = {
   }
 }
 
+/**
+ * Default delimiter regexp
+ * 
+ * @type {RegExp}
+ */
+const defaultDelimiter = /[-!$%^&*()_+|~=`{}\[\]:";'<>?,.\/ ]/
```
jacekkarczmarczyk: maybe this const could be used in mask.js/allowedMasks and maskable.js/data/delimiters?

src/components/VTextField/VTextField.js
```diff
@@ -3,19 +3,22 @@ require('../../stylus/components/_text-fields.styl')
...
 
   inheritAttrs: false,
 
   data () {
     return {
       initialValue: null,
       inputHeight: null,
-      badInput: false
+      badInput: false,
+      oldValue: null,
```
azaars: This can be removed. It's no longer needed in `setSelectionRange()`

src/components/VTextField/VTextField.js
```diff
@@ -84,7 +87,10 @@ export default {
         return this.value
       },
       set (val) {
+        this.oldValue = this.$refs.input ? this.$refs.input.value : ''
```
azaars: This line can be removed. It's no longer needed in `setSelectionRange()`

KaelWD:
There are still issues with the caret. If the mask changes the caret position doesn't update.

```js
computed: {
  mask () { // There are three common formats for AU phone numbers
    return this.phone.length > 1 && this.phone.startsWith('0')
      ? this.phone.startsWith('04')
        ? '#### ### ###' // Mobile
        : '(##) #### ####' // With area code
      : '#### ####' // Local number
  }
}
```

Type `03` and the caret ends up in between the two numbers: 
image

src/mixins/maskable.js
```diff
@@ -0,0 +1,92 @@
...
+        const newValue = this.$refs.input.value
+        if (!this.deleting) {
+          while (newValue.slice(this.selection - 1, this.selection).match(this.delimiters)) {
+            this.selection++
+          }
+        }
+
+        if (this.$refs.input.setSelectionRange) {
+          this.$refs.input.setSelectionRange(this.selection, this.selection)
+        } else if (this.$refs.input.createTextRange) {
```
KaelWD: Is this needed? AFAIK everything IE9+ supports `setSelectionRange`

azaars: I was not so sure about the support when I add the lines. Since it's not needed, you may remove it. Thank you.

azaars:
@KaelWD I made #1896 to allow props.mask swapping and re-positioning of the caret.

src/mixins/maskable.js
```diff
@@ -61,30 +61,22 @@ export default {
...
 
         const newValue = this.$refs.input.value
         if (!this.deleting) {
-          while (isMaskDelimiter(newValue.slice(this.selection - 1, this.selection))) {
+          while (isMaskDelimiter(newValue.substr(this.selection - 1, 1))) {
             this.selection++
           }
+          // TODO: This only helps sometimes
+          if (this.$refs.input.selectionStart - this.selection === 1) {
+            this.selection = this.$refs.input.selectionStart
```
azaars: How does it help?

Please try this:-

1. Type 035
2. Move the caret anywhere in between 3 and 5
3. Type 4
4. The caret moves right after 5 due to this added lines, while it should be in between 4 and 5

Using #1896 and:
```javascript
computed: {
  mask () { // There are three common formats for AU phone numbers
    return this.phone.length > 1 && this.phone.startsWith('0')
      ? this.phone.startsWith('04')
        ? '#### ### ###' // Mobile
        : '(##) #### ####' // With area code
      : '#### ####' // Local number
  }
}
```

KaelWD: Comment it out and you get this instead image

It only checks for a difference of 1, so if more than one character changes then the cursor goes to the end.

evaldobratti:
@johnleider something like https://vuejs-tips.github.io/v-money/
that apply the thousands separator as you type, custom currency symbol and thousands/precision separators

src/mixins/maskable.js
KaelWD: I just realised that it breaks `fill-mask-blanks` -_-

johnleider:
@evaldobratti This is something we can look at for a future release, but not with this one.

src/mixins/maskable.js
```diff
@@ -0,0 +1,110 @@
...
+      'social': '###-##-####',
+      'time': '##:##',
+      'time-with-seconds': '##:##:##'
+    }
+  }),
+
+  props: {
+    fillMaskBlanks: Boolean,
+    mask: {
+      type: [Object, String],
```
azaars: Currently the support on `mask` is for `String` type.

jacekkarczmarczyk: Good question,it's not used now but I think it can be used for creating custom rules  - like now 'A' means `/A-Z/` automaticly converted to upper case, dev could define his own rules, like `'F' => /0-9A-F/`

azaars: This option could be very useful at least to me. I wonder if it could be used as follows.

Handled cases:

1. Predefined : String (existing)
```javascript
    mask="credit-card"
```

2. Predefined with default options : String mapped to Object
```javascript
    mask="numeral"
```

3. Predefined with overidden options : Object
```javascript
    mask="{ formatter: 'numeral', prefix: '$', precision: 4 }"
```

4. Custom mask : String (existing)
```javascript
    mask="(##) - (AA)"
```

5. Custom mask with formatter : Object
```javascript
    mask="{ formatter: function () {}, myOption1: '23', myOption2: '?' }"
```

@johnleider Is this in line with your plan?

johnleider:
I do like that, but not for initial 0.16 release

src/mixins/maskable.js
azaars: Well, I would have to start with something even if it's using git branch in my project. Can I attempt a PR?

azaars:
Well, I would have to start with something even if it's using git branch in my project. Can I attempt a PR?

johnleider:
Yes

Mucaccino:
Hi. How about multiple possible masks, like that:

<the-mask :mask="['###.###.###-##', '##.###.###/####-##']" />

Thanks,

Mucaccino:
Hi. How about multiple possible masks, like that:

`<the-mask :mask="['###.###.###-##', '##.###.###/####-##']" />`

Thanks,

PR #1845, opened 2017-09-23
input/asyncLoading and VProgressLinear enhancements and fixes

jacekkarczmarczyk:
- adds **loadable** mixin that allows to use default progress bar (color taken from `loading` prop or from `color` prop or `primary`, indeterminate, height 2px) or progress bar from `progress` slot
- **v-text-field**/**v-select**/**v-data-table**: adds `progress` slot
- **v-data-table** - fixes `loading` behaviour (setting color didn't work and was supposed to work only for theme colors, now it works for all colors)
- **v-data-table** - changes progress height from 3px to 2px
- **v-text-field**/**v-select** **BREAKING CHANGE**: changes the `asyncLoading` prop to `loading`
- **v-progress-linear** **BREAKING CHANGE**: removes `buffer` prop
- **v-progress-linear** **BREAKING CHANGE**: merges `colorFront`, `primary`, `error` etc props into one `color` prop - linear progress in now _colorable_
- **v-progres-linear** **BREAKING CHANGE**: changes `colorBack` prop to `backgroundColor` - to be more consistent with `backgroundOpacity` prop and `thumbColor` etc props in `v-slider`
- **v-progress-linear**: sets the default value of `bufferValue` to `100`
- **v-progress-linear** - adds `backgroundOpacity` prop (default `1` if `backgroundColor` is provided, otherwise `0.3`, changed from `0.5` that was default before this change)
- **v-progress-linear** - fixes error when the rendered bar had width relative to buffer width instead of full progress width (buffer=50, value=40 - bar had width 20% instead of 40%)
- adds/updates tests for **v-progress-linear**, **v-data-table**, **v-text-field**

**Needs docs update if merged**

### Playground.vue
```vue
<template>
  <v-app class="pa-5">
    <v-content>
      <v-checkbox label="Async loading" v-model="loading"></v-checkbox>
      <v-checkbox label="Custom progress" v-model="custom" :disabled="!loading"></v-checkbox>
      <v-text-field color="cyan darken" label="Password" v-model="val" :async-loading="loading">
        <v-progress-linear v-if="custom" slot="progress" :value="progress" height="7" :color="color"></v-progress-linear>
      </v-text-field>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data: () => ({
    val: 'test',
    custom: true,
    loading: true
  }),
  computed: {
    progress() { return Math.min(100, this.val.length * 10) },
    color() { return ['error', 'warning', 'success'][Math.floor(this.progress / 40)] }
  }
}
</script>
```

image

If the progress bar height is too big it may overlay the details (hint/error) section, but in some cases can be useful

### DIFF
--- a/src/components/VProgressLinear/VProgressLinear.js
+++ b/src/components/VProgressLinear/VProgressLinear.js
@@ -74,7 +74,7 @@ export default {
         return 0
       }
 
-      return Math.min(this.bufferValue, this.value) * 100 / this.bufferValue
+      return this.value * 100 / this.bufferValue
     },
     bufferStyles () {
       const styles = {}
--- a/src/components/VProgressLinear/VProgressLinear.js
+++ b/src/components/VProgressLinear/VProgressLinear.js
@@ -26,8 +26,10 @@ export default {
       type: [Number, String],
       default: 0.3
     },
-    buffer: Boolean,
-    bufferValue: Number,
+    bufferValue: {
+      type: [Number, String],
+      default: 100
+    },
     color: {
       type: String,
       default: 'primary'
@@ -57,20 +59,16 @@ export default {
         styles.height = 0
       }
 
-      if (this.buffer) {
-        styles.width = `${this.bufferValue}%`
-      }
+      styles.width = `${this.bufferValue}%`
 
       return styles
     },
     effectiveWidth () {
-      if (!this.buffer) {
-        return this.value
-      } else if (this.bufferValue) {
-        return Math.min(this.bufferValue, this.value) * 100 / this.bufferValue
-      } else {
+      if (!this.bufferValue) {
         return 0
       }
+
+      return Math.min(this.bufferValue, this.value) * 100 / this.bufferValue
     },
     bufferStyles () {
       const styles = {}
@@ -125,15 +123,11 @@ export default {
 
     const bar = h('div', { class: ['progress-linear__bar'], style: this.styles }, [fade, slide])
     const background = h('div', {
+      staticClass: `progress-linear__background ${this.color}`,
       style: {
-        position: 'absolute',
-        top: 0,
-        left: 0,
-        right: 0,
-        bottom: 0,
-        opacity: this.backgroundOpacity
-      },
-      class: this.color
+        opacity: this.backgroundOpacity,
+        width: `${this.bufferValue}%`
+      }
     })
 
     return h('div', {
--- a/src/components/VProgressLinear/VProgressLinear.js
+++ b/src/components/VProgressLinear/VProgressLinear.js
@@ -59,7 +59,9 @@ export default {
         styles.height = 0
       }
 
-      styles.width = `${this.bufferValue}%`
+      if (!this.indeterminate && this.bufferValue != 100) {
+        styles.width = `${this.bufferValue}%`
+      }
 
       return styles
     },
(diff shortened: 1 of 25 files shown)


src/mixins/input.js
```diff
@@ -194,11 +205,18 @@ export default {
...
-      if (this.asyncLoading) {
+      if (this.asyncLoading !== false) {
         detailsChildren.push(this.$createElement('v-progress-linear', {
           props: {
-            indeterminate: true,
-            height: 2
+            indeterminate: this.asyncLoading === true,
+            value: Number(this.asyncLoading),
+            height: this.asyncLoadingHeight,
+            primary: this.asyncLoadingColor === 'primary',
```
johnleider: Since we are trying to move away from this model, I'd be interested to see if we could modify the color differently.

nekosaur:
I haven't had the chance to look this over, but do we lose the ability to independently set front and back colors, since colorBack is gone?

src/components/VProgressLinear/VProgressLinear.js
```diff
@@ -77,24 +84,39 @@ export default {
       }
 
       return styles
+    },
+    backgroundStyle () {
+      const backgroundOpacity = (this.backgroundOpacity === '' || this.backgroundOpacity == null)
+        ? (this.backgroundColor ? 1 : 0.3)
+        : Number(this.backgroundOpacity)
```
johnleider: Is there a reason to use Number over parseInt here?

src/components/VProgressLinear/VProgressLinear.js
```diff
@@ -77,24 +84,39 @@ export default {
       }
 
       return styles
+    },
+    backgroundStyle () {
+      const backgroundOpacity = (this.backgroundOpacity === '' || this.backgroundOpacity == null)
```
johnleider: Can we simplify this to **this.backgroundOpacity == null**

src/components/VProgressLinear/VProgressLinear.js
```diff
@@ -77,24 +84,39 @@ export default {
...
     },
     genBar (h, name) {
       return h('div', {
-        class: [
-          'progress-linear__bar__indeterminate',
-          name,
-          this.colorFront
-        ]
+        class: {
+          'progress-linear__bar__indeterminate': true,
```
johnleider: this should be moved to the prop **staticClass**

src/components/VProgressLinear/VProgressLinear.js
```diff
@@ -115,12 +137,19 @@ export default {
     const fade = h('v-fade-transition', [this.indeterminate && this.genIndeterminate(h)])
     const slide = h('v-slide-x-transition', [!this.indeterminate && this.genDeterminate(h)])
 
-    const bar = h('div', { class: ['progress-linear__bar', this.colorBack], style: this.styles }, [fade, slide])
+    const bar = h('div', { class: ['progress-linear__bar'], style: this.styles }, [fade, slide])
```
johnleider: this class should change to **staticClass**. (I know it was like this before, but you're in here :))

src/mixins/input.js
```diff
@@ -21,7 +21,18 @@ export default {
...
-    asyncLoading: Boolean,
+    asyncLoading: {
+      type: [Boolean, Number],
+      default: false
+    },
+    asyncLoadingHeight: {
+      type: [String, Number],
+      default: 2
+    },
+    asyncLoadingColor: {
```
johnleider: I would rather expose the progress spot through a scoped slot instead of adding tons of props to mirror to it.

johnleider:
You like to bake big cakes @jacekkarczmarczyk . Is there any documentation updates that coincide with this?

PR #1874, opened 2017-09-25
Auto adjust scrollable dialog's height based on content

d3radicated:
Currently, on scrollable dialogs, you need to set `v-card-text`'s height for it to properly show its contents. Also, where `v-card-text`'s height is too high, it will hide `v-card-actions`. This PR will auto adjust `v-card-text`'s height based on content and it will not hide `v-card-actions`.

### DIFF
--- a/src/stylus/components/_dialogs.styl
+++ b/src/stylus/components/_dialogs.styl
@@ -53,7 +53,7 @@
       .card
         display: flex
         flex: 1 1 auto
-        flex-flow: column
+        flex-direction: column
         
         .card__title
         .card__actions
--- a/src/stylus/components/_dialogs.styl
+++ b/src/stylus/components/_dialogs.styl
@@ -46,8 +46,19 @@
         padding 0 !important
 
   &--scrollable
+    display: flex
     overflow: hidden
-
-    .card__text
-      overflow-y: auto
-      backface-visibility: hidden
+    
+    >
+      .card
+        display: flex
+        flex: 1 1 auto
+        flex-flow: column
+        
+        .card__title
+        .card__actions
+          flex: 1 0 auto
+      
+        .card__text
+          overflow-y: auto
+          backface-visibility: hidden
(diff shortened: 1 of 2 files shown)


johnleider:
Please provide an example markup that will exhibit this behavior without your fix, thank you.

d3radicated:
From docs page: https://codepen.io/dEradicated/pen/OxWzeJ 
With fix: https://codepen.io/dEradicated/pen/MEbQar

Fiddle with the screen's height to see the behavior.

johnleider:
Looks good, only issue I see is the ```v-divider```'s are no longer working with your fix.

d3radicated:
I'll see what I can do with that. 😉

KaelWD:
`.divider` should have `flex: 0 0 1px`

src/stylus/components/_dialogs.styl
```diff
@@ -46,8 +46,19 @@
...
-
-    .card__text
-      overflow-y: auto
-      backface-visibility: hidden
+    
+    >
+      .card
+        display: flex
+        flex: 1 1 auto
+        flex-flow: column
```
KaelWD: This can just be `flex-direction`

d3radicated: Updated.

d3radicated:
I changed `v-divider`'s style by setting its `min-height` instead of just `height`;

KaelWD:
Put it back to `height` and change `flex` instead. Dividers should never be anything other than 1px high, so they shouldn't have `flex-grow` and `flex-shrink` like they do now.

d3radicated:
I changed it to `min-height` because changing it to `flex: 0 0` will break the steppers.

KaelWD:
Oh yeah they can be horizontal as well...  
You should still keep `height` at least because otherwise they'll be too tall in a fixed-height container.

d3radicated:
Done and done.

Issue #1890, opened 2017-09-26
Autocomplete menu only contains selected item when using objects


mirhat:
### Steps to reproduce
Create select with autocomplete option and specify item-text and item-value
Filter items and lose focus
Click on select
### Versions

"vuetify": "0.15.7"
### What is expected ?

All list items should be shown
### What is actually happening ?

Only the selected item is shown
### Reproduction Link

https://codepen.io/anon/pen/aLpaRV

mirhat:
Solution may be to change 
this.searchValue === this.selectedItem to this.searchValue === this.getText(this.selectedItem)

https://github.com/vuetifyjs/vuetify/blob/dev/src/components/VSelect/VSelect.js in 194. line

KaelWD:
> Only one item is shown

That was fixed in 2d4fa52c3ad44ad3cf373d66d9d3d961b20106cb, but looks like it's come back for some reason.

> cursor is in random position

I couldn't reproduce this on dev, it must have been fixed already.

KaelWD:
Fixed with #1891

PR #1891, opened 2017-09-26
Fix autocomplete showing only filtered item if field is not dirty

mirhat:
Based on issue #1890


mirhat:
Ok. I have reverted previous commit.

Issue #1910, opened 2017-09-27
Refactor setSelectionRange?


azaars:
This line appears more than once.

https://github.com/vuetifyjs/vuetify/blob/f7e3c6a3d2e9d6975d600f6a246e79a4a9e4c015/src/mixins/maskable.js#L73

KaelWD:
We can do a refactoring pass once all the functionality is down. That line doesn't really stand out to me as being particularly WET. 

This question is also more suited for discord, github issues are only for bugs and feature requests.

azaars:
@KaelWD I was in code view page when I created this issue. Honestly, I had no idea it's going to end up in the global issue page. That's why I immediately changed the title to be more descriptive. Blame my ignorance ;-)

@jacekkarczmarczyk #1912 is more than what I expected and it's brilliant.

azaars:
#1945

PR #1912, opened 2017-09-27
setSelectionRange refactoring

jacekkarczmarczyk:
fix proposal for #1910

Maybe the current code is not WET, but at least a bit humid, especially when you notice that `input.setSelectionRange()` always comes in pair with `this.selection = ...`, so this is my proposal to keep it together.
Maybe `this.selection` should be synchronized somehow with input selection, now there's a bunch of issues when changes the caret position and continues typing

### DIFF
--- a/src/mixins/maskable.js
+++ b/src/mixins/maskable.js
@@ -63,24 +63,30 @@ export default {
         isMaskDelimiter(oldText[i]) || position++
       }
 
-      this.selection = 0
+      let selection = 0
       for (const char of newText) {
         isMaskDelimiter(char) || position--
-        this.selection++
+        selection++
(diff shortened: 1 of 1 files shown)


src/mixins/maskable.js
```diff
@@ -63,24 +63,30 @@ export default {
...
+        selection++
         if (position <= 0) break
       }
 
+      this.selection = selection
+    },
+
+    selection () {
       this.$refs.input.setSelectionRange(this.selection, this.selection)
     }
```
KaelWD:
```js
selection (val) {
  this.$refs.input.setSelectionRange(val, val)
}
```

PR #1921, opened 2017-09-28
Progress-linear on hide does not shrink background

jacekkarczmarczyk:


src/components/VProgressLinear/VProgressLinear.js
```diff
@@ -85,10 +85,14 @@ export default {
         ? (this.backgroundColor ? 1 : 0.3)
         : parseFloat(this.backgroundOpacity)
 
-      return {
+      const styles = {
         opacity: backgroundOpacity,
         width: `${this.bufferValue}%`
       }
+      if (!this.active) {
```
johnleider: could this not be:

```
return {
  height: !this.active ? 0 : 'auto',
  opacity: backgroundOpacity,
  width: `${this.bufferValue}%`
}
```

jacekkarczmarczyk:
updated

Issue #1918, opened 2017-09-27
Allow adding classes, style and click handler to v-dialog's dialog__content, and its associated overlay


Webifi:
Since there seems to be little interest in making dialogs and their associated overlays stack based on when they were last made visible rather than when they were first added to the dom,  it would be nice to at least be able to target specific and active dialog's  div.dialog__content by allowing us to pass dialog-content-class, dialog-content-style, overlay-class and overlay-style parms to a v-dialog.   

Adding a dialog--content--active class to the parent div.dialog--content when there is a child div.dialog.dialog--active would also be needed.

This would allow using CSS to target specific active dialog containers and their overlays to allow us to set the z-index to what we need.

In addition, allowing some way to hook the click event of a dialog's overlay would be helpful.  Stopping its propagation, etc.

## What will it allow you to do that you can't do today?
Allow more control over dialogs and overlays via css.

## How will it make current work-arounds straightforward?
Current work-arounds to having stacking dialogs require changes to coordinate when dialogs are added to the dom.  This cannot always be easily controlled.

To work-around this with css, we need to be able to target specific  div.dialog__content based on if they contain active div.dialog.dialog--active. 

## What potential bugs and edge cases does it help to avoid?
Trying to deal with multiple open dialogs from different components.  For one example, a notification dialog that's opened by the parent of a component currently showing a dialog.  

Currently, depending on when it was added to the dom,  the parent's notification will not be visible until the child component's dialog is closed, even though the application may need that parent's dialog addressed before operation on the child's dialog can continue.  (Reaffirming expired access, notification of a record lock, etc.)

Regardless of when a dialog was first added to the dom, the associated overlays will always be behind ALL open dialogs.  This stops you from blocking access to other larger dialogs below.

johnleider:
Can you clarify this statement please

> Since there seems to be little interest in making dialogs and their associated overlays stack based on when they were last made visible rather than when they were first added to the dom

Webifi:
The "little interest" refers to previously closed issues that generally center around the behavior of multiple active dialog.  Broadly, the reason given for closing the issues was that opening multiple dialogs at the same time is an edge case. If there is interest in getting multiple active dialog to stack in a more palatable way, great!

[...]

PR #1931, opened 2017-09-28
Add z-index stacking order to vDialog and associated overlays

Webifi:
Sets z-index on dialogs and associated overlays allowing multiple open dialogs to be stacked based on when the dialog was last set active, not when it was added to the dom.

Also stops propagation of click event on overlay if overlay is used by dialog, stopping the dismissal of all open dialogs if the overlay for one dialog is clicked.

Removes need for #1918

### DIFF
--- a/src/components/VDialog/VDialog.js
+++ b/src/components/VDialog/VDialog.js
@@ -55,9 +55,6 @@ export default {
         'dialog--content--active': this.isActive
       }
     },
-    overlayZIndex () {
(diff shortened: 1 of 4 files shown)


johnleider:
Ty for your PR, but the process for overlays and dialogs/drawers and how they interact with each other is being refactored in 1.1.

Webifi:
Well, this would at least make multiple open dialogs usable until then.

PR #1942, opened 2017-09-29
Fixed bug when mask prop is not a string.

azaars:


KaelWD:
Can you give some information about the bug this is supposed to fix?

azaars:
The mask prop is defined to be of type [Object, String]. The Object type is in preparation for upcoming mask enhanced features. But for now, when an Object is passed, the checking on String.length will fail.

KaelWD:
Have you actually tried passing an object? It fails in maskable.js line 46.  
`this.masked` will always be an array, so this check is pointless.

azaars:
`this.masked` will fail when `mask` is passed as object. Therefore a fix is needed there.

Never mind, I'll pass this pull for now.

PR #1945, opened 2017-09-29
Fixed the bug where the caret position is not properly watched.

azaars:
Fixed the bug where the `selection()` is not called at certain situations when the caret position changes. The actual caret position will be moved to the end of input field particularly when the mask changes dynamically, and it should be re-positioned to the value of `this.selection`. However, because `this.selection` is observed (watched) as unchanged, `selection()` is not called to fix the caret position.

### DIFF
--- a/src/mixins/maskable.js
+++ b/src/mixins/maskable.js
@@ -58,27 +58,28 @@ export default {
       const oldText = this.$refs.input.value || ''
       const newText = this.maskText(this.lazyValue || '')
       let position = 0
+      let selection = this.selection
 
-      for (const char of oldText.substr(0, this.selection)) {
+      for (const char of oldText.substr(0, selection)) {
         isMaskDelimiter(char) || position++
       }
 
-      let selection = 0
+      selection = 0
       for (const char of newText) {
         isMaskDelimiter(char) || position--
         selection++
         if (position <= 0) break
       }
 
-      this.selection = selection
-    },
-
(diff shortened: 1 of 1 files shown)


johnleider:
Could you please provide an example that exhibits the behavior without your fix please.

azaars:
Try this:

```javascript
computed: {
  mask () { // There are three common formats for AU phone numbers
    return this.phone.length > 1 && this.phone.startsWith('0')
      ? this.phone.startsWith('04')
        ? '#### ### ###' // Mobile
        : '(##) #### ####' // With area code
      : '#### ####' // Local number
  }
}
```
And this is the bug:

bug-caret-position

johnleider:
I'm going to merge this in, can you please create an example to display within the text-field documentation, ty.

KaelWD:
This still isn't always correct

azaars:
I'll give it a shot. BTW, which docs branch should I use?

johnleider:
dev branch

azaars:
@KaelWD I think since it starts with 1, it follows the `'#### ####' // Local number` rule.

KaelWD:
Yeah, but because I pressed backspace, shouldn't the cursor be at the left of the `1`?

azaars:
Ahh good catch. That's a tricky one.

azaars:
There's a fix on the caret position #1995

PR #1946, opened 2017-09-29
Refactors data table progress markup to follow W3C standard

sindrepm:
As per the HTML 5/W3C [standard](https://www.w3.org/TR/html/tabular-data.html#the-table-element) a table cannot have more than one `thead` element. The current implementation of `v-data-table` adds a separate `thead` element for the progress bar. This PR refactors the markup to be in compliance with the standard. As a result some CSS could be simplified as well.


johnleider:
Does this introduce any breaking changes?

sindrepm:
Nope. It just changes the generated HTML markup.

mirajp:
This breaks tables that redefine the headers slot template with their own header row.
I'm pretty new to Javascript, and Vue/Vuetify, so my fix for now was to just add the progress indicator row right under the header row (note that to make the display conditional, I had to use a **v-if** with **myLoading** -- the slot's **props.loading** that is set by v-data-table's loading directive didn't seem to work even though in the Vue debugger shows that the VDataTable component has loading in the props object):

```
<v-data-table :headers="myHeader" :items="myRows" :loading="myLoading" hide-actions>
    <template slot="headers" slot-scope="props">
        <tr>
            <th v-for="header in props.headers">
                {{ header.text }}
            </th>
        </tr>
        <!-- This custom table row to show the progress when it is loading because Vuetify 0.17.0-beta.1 changes the thead templating -->
        <tr class="datatable__progress">
            <th colspan="100%" class="column">
                <div v-if="myLoading" class="progress-linear" style="height: 2px;">
                    <div class="progress-linear__background primary" style="height: auto; opacity: 0.3; width: 100%;"></div>
                    <div class="progress-linear__bar">
                        <div class="progress-linear__bar__indeterminate progress-linear__bar__indeterminate--active">
                            <div class="progress-linear__bar__indeterminate long primary"></div>
                            <div class="progress-linear__bar__indeterminate short primary"></div>
                        </div>
                    </div>
                </div>
            </th>
        </tr>
    </template>
    ...
</v-data-table>
```

KaelWD:
@mirajp Please create a new issue.

sindrepm:
@KaelWD I have confirmed the bug and created issue #2471

PR #1970, opened 2017-10-01
fixed bug with unchecked property read

nekosaur:
Can be seen on dev docs. The v-list holding the version select throws an error since there is no check to see if `$listeners` actually exists before trying to access `.click`. I simply reused existing isLink computed prop instead.


KaelWD:
IDK how *that* broke so many tests.

PR #1980, opened 2017-10-02
Fix VProgressCircular's error in a-la-carte import

herophuong:
Fix a case where importing VProgressCircular to declare as app local component triggers the arrow function behavior (`this` will be undefined).


herophuong:
It was a quick commit from Github directly. Fixed now.

PR #1994, opened 2017-10-03
fix/v-data-table-expand-sort

### LINKED ISSUE #1991: Table row doesn't expand after sorting
### Steps to reproduce

1. Rows are initailly sorted by 'Calories' column, click 'Calories' header to sort it in a reverse order, try to expand any row
2. Click 'Calories' header again

### Versions
0.16b3

### What is expected ?
1. Row should expand
2. Shound't throw error 

1. Row doesn't expand
2. Error is thrown
```
TypeError: Cannot read property 'style' of undefined
    at afterLeave (vuetify.min.js:6)
```
            
### Reproduction Link
https://codepen.io/anon/pen/ZXXrjG?editors=1010

I'm wondering - maybe expanded row should be added only after expanding (and removed after collapsing)?

nekosaur:
Fixed #1991

### DIFF
--- a/src/components/VDataTable/mixins/body.js
+++ b/src/components/VDataTable/mixins/body.js
@@ -16,25 +16,25 @@ export default {
       return this.$createElement('tbody', children)
     },
     genExpandedRow (props) {
-      const expand = this.$createElement('div', {
-        class: 'datatable__expand-content',
-        key: props.item[this.itemKey],
-        directives: [{
-          name: 'show',
-          value: this.expanded[props.item[this.itemKey]]
-        }]
-      }, [this.$scopedSlots.expand(props)])
+      const children = []
+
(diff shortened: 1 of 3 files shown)


src/components/VDataTable/mixins/body.js
```diff
@@ -16,25 +16,25 @@ export default {
...
-          value: this.expanded[props.item[this.itemKey]]
-        }]
-      }, [this.$scopedSlots.expand(props)])
+      const children = []
+
+      if (this.isExpanded(props.item)) {
+        const expand = this.$createElement('div', {
+          class: 'datatable__expand-content',
+          key: props.item[this.itemKey]
+        }, [this.$scopedSlots.expand(props)])
```
johnleider: I think this doesn't need to be wrapped in an array, could be wrong.

PR #1995, opened 2017-10-03
Fixed caret position on mask change during deletion.

azaars:
Fixed caret position on delete. The caret position stays/jumps to the correct position even if the deletion causes a dynamic switch of mask.

### DIFF
--- a/src/mixins/maskable.js
+++ b/src/mixins/maskable.js
@@ -87,15 +87,10 @@ export default {
       let selection = 0
 
       this.$refs.input.value = newValue
-      if (this.deleting) {
-        for (const char of newValue) {
-          if (this.lazySelection <= 0) break
-          isMaskDelimiter(char) || this.lazySelection--
-          selection++
-        }
-      } else {
-        selection = this.selection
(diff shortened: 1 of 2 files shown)


src/mixins/maskable.js
```diff
@@ -103,13 +114,7 @@ export default {
...
     // caret location is correct
     setSelectionRange () {
-      this.$nextTick(() => {
-        if (!this.$refs.input) return
-
-        this.$refs.input.value = this.maskText(this.lazyValue)
-
-        this.deleting ? this.setCaretPosition(this.selection) : this.updateRange()
-      })
+      this.$nextTick(() => this.updateRange())
```
johnleider: change to `this.$nextTick(this.updateRange)`

azaars:
Guys, I also did some refactoring. It turns out that the same code to handle character delete can also be used when inserting code, making some of the older codes unnecessary. In particular, `this.deleting` is no longer needed, which also means `keydown` event handling should go away. Also the `resetSelection()` groups and resets all needed caret pointers. The codes should look cleaner now.

PR #2005, opened 2017-10-04
adds `disable-hover` prop to disable mouseenter/mouseleave events for tooltip

### LINKED ISSUE #2002: [Feature request] Add possibility to disable showing/hiding tooltip on mouse out
### What will it allow you to do that you can't do today?

It will allow to hide the tooltip only on specific action, like clicking the button for example, otherwise it can be accidentally closed on mouseenter/leave

I'm not aware of any workaround for this (could possibly disable pointer events on the activator, but the activator can have its own hover behaviour which would be disabled as well)

### Example usage

This functionality could be used for example when introducing the application to the new user and showing features explicitely instead of making user to discover it by himself
Example codepen: https://codepen.io/anon/pen/dVZOGM - toggling tooltip on hover is not a desired behaviour

jacekkarczmarczyk:
fixes #2002
Not sure if good prop name

### DIFF
--- a/src/components/VTooltip/VTooltip.js
+++ b/src/components/VTooltip/VTooltip.js
@@ -20,6 +20,7 @@ export default {
       type: [Number, String],
       default: 0
     },
+    disableHover: Boolean,
     fixed: {
       type: Boolean,
       default: true
@@ -144,7 +145,7 @@ export default {
         }
       }, [tooltip]),
       h('span', {
-        on: {
+        on: this.disableHover ? {} : {
           mouseenter: () => {
             clearTimeout(this.leaveTimeout)
 


KaelWD:
I think this would probably be better if it was based on `value` instead of a separate prop. ie. current behaviour if `value == null`, and disable hover otherwise.

KaelWD:
PR: jacekkarczmarczyk/vuetify#1

Webifi:
Since tooltip is toggleable, its "value" is already used for its current toggle state, and it emits "input" on state change.  (v-model compatible.)  I think you'd want to use a different prop, like this PR does.

jacekkarczmarczyk:
Closing in favour of https://github.com/vuetifyjs/vuetify/pull/2164

PR #2006, opened 2017-10-04
renames active prop to inactive to make it false by default

jacekkarczmarczyk:
We don't like boolean props to be true by default

### DIFF
--- a/src/components/VList/VListTile.js
+++ b/src/components/VList/VListTile.js
@@ -18,11 +18,8 @@ export default {
       type: String,
       default: 'list__tile--active'
     },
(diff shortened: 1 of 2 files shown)


KaelWD:
Why not `disabled`?

jacekkarczmarczyk:
Because `disabled` disabled everything
`disableLink` is supposed only disable the hover events, so the tile looks like an item without click/href/to prop
https://codepen.io/anon/pen/veWmZm

KaelWD:
`inactive` then? I just really hate the name.

PR #2008, opened 2017-10-04
Make toggleable's prop a boolean

KaelWD:
This was changed in 1d4e6e9e7cb07b4bad76593a01e283ade8ffc418, but we couldn't think of a proper use case for this so it makes more sense to keep it strict

### DIFF
--- a/src/components/VSelect/mixins/select-generators.js
+++ b/src/components/VSelect/mixins/select-generators.js
@@ -33,7 +33,7 @@ export default {
           offsetOverflow: this.isAutocomplete,
           openOnClick: false,
           value: this.menuIsActive &&
-            this.computedItems.length &&
+            !!this.computedItems.length &&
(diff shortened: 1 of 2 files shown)


src/components/VSelect/mixins/select-generators.js
```diff
@@ -33,7 +33,7 @@ export default {
           offsetOverflow: this.isAutocomplete,
           openOnClick: false,
           value: this.menuIsActive &&
-            this.computedItems.length &&
+            !!this.computedItems.length &&
             (!this.tags || this.tags && this.filteredItems.length > 0),
```
jacekkarczmarczyk: `(!this.tags || this.filteredItems.length > 0)`

johnleider:
I'll have to find the conversation that lead to this change.

PR #2020, opened 2017-10-05
Update head.js to exit early if `hideHeaders` prop is set

CharlesOkwuagwu:


nekosaur:
`hide-headers` is not currently a prop of VDataTable. Or am I missing something?

PR #2042, opened 2017-10-06
fixes breadcrumbs styles for light/dark theme, adds global a:hover styles

### LINKED ISSUE #2041: Breadcrumbs disabled property broken in 16.1
The breadcrumbs disabled property correctly removes the link but still applies the link color (i.e. blue)

### Steps to reproduce
Just add the "disabled" property to a v-breadcrumbs

### Versions
Works correctly in 15.7
Bugged in 16.1

### What is expected ?
The disabled steps should be grey (and not look clickable to the user)

The disabled steps are still blue (and look clickable to the user)

Css classes

### Reproduction Link
https://codepen.io/tsauvajon/pen/wryMYG

(I fixed it by adding a css class)

jacekkarczmarczyk:
fixes #2041

### DIFF
--- a/src/stylus/elements/_global.styl
+++ b/src/stylus/elements/_global.styl
@@ -26,22 +26,6 @@ main
   transition: .3s padding $transition.swing
   will-change: padding-left
 
-.application--light a
-  color: $material-light.text.link
-
-  &:hover,
-  &:focus,
-  &:active
-    color: $material-light.text.link-hover
-
-.application--dark a
-  color: $material-dark.text.link
-
-  &:hover,
-  &:focus,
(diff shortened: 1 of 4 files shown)


src/stylus/elements/_global.styl
```diff
@@ -27,10 +27,20 @@ main
   will-change: padding-left
 
 .application--light a
-  color: $theme.primary
+  color: $material-light.text.link
```
johnleider: should the styles in this file for application go here? https://github.com/vuetifyjs/vuetify/blob/dev/src/stylus/components/_app.styl

PR #2049, opened 2017-10-06
fixes #2031

### LINKED ISSUE #2031: Input Masking - Cursor Issues on Android Oreo
### Steps to reproduce

1. Using Load [the Google Contacts example](https://vuetifyjs.com/examples/layouts/google-contacts) from documentation.
2. Click the FAB in bottom-right corner, to add a new contact.
3. In the phone field, type the number "1234567890"

### Versions

Vue 2.4.2
Vuetify 0.16.1
Android 8.0 / Chrome 59

### What is expected ?
A masked version of the number "1234567890".

Masking characters cause incorrect cursor position. If you don't manually adjust the cursor position in between number-clicks, your final/formatted number will be "(234) 789 - 0651".

### Reproduction Link
https://vuetifyjs.com/examples/layouts/google-contacts (on Android/Chrome)

screenshot_20171005-195022

azaars:
Android stores the caret position to later restore it after the component gets rendered. This exhibits a caret bug when we internally change the input text using masks. We need to delay our caret update right after Android is done with it's restoration.

### DIFF
--- a/src/mixins/maskable.js
+++ b/src/mixins/maskable.js
@@ -44,6 +44,9 @@ export default {
       const mask = preDefined || this.mask || ''
 
       return mask.split('')
+    },
+    isAndroid () {
+      return navigator && /android/i.test(navigator.userAgent)
     }
   },
 
@@ -78,7 +81,13 @@ export default {
   methods: {
     setCaretPosition (selection) {
       this.selection = selection
-      this.$refs.input.setSelectionRange(selection, selection)
+      if (this.isAndroid) {
+        window.setTimeout(() => {
+          this.$refs.input.setSelectionRange(selection, selection)
+        }, 1)
+      } else {
+        this.$refs.input.setSelectionRange(selection, selection)
+      }
     },
     updateRange () {
       if (!this.$refs.input) return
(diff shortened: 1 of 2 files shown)


src/mixins/maskable.js
```diff
@@ -78,7 +81,13 @@ export default {
   methods: {
     setCaretPosition (selection) {
       this.selection = selection
-      this.$refs.input.setSelectionRange(selection, selection)
+      if (this.isAndroid) {
```
johnleider: Add comments as to what this is doing

azaars: That would do.

src/mixins/maskable.js
```diff
@@ -82,6 +82,8 @@ export default {
     setCaretPosition (selection) {
       this.selection = selection
       if (this.isAndroid) {
+        // Android overrides the selection after nextTick().
+        // So defer the call to setSelectionRange().
```
KaelWD: Delay, not defer

azaars:
Hahahaaa…

 

Thanks for that dictation.

 

Now you know why I can’t do documentations.

KaelWD:
Yeah nah I was wrong, they mean basically the same thing.

KaelWD:
This doesn't work, the problem is caused by `resetSelections()` running before the cursor position has updated.

PR #2053, opened 2017-10-07
Adds the `hideHeaders` prop

CharlesOkwuagwu:


src/components/VDataTable/mixins/head.js
```diff
@@ -1,6 +1,8 @@
 export default {
   methods: {
     genTHead () {
+      if (this.hideHeaders === true) { return } // Exit Early since no headers are needed. Do we also need to set this.headers = [] ?
```
KaelWD: Don't ask questions like that in the code, put them in a message in the PR instead.

jacekkarczmarczyk: And since `hideHeaders` is bool, `=== true` can be removed

KaelWD: *And* you don't need the braces either, all they do is look ugly

CharlesOkwuagwu: All noted

KaelWD: And to answer the question, you can't set headers, as it is a prop.

CharlesOkwuagwu: OK. Got it

johnleider:
Add a PR on the docs for the new field

CharlesOkwuagwu:
Done. Do we need an example?

johnleider:
Yes

johnleider:
Will merge with 1.0 release

PR #2059, opened 2017-10-07
Fix cursor position in masked inputs on android

### LINKED ISSUE #2031: Input Masking - Cursor Issues on Android Oreo

KaelWD:
Android chrome selects the last entered character, so we should use
selectionEnd instead of selectionStart.
It also has a slight delay before deselecting the character, so a
timeout of 0 is set to ensure the correct update order.

Better than #2049
Fixes #2031

### DIFF
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -231,7 +231,7 @@ export default {
       this.$nextTick(() => this.$refs.input.focus())
     },
     resetSelections (input) {
-      this.selection = input.selectionStart
+      this.selection = input.selectionEnd
       this.lazySelection = 0
 
       for (const char of input.value.substr(0, this.selection)) {
--- a/src/mixins/maskable.js
+++ b/src/mixins/maskable.js
@@ -78,7 +78,9 @@ export default {
   methods: {
     setCaretPosition (selection) {
       this.selection = selection
-      this.$refs.input.setSelectionRange(selection, selection)
+      window.setTimeout(() => {
+        this.$refs.input.setSelectionRange(selection, selection)
+      }, 0)
     },
     updateRange () {
       if (!this.$refs.input) return


KaelWD:
I'd like if someone could test this fix on a slow device as there is a very slight flicker as the cursor updates. I'm wondering if this might cause additional issues if the user is typing faster than the phone can keep up with.

azaars:
Constraints:
- An old HTC Desire 620G dual sim
- Android 4.4.2 - yep I'm still on KitKat and that's the latest version the phone could live with :'(
- The phone connects to the web server on my laptop and I do a remote debugging.

Given all these, I tried `phone` masking and type 0 and 1 as quickly as I can using both hands. It didn't miss any character although the rendering is not smooth and fluid-like.

oct-07-2017 22-01-18

KaelWD:
Perfect. Did you test your PR like that as well? When I tried it on my phone the cursor would just get stuck at the start.

azaars:
It works equally well. Are you sure you set up your environment correctly?

KaelWD:
100% certain. Here's what it looks like at the first line of `resetSelections()`:
image

Note how the character is selected?
image

Then we get to maskable.js line 91, everything is fine so far
image

But `lazySelection` is `0`, so the cursor gets moved to the start
image

It only happens with the phone keyboard, if I type into the debug view it works like on desktop.

azaars:
Aah, so you're referring to the problem caused by `input.selectionStart`. I have no doubt from the beginning that we must instead use `input.selectionEnd`. I even used it in my very first pull request https://github.com/vuetifyjs/vuetify/blob/90bcf80033b7b922a527f0742d0c30f5b09c98fd/src/components/VTextField/VTextField.js#L92 It was there for such reason. Even if you do text (range) selection backwards or forward, the `selectionEnd` is what matters.

I thought we were talking about the test on the order of cursor updates. That was between:
Line 84-90 of my PR:
https://github.com/azaars/vuetify/blob/a44c5f35881d697ad29ff75fb3e2ea245fd6d7a8/src/mixins/maskable.js#L87

And line 81-83 of your PR:
https://github.com/KaelWD/vuetify/blob/e98a1d414ffebbdbe0eca7ef6c80fd11bddb1a04/src/mixins/maskable.js#L83

Either one would work. I may have misunderstood your request. Shoot.. perhaps my english is confusing :-/

PR #2084, opened 2017-10-08
Make VIcon aria-hidden by default, remove explicit aria-hidden in consumers, update snapshots.  Remove fa and mdi deprecated render logic.

Phlow2001:


johnleider:
merged this into next branch f0546ed2a305ef879b2fb8e0569d1e0c18e2dae5

PR #2099, opened 2017-10-09
Bug fixes on v-text-field

### LINKED ISSUE #2057: Masked inputs emit invalid characters
### Steps to reproduce
1. Type a letter into the input
2. Paste a multi-letter string into the input (`foobar`)
3. Paste a multi-character string starting with a number (`1twothree`)

### Versions
0.16.3

### What is expected ?
(1, 2) Nothing should happen, the value should be rejected and the model remain empty
(3) Only the valid character (`1`) should be accepted and emitted

1. The `input` event fires with the character that was typed, even if it doesn't match the mask
2. The input value doesn't change, but it emits the full string that was pasted in, up to the mask length
3. The input value is set to `1`, but the full string is still emitted as in (2)

### Reproduction Link
https://codepen.io/anon/pen/wryEBb

### LINKED ISSUE #2096: The <v-text-field return-masked-input> is not the same as the masked input field.
### Steps to reproduce

1. Set `mask="phone"` and `return-masked-input`
2. Type numbers
3. Type alphabets

### Versions
0.16.4

### What is expected ?
The returned `input` value must exactly be the same as what is displayed in the `input` field.

1. The returned value appears a few steps behind the `input` value as characters are progressively inserted.
2. The returned value contains illegal characters according to the mask rule.

### Reproduction Link
https://codepen.io/azaars/pen/XeEYpq

### LINKED ISSUE #2097: Illegal value inserted externally into <v-text-field> is not processed correctly.
### Steps to reproduce

1. Set `mask="phone"` and set to update `input` field externally with an illegal value.
2. Type a phone number.

### Versions
0.16.4

### What is expected ?
The value inserted externally should be masked and an input event should emit the correct (possibly modified) value.

No input event is emitted and the incorrect input value remains verbatim.

### Reproduction Link
https://codepen.io/azaars/pen/wrmEYw

### LINKED ISSUE #2136: <v-text-field> doesn't display the number 0.
### Steps to reproduce
1. Create a v-text-field
2. Bind it to a property and give it the value 0 (Number)

### Versions
0.16.4

### What is expected ?
The v-text-field should display the value '0'. Please notice that the string '0' is displayed correctly.

The v-text-field is left blank and the placeholder text also doesn't appear.

### Reproduction Link
https://codepen.io/bspg/pen/yzEXQR

### LINKED ISSUE #2137: <v-text-field> mask doesn't work when combined with .number modifier
### Steps to reproduce
1. Create a v-text-field with the mask "##.##" and set the .number modifier to its v-model.
2. Type '1234' on the text field.

### Versions
0.16.4

### What is expected ?
The text field should display '12.34'

The text field is not formatted and the console in devtools shows errors.

### Reproduction Link
https://codepen.io/bspg/pen/aLKybz

### LINKED ISSUE #2158: VTextField selectionEnd error
@KaelWD 

### Steps to reproduce
Put a type="number" inside v-text-field, like:
`<v-text-field  v-model="example" hide-details type="number"></v-text-field>`

When you change to fill other input, it appears an error: 

**Failed to read the 'selectionEnd' property from 'HTMLInputElement': The input element's type ('number') does not support selection**

### Versions

With android devices, with vue 2.4.4 and at least with vuetify 0.16.4

### What is expected ?
We expected that the content of input is not deleted.

When I want to fill in another entry, the previous entry is empty.

### Reproduction Link

 https://codepen.io/anon/pen/OxwONy

azaars:
1. Resolves #2057
2. Resolves #2096
3. Resolves #2097
4. Resolves #2136 
5. Resolves #2137 
6. Resolves #2158

### DIFF
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -84,12 +84,14 @@ export default {
     },
     inputValue: {
       get () {
-        return this.value
+        return this.lazyValue
       },
       set (val) {
-        this.lazyValue = val
-        this.mask && this.setSelectionRange()
-        this.$emit('input', val)
+        // Inner unmaskText() strips of non-alphanum
+        // maskText() masks and removes dirty alphanum
+        // Outer unmaskText() provides a clean lazyValue
+        this.lazyValue = this.unmaskText(this.maskText(this.unmaskText(val)))
+        this.setSelectionRange()
       }
     },
     isDirty () {
@@ -113,10 +115,15 @@ export default {
     },
     value (val) {
       // Value was changed externally, update lazy
-      this.lazyValue = val
+      const masked = this.maskText(this.unmaskText(val))
+      this.lazyValue = this.unmaskText(masked)
 
       !this.validateOnBlur && this.validate()
       this.shouldAutoGrow && this.calculateInputHeight()
+      this.$nextTick(() => {
+        this.$refs.input.value = masked
+        this.$emit('input', this.lazyValue)
+      })
     }
   },
 
@@ -142,7 +149,7 @@ export default {
     },
     onInput (e) {
       this.resetSelections(e.target)
-      this.inputValue = this.unmaskText(e.target.value)
+      this.inputValue = e.target.value
       this.badInput = e.target.validity && e.target.validity.badInput
       this.shouldAutoGrow && this.calculateInputHeight()
     },
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -90,7 +90,7 @@ export default {
         // Inner unmaskText() strips of non-alphanum
         // maskText() masks and removes dirty alphanum
         // Outer unmaskText() provides a clean lazyValue
-        this.lazyValue = this.unmaskText(this.maskText(this.unmaskText(val)))
+        this.lazyValue = val ? this.unmaskText(this.maskText(this.unmaskText(val))) : val
         this.setSelectionRange()
       }
     },
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -89,7 +89,7 @@ export default {
       set (val) {
         if (this.mask) {
           const value = this.unmaskText(this.maskText(this.unmaskText(val)))
-          this.lazyValue = this.getLazyValue(value)
+          this.lazyValue = typeof val === 'number' ? +value : value
           this.setSelectionRange()
         } else {
           this.lazyValue = val
@@ -136,7 +136,7 @@ export default {
       const masked = this.maskText(this.unmaskText(val))
       const value = this.unmaskText(masked)
 
-      this.lazyValue = this.getLazyValue(value)
+      this.lazyValue = typeof val === 'number' ? +value : value
       this.$nextTick(() => {
         this.$refs.input.value = masked
         this.$emit('input', this.lazyValue)
@@ -145,9 +145,6 @@ export default {
     updateOnValueChange (val) {
       this.lazyValue = val
       this.$emit('input', this.lazyValue)
-    },
-    getLazyValue (val) {
-      return typeof val === 'number' ? +val : val
     },
     calculateInputHeight () {
       this.inputHeight = null
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -90,7 +90,8 @@ export default {
         // Inner unmaskText() strips off non-alphanum
         // maskText() masks and removes dirty alphanum
         // Outer unmaskText() provides a clean lazyValue
-        this.lazyValue = this.unmaskText(this.maskText(this.unmaskText(val)))
+        const value = this.unmaskText(this.maskText(this.unmaskText(val)))
+        this.lazyValue = typeof val === 'number' ? +value : value
         this.mask ? this.setSelectionRange()
           : this.$emit('input', this.returnMaskedValue
             ? this.$refs.input.value : this.lazyValue)
@@ -118,8 +119,9 @@ export default {
     value (val) {
       // Value was changed externally, update lazy
       const masked = this.maskText(this.unmaskText(val))
+      const value = this.unmaskText(masked)
 
-      this.lazyValue = this.unmaskText(masked)
+      this.lazyValue = typeof val === 'number' ? +value : value
       !this.validateOnBlur && this.validate()
       this.shouldAutoGrow && this.calculateInputHeight()
 
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -87,14 +87,14 @@ export default {
         return this.lazyValue
       },
       set (val) {
-        // Inner unmaskText() strips off non-alphanum
-        // maskText() masks and removes dirty alphanum
-        // Outer unmaskText() provides a clean lazyValue
-        const value = this.unmaskText(this.maskText(this.unmaskText(val)))
-        this.lazyValue = typeof val === 'number' ? +value : value
-        this.mask ? this.setSelectionRange()
-          : this.$emit('input', this.returnMaskedValue
-            ? this.$refs.input.value : this.lazyValue)
+        if (this.mask) {
+          const value = this.unmaskText(this.maskText(this.unmaskText(val)))
+          this.lazyValue = this.getLazyValue(value)
+          this.setSelectionRange()
+        } else {
+          this.lazyValue = val
+          this.$emit('input', this.lazyValue)
+        }
       }
     },
     isDirty () {
@@ -118,17 +118,9 @@ export default {
     },
     value (val) {
       // Value was changed externally, update lazy
-      const masked = this.maskText(this.unmaskText(val))
-      const value = this.unmaskText(masked)
-
-      this.lazyValue = typeof val === 'number' ? +value : value
       !this.validateOnBlur && this.validate()
       this.shouldAutoGrow && this.calculateInputHeight()
-
-      this.$nextTick(() => {
-        this.$refs.input.value = masked
-        this.$emit('input', this.lazyValue)
-      })
+      this.mask ? this.maskedUpdateOnValueChange(val) : this.updateOnValueChange(val)
     }
   },
 
@@ -140,6 +132,23 @@ export default {
   },
 
   methods: {
+    maskedUpdateOnValueChange (val) {
+      const masked = this.maskText(this.unmaskText(val))
+      const value = this.unmaskText(masked)
+
+      this.lazyValue = this.getLazyValue(value)
+      this.$nextTick(() => {
+        this.$refs.input.value = masked
+        this.$emit('input', this.lazyValue)
+      })
+    },
+    updateOnValueChange (val) {
+      this.lazyValue = val
+      this.$emit('input', this.lazyValue)
+    },
+    getLazyValue (val) {
+      return typeof val === 'number' ? +val : val
+    },
     calculateInputHeight () {
       this.inputHeight = null
 
@@ -153,7 +162,7 @@ export default {
       })
     },
     onInput (e) {
-      this.resetSelections(e.target)
+      this.mask && this.resetSelections(e.target)
       this.inputValue = e.target.value
       this.badInput = e.target.validity && e.target.validity.badInput
       this.shouldAutoGrow && this.calculateInputHeight()
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -87,11 +87,13 @@ export default {
         return this.lazyValue
       },
       set (val) {
-        // Inner unmaskText() strips of non-alphanum
+        // Inner unmaskText() strips off non-alphanum
         // maskText() masks and removes dirty alphanum
         // Outer unmaskText() provides a clean lazyValue
-        this.lazyValue = val ? this.unmaskText(this.maskText(this.unmaskText(val))) : val
-        this.setSelectionRange()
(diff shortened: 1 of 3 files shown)


azaars:
Ups... It breaks some of the tests.

I think those tests are no longer applicable and should be replaced with something else.

KaelWD:
It breaks *one* of the tests. Have you checked that clearable still works after these changes?

azaars:
Ok, will do so.

src/components/VTextField/VTextField.js
```diff
@@ -84,12 +84,14 @@ export default {
...
       },
       set (val) {
-        this.lazyValue = val
-        this.mask && this.setSelectionRange()
-        this.$emit('input', val)
+        // Inner unmaskText() strips of non-alphanum
+        // maskText() masks and removes dirty alphanum
+        // Outer unmaskText() provides a clean lazyValue
+        this.lazyValue = this.unmaskText(this.maskText(this.unmaskText(val)))
+        this.setSelectionRange()
```
KaelWD: What about non-masked inputs? This will probably break text fields with `type="number"` as they don't support `setSelectionRange()`.

azaars: I did look at non-mask input. I even did some tests on first with a mask then later removed it on the fly.

Yeah you're probably right about `type="number"`. Well, it's not really related to the issues this PR is trying to resolve. There are several other issues that exist even before this PR, which are overlooked yet to surface sooner or later. IMHO lets handle these separately as different issues.

azaars:
Yeay the test passed :-)

src/util/mask.js
```diff
@@ -90,7 +90,7 @@ const maskValidates = (mask, char) => {
  * @return {String}
  */
 export const maskText = (text, masked, dontFillMaskBlanks) => {
-  if (!masked.length || !text.length) return text
+  if (!masked.length || !text || !text.length) return text
```
jacekkarczmarczyk: `if (!masked.length || !text) return text` should be enough

azaars: Really? I'm already too deep into strongly typed language hahaa.

Thanks @jacekkarczmarczyk

johnleider:
Tests need to be written for each issue.

azaars:
I'm sorry. I'm not familiar with the test tools at all. Will surely explore but for now it would take me ages to do that.

johnleider:
Sorry, response was mainly for myself to remember to do it :)

src/components/VTextField/VTextField.js
KaelWD: I mean `type="number"` in general. Currently `setSelectionRange` is only called if a mask is set:
```js
this.mask && this.setSelectionRange()
```

What happens with this patch if you just have a number input without a mask?

```html
<v-text-field v-model="number" type="number">
```

azaars: Now I see what you mean.

KaelWD: You can probably just put the `this.mask &&` back in.

azaars: Placing back merely the `this.mask &&` caused the `clearable` test to fail once again. This time it's expecting an `emit()`.

Admittedly, I did not give much attention to the totally valid `null` value for input field before. So now I've fixed some more lines here and there. For the same reason, the extra check needs to be there once again @jacekkarczmarczyk : 

```javascript
if (!masked.length || !text || !text.length) return text
```

src/mixins/maskable.js
```diff
@@ -60,8 +60,10 @@ export default {
       let position = 0
       let selection = this.selection
 
-      for (const char of oldText.substr(0, selection)) {
-        isMaskDelimiter(char) || position++
+      if (oldText) {
+        for (const char of oldText.substr(0, selection)) {
+          isMaskDelimiter(char) || position++
+        }
```
KaelWD: `oldText` will always be a string, this is redundant

azaars: Well, that's because I  expect `null` to be a valid value, and that `this.lazyValue` has to be able to accept null.

src/util/mask.js
```diff
@@ -90,14 +90,15 @@ const maskValidates = (mask, char) => {
  * @return {String}
  */
 export const maskText = (text, masked, dontFillMaskBlanks) => {
-  if (text == null) return ''
-  text = String(text)
-  if (!masked.length || !text.length) return text
+  if (!masked.length || !text) return text
```
azaars: I just found out that `!'' === true`, so `!text.length` not only is redundant but will fail when `text === null`. Still learning :blush:

src/mixins/maskable.js
KaelWD: `oldText` is directly from the html element, which is *always* a string, even if you set it to null first.

src/components/VTextField/VTextField.js
```diff
@@ -84,12 +84,17 @@ export default {
...
-        this.mask && this.setSelectionRange()
-        this.$emit('input', val)
+        // Inner unmaskText() strips off non-alphanum
+        // maskText() masks and removes dirty alphanum
+        // Outer unmaskText() provides a clean lazyValue
+        const value = this.unmaskText(this.maskText(this.unmaskText(val)))
+        this.lazyValue = typeof val === 'number' ? +value : value
+        this.mask ? this.setSelectionRange()
+          : this.$emit('input', this.returnMaskedValue
+            ? this.$refs.input.value : this.lazyValue)
```
KaelWD: Why are we returning different values based on `returnMaskedValue` when there is no mask?

src/util/mask.js
```diff
@@ -90,14 +90,15 @@ const maskValidates = (mask, char) => {
  * @return {String}
```
KaelWD: This docblock needs updating if you're returning arbitrary values. Does jsdoc have generics?

src/components/VTextField/VTextField.js
KaelWD: Also, this whole function should probably be split by an if statement so none of this casting is happening on non-masked inputs. The same applies for the value watcher too.

src/util/mask.js
azaars: You know better than I do.

azaars:
Please let me know if I miss anything else.

src/mixins/maskable.js
KaelWD: This is still redundant

src/util/mask.js
```diff
@@ -90,14 +90,15 @@ const maskValidates = (mask, char) => {
...
  */
 export const maskText = (text, masked, dontFillMaskBlanks) => {
-  if (text == null) return ''
-  text = String(text)
```
KaelWD: These lines were here for a reason. The value should still be cast to a string before calculating the cursor position, because that's what's going to be set on the input in the end.

src/components/VTextField/VTextField.js
```diff
@@ -128,6 +132,23 @@ export default {
...
+      this.$nextTick(() => {
+        this.$refs.input.value = masked
+        this.$emit('input', this.lazyValue)
+      })
+    },
+    updateOnValueChange (val) {
+      this.lazyValue = val
+      this.$emit('input', this.lazyValue)
+    },
+    getLazyValue (val) {
```
KaelWD: This name is bad, it doesn't accurately describe what the function is doing. Is this method even needed? You only use it in one place.

src/components/VTextField/VTextField.js
```diff
@@ -128,6 +132,23 @@ export default {
...
+
+      this.lazyValue = this.getLazyValue(value)
+      this.$nextTick(() => {
+        this.$refs.input.value = masked
+        this.$emit('input', this.lazyValue)
+      })
+    },
+    updateOnValueChange (val) {
+      this.lazyValue = val
+      this.$emit('input', this.lazyValue)
```
KaelWD: This is causing `input` to be emitted twice, as you're also doing it in both `inputValue.set()` and `updateRange()`

src/components/VTextField/VTextField.js
```diff
@@ -128,6 +132,23 @@ export default {
...
 
   methods: {
+    maskedUpdateOnValueChange (val) {
+      const masked = this.maskText(this.unmaskText(val))
+      const value = this.unmaskText(masked)
+
+      this.lazyValue = this.getLazyValue(value)
+      this.$nextTick(() => {
+        this.$refs.input.value = masked
+        this.$emit('input', this.lazyValue)
```
KaelWD: This one too.

src/mixins/maskable.js
azaars: Indeed it is. `input.value` never becomes null. Even if it's explicitly set to null, it turns to an empty string. At least in the latest Firefox. I'm changing it right away.

src/components/VTextField/VTextField.js
azaars: It's used in 2 places. The other one is at line 92 of `VTextField.js`. Do you prefer not to have this this function? I'll just put the content where it was before.

src/components/VTextField/VTextField.js
azaars: Well, actually I didn't quite get it when you said:

> Also, this whole function should probably be split by an if statement so none of this casting is happening on non-masked inputs. The same applies for the value watcher too.

I understood it as having if statement to call different functions. Maybe because I didn't understand what you meant about casting issue anywhere in that function. Frankly I also don't like the changes I made and in fact I don't like the new function names I gave. Perhaps you would want to elaborate your suggestion.

src/util/mask.js
KaelWD: I think accepting null is fine, but completely removing this means that the cursor position is wrong if the mask changes when the value is falsy (`0` with `v-model.number` for example`). You should be casting `newText` to string (with a *proper* null check) in `updateRange()` and `watch.mask()` before the loop. This will also allow the removal of the ifs, because `for..of` on an empty string just gets skipped.

src/components/VTextField/VTextField.js
azaars: On multiple emit problem, I realize that the `watch.value` has never been completed for what it's supposed to function as. From the comment, it should handle external change but there is nothing in the function to detect whether the change is external or internal. So, there have always been multiple emits from internally. I'm going to address this in a later PR.

src/components/VTextField/VTextField.js
KaelWD: What is it actually doing anyway? It looks like it just casts `val` to a number when it already is one.

src/components/VTextField/VTextField.js
KaelWD: I said that because `this.unmaskText(this.maskText(this.unmaskText(val)))` was being run on every change, even on non-masked inputs.

src/components/VTextField/VTextField.js
azaars: Actually it casts to the type that the dev enters in `props.value`

src/components/VTextField/VTextField.js
KaelWD: > there have always been multiple emits from internally

`<v-text-field :mask="####" v-model.number="num"/>`

Click inside, type `1`, blur. 

On dev:
image

On this branch:
image

src/components/VTextField/VTextField.js
KaelWD: It was [here](https://github.com/vuetifyjs/vuetify/pull/2099/commits/58524576273d9d9f203d8eb1e06d204cddca1d0c#diff-c2cbb2f11cf8c97e6eff0981f9927269L124), but now it does nothing.

azaars:
What exactly does codeclimate means in the last commit?

src/components/VTextField/VTextField.js
azaars: OMG what was I thinking when I did that.

You're damn sharp man. I salute

johnleider:
Don't worry about codeclimate for now

src/util/mask.js
azaars: I can see your point now. All codes that requires the return value of `maskText()` are expecting a string. So, converting the falsy to `'0'` is indeed correct. I'm reverting to your last commit.

KaelWD:
```js
if (text == null) return text
text = String(text)
```
Might work

azaars:
is it always returning string?

src/util/mask.js
azaars: Well, now I prefer your last commit. Like I said, all readers of `maskText()` including `newValue` and `domProps.value` are expecting string.

src/util/mask.js
jacekkarczmarczyk: http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters

azaars:
@KaelWD I've commited the post-review changes.

src/components/VTextField/VTextField.js
```diff
@@ -112,8 +119,25 @@ export default {
...
+
       // Value was changed externally, update lazy
-      this.lazyValue = val
+      if (this.mask) {
+        const masked = this.maskText(this.unmaskText(val))
+        const value = this.unmaskText(masked)
+        this.lazyValue = typeof val === 'number' ? +value : value
+
+        // Emit when the externally set value was modified internally
+        val !== this.lazyValue && this.$nextTick(() => {
```
KaelWD: why the nextTick here?

src/mixins/maskable.js
```diff
@@ -55,23 +55,28 @@ export default {
...
+      if (newValue) {
+        for (const char of newValue) {
+          isMaskDelimiter(char) || position--
+          selection++
+          if (position <= 0) break
+        }
       }
 
-      this.setCaretPosition(selection)
+      this.$nextTick(() => {
```
KaelWD: and here too I guess

PR #2116, opened 2017-10-09
Tooltip shouldn't close dependents

Phlow2001:
Reproduction:
https://codepen.io/jayblanchard/pen/WZJvRG?editors=1010

Hover over button, click to show menu, move mouse onto menu, tooltip goes away and takes menu with it.

I can't think of a scenario where a tooltip should be responsible for closing dependent elements.  So maybe a larger question is, should VTooltip actually implement Dependent mixin at all?


johnleider:
I think that the use-case that covers the reason for having dependents may be a thing, but would definitely be lower than this as far as functionality priority.

Issue #2092, opened 2017-10-09
dialog closes after pressing ENTER in autocomplete


jacekkarczmarczyk:
### Steps to reproduce
Open the codepen https://codepen.io/anon/pen/eGMKyr, click the "Open dialog" button, click the select, type "x", press ENTER

### Versions
0.16.4

### What is expected ?
Dialog should not close

### What is actually happening ?
Dialog closes

### Reproduction Link
https://codepen.io/anon/pen/eGMKyr

Last working version is 0.16.1: https://codepen.io/anon/pen/dVmKLP

barnabas-avalara:
For what it's worth, the behavior does not happen when the dialog is persistent, as demonstrated here: https://codepen.io/anon/pen/mBamLw

jacekkarczmarczyk:
Probably the same issue: https://codepen.io/anon/pen/NaeOdw?editors=1010

PR #2117, opened 2017-10-10
dont close dialog after pressing enter key

spezzino:
solves #2092


KaelWD:
It's actually even worse than that, it prevents the searchValue from clearing when the dialog closes, so the old value is still present. Submit another PR if you manage to fix this issue properly.

PR #2149, opened 2017-10-12
Partially fix #2133: Double Jeopardy, one input should be tested invalid against only one rule

### LINKED ISSUE #2133: [bug]VTextInput validation displays all the error messages
### Steps to reproduce

Goto https://vuetifyjs.com/components/text-fields, scroll to #11 Validation, click the input for email, click other item in the page, you'll get:
image

### Versions

v0.16.4

### What is expected ?

For this case, only the message for required is needed, as the behavior for earlier versions such as v0.15.7.

It's too crowded to show all the error messages. 

### Reproduction Link

https://vuetifyjs.com/components/text-fields
scroll to #11 Validation

KingMario:
One input should be tested invalid against only one rule.
If a rule tells that the input is invalid, other rules should be ignored for testing.

### DIFF
--- a/src/mixins/validatable.js
+++ b/src/mixins/validatable.js
@@ -104,7 +104,7 @@ export default {
 
       this.errorBucket = []
 
-      this.rules.forEach(rule => {
+      return this.valid = this.rules.every(rule => {
         const valid = typeof rule === 'function' ? rule(value) : rule
 
         if (valid !== true && !['string', 'boolean'].includes(typeof valid)) {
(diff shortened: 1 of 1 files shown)


johnleider:
Why do you say it only partially fixes the issue? It appears that it completely resolves it.

KingMario:
@johnleider
Since the computed validations has other sources for error messages, according to [a8fd957](https://github.com/vuetifyjs/vuetify/commit/a8fd957aef113a597c45cd09852d47764c35789d), there could still multi-line error messages.

johnleider:
I'm just going to revert the change I initially made that caused this in the first place.

KingMario:
@johnleider alright, and I guess this fix will improve the performance in some extent.

KingMario:
@johnleider sir, I don't know whether you can reopen a closed pr here, or shall I start another pr for discussion?

If you can't reopen a closed pr, and think the insignificant performance improvement of this pr is valuable, I'll start another pr from the same branch.

Thanks.

PR #1932, opened 2017-09-28
Add z-index stacking order to vDialog, menuable and associated overlays

Webifi:
Sets z-index on dialogs, associated overlays and menuable elements allowing multiple open dialogs to be stacked based on when the dialog was last set active, not when it was added to the dom.

Also stops propagation of click event on overlay if overlay is used by dialog, stopping the dismissal of all open dialogs if the overlay for one dialog is clicked.

Removes need for #1918 
Fixes #1901 
Fixes #2009 
Fixes #2010 
Fixes #2028

Here's an example of using multiple dialogs and menus **with this PR** (using a build of 0.16.0 with this patch applied):
https://codepen.io/webifi/pen/GMMZOd

And, for comparison, here's using the same markup as above, but **without this PR**:
https://codepen.io/webifi/pen/jGaKjo

PR #2150, opened 2017-10-12
Fix/2146

### LINKED ISSUE #2146: Dialog does not open in Edge
### Steps to reproduce
- Visit https://vuetifyjs.com/components/dialogs
- Clock any "Open Dialog" button

### Versions
Vuetify 0.16, Windows 10, build 1703, Edge 40.15063.674.0

### What is expected ?
dialog should render 
(in my memory this feature worked on edge in the previous release)

nothing visible changes. 
If console is open the message is "Object doesn't support property or method 'Symbol(Symbol.iterator)'"

### Reproduction Link
https://vuetifyjs.com/components/dialogs

johnleider:
Fixes #2146

Handles edge case where the Edge browser is unable to iterate an empty HTMLElementList, causing a fatal Symbol.iterator.

Also fixed a bug where the #1932 PR removed the logic that bound click events to `v-app`. This caused an issue on iOS devices not firing **click-outside** events.

### DIFF
--- a/src/directives/click-outside.js
+++ b/src/directives/click-outside.js
@@ -54,14 +54,18 @@ export default {
   bind (el, binding, v) {
     v.context.$vuetify.load(() => {
       const onClick = e => directive(e, el, binding, v)
-      // Use capturing event so we avoid being blocked by 
-      // stopPropagation in bubbling phase
-      document.addEventListener('click', onClick, true)
+      // iOS does not recognize click events on document
+      // or body, this is the entire purpose of the v-app
+      // component and [data-app], stop removing this
+      const app = document.querySelector('[data-app]') ||
+        document.body // This is only for unit tests
+      app.addEventListener('click', onClick, true)
       el._clickOutside = onClick
     })
   },
 
   unbind (el) {
-    document.removeEventListener('click', el._clickOutside, true)
+    const app = document.querySelector('[data-app]')
+    app && app.removeEventListener('click', el._clickOutside, true)
   }
 }
--- a/src/directives/click-outside.js
+++ b/src/directives/click-outside.js
@@ -65,7 +65,8 @@ export default {
   },
 
   unbind (el) {
-    const app = document.querySelector('[data-app]')
+    const app = document.querySelector('[data-app]') ||
+      document.body // This is only for unit tests
     app && app.removeEventListener('click', el._clickOutside, true)
   }
 }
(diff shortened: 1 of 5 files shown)


KaelWD:
Do you have a Playground.vue that can repro the issue? The codepen fails with `"[Vue warn]: Error in render function: 'TypeError: Object doesn't support property or method 'ToString''` which is different to what the issue states, but when I copy the code into my local environment it runs just fine.

src/mixins/stackable.js
```diff
@@ -13,28 +13,33 @@ export default {
...
     getMaxZIndex (exclude = []) {
       const base = this.stackBase || this.$el
-      // start with lowest allowed z-index or z-index of base component's element, whichever is greater
+      // start with lowest allowed z-index or z-index of
+      // base component's element, whichever is greater
       const zis = [this.stackMinZIndex, getZIndex(base)]
       // get z-index for all active dialogs
-      const activeElements = document.getElementsByClassName(this.stackClass)
-      for (const activeElement of activeElements) {
+      const activeElements = [...document.getElementsByClassName(this.stackClass)]
```
KaelWD: This line is what actually fixes the issue, you can still use `for..of` after doing this.

src/stylus/components/_overlay.styl
```diff
@@ -7,6 +7,12 @@
   right: 0
   bottom: 0
   pointer-events: none
+  // The overlay has a dynamically set
+  // z-index, we want the transition
+  // timing to affect its changing
+  // https://github.com/vuetifyjs/vuetify/issues/2146
+  transition: .5s $transition.swing
```
jacekkarczmarczyk: 0.5s here and 150ms below is calculated by trial and error method or are they related to some other props? Maybe they could be made as a consts and resused

johnleider: Trial and error

src/directives/click-outside.js
```diff
@@ -54,14 +54,18 @@ export default {
...
+      const app = document.querySelector('[data-app]') ||
+        document.body // This is only for unit tests
+      app.addEventListener('click', onClick, true)
       el._clickOutside = onClick
     })
   },
 
   unbind (el) {
-    document.removeEventListener('click', el._clickOutside, true)
+    const app = document.querySelector('[data-app]')
```
nekosaur: Should probably do `|| document.body` here as well, so things get removed correctly in tests.

PR #2153, opened 2017-10-12
Option to set v-data-table header width

### LINKED ISSUE #1856: [Enhancement Request] Table header/field with custom width
I'm here asking for a small feature that would help a lot on better information display and visual match/fidelity for data tables.

Change the header size to a certain percentage or pixel size. Simply by adding a attribute to the headers object, like width: '50%' or '50px'.

For example the first column/field of the table should occupy 50% of the table total width, while the other 2 columns/fields should occupy 25% each. Today in the presented data tables component, it splits the header width by the number of columns and you have to "hard" code every <td> to match the size that you want.

It's a simple feature that could add bring a better flexibility to the data tables component and make the code cleaner.

rafaelrenanpacheco:
This PR will allow to set the header width in % or px, for example:

```
headers: [
  { text: 'Dessert (100g serving)', value: 'name', width: '20%'},
  { text: 'Calories', value: 'calories', width: '50%' },
  { text: 'Fat (g)', value: 'fat' },
  { text: 'Carbs (g)', value: 'carbs', width: '200px' },
],
```

Fixes #1856

### DIFF
--- a/src/components/VDataTable/mixins/head.js
+++ b/src/components/VDataTable/mixins/head.js
@@ -49,6 +49,7 @@ export default {
         attrs: {
           role: 'columnheader',
           scope: 'col',
+          width: header.width || null,
           'aria-label': header[this.headerText] || '',
           'aria-sort': 'none'
         }


algil:
Hi, How we can use this? I've added width attribute to headers configuration but nothing happens. Thanks!

KaelWD:
You can't yet, it's still just on the dev branch.

algil:
Ahh ok, I thought was launched in version 0.16.6.

In what version will it be released?

leon-sleepinglion:
This feature will be very helpful, thanks for the changes.

ckitt:
I cannot set the column width with the setting above without setting the `table-layout` property for `table.table` to fixed, but then it requires we set the column width statically for all column

References: https://www.w3schools.com/cssref/pr_tab_table-layout.asp

WhatFreshHellIsThis:
Width still doesn't seem to work in late 2018?!

nekosaur:
It does (see https://codepen.io/anon/pen/ZmxXqK?editors=1010). But there are some scenarios where it breaks (see https://github.com/vuetifyjs/vuetify/issues/5072 for example).

Please refrain from commenting on old and closed issues/PRs. If you have a specific problem, make sure there's not already an issue and then create a new one using the issue creator.

PR #2162, opened 2017-10-13
fix: Fixed select blur in Edge

### LINKED ISSUE #2152: Autocomplete selects need three clicks to blur in edge
### Steps to reproduce
 - Open the page in edge
 - Click in the select to open it
 - Click outside the select

### Versions
Current dev, doesn't work in codepen due to #2146 
Edge 14.14393
Works in chrome 61

### What is expected ?
The select menu should close and it should no longer be focused

The menu closes but the select is still focused
Subsequent clicks will initiate the label animation. but it doesn't blur unless you click again while the animation is still running. 
Once you've double-clicked, the select will be blurred, but the internal text input is still active, requiring a third click to blur.

### Reproduction Link
https://codepen.io/anon/pen/rGrOwr?editors=1011

johnleider:
Fixed a bug where when trying to blur a `v-select` component it would immediately refocus when using
the Edge browser

fixes #2152

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -181,7 +181,7 @@ export default {
      * @return {Number}
      */
     currentRange () {
-      return (this.selectedItem || '').length
+      return this.getText(this.selectedItem || '').length
     },
     filteredItems () {
       // If we are not actively filtering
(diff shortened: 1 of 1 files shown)


src/components/VSelect/VSelect.js
```diff
@@ -175,6 +175,14 @@ export default {
     computedItems () {
       return this.filterDuplicates(this.cachedItems.concat(this.items))
     },
+    /**
+     * The range of the current input text
+     *
+     * @return {Number}
+     */
+    currentRange () {
+      return (this.selectedItem || '').length
```
KaelWD: This is pretty much always `undefined` if `items` are objects. It should use `getText()` to account for this.

PR #2172, opened 2017-10-13
fix: Hide text field in blurred autocomplete v-selects

### LINKED ISSUE #1691: v-select display extra blank line when place in v-flex
### Steps to reproduce

Add `v-select` inside `v-flex` and use data that contain text long enough, after selected value and blur, `v-select` will add some black space below selected value.

### Versions
- Vuetify: 0.15.5
- Vue.js: 2.4.2
- Firefox/Chrome: latest

### What is expected ?

Text must display in single line without blank line below text.

(It would be good if have an option force to display in single line even the text is longer than `v-select` box.)

Text value inside `v-select` has extra space below, if text is too long.

### Reproduction Link

https://codepen.io/mementototem/pen/rGBbNL

KaelWD:
The extra padding on input-group__selections__comma is to prevent a jump
in line height when the element is focused

 - Also scroll to the end when focusing autocomplete selects that have
   overflow

Fixes #1691

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -274,6 +274,7 @@ export default {
 
         requestAnimationFrame(() => {
           this.$refs.input.setSelectionRange(len, len)
+          this.shouldBreak && (this.$refs.input.scrollLeft = this.$refs.input.scrollWidth)
         })
       }
     },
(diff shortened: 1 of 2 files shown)


johnleider:
Code looks good, will have to physically test later. However, I tried this approach before and ran into issues, it's possible that those limitations are no longer present.

PR #2188, opened 2017-10-15
feat: add fixed/absolute props to v-system-bar

### LINKED ISSUE #2186: System bars is always fixed on top
System bars is now always locate fixed on the top of the windows which is not the correct behavior.

Expected, I can be placed anywhere. Ex. if it is wrap in v-card, it should be inside the v-card instead of being on the top.

Reproduction can be found in the code-pen example:
https://codepen.io/Ewocker/pen/WZaGvO?&editors=100
Each is overlap on top of another.

I found this when upgrading from Vuetify@0.15.x to Vuetify@0.16.6

jacekkarczmarczyk:
Additionaly fixes the bug when applicationable was not updated after
changing some properties
Removes unnecessary styles used in the previous version of layout

fixes #2186

Following styles can be removed from docs after merging:
```css
#toolbar-view .system-bar {
    position: relative;
    z-index: auto;
}
```


johnleider:
If we are doing this, we should probably mimic the position behavior of other similar components and add an absolute prop.

Issue #1902, opened 2017-09-26
Issue with stylus block level import


leRenart:
Because Vuetify is using some generic classes like `btn` or `title` is it very hard to use it in an existing environment. However, stylus has the option to block-level imports, that could be very handy to isolate classes into the vue App scope to not collide with existing ones. 

Import `vuetify/src/stylus/app` into a css selector doesn't work as expected

### Steps to reproduce
compile with webpack

### Versions
0.15 à la carte

### What is expected ?
To compile the whole css into a parent selector

### What is actually happening ?
without block-level import it works as expected.

    <style lang="stylus">
        @require '~../node_modules/vuetify/src/stylus/app'
    </style>

with block-level, got a compilation error

    <style lang="stylus">
        #app
          @require '~../node_modules/vuetify/src/stylus/app'
    </style>

` ERROR  Failed to compile with 1 errors                                                                                                                16:14:33

 error  in ./src/App.vue

Module build failed: Error: /Users/xxxx/Workspace/webpack/src/App.vue:18:11
   14| 
   15| #app
   16|   @require '~../node_modules/vuetify/src/stylus/app'
   17| 

Failed to @extend "code"

    at /Users/xxxx/Workspace/webpack/node_modules/stylus/lib/visitor/normalizer.js:406:17
    at Array.forEach (<anonymous>)
    at Normalizer.extend (/Users/xxxx/Workspace/webpack/node_modules/stylus/lib/visitor/normalizer.js:402:17)
    ...`

Zeph33:
Hi
I use vuetify class for use Vuetify in my app 
`<div class='vuetify'> VApp .... </div>`
This i my stylus file :
[vuetify.txt](https://github.com/vuetifyjs/vuetify/files/1337663/vuetify.txt)

Before use this file, edit 
\node_modules\vuetify\src\stylus\elements\_code.styl line 18 and add .vuetify
`  @extend .vuetify code`
\node_modules\vuetify\src\stylus\components\_carousel.styl line 39 and add .vuetify
`     @extend .vuetify .carousel__left`

johnleider:
I am unable to reproduce this issue. Can you provide a set of reproduction steps/minimal reproduction in a repo or something please.

Zeph33:
you can not wrap all styles in a class or id
``` stylus
.myclass
    @import '../node_modules/vuetify/src/stylus/main'
```
or

``` stylus
#myid
    @import '../node_modules/vuetify/src/stylus/main'
[...]

PR #2192, opened 2017-10-15
fix: Remove stylus @extend for stylus block level import (#1902)

Zeph33:


src/stylus/elements/_code.styl
```diff
@@ -13,8 +13,7 @@ code
   &:before
```
jacekkarczmarczyk: these styles (`:after`, `:before`) should not be applied to `kbd`

src/stylus/elements/_code.styl
```diff
@@ -1,4 +1,4 @@
-code
+code, kbd
   background: $grey.base
```
jacekkarczmarczyk: make a separate rule for `background` and `color` for `code` instead of sharing them for `code, kbd` and then overwriting for `kbd`

src/stylus/components/_carousel.styl
```diff
@@ -38,7 +38,6 @@
         background: none
 
   &__right
-    @extend .carousel__left
     left: auto
```
jacekkarczmarczyk:
```styl
&__left, &__right
  ...

&__left
  left: 5px

&__right
  right: 5px
```

Zeph33:
OK, I do not know if style `:after`  and `:before` should be applied to `kdb`.
But this produce the same code see [https://unpkg.com/vuetify@0.16.7/dist/vuetify.css](https://unpkg.com/vuetify@0.16.7/dist/vuetify.css).
It is the role of `@extend`

src/stylus/elements/_code.styl
Zeph33: OK, I do not know if style `:after`  and `:before` should be applied to `kdb`.
But this produce the same code see [https://unpkg.com/vuetify@0.16.7/dist/vuetify.css](https://unpkg.com/vuetify@0.16.7/dist/vuetify.css).
It is the role of `@extend`

jacekkarczmarczyk: You're right, sorry, my bad

PR #2193, opened 2017-10-16
Rippleable (VCheckbox, VRadio, VSwitch) ripple prop with tests

Phlow2001:
VSwitch has to render ripple always, currently, due to the switch knob being part of the ripple styles.

### DIFF
--- a/src/components/VCheckbox/VCheckbox.js
+++ b/src/components/VCheckbox/VCheckbox.js
@@ -73,7 +73,7 @@ export default {
           'icon--checkbox': this.icon === 'check_box'
         },
         key: this.icon,
-        on: Object.assign({}, {
+        on: Object.assign({
           click: this.toggle
         }, this.$listeners)
       }, this.icon)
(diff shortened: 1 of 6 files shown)


src/components/VCheckbox/VCheckbox.js
```diff
@@ -72,7 +72,10 @@ export default {
         'class': {
           'icon--checkbox': this.icon === 'check_box'
         },
-        key: this.icon
+        key: this.icon,
+        on: Object.assign({}, {
```
KaelWD: The empty object isn't needed, it's just a waste of memory.

src/mixins/rippleable.js
```diff
@@ -4,16 +4,23 @@ import Ripple from '../directives/ripple'
 export default {
   directives: { Ripple },
 
+  props: {
+    ripple: {
+      type: [Boolean, Object],
```
KaelWD: Why is this an object too?

Phlow2001: This was copied from VBtn.  I don't know why it was done there, as it's different from Routeable's ripple prop.  I chose to stay consistent with VBtn.  I could go either way.

PR #2203, opened 2017-10-16
feat: (WIP) dynamic themes

### LINKED ISSUE #2155: [Feature] Allow redefinition of `$colors` object
I am creating my own theme, following the guide provided in "[Style-Theme](https://vuetifyjs.com/style/theme)" section. 

As I am already tailoring my CSS package, I would like to reduce the number of colour palettes, as I don't need it all and removing the unneeded ones can reduce my CSS size. Also, I would like to add my own palettes, and use it like the built-in ones, like `class="mybrandname lighten-2"`.

However, in the file [src/stylus/settings/_colors.styl](https://github.com/vuetifyjs/vuetify/blob/dev/src/stylus/settings/_colors.styl) you guys use the overriding attribution operator from stylus, so I can't easily update the colour palettes.

Changing line [318](https://github.com/vuetifyjs/vuetify/blob/dev/src/stylus/settings/_colors.styl#L318) to `$colors := {` would allow that to be done, without any prejudice to the current setup.

johnleider:
fixes #2155

Converted components
- [x] VBtn
- [ ] VBadge
- [ ] VBreadcrumbs
- [x] VChip
- [ ] VDatePicker
- [ ] VTimePicker


src/stylus/settings/_theme.styl
```diff
@@ -73,7 +73,7 @@ $material-light := {
   picker: {
     body: #fff,
     clock: $grey.lighten-2,
-    title: $theme.primary
+    title: 'inherit'
```
jacekkarczmarczyk: This commit breaks the default picker colors, here's the playground for test, compare with dev branch:
```vue
<template>
  <v-app id="inspire" :dark="dark">
    <v-content>
      <v-checkbox v-model="dark" label="Dark"></v-checkbox>
      <v-layout row><v-date-picker v-for="i in 3" style="width: 290px" :light="i === 2" :dark="i === 3"></v-date-picker></v-layout>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data: () => ({ dark: false })
}
</script>
```
The main problem is that dark and light themes use different colors for picker title (primary for light theme, gray for dark theme), otherwise it could be easily fixed in `picker.js` by changing line 32 from `return this.headerColor || this.color` to for example `return this.headerColor || this.color || 'primary'`

Moreover after removing `background: $material.picker.title` from `_pickers.styl` the `$material-*.picker.title` variable is not used any longer and can be removed

src/components/VAlert/__snapshots__/VAlert.spec.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`VAlert.vue should be closed by default 1`] = `
 
-<div class="alert error--light"
+<div class="alert undefined"
```
KaelWD: Failing snapshots should not be updated unless it's caused by an intentional change.

johnleider: Was a mistake.

PR #2204, opened 2017-10-16
Fix for deep nested v-layout's being affected by grid list styling

Phlow2001:
Example of problem that this solves:  https://codepen.io/jayblanchard/pen/LzXEQw?editors=1000

### DIFF
--- a/src/stylus/components/_grid.styl
+++ b/src/stylus/components/_grid.styl
@@ -30,7 +30,7 @@
         padding: $gutter
 
         > .layout
-          .flex
+          > .flex
             padding: ($gutter / 2)
 
         > .layout:only-child
--- a/src/stylus/components/_grid.styl
+++ b/src/stylus/components/_grid.styl
@@ -15,11 +15,11 @@
 
   &.fluid
     max-width: 100%
-    
+
   &.fill-height
     align-items: center
     display: flex
-    
+
     .layout
       height: 100%
       flex: 1 1 auto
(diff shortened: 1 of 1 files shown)


src/stylus/components/_grid.styl
```diff
@@ -29,17 +29,17 @@
       &-{$size}
         padding: $gutter
 
-        .layout
+        > .layout
           .flex
```
KaelWD: I'd probably add one here as well.

johnleider:
This is how it was in the past, I removed it but did not leave a comment, I'll need to dig it back up.

johnleider:
This breaks the ability to do nested grids. https://codepen.io/johnjleider/pen/veMKEL?&editors=101 as an example.

Phlow2001:
It's a grid-list-xl, there's no gutters.

johnleider:
I'm referring to the example codepen that I posted, which is a common functionality that we need to support.

Phlow2001:
Sorry, to be clear, my response "It's a grid-list-xl, there's no gutters." is to your statement "I am not entirely sure what the issue is with the codepen you provided.".

We've talked about it before, I just wanted to record it in the issue in case you forgot.

dohomi:
@Phlow2001 @johnleider I just bumped into this issue as well and fixed it with this PR. Alternative users can set `.ma-0` to the nesting layout as follow:
https://codepen.io/anon/pen/rYyBZN?editors=1000

johnleider:
@dohomi's work around will be the preferred method for this case, as implementing this PR breaks other use-cases that we already support. Thank you for your contribution.

Issue #989, opened 2017-07-14
Feature request: datepicker event indicator


Christilut:
I think it might be a nice addition if you can add an array of dates to the datepicker that will show a small indicator on those dates.
Let's say there is something going on next week and the user can clearly see there is an event because that day has a special marker.

A picture speaks a thousand words, so.. 
![capture](https://user-images.githubusercontent.com/3471363/28207596-8ae3061e-688a-11e7-8692-9001bfd8858a.PNG)

Alternatively: maybe a slot for a day number so total customization is possible

jacekkarczmarczyk:
With https://github.com/vuetifyjs/vuetify/pull/2109 it will be possible to mark events using the following code:

```vue
<template>
  <v-app id="inspire" :dark="dark">
    <v-content>
      <div style="width: 290px">
        <v-date-picker :day-format="d"></v-date-picker>
      </div>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data() {
    return {
      d: date => {
        return '<span class="' + (date.substr(9, 1) === '1' ? 'event' : '') + '">' + date.substr(8, 2).replace(/^0/, '') + '</span>'
      }
    }
  }
}
</script>

<style>
.event {
  position: relative;
}

.event:after {
  display: block;
  content: "";
  width: 6px;
  height: 6px;
  background: blue;
  border-radius: 50%;
  position: absolute;
  bottom: -3px;
  left: calc(50% - 3px);
}
</style>
```

![image](https://user-images.githubusercontent.com/15625235/31582435-b23ec8b8-b1ac-11e7-9067-355a60b8d097.png)

It doesn't mean that this is the final solution though

mknnaren:
Hey @jacekkarczmarczyk , I tried your temp. solution but its not working. Can you please share the version nos. of Vuetify and Vue you used for this. 

Thank you

jacekkarczmarczyk:
0.17 (not released yet)

PR #2211, opened 2017-10-17
vDatePicker: property 'locale-days' which turns on the day numbers localisation

egudova:


jacekkarczmarczyk:
Do not include dist files in the commit

egudova:
I found your answer for #989 issue. So, this request seems to be no longer actual. I will be able to display days as unformatted numbers after 0.17 is released.

jacekkarczmarczyk:
Just out of curiosity - by default picker uses 'en' locale to display days, so it just displays numbers same like you display `day` var in this PR. My question is in what situations you want to change locale but still display days in english locale instead of the selected one?

egudova:
In my project I use ja locale. And days are displayed like "1日". It looks not pretty and also a little strange, so my Japanese colleagues don't like it.

jacekkarczmarczyk:
Ok, thanks for clarification

PR #2217, opened 2017-10-17
Allow flat/raised style options for v-expansion-panel

njouanin:
Like some other components (like v-card) v-expansion-panel should have flat/raised style options to modify the component elevation.

### DIFF
--- a/src/components/VExpansionPanel/VExpansionPanel.js
+++ b/src/components/VExpansionPanel/VExpansionPanel.js
@@ -18,7 +18,9 @@ export default {
     expand: Boolean,
     focusable: Boolean,
     inset: Boolean,
-    popout: Boolean
+    popout: Boolean,
+    raised: Boolean,
+    flat: Boolean
   },
 
   methods: {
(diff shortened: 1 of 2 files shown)


johnleider:
Please provide an example implementation of the markup and possibly an example on codepen showing what this would look like. Thank you for your contribution!

njouanin:
Sorry, I'm not really used to copdepen. How to create a codepen which depends on an unreleased version of vuetify ?
Here is the test codepen : https://codepen.io/anon/pen/aLPRdx. With this PR, it should deplay 3 expansion header: one "normal", one raised, and flat.

johnleider:
For `v-card` it was done out of convenience because the card should, by default, support a raised and flat style. By spec, the `v-expansion-panel` does not support an alternate flat style. However, this can be easily accomplished with the provided **elevation-** helper classes.

https://codepen.io/johnjleider/pen/qPvwbj?editors=1010

njouanin:
OK, thanks for the elevation tip. I thought raised/flat could be a general feature.
The original need was to have an expansion-list with more than one drawer opened. It seems to be not supported, so for now I used an expansion panel with the expansion feature.

PR #2235, opened 2017-10-18
Enhancement: Numeral Mask + External Lib Formatter For v-text-field

### LINKED ISSUE #2138: [Enhancement] v-text-field with decimal amount
Displaying currency is often in a format of `3.50` but as a number it is displayed as `3.5`. 

An option to choose the amount of decimals in a v-text-field that has `type="number"` would be nice here.

Property: `decimals` as number

If you choose 3, then it would always display contents of the text field as `3.500`.

Perhaps it will display `3.5` while typing and `3.500` after blur. I imagine this would make the implementation alot simpler.

azaars:
# Enhancement: Numeral Mask + External Lib Formatter For v-text-field

This PR is an attempt to enhance `v-text-field` component so that it can provide the following missing but highly desirable masking/formatting features:

1. General numeral mask
2. Currency formatter
3. External library plugin

### Design Approach

The existing `mask` property allows developers to specify 2 data types:

1. String
2. Object

The existing mask features work on `mask` of type `String`. This PR extends the new features by leveraging on both types. As such it allows `v-text-field` component to handle the following mask cases:

No. | Case | Type | New | Example
------|------|------|------|------
1.| Predefined | String || `mask="credit-card"`
2.| Predefined with default options | String | :heavy_check_mark: | `mask="numeral"`
3.| Predefined with overridden options | Object | :heavy_check_mark: | `:mask="{ formatter: 'numeral', prefix: '$', precision: 4 }"`
4.| Custom mask | String || `mask="(##) - (AA)"`
5.| Custom mask with external formatter | Object | :heavy_check_mark: | `:mask="{ formatter: function () {}, myOption1: 23, myOption2: '%' }"`

#### Numeral API

Option | Type | Default | Description
-----|-----|-----|-----
blockSeparator | Number/Array | space | Separator between groups of digits
blockSize | Number/Array | 3 | Number of digits per group
decimal | String | . | Decimal mark
formatter | String | | Must specify "numeral"
precision | Number | 2 | Number precision
prefix | String || Prefix inside the input field
positiveOnly | Boolean | false | Non-negative number
positiveSign | Boolean | false | Whether to show '+' on positive number
suffix | String || Suffix inside the input field

##### Example of Numeral Mask Options

``` javascript
mask: {
  positiveOnly: false,        // Positive value only
  positiveSign: false,        // Ex. "+99.00"
  blockSeparator: [',', ' '], // Ex. "9,999,999 99"
  blockSize: [3, 2, 4]        // Ex. "999 999 99 9999"
  decimal: '.',               // Ex. "9 999.00"
  prefix: 'MYR$ ',            // Ex. "MYR$ 9.00"
  precision: 2,               // Ex. "999.00"
  suffix: ' cent',            // Ex. "999.00 cent"
  formatter: 'numeral'
}
```

>For options of type Array the first element in the array becomes the default value on subsequent insertions of digits in the input field.

#### External Library API

1. Options can be specified with data types just like in *numeral API*
2. Unlike *numeral API*, options are not predetermined and you can specify any number of custom options.
3. `formatter` option must be a callback that returns a formatted String that will be shown in the input field.
4. Formatter takes 2 parameters:
  	- The String to format
  	- The options specified in the Object

##### Example of External Library Plugin

This example uses [libphonenumber-js](https://github.com/catamphetamine/libphonenumber-js), a JavaScript implementation that format phone numbers based on Google's famous [libphonenumber](https://github.com/googlei18n/libphonenumber) library, but does not depend on Google Closure library.

Steps to produce:
1. Install libphonenumber-js: `npm install libphonenumber-js`.
2. Write a JavaScript file that returns an Object that contains a `formatter` property specifying a formatter Function from *libphonenumber-js*. Example shown below.
3. Import the Object in your template. Example shown below.
4. Assign the Object to `mask` property.

> The callback function must return a String that will be entered into the input field

###### Example Formatter Function

```javascript
'use strict'

import { asYouType as AsYouType } from 'libphonenumber-js'

const LibphonenumberMask = {
  countryCode: 'MY', // Malaysia
  // formatter takes 2 parameters:
  // 1. String to format
  // 2. Options specified in this object, in this case 'countryCode'
  formatter: (phoneNumber, options) => {
    let result = !options.countryCode
      ? new AsYouType().input(phoneNumber)
      : new AsYouType(options.countryCode).input(phoneNumber)

    return result
  }
}

export default LibphonenumberMask
```

###### Example Vuetify Template

```html
<template>
  <v-content>
    <v-text-field
      :clearable="true"
      v-model="input"
      :mask="mask">
    </v-text-field>
  </v-content>
</template>

<script>
import phoneMask from 'path/libphonenumber-mask'

export default {
  data () {
    return {
      mask: phoneMask,
      input: null
    }
  }
}
</script>
```

> A complete Vuetify template to test the new enhancement is provided **[here](https://gist.github.com/azaars/00eef2afe9551b215ce960f5807fc6d4)** to facilitate reviewing and testing.

Resolves #2138


johnleider:
Thank you for the very thorough PR. This is quite a bit to digest so I've tagged the entire team to take a look.

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      const decimal = text.lastIndexOf('.')
+      const fraction = this.getFraction(text.substr(decimal + 1))
+
+      return text.substr(0, decimal) + '.' + fraction
+    },
+    removeLeadingZeros (text) {
+      let parsed = []
+      let leadingZeros = true
+
+      for (const digit of text) {
```
jacekkarczmarczyk: could be `text.replace(/^0+/, '')`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+        if (leadingZeros) {
+          if (digit === '0') continue
+          leadingZeros = false
+        }
+
+        parsed.push(digit)
+      }
+
+      parsed = parsed.join('')
+      return parsed[0] === '.' ? '0' + parsed : parsed
```
jacekkarczmarczyk: Does it assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+    },
+    removeNonNumeral (text, preserveDecimal = false) {
+      if (text == null) return ''
+      return preserveDecimal ? text.replace(new RegExp(/[^\d.]/, 'g'), '')
```
jacekkarczmarczyk: Does it assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+    addPrefixSuffix (text) {
+      return this.numeralPrefix + this.sign + text + this.numeralSuffix
+    },
+    parseInt (text) {
+      return this.removeNonNumeral(this.removeLeadingZeros(text))
+    },
+    parseFloat (text) {
+      text = this.removeLeadingZeros(text)
+
+      const decimal = text.lastIndexOf('.')
```
jacekkarczmarczyk: Does it (and 3 lines below) assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+        // abs(value) < 1
+        text.substr(this.selection, 1) !== this.decimal && selection++
+        text.substr(this.selection - 1, 1) === this.decimal && selection++
+      }
+
+      return selection
+    },
+    // Correction on arbitrary value entered externally
+    attemptNumeralCorrection (text) {
+      const decimalPoint = text.lastIndexOf('.')
```
jacekkarczmarczyk: Does it assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+    },
+    decimal () {
+      return this.oneChar(this.options.decimal)
+    },
+    blockSeparator () { // Default separator = ' '
+      const separator = this.options.blockSeparator
+
+      if (!Array.isArray(separator)) return [this.oneChar(separator, ' ')]
+
+      switch (separator.length) {
```
jacekkarczmarczyk: Looks like it's forcing block separator, what if I don't want to have a block separator?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      this.blockSeparator.includes(masked[0]) && masked.shift()
+
+      return this.addPrefixSuffix(masked.join(''))
+    },
+    numeralZero () {
+      const zero = 0
+      this.sign = '+'
+      return this.addPrefixSuffix(zero.toFixed(this.precision).toString().replace('.', this.decimal))
+    },
+    oneChar (char, singleChar = '.') {
```
jacekkarczmarczyk: `defaultChar` sound more descriptive than `singleChar`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      return char ? char[0] : singleChar
+    },
+    absVal (value) {
```
jacekkarczmarczyk: `absVal` may be misleading name for function that additionaly restricts the max value

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      // Set the general limit to that of Number.prototype.toFixed()
+      return parsed > 20 ? 20 : parsed
+    },
+    adjustNumeralCaret (selection, text) {
+      // Caret is at prefix
+      if (selection === 0) return this.numeralPrefix.length + this.sign.length
+
+      // Whether it's an integer or a fractional where abs(value) >= 10
+      if (!this.precision ||
+        (String(this.lazyValue).length > this.precision + (this.sign ? 3 : 2))) {
```
jacekkarczmarczyk: What is the meaning of this 2/3 constants?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      if (!text) return text
+
+      this.sign = text = String(text)
+      text = this.internalChange ? this.removeNonNumeral(text) : this.attemptNumeralCorrection(text)
+      if (!text) return text // attemptNumeralCorrection() may return null or falsy
+
+      let number
+      if (this.precision) {
+        number = text.split('')
+        number.splice(-this.precision, 0, '.')
```
jacekkarczmarczyk: Does it assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
...
+      let length = 0
+      let integer = !this.precision
+
+      text = this.sign ? String(text).substr(1) : String(text)
+
+      // Group the digits with specified separators.
+      // The first size and separator element
+      // of the arrays becomes the default value.
+      for (const digit of text.split('').reverse()) {
+        digit !== '.' && masked.push(digit)
```
jacekkarczmarczyk: Does it (and 6 lines below) assume that decimal point is always a dot?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,278 @@
+import {
```
jacekkarczmarczyk: Is the custom formatter mechanism powerful enough to support also numeral formatter? What I'd like to achieve is to keep core files (maskable, VTextField) simple and treat numeral formatter not as an internal part of text input but as a kinda external lib (same like this lib with phone numbers) but which is just bundled into Vuetify.

If this is possible (I'm not sure if it is as the mixin uses vars like `backspace`/`delete` which look tightly coupled with VTextField) we could think about modifying API so that `mask` would accept only mask string ('##AA') or a function, and Vuetify would have some masks (texts or functions) predefined with possibility to override it with dev's own masks:
`:mask="$vuetify.mask.creditCard"`
`:mask="$vuetify.mask.createNumericMask({decimal: ','})"`
`:mask="LibphonenumberMask({ country: 'DE'})"`

```
computedValue: {
  phoneMask: LibphonenumberMask({ country: this.country })
}
<v-text-field :mask="phoneMask">
```

All my comments here are just after quick "visual" scan of the code without using it live, hopefully questions I'm askaing are not very stupid ;)

src/mixins/numeral-formatable.js
azaars: Yes, it's always a dot in `this.lazyValue`. The algorithm used for calculation is all based on `this.lazyValue`, which holds the actual number.

src/mixins/numeral-formatable.js
azaars: Ok, good one.

src/mixins/numeral-formatable.js
azaars: It's enhancing `Math.abs()`. What could be a better name, anyway?

src/mixins/numeral-formatable.js
azaars: The right side of the equality uses `this.precision`.
When there's a sign symbol, we need to add 1 on the right side to match the length in `this.lazyValue`.
When the value equals or is more than 10.0, we need to add 2 on the right side to include 1.0 or 2.0 etc. to match the length of `this.lazyValue`.

All this is because the fall through codes only adjust the caret in situations when we need to maintain a minimum digit count. That is, it could be between 0.0000 and 9.9999. Otherwise, 0.0000 would become 0.

src/mixins/numeral-formatable.js
azaars: It's always a dot in `this.lazyValue`.

src/mixins/numeral-formatable.js
azaars: It's always a dot in `this.lazyValue`.

src/mixins/numeral-formatable.js
azaars: Nice one 👍

src/mixins/numeral-formatable.js
azaars: It's always a dot in `this.lazyValue`.

src/mixins/numeral-formatable.js
azaars: It's always a dot in `this.lazyValue`.

src/mixins/numeral-formatable.js
azaars: I forgot that no-separator condition 😬

src/mixins/numeral-formatable.js
jacekkarczmarczyk: THB I'd jsut use simply `parseInt(value, 10)` (of whatever the var name is) instead of `this.absVal(value)`. Why?

1. why anyone should provide negative numbers for block length or precision? And even if someone provide -3 for precision then probably because he want such a precision (to get only numbers like 1000, 2000, not 1234, not 1234.56) and instead he will get 1234.567

2. if someone will provide precision > 20 and we will call `toFixed` with this value - there will be clear message saying that precision > 20 is not supported, if we only limit it silently to 20, there will be no any message, and the precission will be 20 instead of intended >20. This first of all might lead to confusion and second - I bet it would be a very very rare case

Maybe I had more arguments, but was distracted and forgot. Anyway summing up - I'd just remove the method and use parseInt instead.

src/mixins/numeral-formatable.js
azaars: I'm proposing a design approach in this PR. There could be other approaches and the one you suggested could be a cleaner one. However, I couldn't think of a stronger design that can allow full control over the caret position. To do so, I need to tap on the keystroke event in VTextField and make `maskable` compute the correct function to use. (Perhaps, there's another way?)

The external/plugin approach does have some weaknesses. The user experience isn't that natural. But if the end result is justifiable with such weaknesses, then it does solve some of the end-user problems.

Just to elaborate further, I have a half-cooked mixin that attempts to leverage on features provided by `Number.prototype.toLocaleString()` and I can show it if there's an intention to pursue with `i18n` in VTextField. I have it injected with something like `:mask="{ formatter: 'locale-string': locales: 'ar-EG' }"` and even `<v-text-field locales='ar-EG'>`. I tried with several `locales` and `options`, LTR and RTL. Unfortunately, I solved one `locales` + `options` pair but broke the others. The character sets displayed very well on the input field. But when it comes to controlling the caret and fulfilling the user experience, I still couldn't find a way through. There's a lot of work before it can be usable. All in all, I couldn't get away from doing the internal hack. I couldn't imagine if it can be done as an external/plugin.

src/mixins/numeral-formatable.js
azaars: 1. This is a good feature. I think it could be made available next. Or should it be in this PR?
2. I think the limitation can be removed because the calculations don't use `parseInt()` or `parseFloat()` at all, what more the `toFixed()`.

azaars: Replacing `this.absVal()` entirely with `parseInt(value, 10)` may not work because the later may return NaN. `this.absVal()` is needed to ensure that NaN is treated before proceeding.

I will proceed with item (2), but put item (1) on hold for now. `this.absVal()` may later be gone. Thank you for the suggestions.

src/mixins/numeral-formatable.js
```diff
@@ -90,66 +90,84 @@ export default {
...
-        digit !== '.' && masked.push(digit)
-        if (integer && !(++length % this.blockSize[j])) {
-          masked.push(this.oneChar(this.blockSeparator[i], ' '))
-          i > 0 && i--
-          j > 0 && j--
-          length = 0
-        } else if (digit === '.') {
-          masked.push(this.decimal)
-          integer = true
+        if (!this.maskNumeral(digit, param)) break
```
jacekkarczmarczyk: `maskNumeral` always returns `true`, so this line (as well as `return true` in `maskNumeral`) can be removed

azaars: The return `false` is at line 123.

jacekkarczmarczyk: How did I miss it...

azaars: 😄

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+      return this.oneChar(this.options.decimal)
+    },
+    blockSeparator () { // Default separator = ' '
+      const separator = this.options.blockSeparator
+
+      if (!Array.isArray(separator)) return [this.oneChar(separator, '')]
+
+      switch (separator.length) {
+        case 0: return [' ']
+        case 1: return this.oneChar(separator, '')
```
jacekkarczmarczyk: isn't it just `separator[0]`?
Which leads to the 
```js
return separator.length ? separator : [' ']
```
solution

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+        default: return separator
+      }
+    },
+    blockSize () { // Default size = 3
+      const size = this.options.blockSize
+
+      if (!Array.isArray(size)) return [this.absVal(size)]
+
+      switch (size.length) {
```
jacekkarczmarczyk:
```js
return size.length ? size.map(this.absVal) : [3]
```

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+
+        const negative = val && val.length > 0 && val[0] === '-'
+        const positiveSymbol = this.positiveSign ? '+' : ''
+
+        this.signSymbol = this.options.positiveOnly ? positiveSymbol
+          : negative ? '-' : positiveSymbol
+      }
+    },
+    numeralPrefix () {
+      return this.options.prefix ? this.options.prefix : ''
```
jacekkarczmarczyk: `this.options.prefix || ''`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+    },
+    numeralSuffix () {
+      return this.options.suffix ? this.options.suffix : ''
```
jacekkarczmarczyk: `this.options.suffix || ''`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+
+      return this.parseNumber(text)
+    },
+    maskNumeralText (text) {
+      if (!text) return this.numeralZero()
+      text = this.sign ? String(text).substr(1) : String(text)
+
+      const param = {
+        masked: [],
+        i: this.blockSeparator.length - 1,
```
jacekkarczmarczyk: `i`, `j` are a bit mysterious names, maybe `blockNumber`, `charInBlock` (or whatever is these vars meaning)?

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+        }
+      } else if (digit === '.') {
+        param.masked.push(this.decimal)
+        param.integer = true
+      }
+
+      return true
+    },
+    maskedNumeral (masked) {
+      if (!masked.length) masked.splice(0, 0, '0')
```
jacekkarczmarczyk: `if (!masked.length) masked = ['0']`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+      else if (this.blockSeparator.includes(masked[0])) masked.shift()
+
+      const result = this.addPrefixSuffix(masked.join(''))
+
+      return masked[0] === '.' ? '0' + result : result
+    },
+    numeralZero () {
+      const zero = 0
+      this.sign = '+'
```
jacekkarczmarczyk: 1. Side effects in this kind of methods is not the best practice, I would move `this.sign = '+'` to `maskNumeralText` (or even higher, but that would need my deeper understanding of the code)
2. `zero.toFixed` => `0..toFixed`, remove const
3. `toFixed()`  returns string, no need to to `toString()`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+          : this.numeralCaretOnInsert(selection)
+    },
+    numeralCaretOnInsert (selection) {
+      selection -= this.selection < selection ? 1 : 0
+      selection -= this.selection === selection ? 1 : 0
+
+      return selection
+    },
+    numeralCaretOnDelete (selection, text) {
+      if (text.substr(this.selection, 1) === this.decimal) {
```
jacekkarczmarczyk: `text[this.selection]` (and few more similar occurences below and in `numeralCaretOnBackspace`)

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+    addPrefixSuffix (text) {
+      return this.numeralPrefix + this.sign + text + this.numeralSuffix
+    },
+    parseNumber (text) {
+      if (!this.precision) return this.parseInt(text)
+
+      text = text.split('')
+      text.splice(-this.precision, 0, '.')
+
+      return this.sign + this.parseFloat(text.join(''))
```
jacekkarczmarczyk: maybe stupid question, but what is this function supposed to do?
if I take '3.25' as a text and follow all steps here I get:
line 238: `["3", ".", "2", "5"]`
line 239: (assuming precision = 2): `["3", ".", ".", "2", "5"]`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+    parseInt (text) {
+      return this.removeNonNumeral(this.removeLeadingZeros(text))
+    },
+    parseFloat (text) {
+      text = this.removeLeadingZeros(text)
+
+      const decimal = text.lastIndexOf('.')
+      const fraction = this.getFraction(text.substr(decimal + 1))
+
+      return text.substr(0, decimal) + '.' + fraction
```
jacekkarczmarczyk: isn't it (all this function) just `Number(text).toFixed(this.precision)`? (which makes this method unnecessary as it's only used in one place)

`Number('0003.2').toFixed(2) === '3.20'`
`Number('.2').toFixed(2) === '0.20'`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+    parseNumber (text) {
+      if (!this.precision) return this.parseInt(text)
+
+      text = text.split('')
+      text.splice(-this.precision, 0, '.')
+
+      return this.sign + this.parseFloat(text.join(''))
+    },
+    parseInt (text) {
+      return this.removeNonNumeral(this.removeLeadingZeros(text))
```
jacekkarczmarczyk: `return this.removeNonNumeral(text).replace(/^0+/, '')`, and remove `removeLeadingZeros` (see comment in `parseFloat`)

And maybe remove this function as it's used only in one place and is short enough to inline it

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+        if (this.numeralPrefix) {
+          for (const char of val) {
+            if (allowedMasks['#'].test(char)) break
+            if (char === '+' || char === '-') break
+            i++
+          }
+          val = val.substr(i)
+        }
+
+        const negative = val && val.length > 0 && val[0] === '-'
```
jacekkarczmarczyk: `const negative = val[0] === '-'`

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+    // Correction on arbitrary value entered externally
+    attemptNumeralCorrection (text) {
+      const decimalPoint = text.lastIndexOf('.')
+      let integer = decimalPoint < 0 ? this.removeNonNumeral(text)
+        : this.removeNonNumeral(text.substring(0, decimalPoint))
+
+      integer = integer.length ? integer : '0'
+      if (!this.precision) return integer
+
+      return decimalPoint < 0 ? integer + this.getFraction('')
```
jacekkarczmarczyk: similar to `parseFloat ()` below I think you can also use here built in `toFixed()` function
Probably
```js
return Number(this.removeNonNumeral(text)).toFixed(this.precision)
```
or something similar (and after this change and change in `parseFloat()` `getFraction` could be removed)

src/mixins/numeral-formatable.js
```diff
@@ -0,0 +1,267 @@
...
+
+      return text.substr(0, decimal) + '.' + fraction
+    },
+    removeLeadingZeros (text) {
+      text = text.replace(/^0+/, '')
+      return text[0] === '.' ? '0' + text : text
+    },
+    removeNonNumeral (text, preserveDecimal = false) {
+      if (text == null) return ''
+      return preserveDecimal ? text.replace(new RegExp(/[^\d.]/, 'g'), '')
```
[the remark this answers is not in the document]
nekosaur: I dunno. Making it do `''.replace(re, '')` seems kind of silly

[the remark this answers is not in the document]
jacekkarczmarczyk: `text ? text.replace(re, '') : ''`?

src/mixins/maskable.js
```diff
@@ -40,10 +57,40 @@ export default {
...
 
-      return mask.split('')
+        return text => formatter(text, customOptions)
+      } else { // No mask
+        return text => text
+      }
+    },
+    unmaskText () {
+      if (!this.mask) return text => text
+      return this.isNumeralFormatter ? text => this.unmaskNumeralText(text) : text => unmaskText(text)
```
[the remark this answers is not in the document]
azaars: I got carried away with the multi-parameters call pattern 😁 

The same goes with line 80

src/mixins/numeral-formatable.js
azaars: Well, I think it's probably better not to validate any of the options at all because it's the dev's responsibility to define it correctly if he wants his codes to run well.

src/mixins/numeral-formatable.js
azaars: Hmm.. so index out of bound is not an exception in JS

src/mixins/numeral-formatable.js
azaars: Sometimes thinking too much makes one forget about the KISS rule 😅

src/mixins/numeral-formatable.js
azaars: Blame my C background

src/mixins/numeral-formatable.js
azaars: KISS :sweat_smile:

src/mixins/numeral-formatable.js
azaars: 1. It's an init. Maybe it should be in `mounted ()`

src/mixins/numeral-formatable.js
azaars: What does `.map()` do?

src/mixins/numeral-formatable.js
azaars: `text` is `[0-9]` only when it enters `parseNumber()`

src/mixins/numeral-formatable.js
azaars: 2. If it's an integer only, the dot in `'0.'` should not be displayed.
3. Well, it's a design choice I made. Originally, I used the built-in `parseInt()` and `parseFloat()` everywhere (was why last time you see the hard limit of max 20 for options). Then I dropped them because it's only a computation limit, whereas the input field is free from such limitation. So an app may not want such limitation perhaps because it's not meant to be computed.

src/mixins/numeral-formatable.js
azaars: My rationale above.

src/mixins/numeral-formatable.js
azaars: Same rationale above.

src/mixins/numeral-formatable.js
azaars: Same rationale.

src/mixins/numeral-formatable.js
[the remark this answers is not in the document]
azaars: I'm worried if `text ?` would resolve to `false` for `'0' ?` in some browsers, while `0` is totally a valid digit; hence the `text == null`. Should I or should I not worry?

src/mixins/numeral-formatable.js
```diff
@@ -122,6 +122,9 @@ export default {
       if (isNaN(parsed)) return 0
       return Math.abs(parsed)
     },
+    numeralZero () {
+      return this.addPrefixSuffix(this.precision ? '0.' + this.getFraction('') : '0')
+    },
```
KaelWD: This should probably be a computed property so it can be cached.

src/mixins/numeral-formatable.js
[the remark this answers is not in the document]
nekosaur: I think current code is fine.

davi5e:
Can the tables the OP showed be added to the **Mask Legend Data Table** in https://vuetifyjs.com/components/text-fields

I found this page after I got confused on how to add a custom mask to my text fields, maybe more people don't get how to create their own at first glance...

Regards, DAVI

renatosistemasvc:
Forecast how long this feature will be available? I really need this feature.

KaelWD:
The goal for 1.1 is currently february. This feature might make it into 1.0 in december though if we have time to review and merge it.

nekosaur:
@azaars Are there things left to do on this? If not, please make sure it's properly merged with current dev, and we'll try to find time to review it.

azaars:
Hi @nekosaur . Ok, I'm testing the changes due to conflicts.

KaelWD:
Please don't change lockfiles unless it's related to your PR

jvbianchi:
any news on this PR?

johnleider:
@jvbianchi There is a milestone assignment for 1.1, we will evaluate this then.

johnleider:
Ultimately, this PR is way beyond the scope of what I want to support in the core of Vuetify. I would much rather allow for easier integration/provide steps for integrating with a pre-existing library for more advanced functionality. Double this with the fact that the author of this has disappeared, I will be closing this.

raffaeu:
@johnleider I start to get annoyed by your attitude. The PR is actually brilliant, we are using Vuetify for our internal ERP and you need to understand that if you want to provide a components library your mindset needs to be open enough to accept ideas from us (the developers working with your framework). It is the same problem of other components (see the non-flexibility of the drawer or the tab component).

Have you wonder why @azaars  disappeared? Come'n dude, we are here to enhance Vuetify and make it a better framework, so be more open to accept ideas such this one, which I personally think is great. Now you drop it, so how am I supposed to have proper number formatting into an input box??

johnleider:
We accept ideas all the time @raffaeu. I am the creator but it's not just _my_ framework. We have a team of developers that help guide the direction and progression of the project. Not every decision we make is going to please every developer, and not every decision we make is going to be the right one.

At the end of the day, I think you are misconstruing the reasons for closing this PR. Let's take `v-form` for example. It is a simple implementation that offers basic functionality for validating form controls. We also integrate with **vuelidate** and **vee-validate**. There is no need for the framework to be as robust and feature rich as one of the latter plugins, because we are not a validation framework.

This also applies to many other areas of the framework. It is a balance of providing useful functionality to the developer, while also giving them the freedom (and providing examples) of how to use it with other plugins or to create their own.

Take https://github.com/vuetifyjs/vuetify/pull/3013 for example. The PR was great, but the direction the team was taking with v1.1 was going to make a lot of it incompatible. Yes, I closed the PR, but we brought @sorinsarca into the collaborators to help bring some of those features to the next minor release. This is the same thing that we did with @azaars before he went mia.

Thank you for taking the time to provide feedback and I hope that I have been able to better explain the reasoning for doing things the way we do.

renatosistemasvc:
@johnleider  I agree with @raffaeu  in some points. It's a very important feature.

This is not a validation feature. Vuetifyjs does not allow masking of price (money) field in .
ex: $ 120.00

External money masquerade libraries do not work well with Vuetifyjs. Almost all systems need this feature. You should review the need to do this.

johnleider:
@renatosistemasvc The masking system will continue to be improved upon and we will make sure that integrating more advanced features from other libraries is clear and documented, similar to the validation.

nekosaur:
Here are two examples of external libraries working with vuetify

vue-the-mask: https://codesandbox.io/s/4x8vrmq5yx
v-money: https://codesandbox.io/s/5y55k1klll

paulpv:
As an argument against relying on 3rd party solutions, v-money as the following issue:
https://github.com/vuejs-tips/v-money/issues/44
`Can't programmatically change the value with Vuetify`

`v-money` is recommended by `vue-the-mask` (perhaps written by the same folks) to handle currency

johnleider:
Fair enough. This all will be much easier to manage once v1.1 comes out.

paulpv:
In the end I am using vue-the-mask, which seems to "work", but MMMV.

paulpv:
vue-the-mask, like Vuetify's v-text-field built in mask, requires the masked format.
So, if I set the mask to "###.##", a person cannot enter "0.1", they have to literally enter "000.10".
I literally cannot enter the "." character until after 3 numbers have been entered.
I was also getting some other funky results w/ both vue-the-mask and Vuetify's built in mask.

I am on day 3 of trying to limit my v-text-fields to only a floating point number.
So, I am abandoning getting this to work in Vuetify and will be using a Vue specific solution, or hand rolling my own.

lunchboxsushi:
Awe man, I've been keeping an eye on this PR for a few months to fix our band-aid solution during prototyping. I've tried vue-the-mask and v-money but ended up with vue-numeric, it's the only one that allows us to make it look similar to vue-text-field but the validation is not integrated nicely. 

are there are suggestions for the time being on how to integrate a library with v-text so that the currency input looks & behaviour the same way?

johnleider:
This is something we will look at integrating into the docs, similar to what we did with vuelidate and vee-validate.

paulpv:
You may also try using my implementation:
https://github.com/paulpv/vuetify-number-field

Unfortunately I have not been able package this up in to a working **stand-alone** npm package/module.
The only way I could find out how to get this to work was to fork Vuetify and add code from my repo to Vuetify's components folder.

It would be awesome if anyone can help me find out how to package my repo as a standalone module!

RosilBelli:
Please, is there any estimate for currency mask on text-fields? 
I've tried many external libs but none of them work well with Vuetify.
This is really, really, really important for business applications. 
What other solution do you suggest to overcome the lack of this feature? 
I really need help to achieve this.

nekosaur:
v-money seems to be working fine <https://codesandbox.io/s/5y55k1klll>

RosilBelli:
Yes, I tried it too, but no computed values are displayed in masked fields (just when we type into). 
How did you overcome this issue @nekosaur ?

nekosaur:
Lets not clutter up this issue any more since it's closed. If you want to continue a discussion about this, join us in [chat](https://chat.vuetifyjs.com).

PR #2249, opened 2017-10-20
Fixes #2248 navigation drawer touchless state issue

### LINKED ISSUE #2248: Unable to toggle navigation drawer touchless state
### Versions
Vue 2.5.2
Vuetify 0.16.6

### What is expected ?
I would expect to be able to use `:touchless` and be able to switch between having the touch swipe left and right events toggled on an off.

The touch events only get configured when the component is rendered and the touchless state can't be toggled after that.

### Reproduction Link
https://jsfiddle.net/kkx386w6/1/

If you open the fiddle and use the mobile device emulator you can verify that toggling between the touchless state (using the flash icon on the right of the toolbar) doesn't effect the change the touch state.

mcfarljw:
Fixes #2248 

This allows for toggling the `:touchless` state on the fly without needing to generate the component again. I think the easiest way of accomplishing this is to always push the touch directive and check the touchless state on a swipe event.

### DIFF
--- a/src/components/VNavigationDrawer/VNavigationDrawer.js
+++ b/src/components/VNavigationDrawer/VNavigationDrawer.js
@@ -195,17 +195,16 @@ export default {
         {
           name: 'resize',
           value: this.onResize
+        },
+        {
+          name: 'touch',
+          value: {
+            parent: true,
+            left: this.swipeLeft,
+            right: this.swipeRight
+          }
         }
       ]
-
-      !this.touchless && directives.push({
-        name: 'touch',
-        value: {
-          parent: true,
-          left: this.swipeLeft,
-          right: this.swipeRight
-        }
-      })
 
       return directives
     },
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
This PR solves problem only for v-navigation-drawer, I suppose the proper solution would be to handle properly all necessary 'touch' directive hooks.
At the moment only 'inserted' and 'unbind' hooks are supported (see src/directives/touch.js) while 'update' and 'componentUpdated' are ignored
See https://vuejs.org/v2/guide/custom-directive.html

mcfarljw:
I might have been a bit narrow minded in this PR, but I just checked and it looks like `VTabsItems` also uses the touchless prop. I'm not sure when I'd have to to look into doing that or if I'd cause more harm than good, but it does sound like allowing the touch direction to bind and unbind itself based on a prop state would be the most reusable pattern.

jacekkarczmarczyk:
Yup, look similar

jacekkarczmarczyk:
Closing in favour of more generic solution https://github.com/vuetifyjs/vuetify/pull/2286

PR #2297, opened 2017-10-23
New Component: Data Iterator (DataTable abstraction)

Phlow2001:
- Extracted data related behavior from VDataTable into data-iterable mixin
- Refactored VDataTable to be render and data table specific behavior with data-iterable mixin
- Created new VDataIterator component with data-iterable mixin
- Fixed bugs in VDataTable no-data and no-results, as well as fixing and enhancing data table tests.
- Added VDataIterator tests and snapshots

Example usage of VDataIterator: https://codepen.io/jayblanchard/pen/NvbwMa?editors=1010


src/components/VDataIterator/VDataIterator.js
```diff
@@ -0,0 +1,119 @@
...
+    },
+    genFooter () {
+      const children = []
+
+      if (this.$slots.footer) {
+        children.push(this.$slots.footer)
+      }
+
+      if (!this.hideActions) {
+        children.push(this.$createElement('div', this.genActions())
```
nekosaur: The div you create here should be unnecessary. `genActions` returns a root div.

src/components/VDataIterator/VDataIterator.js
```diff
@@ -0,0 +1,119 @@
...
+        'dataiterator': true,
+        'dataiterator--select-all': this.selectAll !== false,
+        'theme--dark': this.dark,
+        'theme--light': this.light
+      }
+    }
+  },
+
+  methods: {
+    createProps (item, index) {
```
nekosaur: Try to refactor createProps from v-data-table into data-iterable instead.

src/components/VDataIterator/VDataIterator.js
```diff
@@ -0,0 +1,119 @@
...
+    tag: {
+      type: String,
+      default: 'div'
+    }
+  },
+
+  computed: {
+    classes () {
+      return {
+        'dataiterator': true,
```
nekosaur: Use kebab-case for css class names

src/components/VDataIterator/VDataIterator.js
```diff
@@ -0,0 +1,119 @@
...
+
+export default {
+  name: 'v-data-iterator',
+
+  mixins: [DataIterable],
+
+  inheritAttrs: false,
+
+  props: {
+    tag: {
```
nekosaur: Perhaps change this to `content-tag`, and add explicit `content-props` prop (Object) to be used with `v-bind=` on content element?

src/components/VDataIterator/VDataIterator.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        items: [
+          { other: 1, col1: 'foo', col2: 'a', col3: 1 },
+          { other: 2, col1: null, col2: 'b', col3: 2 },
+          { other: 3, col1: undefined, col2: 'c', col3: 3 }
+        ]
+      }
+    }
+  }
+
+  it('should match a snapshot - no matching records', () => {
```
nekosaur: It's probably a good idea to check in these tests that the correct text is actually displayed.

nekosaur:
Try to eliminate the last identical code block that codebeat complains about

src/components/VDataIterator/VDataIterator.js
```diff
@@ -49,38 +49,24 @@ export default {
...
+      }, content)]
     },
     genFilteredItems () {
+      if (!this.$scopedSlots.item) {
+        return null
+      }
+
       const items = []
-      this.filteredItems.forEach((item, index) => {
+      for (let index = 0, len = this.filteredItems.length; index < len; ++index) {
```
Phlow2001: https://jsperf.com/native-map-versus-array-looping

Native loop outperforms by a good bit in most cases, and by a lot in some cases (like latest safari on ios)

PR #2327, opened 2017-10-26
Don't close dialog on synthesized click events

neelance:
When an item of v-select gets selected via the arrow kets and the enter key, HTMLElement.click() gets called for the item. This wrongly gets registered by v-dialog as a click outside of the dialog and it closes.


jacekkarczmarczyk:
It's already fixed in 0.17

neelance:
Okay cool. Sorry for bothering you.

jacekkarczmarczyk:
np, thank you for your contribution though

PR #2329, opened 2017-10-26
Fix for default Google Chrome outline on sortable table headers

lowerends:
This change will prevent the default Google Chrome outline (-webkit-focus-ring-color auto 5px;) being applied to the sortable table headers.

### DIFF
--- a/src/stylus/components/_tables.styl
+++ b/src/stylus/components/_tables.styl
@@ -60,6 +60,7 @@ table.table
 
       &.sortable
         pointer-events: auto
+        outline: none
 
       > div
         width: 100%


Phlow2001:
Tab focus needs some indication of where focus is.  This would break that.

lowerends:
What should the indication look like for tab focus? I didn't find a material design guideline for this.

lowerends:
Just to make sure it's clear what I'm talking about, I pasted a screenshot from the documentation pages:
image

johnleider:
I want to merge this because I think the current way it works looks terrible. @Phlow2001 does have a point though, how do we still support focused states?

johnleider:
While I'm not a fan of the current implementation, until we can support both accessibility **and** good design, I am going to have to close this PR. Thank you for your contribution.

PR #2336, opened 2017-10-27
Add outline support to Alert component

HendrikGrobler:
The button component supports an outline display mode and users may want to display alerts this way for certain designs.

Docs PR: https://github.com/vuetifyjs/docs/pull/353

screenshot_20171030_144202

### DIFF
--- a/src/components/VAlert/VAlert.js
+++ b/src/components/VAlert/VAlert.js
@@ -18,6 +18,7 @@ export default {
   props: {
     dismissible: Boolean,
     icon: String,
+    outline: Boolean,
     type: {
       type: String,
       validator (val) {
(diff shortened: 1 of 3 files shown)


HendrikGrobler:
Added docs PR.

src/components/VAlert/VAlert.js
```diff
@@ -38,10 +39,12 @@ export default {
...
     classes () {
       const colorProp = (this.type && !this.color) ? 'type' : 'computedColor'
-
-      return this.addBackgroundColorClassChecks({
-        'alert--dismissible': this.dismissible
-      }, colorProp)
+      const classes = {
+        'alert--dismissible': this.dismissible,
+        'alert--outline': this.outline
+      }
```
johnleider: add an empty line

src/stylus/components/_alerts.styl
```diff
@@ -17,6 +17,8 @@
     align-self: center
     color: $alert-icon-color
     font-size: $alert-icon-font-size
+  &--outline .icon
+    color: currentColor !important
```
johnleider: Can this be set to just inherit the color and ditch the **!important**?

src/components/VAlert/VAlert.js
HendrikGrobler: Done.

src/stylus/components/_alerts.styl
HendrikGrobler: I'll check and confirm, but from what I remember !important was needed. By inherit do you mean still using `currentColor` or something like `color: inherit`?

johnleider: Either if they will work without important. If it's needed though, just use `color: inherit`

HendrikGrobler: Both and `inherit` and `currentColor` don't work without `!Important` (they appear grey). Changing this to use `color: inherit !Important` then.

PR #2344, opened 2017-10-28
fix: v-text-field realtime validation

### LINKED ISSUE #2342: v-text-field no longer validating on change
### Steps to reproduce
Use a `v-text-field` with one or more validation rules. Start typing into the field.

### Versions
Vue 2.5.2
Vuetify 0.17.0-beta.1

### What is expected ?
The validation state immediately updates whenever the input value changes.

Validation only occurs on blur, even with `:validate-on-blur='false'`

In my example, as soon as you start typing into the first two fields, they're marked as invalid even though they contain a value. Once they're marked as valid, clearing the value only marks them as invalid on blur.

In the second example, adding or removing the number 7 also doesn't trigger an immediate validation state change.

### Reproduction Link
0.17.0-beta.1, not working: https://codepen.io/jaromero/pen/WXNrMZ
0.16.9, working: https://codepen.io/jaromero/pen/ZaEQqq

(FYI, it also isn't working in the example in the beta docs: https://dev.vuetifyjs.com/components/text-fields#example-11)

azaars:
Fix as-you-type validation.

Fixes #2342


KaelWD:
Please update the title to reflect what this PR does.

johnleider:
This doesn't behave as I would expect. If the **validate-on-blur** prop is used, I would expect it to only validate when I leave the field. With this change, that occurs even without the prop. I would expect if **validate-on-blur** is not defined or defined as a falsey value, that it would immediately start to validate.

azaars:
It does that on this statement:

```javascript
!this.validateOnBlur && this.validate()
```

The control flow should run through that line in all conditions regardless of the presence of **validate-on-blur**. In 0.16.9 it did, but in 0.17 beta-1 it doesn't when `this.internalChange === true`

azaars:
@johnleider If you're convinced about the changes, then this won't be necessary https://github.com/azaars/vuetify/commit/0414fc62b9a71fb2f609c55607da5fd12ff5a216#diff-c2cbb2f11cf8c97e6eff0981f9927269R178

PR #2354, opened 2017-10-29
fix: only apply button hover style when supported

### LINKED ISSUE #2178: [Bug Report] Toolbar icon highlight getting stuck after routing on mobile
### Versions
Vue 2.5.2
Vuetify 0.16.6

### What is expected ?
I'm expecting to be able to navigate between routes using icon links without the icon remaining selected or highlighted after the route has changed on mobile devices.

When switching between routes or using button on mobile they seem to be getting stuck in a highlighted state until the user clicks focus somewhere else on the screen.

screen shot 2017-10-14 at 4 22 41 pm

If you load the fiddle up below in Chrome using the mobile device viewer and click the back arrow the issue reveals itself.

### Reproduction Link
https://jsfiddle.net/mcfarljw/83r0bhmh/1/

mcfarljw:
fixes #2178

```
<template>
  <v-app>

    <v-toolbar app>
      <v-spacer></v-spacer>
      <v-btn icon @click="">
        <v-icon>favorite</v-icon>
      </v-btn>
    </v-toolbar>

  </v-app>
</template>
```

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -57,7 +57,6 @@ theme(button, "btn")
 /** Content */
 .btn
   &--active,
-  &:hover,
   &:focus
     .btn__content:before
       background-color: currentColor
@@ -74,6 +73,11 @@ theme(button, "btn")
     padding: $button-padding
     transition: $primary-transition
     white-space: nowrap
+
+  @media not all and (hover: none)
+    &:hover
+      .btn__content:before
+        background-color: currentColor
 
 /** Icons */
 .btn .btn__content .icon
@@ -126,7 +130,7 @@ theme(button, "btn")
     width: $button-floating-width
     padding: 0
     elevation(6)
-    
+
     &.btn--fixed,
     &.btn--absolute
       z-index: 4


jacekkarczmarczyk:
I thought that there's a way to detect 'hover: *' media query using `@support` but can't find the way
However I've added PR for hover support detection https://github.com/vuetifyjs/vuetify/pull/2356 so when it's merged you can use it and apply proper styles, it should work for all browsers including IE
The mixin adds `application--touch-support` class when touch is supported and `application--hover-support` class when hover is supported

jacekkarczmarczyk:
Firefox as well, as I mentioned before with #2356 this could be done like this:
```diff
 +  .application--supports-touch
 +    .btn:hover
 +      .btn__content:before
 +        background-color: currentColor
 +
```
However this might lead to unexpected behaviour when device is supporting both touch and mouse

mcfarljw:
The only bulletproof solution I've found for this is to manually toggle a hover class based on mouse events. Trying to find the perfect combination of supported css rules and javascript detection checks that allow for the use of `:hover` has always been a dead end road for me. Rather than hover detection what is there was a hover directive?

jacekkarczmarczyk:
Unfortunately this might be the only bullet proof solution

mcfarljw:
I'm cool with closing this PR in favor of going that (or another) direction. In the long run I think it would be them most flexible solution.

johnleider:
@KaelWD Are your requested changes to your liking?

KaelWD:
?

There have been no changes, this is basically impossible to do correctly. 
@mcfarljw, what do you mean by "hover directive"?

mcfarljw:
@KaelWD I mean some kind of hover directive that could be added to a component that would completely replace the usage of `:hover` in the stylesheets all together. It would need to toggle a hover classes based on mouse events.

manico:
How about using `:hover` only for devices larger then `960px` using media query?
I would currently cover most of the mobile/tablet devices.
Using directive for this does not seem like a optimum solution.

mcfarljw:
I think it would be best to just fix it rather than making a width based media query assumption. @manico In particular that media query would probably not work on most iPads.

johnleider:
As kael stated above, this breaks IE11 hover states. While what you have presented is an issue, it is a wide spread one that can affect pretty much any framework. Thank you for your contribution.

PR #2356, opened 2017-10-29
feat: add class to application if touch is supported

jacekkarczmarczyk:


src/util/pointerSupport.js
```diff
@@ -0,0 +1,29 @@
...
+    const setupPointerSupport = (propName, eventName) => {
+      const load = () => {
+        const app = document.querySelector('[data-app]') || document.body
+        const className = 'application--' + propName.replace(/([A-Z])/g, '-$1').toLowerCase()
+        app.classList.add(className)
+      }
+
+      addOnceEventListener(window, eventName, () => {
+        this.$vuetify[propName] = true
+        this.$vuetify.load(load)
```
KaelWD: Why wait for load? If the v-app has mounted then it's ready to go, no reason to wait for images.

src/util/pointerSupport.js
```diff
@@ -0,0 +1,24 @@
...
+      })
+    }
+
+    // Adds application--touch-support class
+    // after touchstart event is triggered
+    setupPointerSupport('touchSupport', 'touchstart')
+
+    // Add application--hover-support class
+    // after mouseover event is triggered
+    setupPointerSupport('hoverSupport', 'mouseover')
```
KaelWD: Does this work? Wasn't there something about mobile devices triggering `mouseover` as well?

jacekkarczmarczyk: There was: https://github.com/Modernizr/Modernizr/issues/869
But I like the comment with this 2 letters less in every line :P

jacekkarczmarczyk: Moreover, this PR (and touch detection in general) might be usesell when device has both touch and hover support...

KaelWD: lol fair enough. 
That issue is from a few years ago, so things might have changed since then, but this class should be documented as being inaccurate if that's still the case.

PR #2358, opened 2017-10-29
feat: selects input value when focusing autocomplete

### LINKED ISSUE #2255: Clicking on list does not select the current text
I really like the new changes to v-select but have found a few issues. Keep up the fab work :)

Clicking on list does not select the current text forcing redundant actions and regular misuse for keyboard only actions.

### Reproduction Link
See Scenario 2 on CodePen - https://codepen.io/anon/pen/aLMZPm

### Steps to reproduce
- Focus on select
- Type '2006'

### Versions
Vue 2.4.2
Vuetify 0.16.7
Windows 10
Chrome 61

### What is expected ?
Existing text is selected when the dropdown is displayed so the user can start typing a new value

Immediate typing is appended to existing value. This almost always produces no matches.

jacekkarczmarczyk:
fixes #2255

### Playground
```vue
<template>
  <v-app id="inspire">
    <v-container fluid>
      <input style="border: 1px solid red; padding: 1em; width: 100%">
      <v-select combobox :items="['abc', 'def']" v-model="v1"></v-select>
      <v-select tags :items="['abc', 'def']" v-model="v2"></v-select>
      <v-select autocomplete :items="['abc', 'def']" v-model="v3"></v-select>
      <v-select autocomplete multiple :items="['abc', 'def']" v-model="v4"></v-select>
      <input style="border: 1px solid red; padding: 1em; width: 100%">
    </v-container>
  </v-app>
</template>

<script>
export default {
  data: () => ({
    v1: 'test',
    v2: ['abc'],
    v3: 'abc',
    v4: ['abc']
  })
}
</script>
```

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -288,7 +288,6 @@ export default {
         !this.isMultiple &&
         val
       ) {
-        this.setCaretPosition(this.currentRange)
         this.shouldBreak && this.$nextTick(() => {
           this.$refs.input.scrollLeft = this.$refs.input.scrollWidth
         })
@@ -440,7 +439,7 @@ export default {
 
       if (this.$refs.input && this.isAutocomplete) {
         this.$refs.input.focus()
-        setTimeout(() => this.$refs.input.select(), 20)
+        this.$nextTick(() => this.$refs.input.select())
       } else {
         this.$el.focus()
       }
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -280,19 +280,6 @@ export default {
         }
       })
     },
-    isFocused (val) {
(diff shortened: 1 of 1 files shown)


src/components/VSelect/VSelect.js
```diff
@@ -425,6 +429,7 @@ export default {
 
       if (this.$refs.input && this.isAutocomplete) {
         this.$refs.input.focus()
+        setTimeout(() => this.$refs.input.select(), 20)
```
KaelWD: Where does 20ms come from?

KaelWD: Also you probably don't need the arrow function

jacekkarczmarczyk: from trials and errors

PR #2360, opened 2017-10-29
fix: invalid textfield behaviour with number value

### LINKED ISSUE #2359: [Bug] [v-text-field] interpretes negative numbers as NaN
### Versions and Environment

Vuetify: `v0.17.0-beta1`
Vue: `v2.5.2`
Browsers: `Chrome,Firefox,Safari`
OS: `macOS`

### Expected Behavior
<v-text-field> should allow the entry of negative numbers.

### Actual Behavior

But with v0.17.0-beta.1, as soon as you enter a `minus`, the input will be replaced with a literal string of `NaN`.

### Reproduction Link

https://codepen.io/myleslee/pen/xPxBQY

### Other comments

The issue cannot be reproduced on `v0.16.9` though.

jacekkarczmarczyk:
fixes #2359

### DIFF
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -93,8 +93,7 @@ export default {
           this.lazyValue = typeof val === 'number' ? +value : value
           this.setSelectionRange()
         } else {
-          const type = typeof this.lazyValue
-          this.lazyValue = type === 'number' ? +val : val
+          this.lazyValue = val
           this.$emit('input', this.lazyValue)
         }
       }
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -89,8 +89,7 @@ export default {
       },
       set (val) {
         if (this.mask) {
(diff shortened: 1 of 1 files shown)


KaelWD:
@azaars Does this break anything from #2099?

azaars:
Based on the discussion on discord, this PR is suffice and it won't break anything from #2099 

@jacekkarczmarczyk No need to do `this.lazyValue = String(val)`

src/components/VTextField/VTextField.js
```diff
@@ -89,12 +89,10 @@ export default {
       },
       set (val) {
         if (this.mask) {
-          const value = this.unmaskText(this.maskText(this.unmaskText(val)))
-          this.lazyValue = typeof val === 'number' ? +value : value
+          this.lazyValue = this.unmaskText(this.maskText(this.unmaskText(val)))
```
johnleider: O_o

johnleider:
I went to test this on DEV and the issue does not occur, maybe this can be closed.

PR #2362, opened 2017-10-29
feat: emit event on mobile breakpoint

### LINKED ISSUE #1929: [feature-request] emit mobile-breakpoint event on v-navigation-drawer when isMobile state changes
## What will it allow you to do that you can't do today?
Let parent know when the state of the navigation drawer has changed to be able to react in other ways, like modifying props on other components to suit the change, without having to watch v-resize for the same breakpoint.

## How will it make current work-arounds straightforward?
Simplify watching for a breakpoint

## What potential bugs and edge cases does it help to avoid?
Having parent's mobile breakpoint be different from navigation drawer

NothingWeAre:
fixes #1929

### DIFF
--- a/src/components/VNavigationDrawer/VNavigationDrawer.js
+++ b/src/components/VNavigationDrawer/VNavigationDrawer.js
@@ -187,8 +187,9 @@ export default {
       }
     },
     checkIfMobile () {
+      const savedState = this.isMobile
       this.isMobile = window.innerWidth < parseInt(this.mobileBreakPoint, 10)
-      this.$emit('mobile', this.isMobile)
+      if (savedState !== this.isMobile) this.$emit('mobile', this.isMobile)
     },
     closeConditional () {
       return !this.permanent && (this.temporary || this.isMobile)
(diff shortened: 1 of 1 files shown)


KaelWD:
https://github.com/vuetifyjs/vuetify/issues/1929#issuecomment-340667854

jacekkarczmarczyk:
BTW - this PR emits event on every resize, but it should emit only when mobile state has changed

NothingWeAre:
@jacekkarczmarczyk I agree, fixed

KaelWD:
Why don't you do it in the `isMobile` watcher instead? That is only called when the value actually changes.

NothingWeAre:
@KaelWD 
Sorry, you are right. I was editing commit on the go, from tablet, and didn't notice watcher.
Will fix when i will be able to sit in front of normal monitor and read source more attentively.

NothingWeAre:
moved `$emit` to watcher

johnleider:
As I try to avoid $emit calls when at all possible, I'm leaning towards this should be handled by the developer.

PR #2368, opened 2017-10-30
Add `disable-initial-sort` prop to VDataTable

lukeed:
* Added a new `default-sort` prop to DataTable, which defaults to `true` (no breaking behavior)
* Use new prop to help determine the initial value for `defaultPagination.sortBy`, if any.

_Closes #2366_

### DIFF
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -138,6 +138,10 @@ export default {
         })
       }
     },
+    defaultSort: {
+      type: Boolean,
+      default: true,
+    },
     value: {
       type: Array,
       default: () => []
@@ -314,7 +318,7 @@ export default {
       !('sortable' in h) || h.sortable)
     )
 
-    this.defaultPagination.sortBy = firstSortable
+    this.defaultPagination.sortBy = this.defaultSort && firstSortable
       ? firstSortable.value
       : null
 
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -140,7 +140,7 @@ export default {
     },
     defaultSort: {
       type: Boolean,
-      default: true,
+      default: true
     },
     value: {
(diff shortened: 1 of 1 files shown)


nekosaur:
Thank you for your contribution. However I'm not sold on the prop name. I would suggest something like `disable-initial-sort`, which is more in line with vuetify's style. Thoughts?

That said, I'm also not sold on our current implementation of combining both sorting and pagination under the `pagination.sync` prop. It likely won't get an overhaul until some time post-1.0 though.

lukeed:
Sure. I like your prop name better, I just didn't want to introduce a breaking change. 

I would really like to get (some form of) this in asap. It makes sense to me to rework the pagination at a later point.

jacekkarczmarczyk:
If `disable-initial-sort`'s default is `false` then it's not breaking
And I'd prefer then prop (or other mechanism) that defines which column should be initially sorted (none, first - by default, other) and whether initial sorting is asc or desc

lukeed:
Ah, right, just woke up :zzz:

Do you guys want me to implement the additional prop? Imo, it's a separate PR but I can.

nekosaur:
`And I'd prefer then prop (or other mechanism) that defines which column should be initially sorted (none, first - by default, other) and whether initial sorting is asc or desc`

This is currently done in `pagination.sync`. Like I said we should look it over for a later release but for now it works.

lukeed:
Hey guys, update on this?

lukeed:
Are you waiting for me to update the docs?

nekosaur:
You're more than welcome to do it, otherwise we'll take care of it.

lukeed:
I'm happy to do it if it means it'll land sooner (than 1.0) 🙏 I have a prod release that's pretty much waiting only on this.

nekosaur:
Can't guarantee that.

lukeed:
Thanks! 🎉

PR #2369, opened 2017-10-30
Fix #2359: Set type of Number(input) to String(input) for illegal number

### LINKED ISSUE #2359: [Bug] [v-text-field] interpretes negative numbers as NaN

azaars:
Allow `VTextField.input` to accept any character on non-masked input field even if the passed `v-model` or `v-model.number` expects a number. This seems to be the default behavior of HTML `input` element even if the element is set as `<input type="number">`

### DIFF
--- a/src/components/VTextField/VTextField.js
+++ b/src/components/VTextField/VTextField.js
@@ -89,12 +89,11 @@ export default {
       },
       set (val) {
         if (this.mask) {
-          const value = this.unmaskText(this.maskText(this.unmaskText(val)))
-          this.lazyValue = typeof val === 'number' ? +value : value
+          this.lazyValue = this.getActualValue(typeof val,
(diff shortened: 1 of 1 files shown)


azaars:
Preferred PR: #2360

PR #2375, opened 2017-10-30
feat: contentClass for VSelect

### LINKED ISSUE #2374: [Feature Request] Possibility to add class to a select menu
### New Functionality
It will allow to stylize the select menu, like display list items inline when they are too short, or simply add an option to display items inline

https://codepen.io/talski/pen/MOYEVx?editors=1010

### Improvements
For now, there's no way to do it

As in codepen sample, it will show all the select options at once

jacekkarczmarczyk:
fixes #2374


johnleider:
Could we not use the default implementation of **content-class** that we normally use on `v-menu`?

johnleider:
This needs to be updated in order to target the new file location, `VSelect/mixins/select-computed.js`

Issue #2081, opened 2017-10-08
v-select doesn't show the 'No data available' tile when items is empty


KaelWD:
### Steps to reproduce
Open the select

### Versions
v0.16.3

### What is expected ?
A disabled tile should be shown with the text 'No data available'

### What is actually happening ?
No menu is shown at all

### Reproduction Link
https://codepen.io/anon/pen/bovVOO?editors=1010

guanxudong:
v0.17.3 also has this bug

KaelWD:
The fix for this was reverted in c89ac1034f720d2483fb331412b6331f3f24fdf0

sindrepm:
Can this be fixed by changing the [display logic](https://github.com/vuetifyjs/vuetify/blob/0e81afa413a9c9f0ccb1fde9bb60f0a765e0f7fe/src/components/VSelect/mixins/select-computed.js#L99) to something like this?

```
  menuIsVisible () {
      return this.menuIsActive &&
        (!this.isAutocomplete || this.computedItems.length > 0) &&
        (!this.isAnyValueAllowed || this.filteredItems.length > 0)
    },
```
From what I could see, the reason that the fix was reverted (https://github.com/vuetifyjs/vuetify/commit/c89ac1034f720d2483fb331412b6331f3f24fdf0) was specific to autocomplete selects.

flybayer:
I'm still having this problem in 0.17.6 where selects with and without autocomplete don't show No Data Available when the items array is empty.

lcdss:
Same problem here in v1.0.0-beta.5.

wildan2711:
Thank god I found this, I thought I had the wrong usage.

samuelcolburn:
Same problem in v1.0.0-beta.5, specifically with async autocomplete. 
Searching with no results shows nothing. 
Once a result is finally hit, it will show a 'No Data Available,' but this shows immediately and not only once the search has been complete. Also, it does not use the no-data slot.

https://codepen.io/anon/pen/BYLZWE?&editors=101

kindofone:
Hey guys, is there any workaround for this issue, or plan to either fix or alternatively remove `no-data` slot from the notifications?

johnleider:
This has been resolved in v1.1. Also added a new prop **hide-no-data** which you can conditionally choose to not show the no data slot.

KaelWD:
[...]

PR #2381, opened 2017-10-31
Fix: Show no-data text on non-autocomplete selects

KaelWD:
See #2081
Reverts c89ac1034f720d2483fb331412b6331f3f24fdf0


johnleider:
We discussed this and decided that this is currently the intended functionality.

PR #2402, opened 2017-11-02
fix: picker header color in dark theme

jacekkarczmarczyk:
### Playground
```vue
<template>
  <v-app id="inspire" :dark="dark">
    <v-content>
      <v-container>

        <v-checkbox label="Dark" v-model="dark"></v-checkbox>
        <v-text-field label="Primary color" :value="$vuetify.theme.primary" @input="$vuetify.theme.primary = $event"></v-text-field>
        <v-date-picker></v-date-picker>

      </v-container>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data: () => ({
    dark: false
  })
}
</script>
```

### DIFF
--- a/src/components/VApp/VApp.js
+++ b/src/components/VApp/VApp.js
@@ -40,13 +40,13 @@ export default {
   },
 
   mounted () {
-    this.$vuetify.theme.type = this.type
+    this.$vuetify.dark = this.dark
     window.addEventListener('load', this.runCallbacks)
   },
 
   watch: {
-    type () {
-      this.$vuetify.theme.type = this.type
+    dark () {
+      this.$vuetify.dark = this.dark
     }
   },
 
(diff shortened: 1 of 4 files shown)


src/components/VApp/VApp.js
```diff
@@ -40,13 +40,13 @@ export default {
...
-    this.$vuetify.theme.type = this.type
+    this.$vuetify.dark = this.dark
     window.addEventListener('load', this.runCallbacks)
   },
 
   watch: {
-    type () {
-      this.$vuetify.theme.type = this.type
+    dark () {
+      this.$vuetify.dark = this.dark
```
johnleider: should we put this under the **theme** object in **$vuetify**?

KaelWD: I believe we already discussed that, having it be `$vuetify.currentTheme` or something similar.

PR #2430, opened 2017-11-07
Fix/navigation drawer active states

johnleider:
Resolves complicated states with the **permanent**, **temporary** and **persistent** props.

- removed **persistent**
- a provided explicit boolean value will always determine initial state
- a provided null/undefined value will cause the system to determine initial state
 - permanent will always open by default and not toggle on desktop, regardless of value
 - temporary will always be closed by default
 - no props will open on desktop by default


derwish-pro:
Yes, I noticed that it works when I set 80% scale in the browser. In normal mode, it does not work. How can I reduce the threshold at which it starts to work? I want it to work up to 400px width.

PR #2439, opened 2017-11-08
Enhancement/abstract select

### LINKED ISSUE #2100: V-Select selection slot without multiple option not beign displayed until blur
### Steps to reproduce
Add a V-Select, without multiple option, with slots for item and selection. The value that's shown when the item is selected and has the focus is the value from item-text and not selection slot. Once you blur from the select, the correct slot is shown.

### Versions
Vuetify 0.16.4

### What is expected ?
The template from selection slot should be shown even if the field has the focus

The selection slot is only shown once the field loses focus

### Reproduction Link
https://codepen.io/Eylen/pen/RLMEbQ

### LINKED ISSUE #2254: Select dropdown list does not retrack when null value selected
I really like the new changes to v-select but have found a few issues. Keep up the fab work :)

### Reproduction Link
See Scenario 1 on CodePen - https://codepen.io/anon/pen/aLMZPm

### Steps to reproduce
- Focus on select
- Select 2006
- Select Unspecified

This is possible the same issue as #2200 but hit via a different way

### Versions
Vue 2.4.2
Vuetify 0.16.7
Windows 10
Chrome 61

### What is expected ?
Dropdown is hidden and the model value is null

Dropdown is not hidden although model value is null

### LINKED ISSUE #2401: [Bug Report] VSelect combobox doesn't close menu if you use the arrow to open
### Versions and Environment
Vuetify: v0.17.0-beta.1
Vue: v2.5.2
Browsers: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36, Google Chrome
OS: Windows

### Steps to reproduce
Steps are in codepen.

### Expected Behavior
It should make the selection and close the menu, just as if you click on the text area

### Actual Behavior
When you click Apple, the menu stays open and won't go away.

### Reproduction Link
[https://codepen.io/jayblanchard/pen/JOYrom?editors=1010](https://codepen.io/jayblanchard/pen/JOYrom?editors=1010)

### LINKED ISSUE #2411: [Bug Report] v-select with tags writes to array directly
### Versions and Environment
Vuetify: v0.16.9
Vue: v2.4.2
Browsers: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36
OS: Linux

### Steps to reproduce
Add 2 Tags into the Component in the codepen and watch the console

### Expected Behavior
It should trigger the computed setter and not write the value directly to the array (which may be stored in vuex and therefore will fail)

### Actual Behavior
it triggers the computed setter (which is correct) but also writes to the array directly. This triggers an error in vuex strict mode.

### Reproduction Link
[https://codepen.io/leuprechtroman/pen/ZaQWmR?editors=1010](https://codepen.io/leuprechtroman/pen/ZaQWmR?editors=1010)

johnleider:
WIP

fixes #2254
fixes #2401
fixes #2100
fixes #2411 

- [x] Fix chip highlight on dark theme
- [x] Add prop to remove chip close
- [x] Fix #2411


src/components/VSelect/mixins/select-watchers.js
```diff
@@ -0,0 +1,114 @@
...
+        this.$nextTick(() => this.setMenuIndex(0))
+      }
+
+      this.genSelectedItems()
+    },
+    menuIsActive (val) {
+      if (!val) return
+
+      this.isBooted = true
+      // this.isActive = true
```
johnleider: remove this

src/components/VSelect/mixins/select-helpers.js
```diff
@@ -0,0 +1,34 @@
+// Helpers
```
johnleider: add doc block

src/components/VSelect/mixins/select-generators.js
```diff
@@ -260,6 +255,18 @@ export default {
         props: item
       })
     },
+    genLabel () {
+      const singleLine = this.singleLine || this.isDropdown
```
johnleider: add space between declaration and selection control

dev/Boilerplate.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
nekosaur: What's this doing in here? Does it have anything to do with the PR/v-select?

PR #2440, opened 2017-11-08
fix: TypeScript import error

pharra:
TypeScript require ‘export default class’ when using es2015 module. It can't be 'export = class'  .

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,4 +5,4 @@ declare class Vuetify {
   static install: PluginFunction<never>
 }
 
-export = Vuetify
+export default Vuetify


johnleider:
I am not well versed in TS, can any community members provide feedback on this?

scoobah36:
TL;DR I made the change locally and it works

I ran into the same problem running "typescript": "^2.6.1", "vue": "^2.5.3","vuetify": "^0.17.0" and manually making the change the the vuetify package fixed the import. 

if i try to follow the ts docks and import a packaged that uses the "export =" notation i get:
"[ts] Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format instead."

my testing supports this change.

PR #2446, opened 2017-11-08
allow defining colors for light and dark themes

jacekkarczmarczyk:


KaelWD:
I don't believe this is happening any more, it's easy to just `Object.assign` theme with something custom when you change dark/light

PR #2450, opened 2017-11-09
fix: removed javascript:; declarations

### LINKED ISSUE #2321: v-select fails CSP
### Steps to reproduce
- Use a v-select on a site where the server sets a script-src CSP of ```script-src 'self' https://az416426.vo.msecnd.net/scripts/a/ai.0.js;```
- Click on select to display the options.
- Select an option

Browser reports
```
Refused to execute JavaScript URL because it violates the following Content Security Policy directive: "script-src 'self' https://az416426.vo.msecnd.net/scripts/a/ai.0.js". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution.
```

CSP report payload
```json
{
    "csp-report": {
        "document-uri": "http://localhost:8043/",
        "referrer": "",
        "violated-directive": "script-src",
        "effective-directive": "script-src",
        "original-policy": "default-src 'none'; font-src https://fonts.gstatic.com/s/materialicons/ https://fonts.gstatic.com/s/roboto/; script-src 'self' https://az416426.vo.msecnd.net/scripts/a/ai.0.js; style-src 'self' https://fonts.googleapis.com/css; connect-src 'self' https://dc.services.visualstudio.com/v2/track https://techmentor.auth0.com/ https://techmentorapidev.azurewebsites.net/; img-src 'self' data:; object-src 'none'; report-uri https://techmentors.report-uri.io/r/default/csp/enforce",
        "disposition": "enforce",
        "blocked-uri": "inline",
        "line-number": 1,
        "source-file": "http://localhost:8043/",
        "status-code": 200,
        "script-sample": ""
    }
}
```

### Versions
Chrome 62, Windows 10

```json
 "dependencies": {
    "auth0-js": "^8.9.3",
    "axios": "^0.17.0",
    "es6-promise": "^4.1.1",
    "izitoast": "^1.1.5",
    "marked": "^0.3.6",
    "store": "^2.0.12",
    "tz-ids": "^1.0.0",
    "vee-validate": "^2.0.0-rc.14",
    "vue": "^2.4.2",
    "vue-application-insights": "^1.0.3",
    "vue-class-component": "^6.0.0",
    "vue-property-decorator": "^6.0.0",
    "vue-router": "^3.0.1",
    "vuetify": "^0.16.7",
    "vuex": "^3.0.0",
    "vuex-persistedstate": "^2.0.0"
  },
```

### What is expected ?
Selecting an option should not violate a CSP.

Selecting an option appears to be trying to execute a JavaScript url.
v-select csp failure

The enforced CSP failure does not prevent the select from operating as expected.

### Reproduction Link
I can't use a codepen to produce this because it requires the server to provide the CSP in response headers. You can see this in action at https://techmentorwebdev.azurewebsites.net which uses v-select components on the home page.

johnleider:
routable was rendering anchor tags with **javascript:;**. this fails CSP, removing and adding a
default hover for `<a>` tags

fixes #2321


src/components/VAlert/VAlert.js
```diff
@@ -71,7 +71,6 @@ export default {
     if (this.dismissible) {
       const close = h('a', {
```
jacekkarczmarczyk: Why use an `<a>` tag if this is not a link? It should be a `<button>`

PR #2453, opened 2017-11-09
Allow buttons embedded in other components to still function within VBtnToggle and VBottomNav.

Phlow2001:
Added and fixed existing tests.

The pen shows the problem.  First toggle button is unselectable, as is first botton nav button.  Also note that when you Click to Show Extra Button, the extra buttons don't work.  That bug is fixed as well.

Copy into playground with fixes to see that the toggle buttons work, the bottom nav continues to work, and adding and removing buttons works correctly while preserving mandatory invariant.

https://codepen.io/jayblanchard/pen/ooBXKo?&editors=101

### DIFF
--- a/src/components/VBottomNav/VBottomNav.spec.js
+++ b/src/components/VBottomNav/VBottomNav.spec.js
@@ -1,7 +1,7 @@
 import VBottomNav from './VBottomNav'
 import VBtn from '../VBtn'
 import { test } from '~util/testing'
-import Vue from 'vue/dist/vue.common'
+import Vue from 'vue'
 
 function createBtn (val = null) {
   const options = {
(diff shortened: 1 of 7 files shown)


alfreema:
It's not clear to me if this bug is closed or not, but my code is throwing an error that led me here.   The error message has changed since @Phlow2001 posted.   When you toggle the button off it now shows:

 `[Vue warn]: Error in beforeDestroy hook: "Error: Not implemented !"`

I created a simple pen ([https://codepen.io/alfreema/pen/YOrjNj](https://codepen.io/alfreema/pen/YOrjNj)) to demonstrate that hiding a button that is in a downstream component within a v-bottom-nav is the culprit, though I think @Phlow2001 already described that.

alfreema:
Actually it is exactly related to embedding a button downstream of a v-bottom-nav just like the description.  I posted a simplified codepen to easily replicate it.   I am betting it's the same problem, just now manifesting itself with a different error.  I am happy to open a new issue with that codepen if it's helpful though.

PR #2461, opened 2017-11-09
Fix btn--floating overlapping toolbar--fixed

cassvail:
When scrolling down in a page containing a btn--floating (z-index = 4) with a toolbar--fixed (z-index = 2). The button icon overlaps the toolbar


johnleider:
Can you please provide an example implementation that exhibits the bug without your fix?

cassvail:
https://jsfiddle.net/cwhn9vx3/

johnleider:
This was already resolved with a2bd0f2c9197bd0866b506adff02a1cd56e1c701.

PR #2462, opened 2017-11-09
Fix for small chip visual

Phlow2001:
Problem shown: https://codepen.io/jayblanchard/pen/aVpQLy?editors=101


Phlow2001:
Was fixed by john already.

PR #2473, opened 2017-11-11
fix: added missing exports

### LINKED ISSUE #2472: [Bug Report] Vuetify is not exporting all components for use in a-la-carte
### Versions and Environment
**Vuetify:** 0.17.0-beta.3
**Vue:** 2.5.3
**Browsers:** Chrome 62.0.3202.89
**OS:** Windows 10

### Steps to reproduce
Try to import VCardActions in an SFC

```
import { VCardActions } from 'vuetify/es5/components/VCard'
```

### Expected Behavior
VCardActions should be imported

### Actual Behavior
Component is not found

### Other comments
This applies to other components as well. See https://github.com/vuetifyjs/a-la-carte/issues/4 for initial report.

nekosaur:
When importing a-la-carte components in SFCs, the install function is
not used, which means that we were missing some components that were not
explicitly exported.

closes #2472

### DIFF
--- a/src/components/VCard/index.js
+++ b/src/components/VCard/index.js
@@ -3,18 +3,18 @@ import VCard from './VCard'
 import VCardMedia from './VCardMedia'
 import VCardTitle from './VCardTitle'
 
-export { VCard, VCardMedia, VCardTitle }
+const VCardActions = createSimpleFunctional('card__actions')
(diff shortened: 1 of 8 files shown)


vonWolfehaus:
This is currently throwing me off, is there a chance we can get it in 0.17 rather than 1.0 please?

PR #2468, opened 2017-11-10
next merge

nekosaur:
Did a rebase of dev into next, and hopefully resolved all conflicts correctly..

PR #2482, opened 2017-11-12
Next v2

KaelWD:
#2468 but cleaner

Remember to merge **with a merge commit** NOT A SQUASH


KaelWD:
This was force-pushed in 4c00a61 because john temporarily forgot how to read

PR #2487, opened 2017-11-12
Fix v-btn modifier triggering twice

### LINKED ISSUE #2424: [Bug Report] "@click.once" on v-btn triggers twice
### Versions and Environment
Vuetify: v0.16.9
Vue: v2.5.2
Browsers: Google Chrome
OS: Windows

### Steps to reproduce
Add the .once modifier to a click listener on a Vuetify button.

### Expected Behavior
Event fires once.

### Actual Behavior
Event fires twice.

### Reproduction Link
[https://codepen.io/mornir0/pen/bYeNVM](https://codepen.io/mornir0/pen/bYeNVM)

### Other comments
@click.native.once on v-btn works fine.
@click.once on a normal button tag works as expected.

ergun1017:
Fixes #2424

Technical details: 
Vue prepends !, ~, & to event's name for 'capture', 'passive' and 'once' modifiers so overriding 'click' key in vm.$listeners doesn't change '!click' or '~click' event. Because of this, listener with modifier is added twice.

I have solved this issue by first normalizing the event name (deleting the prepended chars) then overriding the key.


johnleider:
Is a utility method required for this? The only logic is essentially a **.replace**. It would make sense if we were passing it arguments for everything and it returned the final object that we are going to pass through to the **on** property.

ergun1017:
Ah, you are very right. Firstly, I thought we need to get also what modifier the event has but then I realized we don't need that but forgot to change the function. So sorry about that.

I'm gonna only replace the chars.

src/mixins/routable.js
```diff
@@ -25,7 +25,9 @@ export default {
     generateRouteLink () {
       let exact = this.exact
       let tag
-
+      const normalizedListeners = Object.keys(this.$listeners).map(event =>
+        Object.create({ [event.replace(/(&|!|~)/g, '')]: this.$listeners[event] })
```
KaelWD: Why `Object.create()`?

The regex can just be `event.replace(/[&!~]/, '')`

src/mixins/routable.js
```diff
@@ -35,11 +37,10 @@ export default {
           value: this.ripple || false
         }],
         on: {
-          ...(this.$listeners || {}),
+          ...(normalizedListeners || {}),
```
KaelWD: The `|| {}` isn't needed any more, use `Object.assign` instead.

src/mixins/routable.js
ergun1017: Uh, I was on a rush; couldn't find a way to return an object from an arrow function that time. But I fixed it know. 😅

KaelWD:
I don't think you can without completely refactoring the function, it was already pretty close to the threshold for all of those warnings.

PR #2525, opened 2017-11-16
Register nav-drawer watchers only if needed

Frondor:
The watcher was being registered whether its functionality was disabled or not via component's props.
So I created a `registerWatchers` method (and its counterpart for unregistering), which can be used for opt-in functionality.
More watchers can be moved to this implementation in the future


Frondor:
No idea what happened with package.json

PR #2526, opened 2017-11-16
Register nav-drawer watchers only if needed

Frondor:
The watcher was being registered whether its functionality was disabled or not via component's props.
So I created a registerWatchers method (and its counterpart for unregistering), which can be used for opt-in functionality.
More watchers can be moved to this implementation in the future

### DIFF
--- a/src/components/VNavigationDrawer/VNavigationDrawer.js
+++ b/src/components/VNavigationDrawer/VNavigationDrawer.js
@@ -8,6 +8,11 @@ import ClickOutside from '../../directives/click-outside'
 import Resize from '../../directives/resize'
 import Touch from '../../directives/touch'
 
+const watchers = {
+  unwatchRoute: Function,
+  unwatchResize: Function
+}
+
 export default {
   name: 'v-navigation-drawer',
 
@@ -136,11 +141,6 @@ export default {
   },
 
   watch: {
-    $route () {
-      if (this.reactsToRoute) {
-        this.isActive = !this.closeConditional()
-      }
-    },
     isActive (val) {
       this.$emit('input', val)
 
@@ -194,6 +194,7 @@ export default {
 
   beforeMount () {
     this.checkIfMobile()
+    this.registerWatchers()
 
     if (this.permanent) {
       this.isActive = true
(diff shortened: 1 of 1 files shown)


KaelWD:
Please specify what problem this solves, along with some code that will reproduce it.

Frondor:
I already explaimed the problem if you read the first line. This is not a bugfix, but an enhancement.
From discord chat:

> Looking at the watchers in v-navigation-drawer component I see `$route` which calls the computed property `reactsToRoute`, and this one uses prop `disable-route-watcher` **which actually is not disabling the watcher** (the watcher is registered whether or not that prop is true). I think it would be a better approach to use `unwatchRoute = this.$watch(...)` prototype function on `mounted()` and unwatch it on `destroy()`. But this time, checking first if  `disable-route-watcher` is true.

KaelWD:
> The watcher was being registered whether its functionality was disabled or not via component's props.

I don't understand how this is a problem. In discord you were saying something about performance but I don't see any performance problems caused by this listener.

Frondor:
If you don't want to accept it, go ahead and close this PR. 
@Phlow understood the point on this, you're probably confussing many discussions we had in pararell about useless watchers this component has. This pr is not addressing that.

That watcher is being registered no matter if the dev disabled its functionality or not, every time route changes **it fires** and then checks if its disabled original not.
If you read the code, all it does is movimiento the watcher to a method, conditionally register it if needed.

KaelWD:
Thanks for your contribution, but I'm going to reject this on the fact that it prevents the temporary disabling of the route watcher. Also `const watchers` is global state, so this will break with more than one drawer.

Frondor:
Doesn't prevents anything, the functionality is still the same. And the `const` is not global if you set a `'use strict'` at the top of the file. The PR was a starting point, so it was subject to changes.
You didn't even paid attention to the code it seems.
There's no problem anyway, since I replaced this library in the moment I started to find all these inconsistencies in the code, leading to poor performance of most layout components.

KaelWD:
> the functionality is still the same

You can't temporarily disable the watcher with this change, it is created when the component mounts. You'd need another watcher to remove the first one, which defeats the whole point of this. 

> the `const` is not global if you set a `'use strict'` at the top of the file

ES modules are implicitly strict already. The `watchers` object isn't global in the sense that it's available under `window`, but it *will* be shared among any other instances of the component.  

I even made an example because I don't like to be wrong - https://codesandbox.io/s/5wv2ypv00k  
Type something in one of the fields and click the button next to the other one.

PR #2567, opened 2017-11-19
feat(theme): Create util file with color variables

### LINKED ISSUE #2564: [Feature Request] Access material colors palette in JS
### New Functionality
Obtain the HEX of a color in JS like in stylus, for example this.$vuetify.colors.red.base

### Improvements
I need it for defining the colors of Vuetify theme and for other parts of the app, like coloring charts using material palette.

When I need colors in JS I have to copy paste from https://material.io/guidelines/style/color.html#color-color-palette

Using raw HEX colors is really messy, when you read the src code it's impossible to understand which shade you used and when you need a color you need to copy paste the HEX from material.io website

KaelWD:
This adds about 1.7kB to the gzipped js bundle. 

Resolves #2564

UPDATE: This no longer adds colors to $vuetify, but you can import them manually from `'vuetify/es5/util/colors'`

### DIFF
--- a/src/components/Vuetify/index.js
+++ b/src/components/Vuetify/index.js
@@ -1,5 +1,6 @@
 import application from './mixins/application'
 import theme from './mixins/theme'
+import colors from '../../util/colors'
 
 const Vuetify = {
   install (Vue, opts = {}) {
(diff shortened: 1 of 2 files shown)


johnleider:
I like this, but should it be included by default?

PR #2590, opened 2017-11-21
fix: table items should not be filtered if the search if empty

jacekkarczmarczyk:


src/components/VDataTable/VDataTable.js
```diff
@@ -99,6 +99,8 @@ export default {
       type: Function,
       default: (items, search, filter, headers) => {
         search = search.toString().toLowerCase()
+        if (search.trim() === '') return items
```
Phlow2001: Should spaces not be a valid search by default?

jacekkarczmarczyk: Probably 99,9% of search with only spaces is by accident [citation needed :P]. I can't image a situation where you would want to search for spaces unless it's something very specific (and in this case you'll probably need to write your own custom filter anyway)
I would go even further and pass trimmed `search` to filtering method

PR #2630, opened 2017-11-24
fix/2551 missing keys

### LINKED ISSUE #2551: [Bug Report] Keys need to be added to rendered output that iterates
### Versions and Environment
**Vuetify:** 0.17.1
**Vue:** 2.5.4
**OS:** Windows 10

### Steps to reproduce
Run reproduction, notice that the scoped CSS does not apply to the first node of the data table. (code is in components/ContourTable.vue)

### Expected Behavior
https://codesandbox.io/s/9jnz0mn24y
Here is the same code, with unique keys on the td's of the data table.

### Actual Behavior
See repro

### Reproduction Link
[https://codesandbox.io/s/7jqkxrom01](https://codesandbox.io/s/7jqkxrom01)

### Other comments
This is due to previous render executions creating the "no-data" row prior to scoped CSS data attrs being applied, and the absence of a unique key on the nodes to tell Vue that it can't reuse it, so it does.
We should make sure we're applying keys appropriately, to areas that render content that change with reactivity, throughout the codebase.  Most notably for-loops.

nekosaur:
closes #2551

### DIFF
--- a/src/components/VBreadcrumbs/VBreadcrumbs.js
+++ b/src/components/VBreadcrumbs/VBreadcrumbs.js
@@ -41,7 +41,7 @@ export default {
     /**
      * Add dividers between
      * v-breadcrumbs-item
-     * 
+     *
      * @return {array}
      */
     genChildren () {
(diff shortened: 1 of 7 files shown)


onlyurei:
Causes #2707.

PR #2033, opened 2017-10-06
WIP: Add disabled prop to v-form

KaelWD:
This would probably be better off as an injection once vuejs/vue#6322 is released, but I'm not sure how to make them reactive. 

Closes #1707

PR #2632, opened 2017-11-24
exclude nested forms from validation

neelance:
If a v-form contains a v-dialog which in turn contains another v-form, then it breaks the validation of the outer form, since inputs of the inner form affect the outer one.

Is this an acceptable solution to the issue?

### DIFF
--- a/src/components/VForm/VForm.js
+++ b/src/components/VForm/VForm.js
@@ -34,6 +34,9 @@ export default {
 
       const search = (children, depth = 0) => {
         for (const child of children) {
+          if (child.$options.name === 'v-form') {
+            continue
+          }
           if (child.errorBucket !== undefined) {
             results.push(child)
           } else {


johnleider:
Please do not push with **--no-verify**.

neelance:
I did not use any `--no-verify`.

johnleider:
The test is re-running, but it had previously failed with lint errors, you shouldn't be able to commit or push with test/lint errors without **--no-verify**.

neelance:
I used the GitHub web UI for this small changed. Did a full checkout now since I saw the linter errors.

neelance:
CI is happy now, but codecov isn't. Unfortunately I don't have the time to get to know your testing setup and a test case with a nested v-form and triggering validation seems non-trivial. What's your policy on this?

KaelWD:
You can ignore that, it's thinking there's changes in VTimePicker for some reason.

neelance:
Okay, thanks. So is this a good solution and ready to merge?

nekosaur:
It's a solution for now. Forms will most likely be refactored to use inject/provide for 1.1. Although I am actually not sure how to solve this problem when we do..

johnleider:
Forms are currently being refactored to use provide/inject for 1.1. Was already mentioned here https://github.com/vuetifyjs/vuetify/pull/2033. Thank you for your contribution.

PR #2633, opened 2017-11-24
fix/2558 v-btn icon padding

### LINKED ISSUE #2558: [Bug Report] <v-btn icon large> has too much padding
### Versions and Environment
**Vuetify:** 0.17.1
**Vue:** 2.5.4
**Browsers:** Chrome 62.0.3202.89
**OS:** Windows 10

### Steps to reproduce
Hover the left button, then slowly move the mouse to the right

### Expected Behavior
The large button should begin hovering halfway between them

### Actual Behavior
The large button hover begins as soon as you leave the visible area of the first icon, due to the `    padding: 0 32px` on its `.btn__content`

### Reproduction Link
[https://codepen.io/anon/pen/POEbKX?editors=1010](https://codepen.io/anon/pen/POEbKX?editors=1010)

MisRob:
Fixes #2558

Ignore size props (`small`, `large`) padding when
using it together with `icon` prop.

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -102,7 +102,9 @@ theme(button, "btn")
     min-width: 0
 
     .btn__content
-      padding: 0
+      // overwrite size padding when using `icon` prop together
+      // with one of size props (`small`, `large`)
+      padding: 0 !important
 
       &:before
         border-radius: 50%


MisRob:
Hi, can anyone advice me how could I fix not passing code coverage checks? Maybe @jacekkarczmarczyk or @KaelWD?This commit is really simple, I am not sure what should I do now. Thanks

jacekkarczmarczyk:
No idea, but i wouldn't worry about this if it's just a css change

MisRob:
OK, I thought so, but this is my first commit so I just wanted to be sure :) Thank you

KaelWD:
Thank you for your contribution, but I've decided to fix this without using `!important`: 8d6c8b7a782a5eba341cdf5aee172940cdf4bf16

MisRob:
Hi, ok, no problem at all :)!  I considered more solutions without using `!important` too but it seemed a little bit nicer to let the attribute where it was originally to me, for the sake of clarity. I had a look at the commit 8d6c8b7, why is it placed in /* Sizes */ section now? I would like to understand this more as I plan to try to commit more in the future :) Thanks!

KaelWD:
I put it below the sizes so it can override them

MisRob:
Yeah, sure, but the question was meant more like why do you prefer this (for example you are sure it won't be overwritten somewhere below in the future or something like that), not what is happening there 'technically' 😉 thanks

KaelWD:
I'm not sure it won't be overridden somewhere else, but `!important` is a last resort. If we use `!important` here, the only way someone can override that in their own code is by also using `!important`. Simply moving the rule like I did may make it less readable, but it also keeps it flexible.

MisRob:
Ok, it's clear now, thanks 🙂

PR #2643, opened 2017-11-26
fix/2641 clearable disabled input

### LINKED ISSUE #2641: [Bug Report] Clearable v-text-field shouldn't delete text when disabled
### Versions and Environment
**Vuetify:** 0.17.1
**Vue:** 2.5.3
**Browsers:** Chrome 63.0.3239.52
**OS:** Mac OS 10.13.1

### Steps to reproduce
Create v-text-field with some initial text. Make it clearable and disabled. 

### Expected Behavior
When user clicks on the clear ("X") button the current text should not be deleted if the field is disabled.

### Actual Behavior
When the user clicks on clear button the current text is deleted even when the field is disabled.

### Reproduction Link
[https://codepen.io/johnjleider/pen/bgJOrX](https://codepen.io/johnjleider/pen/bgJOrX)

manico:
This PR fixes issue in #2641.
This is solved by creating relation between disabled and clearable prop in input mixin, since it does not make sense that clearable is visible when input (or select) is disabled.

I have created this against dev branch since it does not seem as a high priority fix and issue can be solved by not using clearable prop along with disabled.

Fixes #2641


src/mixins/input.js
```diff
@@ -63,6 +63,9 @@ export default {
         'theme--light': this.light
       }, this.classes)
     },
+    isClearable () {
+      return this.disabled ? false : this.clearable
```
KaelWD: `return this.clearable && !this.disabled`

manico: No problem, will include.
I did npm install so lockfile got updated.
Should I revert lock file or will you handle that on merge?

jacekkarczmarczyk: Yes, remove it please, you can push --force

Dylan-Chapman:
Shouldn't ```readonly``` and ```clearable``` be allowed to be used in conjunction? For example if I have a field ```readonly``` because it's displaying the value for a date or time picker, but it's not required and is thus ```clearable```?

KaelWD:
Good point @Dylan-Chapman, I didn't think about that one.

manico:
Yup. So, back to just handling disabled and clearable?

johnleider:
I do not agree with the above mentioned. A read-only field is not editable, and should be sent when the form is submitted. Being able to clear that would break that logic.

KaelWD:
How should pickers work with clearable read-only fields then?

manico:
There is a logic for pickers, so why not just leave this decision to developer? If it is used for pickers clearable should be set to false.

jacekkarczmarczyk:
dev could use `append-icon` + `append-icon-cb`

KaelWD:
Or dev could not use clearable and readonly if they don't want it to be cleared?

jacekkarczmarczyk:
You may want to block editing (typing) but not clearing
https://codepen.io/anon/pen/Ebrpvo

johnleider:
We will handle this bug, but not until v1.1 and it will be handled differently due to some changes. For now, the dev can do this workaround: **:clearable="!disabled"** where **disabled** is the property disabling the input.

PR #2649, opened 2017-11-27
fix/2617 wrapped expansion panel

### LINKED ISSUE #2617: [Bug Report] Expansion panel does not open when wrapped in custom component
### Versions and Environment
**Vuetify:** 0.17.2
**Vue:** 2.5.7
**OS:** Windows 7

### Steps to reproduce
I am unsure whether this should be filed as a bug or feature request, but it appears to me as a bug.

Basically I would like to wrap my expansion panels in custom components in order to add some extra functionality and keep it clean and separate. However, when doing this the panel won't open when I click on the header.

See provided codepen.

### Expected Behavior
The panel should open

### Actual Behavior
The panel does not open

### Reproduction Link
[https://codepen.io/anon/pen/RjydZy](https://codepen.io/anon/pen/RjydZy)

### Other comments
The resulting markup is the same as when just using the Vuetify components directly, i.e. my component is not adding any additional nesting.

As a workaround I have added by own @click handler to the header slot element which simply toggles the v-model. This kind of works, but has some issues, mainly:

- Not firing when clicking on the up/down arrow
- Not collapsing on "click outside"
- Header padding cause the clickable area to be less than the visible header area

Is this something that can be fixed or is it a limitation of Vue?

manico:
Not sure if this should even be handled, but in the issue reported problem is that even though DOM seems like same, virtual dom is not and vue is wrapping (as expected) expansion panel inside panel wrapper component.

I have created small fix/hack to deal with this.

Anyways, this part of the code should be refactored.
Do not like the part where name is compared `c.$options.name === 'v-expansion-panel-content'`. Maybe it is better to check for existence of toggle method, or something like that.

Fixes #2617

### DIFF
--- a/src/components/VExpansionPanel/VExpansionPanel.js
+++ b/src/components/VExpansionPanel/VExpansionPanel.js
@@ -24,8 +24,8 @@ export default {
   methods: {
     getChildren () {
       return this.$children.reduce((toggleables, child) => {
-        const component = child.$children[0] || child;
-        const options = component.$options;
+        const component = child.$children[0] || child
+        const options = component.$options
         if (options && options.name === 'v-expansion-panel-content') {
           toggleables.push(component)
         }
--- a/src/components/VExpansionPanel/VExpansionPanel.js
+++ b/src/components/VExpansionPanel/VExpansionPanel.js
@@ -23,11 +23,15 @@ export default {
 
   methods: {
     getChildren () {
-      return this.$children.filter(c => {
-        if (!c.$options) return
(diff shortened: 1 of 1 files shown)


Phlow2001:
VExpansionPanel and VExpansionPanelContent should be refactored to use registrable mixin so dom searches and name checks aren't needed.

johnleider:
Phlow is correct. We are moving away from iterating children. Instead, check out the **registrable** mixin. If you want to create another PR using that, we will review it. Thank you for your contribution.

PR #2650, opened 2017-11-27
fix(speed-dial): Allow any element inside the activator button

### LINKED ISSUE #1318: Speed-dial activator button should also work with avatars
### Steps to reproduce
I have this codepen https://codepen.io/epam20/pen/dzzZyN
On version 0.13.1 I could use this code

```
<v-icon>
    <v-avatar>
        <img v-bind:src="avatar" class="user-avatar">
    </v-avatar>
</v-icon>
<v-icon>close</v-icon>
```

to control the activator button's two states and the animations where working
properly.

On version 0.14.8 the same code breaks with a console warning:

[Vue warn]: Error in render function: "TypeError: Cannot read property 'indexOf' of undefined"

I know this is because v-icon is supposed to work only for icons like fa, mdi and so on, but 
the speeddial activator states should not only work for any other arbitrary element too, such
as avatars.

On 0.14.8 I could fix this, thanks to @nekosaur with this code

```
<!--<v-fab-transition>-->
<transition name="fab-transition" mode="out-in">
    <v-avatar v-if="!fab">
        <img v-if="signedIn" v-bind:src="avatar" class="user-avatar">
        <img v-else src="images/guest_profile.png" class="user-avatar">
    </v-avatar>
    <v-icon v-else>close</v-icon>
</transition>
<!--</v-fab-transition>-->
```

### Versions
Codepen is in 0.13.1 with a working example
Issue exists on both versions

### What is expected ?
The behaviour, as is, is working fine. What I would like to achieve is to use images or other
elements as activators on the speed-dial, while maintaining the activator's transition animation

Maybe instead of using two icons in the activator button we could use two generic elements that we
can add avatars, or icons or whatever, such as two v-containers (tried with those and they don't
work)

Currently, if you have a combination of avatar and icon as activators, i.e. see code below

```
<v-avatar>
    <img src="http://lorempixel.com/200/200" class="user-avatar">
</v-avatar>
<!--<v-icon>account_circle</v-icon>-->
<v-icon>close</v-icon>
```

in https://codepen.io/epam20/pen/ayyEwY

the fab transition animation is not working, the avatar and the icon are shown side by side

### Reproduction Link
Reproduction links are given in the question above but for convenience here they are again:

The one that worked with 0.13.1:	https://codepen.io/epam20/pen/dzzZyN
The one that doesn't:				https://codepen.io/epam20/pen/ayyEwY

KaelWD:
- Uses `:last-child` instead of `.icon:last-of-type` to allow non-icon elements

Fixes #1318

Playground.vue:
```html
<template>
  <boilerplate>
    <v-card>
      <v-card-text>
        <v-speed-dial
          v-model="fab"
        >
          <v-btn
            slot="activator"
            fab
            hover
            v-model="fab"
          >
            <v-avatar>
              <img src="https://unsplash.it/200/200/?random" alt="">
            </v-avatar>
            <v-icon>close</v-icon>
          </v-btn>
          <v-btn
            fab
            dark
            small
            color="green"
          >
            <v-icon>edit</v-icon>
          </v-btn>
          <v-btn
            fab
            dark
            small
            color="indigo"
          >
            <v-icon>add</v-icon>
          </v-btn>
          <v-btn
            fab
            dark
            small
            color="red"
          >
            <v-icon>delete</v-icon>
          </v-btn>
        </v-speed-dial>
      </v-card-text>
    </v-card>
  </boilerplate>
</template>

<script>
export default {
  data: () => ({
    fab: false
  })
}
</script>
```

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -142,24 +142,27 @@ theme(button, "btn")
     &:after
       border-radius: 50%
 
-    .icon
-      &:not(:only-of-type):last-of-type
-        opacity: 0
-        position: absolute
-        transform: rotate(-45deg)
-        left: calc(50% - 12px)
-        top: calc(50% - 12px)
+    .btn__content
+      :not(:only-child)
+        transition: $primary-transition
+
+        &:last-child
+          opacity: 0
+          position: absolute
+          transform: rotate(-45deg)
+          left: calc(50% - 12px)
+          top: calc(50% - 12px)
 
     &.btn--active
-      .icon
-        &:not(:only-of-type)
-          &:first-of-type
+      .btn__content
+        :not(:only-child)
+          &:first-child
             opacity: 0
             transform: rotate(45deg)
             left: 'calc(50% - %s)' % (24 / 2)
             top: 'calc(50% - %s)' % (24 / 2)
 
-          &:last-of-type
+          &:last-child
             opacity: 1
             transform: none
 
@@ -258,14 +261,14 @@ theme(button, "btn")
   &--small
     font-size: $button-small-font-size
     height: $button-small-height
-    
+
     .btn__content
       padding: $button-small-padding
 
   &--large
     font-size: $button-large-font-size
     height: $button-large-height
-    
+
     .btn__content
       padding: $button-large-padding
 


src/stylus/components/_buttons.styl
```diff
@@ -142,24 +142,27 @@ theme(button, "btn")
...
-      .icon
-        &:not(:only-of-type)
-          &:first-of-type
+      .btn__content
+        :not(:only-child)
+          &:first-child
             opacity: 0
             transform: rotate(45deg)
             left: 'calc(50% - %s)' % (24 / 2)
             top: 'calc(50% - %s)' % (24 / 2)
```
KaelWD: Also, this generates invalid CSS (`calc(50% - 12)` instead of `calc(50% - 12px)`)

johnleider:
When I tested this, the x icon was always shifting position after the animation
bug

KaelWD:
That shouldn't happen, the avatar is supposed to rotate and fade out as if it was an icon. Is that with the playground I posted?

johnleider:
It is with your playground. All I did was add the **dark** prop to the `v-btn` in the activator slot.

KaelWD:
I couldn't reproduce that issue. There is a slight (0.5px) shift that's also present on dev, but the avatar works as expected.

PR #2662, opened 2017-11-28
Add keys to each top level TR that comes from items slot.

Phlow2001:


src/components/VDataTable/mixins/body.js
```diff
@@ -36,17 +36,23 @@ export default {
...
             attrs: { active: this.isSelected(item) }
           })
-          : row)
+        } else {
+          for (let childIndex = 0; childIndex < row.length; ++childIndex) {
+            const rowChild = row[childIndex]
+            rowChild.key = rowChild.key || `${index}:${childIndex}`
+          }
+        }
+        rows.push(row)
```
KaelWD: If you want speedy fast, `rows` should be initialised to the length you need, and use `rows[i]` instead of push.

nekosaur:
@Phlow2001 codebeat issues need to be resolved

KaelWD:
Closing in favour of #3968

PR #2665, opened 2017-11-29
support font awesome 5 icons

BePsvPT:
Font Awesome 5 introduces new type of icons, like regular, light, brands and logos. When using these icons, each prefix name is different. For example, when using `facebook` icon, the class name should be `fab fa-facebook` instead of `fa fa-facebook`.

Currently, when using font awesome 5 icons and the icon is not solid type(solid is default type, prefix with `fa`), the icon class name is incorrect. This pull request add the function that support font awesome 5 icons.

Icon class name comparison:

| icon | before | after |
|-----|---------|---------|
| fa-check | fa icon fa-check| fa icon fa-check |
| far-check | far icon far-check | far icon fa-check |
| fab-facebook | fab icon fab-facebook | fab icon fa-facebook |

### DIFF
--- a/src/components/VIcon/VIcon.js
+++ b/src/components/VIcon/VIcon.js
@@ -59,7 +59,11 @@ export default {
 
     let iconType = 'material-icons'
     const thirdPartyIcon = iconName.indexOf('-') > -1
-    if (thirdPartyIcon) iconType = iconName.slice(0, iconName.indexOf('-'))
+
+    if (thirdPartyIcon) {
+      iconType = iconName.slice(0, iconName.indexOf('-'))
+      if (['fas', 'far', 'fal', 'fab'].some(val => iconType.includes(val))) iconType = ''
+    }
 
     data.staticClass = (`${iconType} icon ${data.staticClass || ''}`).trim()
     data.attrs = data.attrs || {}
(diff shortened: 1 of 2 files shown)


Phlow2001:
It would be preferable to keep the icon name unmodified.
I would prefer changing it "fab fa-facebook" and have the code check for "fab " and "far " before assuming '-' is important.

johnleider:
I agree with @Phlow2001 on this one. Thoughts @BePsvPT ?

BePsvPT:
In font awesome 5, use `<v-icon>fab fa-facebook</v-icon>` instead of `<v-icon>fab-facebook</v-icon>` and in background, we just do nothing and return it when find `fab` or `far`. Am I understanding correctly?

dwyer14:
+1, this would be a great addition to the 1.0 release!

BePsvPT:
In latest font awesome 5, it seems `fa` prefix has been deprecated.

https://fontawesome.com/how-to-use/svg-with-js#styles-and-prefixes

BePsvPT:
@Phlow2001 I have changed the code to check iconType instead of modified the icon name.

nekosaur:
@BePsvPT Please fix the linting error and we'll merge this

BePsvPT:
@nekosaur How do I fix this error?

/home/travis/build/vuetifyjs/vuetify/src/components/VIcon/VIcon.js
error  Method 'render' has too many statements (25). Maximum allowed is 24  max-statements

nekosaur:
render function is too long. max number of lines for functions is 24. it is currently 25

johnleider:
Any update on this?

PR #2666, opened 2017-11-29
Bugfix: Changing VDataTable mixin to reset expanded value to object instead o…

GregYankovoy:
Clicking a VDataTable header triggers a reset of the expanded property.  The functionality requires an object, but this mixin sets it to an array. I believe this was missed in a previous bugfix.

The problematic code causes VDataTable numeric string item-key values to create an array the size of the item-key value. What we want is an object with an item-key to bool mapping.  This change is tested and working.


GregYankovoy:
How do I make sure someone merges this in?

PR #2672, opened 2017-11-29
Add events array to VDatePicker

nekifirus:
Hello! This is my first contribution to Vuetify. In my project I need event calendar, but I don't find working solutions with Material Design and Nuxt. 

But all I need - mark some days as events dates. VDatePicker has functions for allowedDates, I'm add analogy for eventDates.

Maybe this thing will help someone.

### DIFF
--- a/src/components/VDatePicker/VDatePicker.js
+++ b/src/components/VDatePicker/VDatePicker.js
@@ -54,7 +54,7 @@ export default {
     // array of dates to mark days in picker
     eventDates: {
       type: Array,
-      default: []
+      default: () => (null)
     },
     // Function formatting the day in date picker table
     dayFormat: {
@@ -270,9 +270,9 @@ export default {
 
       return true
     },
-    isEventDate(date) {
+    isEventDate (date) {
       // function to check date in eventDates array or not
-      if (!this.eventDates.length) return false
+      if (!this.eventDates) return false
       return this.eventDates.indexOf(date) > -1
     },
     genTableTouch (touchCallback) {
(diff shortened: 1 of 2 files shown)


KaelWD:
Have you seen #989? I don't like this solution as it just hijacks the active class.

jacekkarczmarczyk:
1. There is a massive picker refactoring on the way https://github.com/vuetifyjs/vuetify/pull/2664 so it might be better to use the `refactor/picker` branch as a base instead of `master`

2. Agree with Kael that it should hijack the `btn--active` class, use `date-picker-table__event` class on `td` instead of `.btn` (it will give us and devs more styling possibilities) and apply similar styling as in #989, `eventColor` prop would be nice

3. In the `refactor/picker` branch `allowedDates` prop uses `isValueAllowed` util which can take array/function/object as an argument, use the same mechanism also for events

4. I'd call the prop just `events` instead of `eventDates`

nekifirus:
@jacekkarczmarczyk 

1. Ok - switching to your branch
2. Now it looks [like this](https://imgur.com/X7tuIt0)
I think this is fine as a default solution. But eventColor prop is better 
3. What a structure of object will be to mark days as event days? The best solution - array of objects each over has a 'date' key.
4. Ok - I'm simply try to repeat allowedDates.

jacekkarczmarczyk:
2. imho it looks too disturbing and when the selected date is an event date you won't be able to tell that there's an event on that day

3. object needs to have `min` and `max` properties, but you don't need to worry about this, just use it same as it's used for checking allowed dates

johnleider:
I do like this functionality, but would only consider exploring it after @jacekkarczmarczyk 's refactor is in place.

PR #2676, opened 2017-11-30
Refactor/picker events

nekifirus:
#2672


src/util/isValueEvent.js
```diff
@@ -0,0 +1,17 @@
+export default (value, events) => {
```
jacekkarczmarczyk: No need to duplicate functionality of `isValueAllowed`, just use the existing function (probably method name could be changed to more generic, but that can be done later unless you can suggest something now)

src/components/VDatePicker/VDatePickerDateTable.js
```diff
@@ -50,13 +51,14 @@ export default {
...
-    genButtonClasses (day, disabled) {
+    genButtonClasses (day, disabled, event) {
       const isActive = this.isActive(day)
       const isCurrent = this.isCurrent(day)
       const classes = Object.assign({
         'btn--active': isActive,
         'btn--outline': isCurrent && !isActive,
-        'btn--disabled': disabled
+        'btn--disabled': disabled,
+        'date-picker-table__event': event
```
jacekkarczmarczyk: add class to `td` instead of button

src/util/isValueEvent.js
nekifirus: completed in e967154

jacekkarczmarczyk: Hrm, what I meant is that you can just use `isAllowedDate` in the component: https://github.com/vuetifyjs/vuetify/pull/2676/files#diff-5f403996a38f48f5de2879d0f0471a28R72

`const isEvent = isValueAllowed(date)` instead of `const event = isValueEvent(date)`

There's no need to create a separate util that doubles the functionality of `isValueAllowed`, test is also not necessary as there is a test for `isValueAllowed`, sorry if it wasn't clear before

johnleider:
@jacekkarczmarczyk Is this still relevant with your refactor?

jacekkarczmarczyk:
Still changes required (remove isValueEvent function, apply class to td instead of button and use different styling)

nekifirus:
@jacekkarczmarczyk No need to to remove isValueEvent function!

johnleider:
At this time we are not exploring any new functionality to v-date-picker until @jacekkarczmarczyk 's refactor is merged.

PR #2678, opened 2017-11-30
Fix ripple directive binding issues

### LINKED ISSUE #2677: [Bug Report] Checkbox ripple is not centered if checkbox starts out disabled
### Versions and Environment
**Vuetify:** 0.17.3
**Vue:** 2.5.9
**OS:** Linux x86_64

### Steps to reproduce
1. Click "disabled" to uncheck it and enable the next checkbox
2. Click "check"

### Expected Behavior
The ripple effect should be centered on the checkbox

### Actual Behavior
The ripple effect is centered on the mouse pointer

### Reproduction Link
[https://codepen.io/anon/pen/jaQvmQ](https://codepen.io/anon/pen/jaQvmQ)

nkovacs:
Unbind was incorrect, it did not remove the registered event handlers.
Update did not update the value used by the event handler,
so a checkbox that is initially disabled would have an event
handler without the center: true option, and enabling the
checkbox would cause the ripple to appear at the mouse pointer's
location.

Fixes #2677


Phlow2001:
While I don't see the problem with the ripple at the pointer, you're right about the removal of handlers.

That being said, I think this should be done differently.  Based on [best-practices](https://vuejs.org/v2/guide/custom-directive.html#Directive-Hook-Arguments), using data attrs to pass data between hooks, we should change the handlers to global, not store them on the dom node, and instead use data attrs to store 'center' and 'class' values, then get the el from the event.

nkovacs:
Should I change `ripple.show` and `ripple.hide` to be the handler, or should I add new handler functions that wrap them?

nkovacs:
I looked into dataset, and I really don't like it, for several reasons:
- it's a stringmap, so everything has to be a string
- apparently on Safari 10, you can't delete properties: https://stackoverflow.com/questions/9200712/how-to-remove-data-attributes-using-html5-dataset#comment78166546_9201264
- it's not supported on IE9-10 (I know, vuetify doesn't support those anyway)
- It's 90% slower (probably because it has to change the dom): https://jsperf.com/data-dataset

The migration guide recommends setting properties on the element, and Evan You said it was fine: https://github.com/vuejs/vue/issues/5251#issuecomment-288290410

Also there's another bug. If you disable the ripple directive while there's an active ripple, it won't be hidden: https://codepen.io/anon/pen/LOMdRg

nkovacs:
Also, dataset doesn't work in tests, so all the tests just broke.

Phlow2001:
Sounds good.  If Evan says it's without side-effect (like SSR), great!

nkovacs:
@johnleider Your merge broke the tests, so I rebased the branch against dev again.
It shouldn't break SSR, the difference is that the server-rendered html won't have data-ripple="true", but
the bind function has to run in the browser to set the event handlers anyway, so it doesn't depend on that.

src/directives/ripple.js
```diff
@@ -80,41 +78,80 @@ const ripple = {
...
 
 function unbind (el, binding) {
-  el.removeEventListener('touchstart', e => ripple.show(e, el, binding), false)
-  el.removeEventListener('mousedown', e => ripple.show(e, el, binding), false)
-  el.removeEventListener('touchend', () => ripple.hide(el), false)
-  el.removeEventListener('touchcancel', () => ripple.hide(el), false)
-  el.removeEventListener('mouseup', () => ripple.hide(el), false)
-  el.removeEventListener('mouseleave', () => ripple.hide(el), false)
-  el.removeEventListener('dragstart', () => ripple.hide(el), false)
+  delete el.__ripple
```
johnleider: Is there anyway we could make this `_ripple` as opposed to 2 underlines? Mainly for continuity between the syntax we use for other directives for the same thing.

nkovacs: I'm not sure why I used two underscores, but sure.

src/directives/ripple.js
```diff
@@ -80,41 +78,80 @@ const ripple = {
...
+
+function updateRipple (el, binding, wasEnabled) {
+  const enabled = isRippleEnabled(binding.value)
+  if (!enabled) {
+    ripple.hide(el)
+  }
+  el.__ripple = el.__ripple || {}
+  el.__ripple.enabled = enabled
+  const value = binding.value || {}
+  if (value.center) {
```
jacekkarczmarczyk: I guess this could be just:
```js
  el.__ripple.centered = true
  el.__ripple.class = binding.value.class
```

Issue #2686, opened 2017-12-01
[Bug Report] Checkbox label is unreadable


kreischweide:
### Versions and Environment
**Vuetify:** 0.17.3
**Vue:** 2.5.9
**Browsers:** Chrome 62.0.3202.94
**OS:** Windows 10

### Steps to reproduce
View any v-checkbox element on a mobile device with limited width.

### Expected Behavior
The user can read the full label text.

### Actual Behavior
Label text gets truncated and can a human undestandable context easily.

### Reproduction Link
[https://codepen.io/kreischweide/pen/Vrqddx](https://codepen.io/kreischweide/pen/Vrqddx)

### Other comments
The user has to be able to read the full text that he needs to check/uncheck. A checkbox label should never be considered something that can be truncated or applied an ellipsis to. This is also important in terms of accessibility, which gets much worse if people rely on a higher contrast / larger font to understand the content.

KaelWD:
I created this codepen for one of the patrons a while ago, I think it should probably be the default styling for labels: https://codepen.io/anon/pen/yPxYOR

frutality:
This is workaround for checkboxes: https://codepen.io/frutality/pen/BYdOQm
Didn't check it with all possible states (errors, hints etc). Waiting for 1.1 where it should be addrressed.

PR #2688, opened 2017-12-01
fix(v-checkbox): label overflow now readable

kreischweide:
Just saw the reply from @KaelWD and the codepen workaround for #2686. I tried to come up with a fix for it and found a different way, not sure if its worth the pull request though. First time looking into the internals of vuetify and the element creation (oh and flexbox). Please close if inappropriate!

### DIFF
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -173,10 +173,6 @@ export default {
         }
       }, data)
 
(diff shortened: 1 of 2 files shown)


johnleider:
This will be addressed in 1.1. Due to the types of changes that will be happening because of it, this PR will unfortunately become obsolete.

PR #2703, opened 2017-12-03
fix(application): abstracting applicationable functionality

### LINKED ISSUE #2679: [Bug Report] Navigation Drawer right-watch always resets application padding
### Versions and Environment
**Vuetify:** 0.17.3
**Vue:** 2.5.9
**Browsers:** Firefox 57.0
**OS:** Windows 10

### Steps to reproduce
I installed the webpack-advanced template with vue-cli and had problems with the right temporary Navigation Drawer.
- Click on the top-right button to open the temporary drawer
- Now use the 'Switch drawer' button

### Expected Behavior
Because the right Navigation drawer is temporary, you would expect it to switch from right to left and back, without changing the application paddings, because it is temporary.

### Actual Behavior
You'll see that the toolbar, content and footer lose their left padding. Now the toolbar is behind the left Navigation drawer and it will not come back.

### Reproduction Link
[https://codepen.io/anon/pen/EbOeVq](https://codepen.io/anon/pen/EbOeVq)

### Other comments
The problem is not only in the webpack-advanced template, I think it is also a small mistake in the source of the Navigation Drawer. 

To match the *webpack-advanced* template with the *webpack* and *webpack-ssr* templates, the **app** property of the right Navigation Drawer should probably be changed to **fixed**.

To fix the Navigation Drawer:
*vuetify/src/components/VNavigationDrawer/VNavigationDrawer.js*
In the **watch** of property **right** on line **192**
Original: `if (prev != null)`
I think it should be: `if (prev != null && this.app)`

After these changes all works well and as expected.

johnleider:
Turned applicationable into a factory function, added ability to have a `v-bottom-nav` and `v-footer`. Using the **app** prop now automatically adds **fixed** as we will not support any other behavior. Added **height** prop to `v-bottom-nav`, `v-footer`, and `v-system-bar`. Add tons of new tests. Added missing directive **scroll** in `v-toolbar`. Normalized property names in layout files. Added **inset** prop to `v-footer`. Updated style declarations to be lazy.

fixes #2679

### DIFF
--- a/src/components/VToolbar/VToolbar.js
+++ b/src/components/VToolbar/VToolbar.js
@@ -199,8 +199,8 @@ export default {
     },
     /**
      * Update the application layout
-     * 
-     * @return {integer}
+     *
+     * @return {number}
      */
     updateApplication () {
       return this.invertedScroll
--- a/src/components/VToolbar/VToolbar.js
+++ b/src/components/VToolbar/VToolbar.js
@@ -1,19 +1,31 @@
+// Styles
 require('../../stylus/components/_toolbar.styl')
 
+// Mixins
 import Applicationable from '../../mixins/applicationable'
 import Colorable from '../../mixins/colorable'
 import Themeable from '../../mixins/themeable'
 import SSRBootable from '../../mixins/ssr-bootable'
 
+// Directives
+import Scroll from '../../directives/scroll'
+
 export default {
   name: 'v-toolbar',
 
   mixins: [
-    Applicationable,
+    Applicationable('top', [
+      'clippedLeft',
+      'clippedRight',
+      'height',
+      'invertedScroll'
+    ]),
     Colorable,
     SSRBootable,
     Themeable
   ],
+
+  directives: { Scroll },
 
   data: () => ({
     activeTimeout: null,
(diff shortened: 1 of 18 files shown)


src/components/VApp/VApp.js
```diff
@@ -32,6 +32,9 @@ export default {
   },
 
   computed: {
+    application () {
+      return this.$vuetify.application
+    },
```
johnleider: Should this be removed? I added it but I'm not sure.

KaelWD: I don't see any reason to leave it in.

src/mixins/applicationable.spec.js
```diff
@@ -3,26 +3,102 @@ import { test } from '~util/testing'
 import Applicationable from '~mixins/applicationable'
 
 test('applicationable.js', ({ mount }) => {
-  it('should update application on mount', async () => {
+  it.skip('should update application on mount', async () => {
```
KaelWD: Add a `// TODO: reason` when skipping tests

src/mixins/button-group.js
```diff
@@ -8,6 +8,8 @@ export default {
 
   methods: {
     getValue (i) {
+      if (this.buttons.length < 1) return i
+
```
KaelWD: I don't think this is relevant

src/components/VToolbar/VToolbar.js
```diff
@@ -241,6 +242,10 @@ export default {
       }, this.$slots.extension))
     }
 
-    return h('nav', data, children)
+    return h('nav', data, [
+      h('div', {
+        staticClass: 'toolbar__wrapper'
```
KaelWD: What's this for? I don't see it in any stylesheets

src/mixins/applicationable.js
```diff
@@ -1,15 +1,51 @@
...
+  return {
+    props: {
+      absolute: Boolean,
+      app: Boolean,
+      fixed: Boolean
+    },
+
+    computed: {
+      applicationProperty () {
+        return value
```
KaelWD: This should probably be data instead of a computed property

src/mixins/applicationable.js
```diff
@@ -1,15 +1,51 @@
...
-  props: {
-    app: Boolean
-  },
-
-  watch: {
-    app () {
-      this.updateApplication()
-    }
-  },
+export default function (value, events = []) {
```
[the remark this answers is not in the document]
KaelWD: And that somehow ticked off codebeat 😄

PR #2704, opened 2017-12-03
add vue storybook boilerplate

Maxhodges:
Issue: Storybook for Vue with Vuetify Material Component Framework takes some time to setup. Also new users to Vue may not understand how to wrap a layout around a component in a story (which isn't addressed in the Storybook Vue docs).

What I did

Created a boilerplate Storybook-for-Vue project using Vuetify Material Component Framework. Having something like this would have helped me get started much faster. So I'm sharing for others. Includes stories for a component with two states and a layout which wraps the component.


KaelWD:
> new users to Vue may not understand how to wrap a layout around a component in a story

Looks like you don't either tbh

PR #2709, opened 2017-12-04
enh: add possibility to define previous/next date icons in pickers

### LINKED ISSUE #2123: [Feature Request] Other icons libraries in predefined Components
Feature Request

- Answer those questions:
  **- What will it allow you to do that you can't do today?**
Using mdi icons or other icons supported by the Icon component in case of Material Icons is not installed in the application
Components like Steppers, Selects, Pickers, Data table, Selection Controls : Checkboxes, Radios and Carousels all needs Material Icons.
  **- How will it make current work-arounds straightforward?**
Reduce application size if we don't need Material Icons in the rest of the application
  **- What potential bugs and edge cases does it help to avoid?**

Thank you in advance

jacekkarczmarczyk:
fixes #2123

docs: https://github.com/vuetifyjs/vuetifyjs.com/pull/105

### DIFF
--- a/src/components/VDatePicker/VDatePicker.date.spec.js
+++ b/src/components/VDatePicker/VDatePicker.date.spec.js
@@ -403,4 +403,17 @@ test('VDatePicker.js', ({ mount, compileToFunctions }) => {
     wrapper.setProps({ type: 'date' })
     expect(wrapper.vm.activePicker).toBe('YEAR')
   })
+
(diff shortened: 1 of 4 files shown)


johnleider:
Can we change the terminology to match other areas such as carousel/input groups? **append-icon** and **prepend-icon**

jacekkarczmarczyk:
@johnleider done

PR #2710, opened 2017-12-04
feat(VSelect): allow the use of custom list tiles in items slot

Fl0pZz:
This PR adds an ability for using custom `v-list-tile` as a select item.

~~Look at the example: [Playground.vue](https://gist.github.com/Fl0pZz/2f9edd53d283b8ac1996e74be6de7ec9)~~

Updated example: https://pastebin.com/qrNap6L8

How it looks: 
screenshot_20171204_150548

What is it done for? For example, for adding classes or custom listeners.


jacekkarczmarczyk:
Please add test for this PR

Fl0pZz:
But I can't find any test with scoped slots in the project

jacekkarczmarczyk:
https://github.com/vuetifyjs/vuetify/blob/master/src/components/VSelect/VSelect.spec.js#L119 for example
Search for "scopedSlots" in *.spec.js files

johnleider:
If we were to do this, it would need to be in line with implementations in other components, e.g. https://github.com/vuetifyjs/vuetify/blob/dev/src/components/VDataTable/mixins/body.js#L44

Fl0pZz:
ok, I will try to do it

Fl0pZz:
@johnleider, this can not be done, because table use `slot`, and the select uses the `scoped slot` (because `on`: https://github.com/vuetifyjs/vuetify/blob/master/src/components/VSelect/mixins/select-generators.js#L274)

johnleider:
The table definitely uses a scopedSlot, the logic in the above mentioned **mixins/body.js** would translate just as easy.

Fl0pZz:
[Playground.vue](https://gist.github.com/Fl0pZz/f26a5f4ddc50d02f0fa0c3a671938f8e)

src/components/VSelect/VSelect.js
```diff
@@ -116,6 +116,9 @@ export default {
   },
 
   methods: {
+    needTile (tile) {
```
johnleider: make this plural

src/components/VSelect/mixins/select-generators.js
```diff
@@ -272,14 +272,11 @@ export default {
...
-      if (this.$scopedSlots['item-tile']) {
-        return this.$scopedSlots['item-tile']({ parent: this, item, listTile: data })
-      }
-
       if (this.$scopedSlots.item) {
-        return this.$createElement('v-list-tile', data,
-          [this.$scopedSlots.item({ parent: this, item })]
-        )
+        const tile = this.$scopedSlots.item({ parent: this, item, tile: data })
+        return this.needTile(tile)
```
johnleider: update to plural

src/components/VSelect/VSelect.js
```diff
@@ -116,6 +116,9 @@ export default {
   },
 
   methods: {
+    needsTile (tile) {
+      return !/v-list-tile$/.test(tile.tag)
```
johnleider: Doing some research on this, it appears that a comparison operator is [indefinitely](https://jsperf.com/matchgroup-vs-equal) faster than running .test. Since we have no direct need to compare an expression, can we change this to an absolute **equal**?

Fl0pZz: Definitely, for `tile.tag` we can not seriously change this check, because by my tests it looks like `vue-component-40-v-list-tile`. On the other hand, we can access the property `tile.componentOptions.tag` and use the comparison operator:
`tile.componentOptions.tag === 'v-list-tile'`. However, I do not know whether it is legal

johnleider: I sure hope it's legal, Vuetify has been using componentOptions.tag in various components for quite a long time.

johnleider:
Everything looks good, bonus points if you add a unit test, if not let me know and I'll add one.

Fl0pZz:
Today's tests will not be exactly

src/components/VSelect/VSelect.js
KaelWD: Use name instead of tag, there was a PR a while ago to do that because someone was trying to rename components and tag caused some to not function correctly.

Fl0pZz: @KaelWD i cant find the `name` property in the instance of `v-list-tile`

KaelWD: It's in `$options`: #2312

I'm not sure if that issue will affect this actually, but it might be worth looking into.

nekosaur:
This still needs a test @Fl0pZz

KaelWD:
And conflicts reolved

johnleider:
Looking for another @vuetifyjs/dev-team approval.

PR #2716, opened 2017-12-05
Enhancement/absolute menus

### LINKED ISSUE #795: "v-menu" appears as fixed if parent component is inside a scrolling container

### LINKED ISSUE #1300: Allow me to set the target element for menu content and dialog content
When I use `this.$el.requestFullscreen()` in a component that uses menus or dialogs, their contents won't be shown because they are a few levels up the DOM tree.

An easy way to solve this issue would be to have a prop that takes a dom element instance so you can set a different element to use when in fullscreen mode. Example:

```vue
<template>
<div>
  <v-toolbar>
    <v-menu :target="target">
      <v-btn slot="activator">Open Menu</v-btn>
      <!-- content ...  -->
    </v-menu>
    <v-spacer></v-spacer>
    <v-btn icon @click="setFullscreen"><v-icon>fullscreen</v-icon></v-btn>
  </v-toolbar>

  <div>
    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. Ad asperiores quo impedit nihil blanditiis fugiat ipsum iste aliquid explicabo saepe expedita, optio voluptas veritatis iusto, quas praesentium dolores laboriosam excepturi.</p>
  </div>
</div>
</template>

<script>
export default {
  data() {
    return {
      target: document
    }
  },
  methods() {
    setFullscreen() {
      this.target = this.$el
      this.$el.requestFullscreen()
    }
  }
}
</script>
```

### LINKED ISSUE #2202: [v-menu] Content position not caculated properly with margin and relative position
### Versions
v0.16.7

### What is expected ?
[v-menu] content should shown properly underneath the activator.

The [v-menu] content isn't shown properly when a parent element has the position set to relative and margin attribute.

### Reproduction Link
https://codepen.io/anon/pen/eGQOpa

PS: It is the same to position absolute Elements with a left value: https://codepen.io/anon/pen/eGQOeO

johnleider:
Added new **absolute** prop (want to change, need ideas) that will keep the `v-menu` content within its container (unless specified different from **[data-app]**). Added ability to specify a detachable target, defaults to **[data-app]**.

@Phlow2001 has proposed the prop name of **attach** which I like.

fixes #2202
fixes #1300
fixes #795

### DIFF
--- a/src/components/VDataIterator/VDataIterator.spec.js
+++ b/src/components/VDataIterator/VDataIterator.spec.js
@@ -32,7 +32,7 @@ test('VDataIterator.js', ({ mount, compileToFunctions }) => {
     const content = wrapper.find('.data-iterator div div')[0]
     expect(content.element.textContent).toBe('No matching records found')
 
-    expect('Application is missing <v-app> component.').toHaveBeenTipped()
+    expect('Unable to locate target [data-app]').toHaveBeenTipped()
   })
 
   it('should match a snapshot - hideActions and no footer slot', () => {
@@ -51,7 +51,7 @@ test('VDataIterator.js', ({ mount, compileToFunctions }) => {
     const wrapper = mount(VDataIterator, data)
 
     expect(wrapper.html()).toMatchSnapshot()
-    expect('Application is missing <v-app> component.').toHaveBeenTipped()
+    expect('Unable to locate target [data-app]').toHaveBeenTipped()
   })
 
   it('should match a snapshot - no data', () => {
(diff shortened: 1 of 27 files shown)


johnleider:
@Phlow2001 has proposed the prop name of **attach** which I like.

PR #2721, opened 2017-12-05
fix(menuable): Enforce the check that $refs.content exist in sneakPeek

Seraf:
If `$refs.content` is removed between `sneakPeek` being called and the animation frame fires, it becomes undefined in the callback.

Then it results with a `Uncaught TypeError: Cannot read property 'style' of undefined`

This fix ensure that $refs.content exists.

===
Thanks Kael for the help to debug it :-)


KaelWD:
For future reference: 879277771ff201269e6c07cc5d5d6e7588250d08

PR #2725, opened 2017-12-05
perf: Rewrite filteredDuplicates to be O(n) rather than O(n*n)

Phlow2001:
Basic comparison, old vs new (Set) vs new (Map)
https://codepen.io/jayblanchard/pen/qVwpGz?editors=1010


johnleider:
I really enjoyed working with you on this, 54x increase in speed!

PR #2732, opened 2017-12-06
Update VTabsContent.js for adding option touchless

dbweb-io:
i've added the prop and the method/logic in the render function


dbweb-io:
dev

PR #2738, opened 2017-12-07
Enh/foreach optimizations

jacekkarczmarczyk:
Replaced `forEach` with `for(;;)` or `for ... of` loops (everywhere, just for consistency)


smares:
Just a side-note: According to https://jsperf.com/foreach-vs-for-of only Chrome performs better with for...of vs. forEach, while IE 11, Edge 16 and Firefox 57 are all much worse.

KaelWD:
Vuetify uses babel, yo you should compare https://jsbench.me/7ejagsewtr instead.

jacekkarczmarczyk:
With babel it's even worse :/

Firefox:
for(;;) - 15.000.000 op/s
forEach - 4.000.000 op/s
for of - 450.000 op/s
babel - 170.000 op/s

Chrome:
for(;;) - 9.000.000 op/s
forEach - 710.000 op/s
for of - 3.000.000 op/s
babel - 3.100.000 op/s

smares:
Yup, I can confirm, Firefox for of babel 243,757 ops/s, forEach 6,573,502 ops/s. Edge won't run so I opened https://jsperf.com/foreach-vs-for-loop-vs-for-in-vs-for-of-vs-babel-for-of in Edge and got 6,520 for of babel and 331,662 forEach.

KaelWD:
The jsperf test is broken, that why I moved it to jsbench in the first place.

smares:
I see. Anyways, the test with Firefox was ran on jsbench and there forEach is faster than for of babel. :/

smares:
Just updated from Chrome 62 to 63 and now I am getting better results with forEach even in Chrome.

for(;;) - 11,181,912 ops/s
forEach - 4,298,081 ops/s
for of - 3,903,166 ops/s
babel - 3,919,877 ops/s

jacekkarczmarczyk:
Just in time @smares , I was about to go for `for(;;)`...
I guess with Chrome 63 it's ok to stay with `forEach`, isn't it?

smares:
for(;;) is fastest with every browser, but forEach has decent performance now in Chrome, too.

Issue #2335, opened 2017-10-27
Horizontal scrolling of overflowing content in tabs changes the tab


jacekkarczmarczyk:
### Steps to reproduce
Create swipeable tabs with content wider than the tab width, scroll the tab content

### Versions
0.17

### What is expected ?
Content should be scrolled, tab should not change

### What is actually happening ?
Content is scrolled, later tab is changed

### Reproduction Link
Go to the API section https://dev.vuetifyjs.com/components/tabs in mobile and try to scroll the content to see the props description

KaelWD:
tl;dr use `touchless`

PR #2740, opened 2017-12-07
feature(v-tabs-bar): add touchless prop

abalabahaha:
Feature request by @johnleider in the Discord server

image

Looks like [the touch directive](https://github.com/vuetifyjs/vuetify/blob/bd0ce95/src/components/VTabs/VTabsBar.js#L114) is in `v-tabs-bar` and not `v-tabs-content`, so this PR assumes that was just a typo in chat

### DIFF
--- a/src/components/VTabs/VTabsBar.js
+++ b/src/components/VTabs/VTabsBar.js
@@ -43,6 +43,10 @@ export default {
     }
   },
 
+  props: {
+    touchless: Boolean
+  },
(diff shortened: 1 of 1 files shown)


KaelWD:
Does this fix #2335?

abalabahaha:
Apparently this is the wrong approach, touchless should be targeting the tab content area, not the tab bar.

Issue #2759, opened 2017-12-08
[Enhancement] Snackbar animation is incorrect


KaelWD:
### New Functionality
N/A

### Improvements
The snackbar body should slide in, while only the contents fade. When exiting, it should just slide out without any fade at all. 
https://material.io/guidelines/components/snackbars-toasts.html#specs-figure-caption-0

Consective snackbars should wait until the animation is complete before beginning the next one: https://material.io/guidelines/components/snackbars-toasts.html#usage-figure-caption-6

### Bugs or Edge Cases it Helps Avoid
N/A

PR #2768, opened 2017-12-10
fix/2759 snackbar material design

manico:
To solve #2759 I ended up with quite a refactoring.
Now, I have a few questions to see if I'm going in the right direction.

- Snackbar is rather strict with allowing only message and action button. I have created new props for this to replace default slot as button added to snackbar does not conform to material design (no ripple, no padding, no hover). Would these new props make sense?

- I have added transitions to `snackbar.styl` as it is very custom containing transitions to container and content. Is there a reason that transitions are separated from component style file or is it ok that they are inside like in this example?

- As for #2759 and stacking of snackbars. This requires more refactoring as snackbars should be aware of one another (some global "storage" for this). Any suggestions?


manico:
Seems like wrong branch. Will create another cleaner PR.

Issue #2384, opened 2017-10-31
[Feature Request] Snackbar queue


mitar:
### New Functionality
Currently it is not easy to handle interaction of multiple snackbar messages. They can overlap and hide each other. Moreover, because they are added as components inside other components, if parent component is removed (for example, because of a route change), snackbar message also disappears or is even not shown.

### Improvements
I would propose that there is a global queue of snackbar messages. Maybe by sending a Vue event. Some options could then be to display messages one after the other, or display them stacked, up to some number.

### Bugs or Edge Cases it Helps Avoid
Currently it seems one has to manually implement whole logic how to handle interaction between multiple snackbar messages.

EDIT: I have made a NPM package with this functionality [@tozd/vue-snackbar-queue](https://gitlab.com/tozd/vue/snackbar-queue).

nekosaur:
The reason for not currently supporting multiple snackbar messages is that MD spec specifies that only one snackbar should be displayed at a time.

However, several people have made similar requests previously. And I think there is some merit to at least include basic support for displaying consecutive snackbar messages. Not a fan of providing a global queue in vuetify though.

Will leave this open for discussion.

tafel:
One could create a component (say SnackDisplayer) and, using Vuex and stores, displaying the snackbar outside of any component.

I tried this with a quite successful result. Here is a fiddle that demonstrate this:
http://jsfiddle.net/13uc6mu5/

With this configuration (check fiddle source), one can call `this.$store.commit('msg/set', { message: 'hello' })` in any component and the snack pop up.

One thing that can be improved is indeed the possibility to have consecutive snackbars.

KaelWD:
I think it'd make sense to allow the `v-snackbar` to handle the message queue internally.

mitar:
> The reason for not currently supporting multiple snackbar messages is that MD spec specifies that only one snackbar should be displayed at a time.

So one way could be that they are queued and displayed one after another. But optional stacking would be cool as well. (Or maybe one message, but with a badge in the corner showing that there are many other pending behind this one.)

Is there some other more MD-approved approach to display multiple messages to the user?

peluprvi:
[...]

PR #2769, opened 2017-12-10
Fix snackbar animation and refactor

manico:
To solve #2759 I ended up with quite a refactoring.
Now, I have a few questions to see if I'm going in the right direction.

- Snackbar is rather strict with allowing only message and action button. I have created new props for this to replace default slot as button added to snackbar does not conform to material design (no ripple, no padding, no hover). Would these new props make sense?

- I have added transitions to `snackbar.styl` as it is very custom containing transitions to container and content. Is there a reason that transitions are separated from component style file or is it ok that they are inside like in this example?

- As for #2759 and stacking of snackbars. This requires more refactoring as snackbars should be aware of one another (some global "storage" for this). Any suggestions?


manico:
Added small vuetify store to handle snackbar queue.
Would this approach be ok for the library?
If it is ok, I can create more tests to cover refactored snackbar.

KaelWD:
The queue is a separate issue to the animation: #2384. I don't think I'd handle it the way you have here.

manico:
In #2759 it is mentioned to create sequential snackbars and since animation fix requires refactoring I figured it would be good to implement that too. Any suggestions on how you would handle queue?

jacekkarczmarczyk:
I think that the queue implementation should be done in another PR. I'm also not sure if it should be done in the `VSnackbar.js`, the component itself should be responsible just for displaying the snackbar and queing should be handled by a separate mixin/component/helper/whatever

johnleider:
Chiming in, queue implementation should not be built in to the component.

KaelWD:
@manico can you merge in the changes from 3392d13b9726509f86cbeba9071a97007878b29e?

manico:
@KaelWD Sorry, I totally missed this. :(
Do you still need me to merge 3392d13?

KaelWD:
I already did it. I'll finish reviewing this today, it's blocking another task that I need to do.

johnleider:
After reviewing this, I've come to the conclusion that we should revert this PR back to the original scope of #2759 and modify the animation only.

PR #2780, opened 2017-12-12
fix(detachable): Hide when deactivated

### LINKED ISSUE #2480: [Bug Report] Tooltip on focusable elements does not close automatically when changing route
### Versions and Environment
**Vuetify:** 0.16.9
**Vue:** 2.4.3
**Browsers:** Chrome 62.0.3202.75
**OS:** Windows 10

### Steps to reproduce
Using vue-router, having the `router-view` wrap with `keep-alive`.
Changing of route using elements such as buttons or &lt;a&gt; tags causes this behavior. The tooltip still appear even the `activator` is now hidden.

Check the reproduction link to see it in action.

### Expected Behavior
The tooltip must be hidden after changing route.

### Actual Behavior
The tooltip still visible after changing route.

### Reproduction Link
[https://jsfiddle.net/ott3jgah/4/](https://jsfiddle.net/ott3jgah/4/)

### Other comments
I recommend also to have a functionality of showing tooltip when a focusable elements such buttons, inputs and &lt;a&gt; tags has focus. This feature/functionality will automatically fix this bug.

We can add this code on the tooltip component:

```
focusin: () => {
    this.runDelay('close', () => (this.isActive = true))
},
focusout: () => {
    this.runDelay('close', () => (this.isActive = false))
}
```

stormy251:
fix: Added a deactivated Life cycle hook that will simple set isActive to false when the element is deactivated

fixes #2480

### DIFF
--- a/src/components/VTooltip/VTooltip.js
+++ b/src/components/VTooltip/VTooltip.js
@@ -130,6 +130,10 @@ export default {
     this.value && this.callActivate()
   },
 
+  deactivated () {
+    this.isActive = false
+  },
+
   render (h) {
     const tooltip = h('div', {
       staticClass: 'tooltip__content',
--- a/src/components/VTooltip/VTooltip.js
+++ b/src/components/VTooltip/VTooltip.js
@@ -130,10 +130,6 @@ export default {
     this.value && this.callActivate()
   },
 
-  deactivated () {
-    this.isActive = false
-  },
-
   render (h) {
     const tooltip = h('div', {
       staticClass: 'tooltip__content',
--- a/src/mixins/detachable.js
+++ b/src/mixins/detachable.js
@@ -13,6 +13,10 @@ export default {
     this.initDetach()
   },
 
+  deactivated () {
+    this.isActive = false
+  },
+
   beforeDestroy () {
     if (!this.$refs.content) return
 


KaelWD:
I've moved it to the detachable mixin instead so it works for menus and dialogs too. 

Playground.vue:
```html
<template>
  <boilerplate>
    <keep-alive>
      <v-tooltip top v-if="show">
        <v-btn @click="show = false" slot="activator">tooltip</v-btn>
        <span>this is a tooltip</span>
      </v-tooltip>
    </keep-alive>
    <keep-alive>
      <v-menu v-if="show" :close-on-content-click="false">
        <v-btn slot="activator">menu</v-btn>
        <v-list>
          <v-list-tile v-for="i in 10" :key="i" @click="show = false">
            <v-list-tile-title>{{ i }}</v-list-tile-title>
          </v-list-tile>
        </v-list>
      </v-menu>
    </keep-alive>
    <keep-alive>
      <v-select v-if="show" :items="['foo', 'bar', 'baz']">
        <template slot="item" slot-scope="data">
          <v-list-tile-content @click.stop.prevent="show = false">
            {{ data.item }}
          </v-list-tile-content>
        </template>
      </v-select>
    </keep-alive>
    <v-btn @click="show = true">show</v-btn>
  </boilerplate>
</template>

<script>
  export default {
    data: () => ({
      show: true
    })
  }
</script>
```

PR #2811, opened 2017-12-16
refactor(v-snackbar): change prop 'timeout' to 'closeDelay' 

adolt:
Change prop 'timeout' to 'closeDelay'  to match API in delayable.js.

### DIFF
--- a/src/components/VSnackbar/VSnackbar.js
+++ b/src/components/VSnackbar/VSnackbar.js
@@ -7,6 +7,7 @@ import {
 
 import Colorable from '../../mixins/colorable'
 import Toggleable from '../../mixins/toggleable'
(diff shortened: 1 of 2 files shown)


KaelWD:
Looks like you've also changed every single newline in the file too. 

Also an FYI - just because there's a TODO doesn't mean it's still valid. You should check with john to see if he still wants this changed.

johnleider:
I have no issue with this, but you need to fix the commit. As @KaelWD pointed out, you've changed majority of the lines for no reason.

KaelWD:
Closing due to inactivity. This would be a breaking change for little benefit anyway, so we'll just leave it as-is for now.

PR #2842, opened 2017-12-20
improve code in components/vuetify/index.js

nicholaslee119:
I think it is better to use Object.values() instead of Object.keys() because of fewer code


smares:
Is this polyfilled for IE?

nicholaslee119:
@smares 
yes, babel-polyfill have supported it

PR #2846, opened 2017-12-20
No style size on VIcon by default

### LINKED ISSUE #2845: [Bug Report] Default style on v-icon can't be overridden by CSS
### Versions and Environment
**Vuetify:** 1.0.0-alpha.3
**Vue:** 2.5.2
**Browsers:** Chrome 63.0.3239.108
**OS:** Mac OS 10.13.2

### Steps to reproduce
Visit the codesandbox link.

### Expected Behavior
The "home" icon and the "event" icon should both be 48px (set by their CSS classes). If no size property is provided to the VIcon it should not set a default, but rather use the 24px default set by the .icon CSS class. This will allow the font-size to be overridden without using !important.

### Actual Behavior
The "home" icon has a style attribute setting its size to 24px. The "event" attribute has its size attribute overridden by using a !important CSS rule.

### Reproduction Link
[https://codesandbox.io/s/01v6ooxyjl](https://codesandbox.io/s/01v6ooxyjl)

### Other comments
VIcon is adding a style: 24px attribute to icons, overriding any CSS font-size styles on the icon - including the 24px font-size defined in the .icon class.

mattpjohnson:
Fixes #2845

### DIFF
--- a/src/components/VIcon/VIcon.js
+++ b/src/components/VIcon/VIcon.js
@@ -44,8 +44,8 @@ export default {
         break
     }
 
-    if (fontSize !== undefined) {
-      data.style = { fontSize, ...data.style }
+    if (fontSize) {
+      data.style.fontSize = fontSize;
     }
 
     let iconName = ''
--- a/src/components/VIcon/VIcon.js
+++ b/src/components/VIcon/VIcon.js
@@ -45,7 +45,7 @@ export default {
     }
 
     if (fontSize) {
-      data.style = { fontSize, ...data.style };
+      data.style = { fontSize, ...data.style }
     }
 
     let iconName = ''
--- a/src/components/VIcon/VIcon.js
+++ b/src/components/VIcon/VIcon.js
@@ -25,8 +25,7 @@ export default {
     medium: Boolean,
     right: Boolean,
     size: {
-      type: [Number, String],
-      default: SIZE_MAP.default
+      type: [Number, String]
     },
     small: Boolean,
     xLarge: Boolean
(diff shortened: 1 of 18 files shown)


src/components/VIcon/VIcon.js
```diff
@@ -44,7 +43,10 @@ export default {
       case props.xLarge: fontSize = SIZE_MAP.xLarge
         break
     }
-    data.style = { fontSize, ...data.style }
+
+    if (fontSize !== undefined) {
```
johnleider: I think it's safe to assume `if (fontSize)`

src/components/VIcon/VIcon.js
```diff
@@ -44,7 +43,10 @@ export default {
...
+      data.style = { fontSize, ...data.style }
```
johnleider: change this to data.style.fontSize = fontSize, no reason to use spread operator here

src/components/VIcon/VIcon.spec.js
```diff
@@ -25,10 +25,16 @@ test('VIcon.js', ({ mount, compileToFunctions }) => {
     expect(wrapper.element.classList).toContain('icon--disabled')
   })
 
+  it('should render no size by default', () => {
+    const context = functionalContext({}, 'add')
+    const wrapper = mount(VIcon, context)
+
+    expect(wrapper.element.style.fontSize).toBe('')
```
johnleider: should not set font size if none provided

src/components/VIcon/VIcon.js
mattpjohnson: I think the spread operator is being used in case style is not a property of data.

src/components/VIcon/VIcon.js
mattpjohnson: I've updated my PR with this change.

src/components/VIcon/VIcon.spec.js
mattpjohnson: I've updated the spec name

PR #2858, opened 2017-12-22
[Feature Request] #2857 introduce new stylus variable $tab-text-transform and use it in tab

### LINKED ISSUE #2857: [Feature Request] Allow configuration of text-capitalization in V-TAB component via stylus/settings/_variables.styl
### New Functionality
Setting the text-capitalization for `v-tab` component via Stylus-Variable. The same is already availalbe vor `v-btn` component. 

### Improvements
Currently, when i want to get rid of text-capitalization in vuetify, i have to set the  `$button-text-transform := uppercase` in `stylus/settings/_variables.styl` which will affect the buttons but not the v-tab component. For v-tab I have to add custom css code which feels inconvinient to me.

After fixing the Feature Request, it is possible to set the v-tab behaviour via config in _variables.styl

It helps from getting bad UX feedback because of text capitalization (http://uxmovement.com/content/all-caps-hard-for-users-to-read/).

julkuh:
In `stylus/settings/_variables.styl`a new variable `$tab-text-transform` can be set to configure whether to use or not to use text-capitalization in `v-tab-items`.

fixes #2857

### DIFF
--- a/src/stylus/components/_tabs.styl
+++ b/src/stylus/components/_tabs.styl
@@ -129,7 +129,7 @@ theme(tabs, "tabs__bar")
     opacity: .7
     text-align: center
     text-decoration: none
-    text-transform: uppercase
+    text-transform: $tab-text-transform
     text-overflow: ellipsis
     white-space: nowrap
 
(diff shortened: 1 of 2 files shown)


julkuh:
My PR does not include any of the failing files. What to do?

PR #2868, opened 2017-12-25
Add the capability for creating fixed header with VDataTable

### LINKED ISSUE #1547: [Enhancement] Fixed DataTable Headers
### What will it allow you to do that you can't do today
It will allow you to have scrollable tables with fixed / sticky header, which is becoming more and more popular. It'll also give a polished feel to the application. 
See https://codepen.io/anon/pen/qbLaMb?editors=101

To fix the headers in datatables, some hacks are likely needed right now. This will reduce the need for hacks 

Something like a "fix-header' prop would be amazing:
` <v-data-table
      v-bind:headers="headers"
      :items="items"
      fix-header
    ></v-datatable>`

ckitt:
Add the capability for creating fixed header with VDataTable

## Description
- the fixed header is actually a separate table in DOM tree, with `position` to be set as `sticky`
- since the displaying width of each column is determined by the browser automatically, it is difficult to ensure the two "tables" (i.e. the fixed header and the table itself) are having same column width, `table-layout` of the table should be changed to `fixed` to have better control on the width of each column (in px), then we can align the width of each of the columns in both the header and the content
- implement logic to determine column width automatically, ensuring there is enough width for each of the header text

## Motivation and Context
Resolves #1547 

## How Has This Been Tested?
unit test for matching snapshot

## Markup
```vue
<template>
  <v-app>
    <div class="wrapperbox">
      <v-data-table
          v-bind:headers="headers"
          :items="items"
          :height="300"
          hide-actions
          class="elevation-1"
        >
          <td v-for="item in props.item">{{ item }}</td>
        </template>
      </v-data-table>
    </div>
  </v-app>
</template>

<style>
.wrapperbox {
  margin: 20px;
}
</style>

<script>
  export default {
    computed: {
      computedHeaders: function() {
        const headers = [];
        for (let col = 0; col < this.totalColumn; col++) {
          const header = {
            align: "left",
            sortable: false,
            text: `Column ${col}`,
            value: `column_${col}`
          };
          if (col !== 2 && col !== 4) {
            header["width"] = 300;
          }
          headers.push(header);
        }
        return headers;
      },
      computedItems: function() {
        const items = [];
        for (let row = 0; row < this.totalRow; row++) {
          const genRow = {};
          for (let col = 0; col < this.totalColumn; col++) {
            if (col % 2 === 1) {
              genRow[`column_${col}`] = `r${row}c${col} This column will be long`;
            } else {
              genRow[`column_${col}`] = `r${row}c${col}`;
            }
          }
          items.push(genRow);
        }
        return items;
      }
    },
    data() {
      return {
        headers: [],
        items: [],
        totalRow: 10,
        totalColumn: 5,
      };
    },
    created: function() {
      this.headers = this.computedHeaders;
      this.items = this.computedItems;
    }
  };
</script>
```

## Types of changes

## Checklist:


KaelWD:
Mentioning an issue # isn't sufficient to close it - see [closing issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/)

jacekkarczmarczyk:
Can it be splitted to 2 PRs? 1 for fixed headers and 1 for the rest?

laonawuli:
This is super exciting, may I have an update on this ticket?

johnleider:
A few issues immediately whenever I looked at the examples in the docs. Do you have the vuetifyjs.com repository locally running and linked with your changes? There are quite a few irregularities.

- Columns are overlapping
- Without using **hide-details** the **height** prop doesn't work properly

Are 2 examples

ckitt:
overlapping columns because some problems when determining the column width,  i.e.  `table-layout` is changed to `fixed`, we must calculate the exact column width based on the header text given in the header props.

Thus, I solve the problem by wrapping each of the element inside the `th` tags with a `span` element to allow me to calculate the minimum width of a column need.

As a result, the newly created props `minColumnWidth` previously is no longer needed

I am not sure if it is the best way to achieve what we want, would be great if there is better suggestion in implementing it.

johnleider:
I'd like to revisit this if you could resolve the conflicts.

ckitt:
@johnleider resolved conflict 👍

nekosaur:
Please add sample markup to show off the changes made and an explanation of any new or changed props.

ckitt:
updated sample markup for demo

ais-one:
Is this PR going to be merged anytime soon? is it failing because of the code coverage? What can I do to help speed this up?

test/unit/components/VDataTable/VDataTable.spec.js
```diff
@@ -339,10 +339,10 @@ test('VDataTable.vue', ({ mount, compileToFunctions }) => {
 
   it('should render correct colspan when using headers-length prop', async () => {
     const data = dataTableTestData()
-    data.propsData.headersLength = 11
+    data.propsData.headersLength = 3
```
KaelWD: If 11 was failing then there's probably something wrong.

ckitt: need to add 11 items in the `headers` array if we insist to use 11, because the program depends on the computed properties `headerColumns` for width calculation

ps. Feel like the value of `headersLength` should align with the length of the headers array

```
  function dataTableTestData () {
    return {
      propsData: {
        pagination: {
          descending: false,
          sortBy: 'col1',
          rowsPerPage: 1,
          page: 1
        },
        headers: [
          { text: 'First Column', value: 'col1', class: 'a-string' },
          { text: 'Second Column', value: 'col2', sortable: false },
          { text: 'Third Column', value: 'col3', class: ['an', 'array'] }
        ],
        items: [
          { other: 1, col1: 'foo', col2: 'a', col3: 1 },
          { other: 2, col1: null, col2: 'b', col3: 2 },
          { other: 3, col1: undefined, col2: 'c', col3: 3 }
        ]
      }
    }
  }
```

ckitt:
added a unit test on the width distribution process

ais-one:
able to merge this?

ais-one:
able to merge this?

nekosaur:
@ckitt What's the support for IE11? According to caniuse, sticky is not available on IE11.

ckitt:
@nekodaur Oh, I didn’t aware of this.... would like to know if anyone have better suggestions for that?...  :(

ais-one:
Possible to use polyfills?

https://github.com/dollarshaveclub/stickybits
https://github.com/wilddeer/stickyfill



*Warmest Regards,*

*Aaron*


The information herein is given on a general basis without obligation and
is strictly for information purposes only. We do not warrant the accuracy
or completeness of the information, graphics, links, or other items
contained herein, and expressly disclaim liability for errors or omissions
in these materials.

ckitt:
wonder if it's a good idea for introducing new lib?

ais-one:
The other option would be to ignore IE11 and below users... or have a note
that fixed headers does not work with IE...



*Warmest Regards,*

*Aaron*


The information herein is given on a general basis without obligation and
is strictly for information purposes only. We do not warrant the accuracy
or completeness of the information, graphics, links, or other items
contained herein, and expressly disclaim liability for errors or omissions
in these materials.

ckitt:
might need to play some css trick with overflow and position with some wrapper to get this done, need to think about the arrangement between the fixed header and the table body...

ckitt:
Fix some issue that auto column width adjustment will not include the "select" checkbox column, it will always take the minimum possible width. Refactor the code base a little bit to reduce the overall complexity

ais-one:
Hopefully this makes its way to 1.1.0 or earlier

ckitt:
@johnleider @nekosaur can help review? many thanks~

trueChazza:
@ckitt thank you for adding this feature - greatly needed

johnleider:
I'm not sure where this plays into @nekosaur 's refactor. Bottom line, I want this functionality in the framework, but I do know there are refactor efforts in play.

ckitt:
May I know what refactoring you are refering to? Is there something that I might have missed? Thanks

ais-one:
Basically, from what I see in the commits, there are quite a number of changes in the code...

Those changes could have impact...

If the impact could be further reduced, I think it would speed up the integration.

This feature should really be targeted for the 1.1.X milestone

nekosaur:
I have a data-table refactor I've been working on (https://github.com/vuetifyjs/vuetify/pull/3833), which includes the ability to have fixed headers. However it is using flexbox instead of html tables and includes a number of breaking changes, which is why it's targeted for 2.0.

We recognize that this is a desired feature and as such it has been added to the 1.2 milestone, and I will try to review the code soon-ish.

ckitt:
Let’s see if this will be included in the coming released? or should we wait for @nekosaur’s implementation instead?

johnleider:
With the introduction of #3833 I'm going to close this. Thank you for your contribution.

Issue #2878, opened 2017-12-27
[Bug Report] v-select with combobox option uses searchValue when blurring


rokkie:
### Versions and Environment
**Vuetify:** 1.0.0-alpha.3
**Vue:** 2.5.13
**Browsers:** Chrome 63.0.3239.108
**OS:** Linux x86_64

### Steps to reproduce
1. Create a v-select with combobox option
2. Make sure items are _objects_ (not just primitives)
3. Select an item from the list
4. Blur the field by e.g. clicking outside

### Expected Behavior
The `itemValue` to be the the `value` option of the selected item.

### Actual Behavior
The `itemValue` is the `text` option of the selected item.

### Reproduction Link
[https://codepen.io/rokkieb/pen/XVpOEg](https://codepen.io/rokkieb/pen/XVpOEg)

johnleider:
Not a bug, I mean, technically. Yes, it does not work, but it is outlined in the docs that the current implementation only accepts primitive values.

rokkie:
Do I need to change/update the issue then?

johnleider:
I'm going to keep it around and tag for 1.1 as that is the form components refactor.

PR #2879, opened 2017-12-27
v-select with combobox option uses search value when blurring

rokkie:
## Description
When using v-select with the `combobox` option and items as objects (not just primitives), and you select an option and then blur the field, the `itemValue` becomes the text value of the selected item object.

## Motivation and Context
See: [Issue](https://github.com/vuetifyjs/vuetify/issues/2878)

## How Has This Been Tested?
Added a new test

## Markup:
See: [Codepen](https://codepen.io/rokkieb/pen/XVpOEg)

## Types of changes

## Checklist:

- [ ] The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).

## Comments:
PR submitted to `dev` since the patch applies to code that is not in `master`.


src/components/VSelect/VSelect.js
```diff
@@ -237,7 +237,10 @@ export default {
         }
       }
     },
-    findExistingItem (item) {
+    findExistingItem (val) {
+      return this.items.find(i => this.valueComparator(this.getValue(i), this.getValue(val)))
```
johnleider: Store `this.getValue(val)` in a variable. It should not be re-evaluated for every find iteration.

src/components/VSelect/mixins/select-watchers.js
```diff
@@ -12,6 +12,11 @@ export default {
       this.$refs.menu && this.$refs.menu.updateDimensions()
     },
     inputValue (val) {
+      // Search for an existing item when a value was selected from the menu
```
johnleider: split comment into two lines. bonus points for having second line 2 less characters than the first

src/components/VSelect/VSelect-combobox.spec.js
```diff
@@ -77,6 +77,31 @@ test('VSelect - combobox', ({ mount }) => {
...
+    await wrapper.vm.$nextTick()
+
+    wrapper.vm.selectItem(item)
+    await wrapper.vm.$nextTick()
+
+    wrapper.setData({ isActive: false })
+    await wrapper.vm.$nextTick()
+
+    expect(event).toBeCalledWith(item)
+    expect('Unable to locate target [data-app]').toHaveBeenTipped()
```
johnleider: Merge in the most recent updates from dev. @jacekkarczmarczyk added better error/warning messages and this test will fail.

src/components/VSelect/mixins/select-watchers.js
rokkie: Fixed in [f94675dc](https://github.com/rokkie/vuetify/commit/f94675dcaa7163596015a1595831e00255094cf7)

src/components/VSelect/VSelect.js
rokkie: Fixed in [7819c6f9](https://github.com/rokkie/vuetify/commit/7819c6f960e02deb5d21782b82dd7d027d00214d)

src/components/VSelect/VSelect-combobox.spec.js
rokkie: I just merged in the most recent (at time of writing) `dev` branch ([0d75322e](https://github.com/rokkie/vuetify/commit/0d75322ebedc87d58f503c8d70cbb7d0ed592db0)) but as far as i can tell the test passes (at least on my machine..). Am i doing it wrong or did things change in the meantime?

jacekkarczmarczyk: My change add the "[Vuetify]" prefix to the error message, `toHaveBeenTipped` function checks if the console messages **includes** the text from `expect` so it still passes. No need imho to change anything here

PR #2969, opened 2018-01-09
fix: apply will-change to btn only when necessary

jacekkarczmarczyk:
## Motivation and Context
v-btn shadow won't change if the button is flat/icon/disabled, and
it will change for other buttons not only when depressed
Maybe it will speed up things by 0.0003s?

## How Has This Been Tested?
thoroughly

## Markup:
```vue
<v-btn icon>X</v-btn>
<v-btn disabled>X</v-btn>
<v-btn flat>X</v-btn>
```

## Types of changes

- [x] Improvement

## Checklist:

### DIFF
--- a/src/stylus/components/_buttons.styl
+++ b/src/stylus/components/_buttons.styl
@@ -85,8 +85,10 @@ theme(button, "btn")
     background-color: transparent !important
     box-shadow: none !important
 
-  &:not(.btn--depressed)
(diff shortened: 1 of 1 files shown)


KaelWD:
That will-change doesn't have an effect anyway, it should probably just be removed entirely. The correct way to transition box shadows is to have different ones on ::before and ::after, fading between them with opacity instead.

KaelWD:
We'll revisit this at some point, it's fine for now though.

PR #2987, opened 2018-01-11
Modify the aspect of the searchValue be confirmed 

huyang2229:
Maybe I think that when the user blur on the input, the searchValue you input should not be confirmed.Because I think that only when the user press the enter or tab means the user want to comfirm the tag.So I modify the code that  empty the searchValue when the user blur the input.Maybe you think the way is not suitable, but I think you should close the menu when the user blur the input and the searchValue is confirmed, or it will make the user confused.It just my humble option and I hope you can consider my options and give me a response.Thank you!

### DIFF
--- a/src/components/VSelect/VSelect.js
+++ b/src/components/VSelect/VSelect.js
@@ -99,6 +99,7 @@ export default {
   },
 
   mounted () {
+    console.log('Fire!!!')
     // If instance is being destroyed
     // do not run mounted functions
     if (this._isDestroyed) return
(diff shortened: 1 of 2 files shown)


KaelWD:
You **must** follow the contributing guidelines when submitting changes to the project. I'm rejecting this because:

- You clearly haven't tested this at all, all your commits were made through the github website
- You haven't provided adequate justification for this change
- You didn't fill out the provided PR template

Please read this then try again: https://github.com/vuetifyjs/vuetify/blob/dev/.github/CONTRIBUTING.md#pull-requests

PR #2992, opened 2018-01-11
fix: pagination for server-side data-table with search val

### LINKED ISSUE #2761: [Bug Report] Server-side data-table with search val breaks pagination
### Versions and Environment
**Vuetify:** 0.17.4
**Vue:** 2.5.3
**OS:** Mac OS 10.12.6

### Steps to reproduce
Enter something in the search field.

### Expected Behavior
The totalItems set after data fetch would still be used for paging.

### Actual Behavior
Paging breaks.

### Reproduction Link
[https://codepen.io/kdholt34/pen/mqNbQP](https://codepen.io/kdholt34/pen/mqNbQP)

### Other comments
I did check v1-alpha.2 and the bug seems to still exist there.  Related to the fact that computed prop itemsLength returns searchLength if there's a search val without first checking if there's a totalItems prop (server-side processing) and elsewhere searchLength is never set if there's totalItems.

kdholt34:
## Description

Fixed pagination for server-side data-table with search val, so computed prop `itemsLength` will use `totalItems` if available.

## Motivation and Context

fixes #2761 

## How Has This Been Tested?

Snapshot tests

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -187,7 +187,7 @@ export default {
       return this.selectAll !== undefined && this.selectAll !== false
     },
     itemsLength () {
-      if (this.search) return this.searchLength || this.totalItems
+      if (this.search) return this.totalItems || this.searchLength
       return this.totalItems || this.items.length
     },
     indeterminate () {
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -187,7 +187,7 @@ export default {
       return this.selectAll !== undefined && this.selectAll !== false
     },
     itemsLength () {
-      if (this.search) return this.searchLength
+      if (this.search) return this.searchLength || this.totalItems
       return this.totalItems || this.items.length
     },
     indeterminate () {


nekosaur:
I'm not sure I understand what this is supposed to fix, even after looking at the related issue. 

If you're using server side data, you shouldn't be using the `search` prop, since it would only filter the local data only, which is not the source of truth. With that said it won't actually do anything (as evidenced by the codepen). If `total-items` is set then all local sorting, pagination and search filtering is disabled.

kdholt34:
For simplicity, the related codepen doesn't actually filter, but the desired use is for the server to do the filtering.  So I want to utilize the search prop which already has code in place to reset the pagination page (and emits an update event) and just send that search val to the server.

nekosaur:
If you want the server to do filtering, don't use `search` prop. Do something like this

https://codepen.io/anon/pen/Zvovwz?editors=1010

kdholt34:
Generally when adding a search/filter, the page is reset to 1.  To do that and not cause an endless loop issue with the watchers, I'd literally be duplicating the code from the existing `search` prop watcher.  What I'm wanting to do is essentially the same as using the existing `pagination` object for server-side processing, so I'm not sure I understand the issue or difference.  All the other places in the code that might use the val from the `search` prop (like` filteredItems`) use a return early pattern that checks for `totalItems`.

johnleider:
Closing due to inactivity.

milosbarlov:
@nekosaur  in this example https://codepen.io/anon/pen/Zvovwz?editors=1010 when you type something in search box you will have two request to the server. The issue is in  
 watch: {
      search () {
        this.fetch()
      },
      pagination: {
        handler () {
          this.fetch()
        },
        deep: true
      }
    },
First will be called search function then pagination function . We need only one request to the server. Can we prevent this situation?

nekosaur:
See open issue https://github.com/vuetifyjs/vuetify/issues/3585

PR #3002, opened 2018-01-13
Fixed stylus path for es5/index.js

sorinsarca:
In `es5/index.js` the path to stylus is wrong
```js
require('./stylus/app.styl');  // should be ../src/stylus/app.styl
```
This patch fixes it.


nekosaur:
The e5 folder is generated so this won't work

KaelWD:
did you even look at the change @nekosaur?

nekosaur:
haha no :) was on my phone

PR #2882, opened 2017-12-28
Fixed VSelect genSelectedItems method

sorinsarca:
## Description

`genSelectedItems` method doesn't use `valueComparator` property when multiple is false, so initial selection is empty if value is an object.

## Motivation and Context
Bugfix.

## How Has This Been Tested?
Manually tested.

## Markup:

```js
let items = [
	{id: {x: 1, y: 1}, text: "A"},
	{id: {x: 2, y: 2}, text: "B"}
];
let selected = {x: 2, y: 2};

<v-select :items="items" v-model="selected" item-value="id"></v-select>
```
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes.

PR #3013, opened 2018-01-14
Bugfixes & enhancements: VSelect, VRadioGroup and data iterables

### LINKED ISSUE #2111: v-select does not show the select items when in fullscreen v-dialog on iPhone
**Reproduction Link:**

https://codepen.io/kimsanka/pen/zEWgJz

**Versions:**

Vuetify: 0.16
iPhone 7+: iOS 11.0.1

**What is expected?**

For the select to show the select items on tab

**What is actually happening?**

The select changes color but does not show the select items

**Bug noticed on:**

iPhone 7+, Chrome + Safari

### LINKED ISSUE #2337: v-checkbox need compare objects by a attribute and return object.
### Versions
Vue https://unpkg.com/vue/dist/vue.js
Vuetify https://unpkg.com/vuetify/dist/vuetify.js

### What is expected ?
I expect the 'v-checkbox' component to be checked when the 'v-model' is initialized with an already existing object among the options. And I want the 'v-model' to save as selected objects rather than primitive types. This expected behavior already happens with 'v-select' using 'item-value' and 'return-object'

Currently the 'v-checkbox' is not getting marked when using it with object.
example:

when i set :value = "item.id" the comparator works however the object returned is the id
when i set :value = "item" the comparator does not work because, although the instances have the same content, they have different reference

### Reproduction Link
https://codepen.io/thiagoaos/pen/eeOoWj

I think using return-object with 'v-checkbox' will resolve the problem. But currently 'return-object' has no effect on 'v-checkbox'

### LINKED ISSUE #2943: [Feature Request] Support dot notation in item-key
### New Functionality
refer a property in object like as:
 *item-value="value.matricula"*
*return-object*
from component<v-select>

### Improvements
more easy to manipulate array, 

https://codepen.io/felipemuner/pen/zpPOpb

sorinsarca:
## Description
* Added a new helper `looseEqual` which checks if two values are loosely equal
* Fixed `VSelect.genSelectedItems()` when multiple is false.
* `VSelect.itemText` and `VSelect.itemValue` can now use a function.
* Added a new helper `getNestedValue` which behaves like `getObjectValueByPath` helper but the second argument is an array. `getObjectValueByPath` now uses `getNestedValue`.
* Added `getItemKey()` method in `data-iterable.js`. 
* `itemKey` property can now be a function, also if is a string can be in dot notation `path.to.value` (uses `getObjectValueByPath` helper)
* Updated `VDataTable` to use `getItemKey()` method
* Added `valueComparator` property for `VRadioGroup` (behaves like `VSelect.valueComparator`) and fixed the comparisons.
* Added `valueComparator` to `selectable` mixin and also added `multiple` property to `selectable` so that `valueComparator` to work when `value` is also an array
* Added a common class `stackable__content__active` for stackable elements to fix z-index issues
* Fixed and refactored `stackable` mixin to support selectors not only classes

## Motivation and Context

* `looseEqual` is something generic that can be reused (now is used by `VSelect` and `VRadioGroup`).
* `VSelect.genSelectedItems()` has bug if multiple is false and the value is an object - #2882 
* `getNestedValue` can be reused with any format (`getObjectValueByPath` only understands `a.b.c` but you can use other function to parse `a/b/c` and reuse `getNestedValue`)
* `getItemKey()` method was added to support `itemKey` property as a function or dot notation - Fixes #2943
* `VRadioGroup` is similar to `VSelect[multiple=false]` so it must have a `valueComparator`, otherwise using objects as values won't work.
* `selectable` mixin is used by `VChecbox` and `VSwitch`, the `valueComparator` and `multiple` properties makes them work even if value is array or object - Fixes #2337 
* `stackable` mixin now uses a common selector (`stackable__content__active` which replaces `menuable__content__active` and `dialog__content__active`) to fix z-index issues and respects `stackMinZIndex` property - Fixes #2111 

## How Has This Been Tested?

* Added a new test for `VSelect.itemValue` as function.
* Others were tested manually.
LE:
* Added a new test for `VDataTable.itemKey` as function
* Fixed three tests for `VSelect` 
* Added tests for `VCheckbox`
* Updated snapshots for stackable elements

## Types of changes

## Checklist:

If PR is accepted I'll add relevant informations to documentation.


johnleider:
Is there a reason for so many different items in the same PR. Is it possible to split them up?

KaelWD:
They all seem to be related to value comparison, so it makes sense to do it all at once. 

@sorinsarca Can you update the PR body to [close the relevant issues](https://help.github.com/articles/closing-issues-using-keywords/)?

sorinsarca:
@johnleider the reason is simple: if the first PR had only the `looseEqual` helper then I'm sure it would be rejected. If with a simple helper and some non-breaking additions I can solve at least three bugs (I hope I didn't introduce other, the tests ran ok) then it is more likely to be accepted (I hope so). Also I need these fixes asap :)

src/components/VRadioGroup/VRadioGroup.js
```diff
@@ -3,6 +3,9 @@ require('../../stylus/components/_input-groups.styl')
 require('../../stylus/components/_selection-controls.styl')
 require('../../stylus/components/_radio-group.styl')
 
+// Helpers
+import {looseEqual} from "../../util/helpers";
```
johnleider: change to `{ looseEqual }`

sorinsarca:
@KaelWD PR body is now updated

src/components/VSelect/mixins/select-props.js
```diff
@@ -1,3 +1,5 @@
+import {looseEqual} from "../../../util/helpers"
```
johnleider: change to `{ looseEqual }`

src/mixins/data-iterable.js
```diff
@@ -240,11 +240,18 @@ export default {
         this.defaultPagination = updatedPagination
       }
     },
+    getItemKey (item) {
+      const key = this.itemKey
+      if (typeof key === 'function') {
+        return key(item)
+      }
+      return getObjectValueByPath(item, key)
```
johnleider: change to:
```
const key = this.itemKey

return typeof key === 'function'
  ? key(item)
  : getObjectValuesByPath(item, key)
```

src/mixins/data-iterable.js
```diff
@@ -285,46 +292,47 @@ export default {
...
         set: (value) => {
           if (itemKey == null) {
-            consoleWarn(`"${keyProp}" attribute must be defined for item`, this)
+            consoleWarn(`"itemKey" attribute must be present for item`, this)
           }
 
           let selected = this.value.slice()
           if (value) selected.push(item)
-          else selected = selected.filter(i => i[keyProp] !== itemKey)
+          else selected = selected.filter(i => this.getItemKey(i) !== itemKey)
```
johnleider: add space between conditional

src/mixins/selectable.js
```diff
@@ -1,5 +1,6 @@
 import Colorable from './colorable'
 import Input from './input'
+import {looseEqual} from "../util/helpers"
```
johnleider: same as above

src/mixins/selectable.js
```diff
@@ -19,23 +20,40 @@ export default {
...
     isActive () {
-      if ((Array.isArray(this.inputValue))
-      ) {
-        return this.inputValue.indexOf(this.value) !== -1
+      const value = this.value
+      const inputValue = this.inputValue
+      if (this.multiple) {
+        if (!Array.isArray(inputValue)) {
+          return false
+        }
```
johnleider: one line this with space after

src/mixins/selectable.js
```diff
@@ -19,23 +20,40 @@ export default {
...
+        for (let i = 0, m = inputValue.length; i < m; i++) {
+          if (this.valueComparator(inputValue[i], value)) {
+            return true
+          }
+        }
```
johnleider: add space between iteration and return statement

src/mixins/selectable.js
```diff
@@ -19,23 +20,40 @@ export default {
...
   },
 
   computed: {
     isActive () {
-      if ((Array.isArray(this.inputValue))
-      ) {
-        return this.inputValue.indexOf(this.value) !== -1
+      const value = this.value
+      const inputValue = this.inputValue
+      if (this.multiple) {
```
johnleider: add space between declaration and conditional

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
       let input = this.inputValue
-      if (Array.isArray(input)) {
-        input = input.slice()
-        const i = input.indexOf(this.value)
-
-        if (i === -1) {
-          input.push(this.value)
-        } else {
-          input.splice(i, 1)
+      if (this.multiple) {
```
johnleider: add space between declaration and conditional

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
+          return
         }
-      } else if (this.trueValue || this.falseValue) {
-        input = input === this.trueValue ? this.falseValue : this.trueValue
-      } else if (this.value) {
-        input = this.value === this.inputValue
-          ? null
-          : this.value
-      } else {
+        let removed = false
```
johnleider: add space between declaration and iteration

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
+        for (let i = 0, m = input.length; i < m; i++) {
+          if (this.valueComparator(input[i], value)) {
+            removed = true
+            input.splice(i, 1)
+            i--; m--
+          }
+        }
+        if (!removed) {
```
johnleider: add space between iteration and conditional

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
+          input.push(value)
+        }
+      }
+      else if (this.trueValue || this.falseValue) {
```
johnleider: this should be on line 101

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
+        input = this.valueComparator(input, this.trueValue) ? this.falseValue : this.trueValue
+      }
+      else if (value) {
```
johnleider: this should be on line 104

src/mixins/selectable.js
```diff
@@ -62,23 +80,32 @@ export default {
...
+        input = this.valueComparator(value, input) ? null : value
+      }
+      else {
```
johnleider: this should be on line 107

src/mixins/data-iterable.js
jacekkarczmarczyk: This could be also a computed prop
`return typeof key === function ? key : item => getObjectValuesByPath(Item, key)`

src/mixins/data-iterable.js
```diff
@@ -240,11 +240,18 @@ export default {
         this.defaultPagination = updatedPagination
       }
     },
+    getItemKey (item) {
```
jacekkarczmarczyk: I'll repeat my comment in case it was missed: `getItemKey` could be a computed property
`return typeof key === 'function' ? key : item => this.getObjectValueByPath(item, key)`

src/components/VRadioGroup/VRadioGroup.js
```diff
@@ -3,6 +3,9 @@ require('../../stylus/components/_input-groups.styl')
 require('../../stylus/components/_selection-controls.styl')
 require('../../stylus/components/_radio-group.styl')
 
+// Helpers
+import { looseEqual } from "../../util/helpers";
```
nekosaur: remove `;`

src/mixins/data-iterable.js
sorinsarca: I don't see any reason to be a computed property, `getItemKey` must always be a function so a method makes more sense. If it is a computed property there is an unnecessary proxy function to `getObjectValueByPath`. While debugging (or just reading code), if you find something like `this.something()` you'll think that it is a method not a prop. Also there is a semantic problem, if I make `getItemKey` a computed prop then I should rename it to something like `getItemKeyFunc` (to know that this prop will always give me a function).

So, even if it can be a computed prop, my opinion is not to make it one.

sorinsarca:
@nekosaur tests for helpers were added

src/mixins/data-iterable.js
jacekkarczmarczyk: > If it is a computed property there is an unnecessary proxy function to getObjectValueByPath

Now the `getItemKey` function itself is a proxy to `getObjectValueByPath` so one proxy will be just replaced with another one, but it will remove the proxy to the `key` function. And for both cases it will remove some code that is run every time the `getItemKey` is called like checking whether `key` is a function or not

>  I should rename it to something like getItemKeyFunc

I don't think it's a bigger problem than having a `key` prop which could be a function but you can't guess it just by looking at the name. And `get*` name clearly shows that this is a function.

Not sure if this solution would be debugging-friendly though

sorinsarca: I think that in most cases `itemKey` will be a string, and in some edge cases you'll need a function for custom behaviour. But you still need to allow a function to prop because a developer works with 3rd party APIs or have custom data structures, otherwise vuetify is good just for demos and examples, not for production apps. 
Thats why `itemValue` and `itemText` from `VSelect` also support functions.

Here is an example for `VSelect.itemValue`

```js
items = [
   {
       type: 'car',
       carTitle: '1st car',
       carId: 'car-1'
   },
   {
      type: 'bike',
      bikeTitle: "1st bike",
      bikeId: 'bike-1'
   }
]

// let's say we only have cars and bikes
itemValue = item => item.type === 'car' ? item.carId : item.bikeId
itemTitle = item => item.type === 'car' ? item.carTitle : item.bikeTitle
```

src/components/VSelect/mixins/select-props.js
```diff
@@ -38,11 +40,11 @@ export default {
       default: 'disabled'
     },
     itemText: {
-      type: String,
+      type: [String, Function],
       default: 'text'
     },
     itemValue: {
-      type: String,
+      type: [String, Function],
```
KaelWD: There's also `itemDisabled` and `itemAvatar` that could use this.

src/mixins/data-iterable.js
jacekkarczmarczyk: > But you still need to allow a function to prop because

I've never disagreed with it. I've only said that it's possible to call `key` (if it's a function) directly without any proxy

sorinsarca: > it's possible to call key (if it's a function) directly without any proxy

Yes, but since in most cases `itemKey` is a string and not a function, a proxy function is used anyway. In method there is typeof check but in a computed prop I think there is a bigger overhead since it has to be cached and in order to cache vue must watch changes in dependencies. For me looks like a performance battle between `typeof` operator and `vue`'s computed properties, and I'll bet on `typeof`.

Anyway, I'll leave it as a method for now, if anyone wants it as a computed prop can also send a PR.

johnleider:
Last request. Fix the conflicts and we'll merge this in.

sorinsarca:
@johnleider Conflicts were fixed

johnleider:
There were some linting issues that made it through.

sorinsarca:
Fixed the `return-object` problem that caused #2337

johnleider:
I haven't been able to figure out exactly why yet, but with the recent changes, the documentation throws this error:

image

The stack is less than useful which is why I'm not posting it here, but you could easily see it locally.

sorinsarca:
@johnleider the problem is that `document` is not defined on server side. There must be a check in `stackable` mixin in method `getMaxZIndex` before using `document`, something like this:
```js
const activeElements = typeof document !== 'undefined' ? [...document.querySelectorAll(this.stackSelector)] : []
```

or like this

```js
const activeElements = this.$isServer ? [...document.querySelectorAll(this.stackSelector)] : []
```

I'll push the changes to `getMaxZIndex` soon.

src/components/VSelect/mixins/select-generators.js
```diff
@@ -266,7 +265,7 @@ export default {
           }
         },
         props: {
-          avatar: item === Object(item) && this.itemAvatar in item,
+          avatar: item === Object(item) && this.getAvatar(item),
```
johnleider: avatar is a boolean prop, this throws errors

src/mixins/stackable.js
```diff
@@ -42,7 +42,9 @@ export default {
       // Convert the NodeList to an array to
       // prevent an Edge bug with Symbol.iterator
       // https://github.com/vuetifyjs/vuetify/issues/2146
-      const activeElements = [...document.getElementsByClassName(this.stackClass)]
+      const activeElements = typeof document !== 'undefined'
```
johnleider: extra space at the end, should of been caught by linter, did you push with --no-verify?

sorinsarca: that change was made directly on github :)

KaelWD:
Which of these are breaking changes?

sorinsarca:
@KaelWD 

- for elements using `selectable` mixin, such as `VSwitch` and `VChecbox`, you must explicitly use `multiple` property if you want multiple options
```js
<v-checkbox multiple label="John" v-model="model" value="John"></v-checkbox>
<v-checkbox multiple label="Jacob" v-model="model" value="Jacob"></v-checkbox>

// ....

model = ["John", "Jacob"]
```

this is because a selection control can also have array as value:
```js
<v-checkbox multiple label="John & Jacob" v-model="model" :value="['John', 'Jacob']">
</v-checkbox>
<v-checkbox multiple label="John" v-model="model" value="John"></v-checkbox>
<v-checkbox multiple label="Jacob" v-model="model" value="Jacob"></v-checkbox>

// ....

// "John & Jacob" checkbox it is NOT checked
model = ["John", "Jacob"] 
// "John & Jacob" checkbox it is checked
model = ["John", "Jacob", ["John", "Jacob"]] 
```

- in `stackable` mixin `stackClass` was replaced by `stackSelector` (but I don't think this affects someone)

KaelWD:
ty

> `stackClass` was replaced by `stackSelector`

That's fine, it's internal data so it doesn't count as breaking. 

> `v-checkbox multiple`

I don't think this is OK. Checkboxes should be multiple by design.

johnleider:
Can we not mimic the behavior of Vue with these input types?

sorinsarca:
> I don't think this is OK. Checkboxes should be multiple by design.

Then I should use `true` as default value for `multiple`. But even before it wasn't multiple by design, it was multiple because (in implementation) no one thought (at that time) that checkbox can hold an array/object as value. Since the value of a checkbox can also be an array, `Array.isArray` is not enough to mark it as multiple, there must be an explicit flag, hence the `multiple` property. 
So, in my opinion `multiple` it should not be a design aspect, it should be an implementation one because the developer will choose the final/desired behaviour.
Anyway, whatever `multiple` is true or false by default it is up to you (for me it doesn't matter).

johnleider:
How does Vue handle this with a native checkboxes?

sorinsarca:
> How does Vue handle this with a native checkboxes?

Buggy :). If the model is an array then it thinks it is multiple. Take a look at following example to see why it isn't ok in all cases:

```js
<input type="checkbox" :value="['A', 'B']" v-model="model">

// ...
model: ['A', 'B']
```

In this case the checkbox is not checked.

Anyway, I can make the `multiple` property optional, and if not defined just guest it by model type. In this case we also mimic Vue's behaviour and also have the power to explicitly set `multiple` as `true/false` when we work with complex data (especially when values are arrays).

KaelWD:
> In this case the checkbox is not checked.

And there's nothing wrong with that. If you want it checked then it needs to be an array item, eg:

```js
model: [['A', 'B']]
```

sorinsarca:
@KaelWD no, it is wrong. Vue or Vuetify should not care about my data structure nor make assumptions. Here is another example:

```js
<input type="checkbox" :value="value" v-model="model">

// ...

var myData = "Some string"
// in this case, the checkbox is checked

var myData = ["A", "B", "C"]
/*
in this case, the checkbox is not checked, 
because it thinks that it is multiple just because it is an array (assumption). 
But I don't want that because myData is something dynamic, 
from a 3rd party or something out of my control (such as user input)
Anyway, with a flag I can force the desired behaviour, hence (again) the multiple prop.
*/

data() {
   return {
     model: myData, 
     value: myData
   }
}

```

johnleider:
We are not going to break from Vue's model. That doesnt mske any sense.

sorinsarca:
@johnleider and we will not, the next commit will make `multiple` optional, and can be used for edge cases (like when value is an array). So the `multiple` prop is no longer a breaking change.

KaelWD:
Looks better, thanks. I'll try to give this a proper run though tomorrow and see if there's anything else that should be amended.

johnleider:
I do not think this is going to make it for v1.0. I would rather take these changes and work with @sorinsarca  during the v1.1 form refactor.

sorinsarca:
> I do not think this is going to make it for v1.0. I would rather take these changes and work with @sorinsarca during the v1.1 form refactor.

I don't know when is planned the release of v1.0 since I wasn't able to quickly find a release roadmap (it would be really nice if you point me there), but:
- this PR has no breaking changes
- the z-index problem is really annoying because it also affects nested fullscreen dialogs (the transition is not visible because it happens behind the first dialog)
- the code has tests and it was also tested manually in my projects
- and, as I promised, I'll create a PR for documentation including examples where needed

I can work just fine with my fork until these changes will be merged and I don't insist to merge these changes, but I'm just confused about the roadmap because it looks like v1.0 will be released tomorrow and it's not enough time to review the changes :)

Again, if you can point me to roadmap's page (or similar) it would be great. Thanks!

johnleider:
Https://next.vuetifyjs.com/getting-started/roadmap

V1.0 release is Feb 13th now though

mathieumagalhaes:
I suppose this did not get added to 1.0.0 yesterday. Any ETA?

johnleider:
@sorinsarca Can you come to the community and msg me please, https://community.vuetifyjs.com

johnleider:
We will be working with @sorinsarca to get some of the QOL changes of this PR into v1.1. This is all being refactored for that release.

PR #3018, opened 2018-01-15
use eslint-config-standard

KaelWD:
## Description
I think this should be good, but I had to make some logic changes to appease eslint. Please triple-check all of these as some of them are rather convoluted. 

## Types of changes


src/components/VDataTable/mixins/head.js
```diff
@@ -58,7 +58,7 @@ export default {
         }
       }
 
-      if ('sortable' in header && header.sortable || !('sortable' in header)) {
+      if (header.sortable === true || header.sortable == null) {
```
KaelWD: `header.sortable || header.sortable == null`

src/components/VDialog/VDialog.js
```diff
@@ -171,9 +171,10 @@ export default {
     }, [dialog]))
 
     return h('div', {
-      'class': 'dialog__container',
+      staticClass: 'dialog__container',
       style: {
-        display: !this.$slots.activator && 'none' || this.fullWidth ? 'block' : 'inline-block'
+        // TODO: double-check this logic, it might need 'none'
+        display: (!this.$slots.activator || this.fullWidth) ? 'block' : 'inline-block'
```
KaelWD: 99c586cb056b71188d8b558426e34554c7b4d348

src/mixins/input.js
```diff
@@ -85,10 +85,8 @@ export default {
     genMessages () {
       let messages = null
 
-      if ((this.hint &&
-            this.isFocused ||
-            this.hint &&
-            this.persistentHint) &&
+      if (((this.hint && this.isFocused) ||
+          (this.hint && this.persistentHint)) &&
```
KaelWD: `this.hint && (this.isFocused || this.persistentHint)`

src/components/VDialog/VDialog.js
johnleider: reverted to original logic

PR #3029, opened 2018-01-16
feat: min/max props for date/time picker

### LINKED ISSUE #2991: [Feature Request] Date pickers - Allowed dates optimization
### New Functionality
There is a need to fully restrict user input in datepicker to a range specified in props.

Currently there is a way to set allowed dates for the picker, as per
https://vuetifyjs.com/components/pickers#example-6
however, this works only for days select, year and month selects are not affected.

Steps to reproduce:
1. open https://codepen.io/pen/?&editors=101
2. select "object" option. following app's logic only dates between *today*  and *today -5* should be selectable
3. try to select any date outside allowed range, e.g. in 2016
4. see that year and month for that date are selectable, and days are not

It would be nice to:
- limit years visible/scrollable in year picker, or disable years that are out of range
- in monthpicker, disable months that are out of range

### Improvements
I am not aware of any work-around for this case, except for explicitly informing the user about restricted dates range before he reaches the datepicker via some text field, or in worst scenario via validation error. Making year/month selects restrictable will significantly improve UX of the datepicker component.

Currently user is able to select year and month that are out of the allowed range, making 2 steps out of 3 possible before "hitting the wall". Then he has to go back and select a value once again. This can cause some frustration on user's end.

jacekkarczmarczyk:
## Description
Adds **min** and **max** props for date picker
Adds **min** and **max** props for time picker
Changes `v-date-picker`'s **allowedDates** prop to accept only `Function`
Changes `v-time-picker`'s **allowedHours** and **allowedMinutes** props to accept only `Function`
Disables navigation to dates outside of the allowed range
Doesn't show years outside of the allowed range

## Motivation and Context
Fixes #2991

## How Has This Been Tested?
jest

## Markup:
`<v-time-picker  min="9:30" max="19:20" v-model="time"></v-time-picker>`
`<v-date-picker min="2013-02-03" max="2020-01-01" v-model="date"></v-date-picker`

## Types of changes

## Checklist:


src/components/VDatePicker/mixins/date-picker-table.js
```diff
@@ -61,6 +57,11 @@ export default {
   },
 
   methods: {
+    isDateAllowed (value) {
```
johnleider: this method is duplicated between components, can we normalize this?

PR #3052, opened 2018-01-20
fix issue with colspan=100% when table has table-layout:fixed

nicocevallos:
## Motivation and Context

`v-data-table` component generates a table with second `tr` tag to place the progress bar. That tag has a `th` tag inside with the attribute `colspan="100%"`
When the CSS `table-layout:fixed` is applied to the table it causes the browser interprets the table has 100 columns, despite the `%` mark.
So, to fix this issue the value of the `colspan` attribute should be equal to the number of headers assigned to the prop `headers` of the `v-data-table` component

## How Has This Been Tested?

I created 2 tables, one without data to show the cell that says there is no data, the other one with data and an option to filter the values and test when the filter doesn't match with any result.
Then I assigned the CSS class `example-table-layout` to the `class` attribute of `v-data-table` to apply the `table-layout:fixed;` to the table into it.
Each table has 4 columns, each column has the class `example-column` except the last one that has the class `example-column-fixed`. Simulating the use case I used when I found the issue.
The `.datatable__progress th` selector applies the red background color to check the progress bar has the correct width.
NOTE: Please check if I missed any use case for this kind of cells with `colspan="100%"`
```
<style>
.example-table-layout table {
  table-layout: fixed;
}
.datatable__progress th {
  background-color: red;
}
.example-column {
	width: calc(33% - 52px);
}
.example-column-fixed {
	width: 156px;
}
</style>
```
## Markup:

[Playground.txt](https://github.com/vuetifyjs/vuetify/files/1648539/Playground.txt)
It's a `.vue` file.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -335,13 +335,15 @@ export default {
   },
 
   render (h) {
+    var headersLength = this.headers.length
+
     return h('v-table-overflow', {}, [
       h('table', {
         'class': this.classes
       }, [
-        this.genTHead(),
-        this.genTBody(),
-        this.genTFoot()
+        this.genTHead(headersLength),
+        this.genTBody(headersLength),
+        this.genTFoot(headersLength)
       ])
     ])
   }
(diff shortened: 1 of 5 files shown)


KaelWD:
Please rebase this onto the dev branch for 1.0, 0.17 is no longer receiving updates.

nicocevallos:
Should I close this PR?

KaelWD:
No, you can just rebase the branch

PR #3073, opened 2018-01-22
feat: add options to $vuetify.goTo

zaken:
## Description
Accept options object as second argument. General code style improvements.

## Markup:
```
$vuetify.goTo(123)
$vuetify.goTo('#id')
$vuetify.goTo('.class')
$vuetify.goTo($ref)

$vuetify.goTo(target, { duration: 10000 })
$vuetify.goTo(target, { offset: 300 })
$vuetify.goTo(target, { easing: 'linear' })
$vuetify.goTo(target, {
  offset: -50,
  duration: 300,
  easing: 'easeInOutQuint'
})
```

## Playground:
https://hastebin.com/amimacovoh.xml

## Types of changes

## Checklist:


src/components/Vuetify/util/goTo.js
```diff
@@ -0,0 +1,92 @@
...
+  const startTime = performance.now()
+  const startLocation = window.pageYOffset
+  const targetLocation = getTargetLocation(target, settings)
+  const distanceToScroll = targetLocation - startLocation
+
+  if (isNaN(targetLocation)) return consoleError(`Target must be a Selector/Number/DOMElement, received ${target.constructor.name} instead.`)
+
+  function step (currentTime) {
+    let progressPercentage = Math.min(1, ((currentTime - startTime) / settings.duration))
+    let easeFunction = easingPatterns[settings.easing](progressPercentage)
```
jacekkarczmarczyk: Would be nice if `easing` could be a user provided function as well. And maybe `easingPatterns` could be extracted to a separate helper so that it could be used in other context. And maybe `easing` could be just a function instead of function/string? That it will also easier to do code autocompletion
```js
import { easeInQuad } from '.../easing'
import goTo from '.../goTo'

goTo(target, {
  easing: easeInQuad
})
```
 
`easeFunction` suggests that it's a function, but it's a number - however `easeFunction` could be a function, just evaluate it outside of the `step()` and then do `distanceToScroll * easeFunction(progressPercentage)`. Of course we allow only functions then there's need to do it, but still the var name should be changed

src/components/Vuetify/util/goTo.js
```diff
@@ -0,0 +1,92 @@
...
+    let targetPosition = Math.floor(startLocation + distanceToScroll * easeFunction)
+
+    window.scrollTo(0, targetPosition)
+    if (window.pageYOffset === targetLocation) return
```
jacekkarczmarczyk: maybe `>=`/`<=` (depending on the scroll direction) instead of `===` could be safer if something goes wrong with calculations (especially with user provided easing function)?

src/components/Vuetify/util/goTo.js
```diff
@@ -0,0 +1,92 @@
...
+    document.documentElement.scrollHeight,
+    document.body.offsetHeight,
+    document.documentElement.offsetHeight,
+    document.body.clientHeight,
+    document.documentElement.clientHeight
+  )
+}
+
+function getTargetLocation (target, settings) {
+  const documentHeight = getDocumentHeight()
```
jacekkarczmarczyk: so it works only for scrolling the whole page? could it be extended to scroll a scrollable divs as well?

zaken:
Yeah, I can do that. Was trying to keep things super simple.

src/components/Vuetify/util/goTo.js
zaken: So like accept another argument with the element we want to scroll? Or put it in options? Not sure how should this be done.

src/components/Vuetify/util/goTo.js
zaken: I can add a `customEasing` prop inside the options object. Externalizing the easing functions as you did adds an extra importing step. Isn't that needless complication?

src/components/Vuetify/util/goTo.js
jacekkarczmarczyk: Me too, I was just asking, if you think it will complicate things too much then skip it (we can always extend the functionality later)

src/components/Vuetify/util/goTo.js
jacekkarczmarczyk: Definitely don't want another prop.
And I don't think that one more import is issue, at least the possibility of reusing these functions is more important (of course you could export them in `goTo.js` but importing easing fn from `goTo` doesn't sound right for me)

src/components/Vuetify/util/goTo.js
zaken: Since `targetLocation` is Math.round'ed maybe we should round `window.pageYOffset` just in case and compare those with `===`?

jacekkarczmarczyk: I was more concerned about the situation where calculating the easing will result with `value > target Location` (becuase of bug in user provided function or bug in our code) so the scrolling will continue ad infinitum.
But rounding `pageYOffset` sounds good too

src/components/Vuetify/util/goTo.js
nekosaur: While I think it's a good idea to break out the easing functions to a separate helper, it should be goTo that imports them, not the user. imo `easing´ should take either a string (look up in imported easings) or a function to provide a custom one. I don't see a problem with goTo importing the easings, as it does depend on them to function properly.

KaelWD:
Update the type definition too with the available option once you've figured out what the API should look like.

src/components/Vuetify/util/goTo.js
zaken: I agree with neko. If `easing` is a string -> look it up in the patterns, if it's a function use it.

jacekkarczmarczyk: Fine for me, but move the evaluation of easing function outside of `step` function

```js
const easingFunction = easing is function ? easing : patterns[easing]

function step() {
  easing = easignFunction(percent)
}
```

jacekkarczmarczyk: and `consoleError('Easing function ... not found')` if the function with provided name doesn't exist

zaken:
@KaelWD Can you help with those type definitions? I have no idea about TS.

KaelWD:
I can write it if you provide a list of every available option

zaken:
Should be it

```
/**
 * @param {number|string|HTMLElement} target
 * @param {Object}                    [options]
 * @param {number}                    options.duration
 * @param {number}                    options.offset
 * @param {string/Function}           options.easing
 */
```

PR #3075, opened 2018-01-22
Fix/2791 v select focus

### LINKED ISSUE #2791: [Bug Report] VSelect input loses focusability/caret after chip is removed or focused
### Versions and Environment
**Vuetify:** 1.0.0-alpha.2
**Vue:** 2.5.9
**OS:** Mac OS 10.13.1

### Steps to reproduce
- Select some options from dropdown menu
- While menu is still open, select or remove some of the chips in selection by clicking on it
- Try to focus (click) select input field and type in it

### Expected Behavior
Input field should be focused, caret shown and it should accept inputs for autocomplete.
This works if you select/remove chip using keyboard.

### Actual Behavior
Input field doesn't react to clicks nor typing, though it looks focused so basically autocomplete functionality is lost. It is restored only after clicking somewhere else in the app and then back to VSelect.

### Reproduction Link
[https://codepen.io/anon/pen/dJoGdM?editors=1010](https://codepen.io/anon/pen/dJoGdM?editors=1010)

### Other comments
This bug is not active in [documentation example](https://next.vuetifyjs.com/components/selects#example-scoped-slots) for scoped slots in VSelect, where selection slot is applied.

This is also active on 0.17.4.

dalevfenton:
## Description
After a user deletes a selected chip, the component can fall out of the correct state which will wipe out a user's subsequent input when they select an item or hit the tab key. Furthermore, once this happens the enter key will not add the user's input either.

## Motivation and Context
fixes #2791 

## How Has This Been Tested?
Tested in playground

## Types of changes

## Checklist:


johnleider:
This should also be set in `select-watchers.js` in the **searchValue** handler.

src/components/VSelect/VSelect-tags.spec.js
```diff
@@ -268,4 +268,75 @@ test('VSelect - tags', ({ mount, compileToFunctions }) => {
     expect(change).toBeCalledWith(['bar'])
     expect('Unable to locate target [data-app]').toHaveBeenTipped()
   })
+
+  it('should be able to add a tag from user input after deleting a tag with delete', async () => {
```
johnleider: This passes even without the changes in this PR

dalevfenton:
I think this one should work now finally, thanks for the help along the way.

PR #3099, opened 2018-01-25
fix: 'firstChild' of undefined" on ssr

hareku:


johnleider:
This is being handled differently for https://github.com/vuetifyjs/vuetify/pull/3126

PR #3054, opened 2018-01-20
fix(babelrc): transform harmony exports

ishitatsuyuki:
There is no export syntax in ES5, thus `export default` makes the artifact non-importable. CommonJS has become the defacto and this is the correct way to get imports working.

## Motivation and Context

Parsing of the imported es5 module, for example `import colors from 'vuetify/es5/util/colors'` fails. This is mentioned in [doc](https://next.vuetifyjs.com/style/theme#customizing).

This issue at least occurs as a Nuxt.js plugin and probably affect more cases. In some cases this can silently work because babel+webpack is used very widely around the Vue.js community.

## How Has This Been Tested?

Nothing was performed because there is currently no integration tests / doctests.

## Types of changes

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- The PR title is no longer than 64 characters.
- The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).
- My code follows the code style of this project.
- (Not applicable) My change requires a change to the documentation.
- (Not applicable) I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes.

PR #3112, opened 2018-01-27
feat: migrate to babel-preset-env

### LINKED ISSUE #3103: [Feature Request] Use Babel transform-runtime
### New Functionality
Using the transform-runtime will eliminate the need of babel-polyfill which pollutes the global environment.

Also, this is the default in Nuxt.js and babel-preset-vue-app, so integrating the transform runtime will make it easier to integrate with the framework build system.

### Improvements
babel-polyfill needs an additional plugin/configuration in Nuxt.js. Using the transform runtime makes it obsolete.

Nothing. This is a somewhat breaking change requiring optional migration (shouldn't break builds).

ishitatsuyuki:
## Description

This PR migrates to the env preset.

Also fixes the tree shaking regression from #3054.

## Motivation and Context

The old presets are no longer recommended, and env preset provides better browser targeting.

## How Has This Been Tested?

The tests passes. Roughly inspected the build outputs and they are working as intended.

## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Checklist:

- The PR title is no longer than 64 characters.
- The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).
- My code follows the code style of this project.


KaelWD:
> Also fixes the tree shaking regression from #3054

Can you explain what this was?

ishitatsuyuki:
The tree shaking regression was due to `{modules: false}` removed in production configs. It is because webpack only supports the shaking with ES2017 import/export syntax.

+2KB gzipped. Is that acceptable? Basically all polyfill references are replaced with core-js reference.

KaelWD:
> webpack only supports the shaking with ES2017 import/export syntax

And it has *never* worked with the single vuetify.js file afaik, which is why the es5 folder exists. Do you have an actual reproduction of this, or is it just speculation?

> +2KB gzipped

Yeah that's fine, and it's probably just from the module change anyway. 

> all polyfill references are replaced with core-js reference

Had a quick look into this, and it seems as though consumers will now have to install `babel-runtime`? I'd like you to check this change against the following (both chrome and IE11):
 - existing webpack projects that *don't* already use `babel-runtime`
 - the docs repo with babel-polyfill removed (https://github.com/vuetifyjs/vuetifyjs.com)
 - standalone (ie. CDN) usage. Example: https://github.com/vuetifyjs/vuetify/blob/10a7e39d392da86601e11d19df278e334c3a97c5/dev/test.html

ishitatsuyuki:
> And it has never worked with the single vuetify.js file afaik, which is why the es5 folder exists. Do you have an actual reproduction of this, or is it just speculation?

It's just speculation, but a search says that it does matter ([reference](https://github.com/babel/babel-loader/issues/521)). If tree shaking doesn't work anyway, this doesn't regress anything either.

> Had a quick look into this, and it seems as though consumers will now have to install babel-runtime?

Yes. I'm going to check those and I'm already working on the docs update.

`babel-runtime` is now bundled in Vuetify itself, and it should be okay for the projects who doesn't use it. For those already use it, I think webpack does dedupe on dependencies (not sure).

ishitatsuyuki:
Hmm, so the codebase is extensively using `Array.prototype.includes` and `Array.prototype.find` (and others). These only works with polyfills.

There are two ways:
1. Keep it intact, revert to the polyfills method.
2. Use a helper method like in [bootstrap-vue](https://github.com/bootstrap-vue/bootstrap-vue/blob/142d517ecad405c6fcb73c112ba82998ee23d63a/src/utils/array.js)

Which sounds better?

jacekkarczmarczyk:
Another option to consider:
3. remove polyfill from the Vuetify bundle and add option in webpack templates whether dev wants do add it to packages.json and main.js or not (for those who don't care about shit browsers)

ishitatsuyuki:
@jacekkarczmarczyk polyfill is currently not included in Vuetify at all.

jacekkarczmarczyk:
Oh sry, mistaken it with `babel-runtime`

KaelWD:
I'd go with \#&#xfeff;1, it's what we're already doing and there's not really any big downsides to it.

package.json
```diff
@@ -42,11 +42,9 @@
     "babel-plugin-detective": "2.0.0",
     "babel-plugin-module-resolver": "3.0.0",
     "babel-plugin-transform-async-to-generator": "6.24.1",
-    "babel-plugin-transform-runtime": "6.23.0",
-    "babel-polyfill": "6.26.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-polyfill": "^6.26.0",
```
KaelWD: Use exact versions (`"6.26.0"`)

.babelrc
```diff
@@ -2,11 +2,15 @@
...
     "test": {
       "presets": [
-        ["es2015"],
-        ["stage-2"]
+        [
+          "env",
+          {
+            "targets": {"ie": 11}
+          }
+        ]
```
KaelWD: Same for the others too:
```js
"presets": [
  ["env", {
    "targets": { "ie": 11 },
    "modules": false // where needed
  }]
],
```

package.json
ishitatsuyuki: I don't understand why you're doing that. yarn.lock exists for stability purpose and semver exists for easier upgrade.

.babelrc
ishitatsuyuki: I'm not sure if I've got it. ES modules is not recommended in Jest and using it for the ES5 version makes it no longer ES5.

KaelWD: Just the formatting, keep the modules as you have it.

ishitatsuyuki:
I'm aware of the test failure, I will upload the fix soon.

package.json
KaelWD: So we know at a glance what version we're actually using, updates are always done manually in package.json anyway. We've had a few problems with packages not respecting semver before.

KaelWD:
I think adding `node: latest` to the targets should fix it.

ishitatsuyuki:
No I missed a transform, please wait a few minutes patiently.

ishitatsuyuki:
The altered runtime transform options brought -20KB uncompressed. Enjoy!

KaelWD:
- I'm seeing +7KB gzipped, still not a big issue though
 - The main thing is that [the docs repo](https://github.com/vuetifyjs/vuetifyjs.com) now fails to build:
image
It works fine with the dev branch, so that's caused by this PR. Removing  `"modules": false` from the production config didn't make a difference.

johnleider:
I just tested this extensively local and I was unable to reproduce your error @KaelWD .

ishitatsuyuki:
| vuetify.min.js | Raw size | gzip -9 |
|----------------|----------|---------|
| This PR        | 250 kB   | 66.4 kB |
| HEAD~1         | 283 kB   | 66.9 kB |

I have also tested the docs site. No errors happened. What's your node version @KaelWD?

ishitatsuyuki:
Rebased. Still working good.

KaelWD:
I'm on node 8.9.4. I'll try it again, might have been user error.

KaelWD:
Same problem:
image
image

 - `rm -rf node_modules` in both projects
 - ensured I had a fresh copy of both branches
 - `yarn`
 - `yarn link` on this branch, `yarn link vuetify` on vuetifyjs.com#master
 - `yarn build`
 - `yarn dev` on vuetifyjs.com

Killing webpack, switching back to dev, and rebuilding works fine.

ishitatsuyuki:
I did the same thing on my server with Node v8.9.4, and both the dev server and prebuilt server ran correctly without errors.

I didn't use yarn link but specified absolute path in package.json; can you try that?

KaelWD:
I went one step further and installed vuetify from npm as usual then copied the built /dist directly into node_modules. That worked, but isn't an acceptable solution as we need yarn link for development. No idea why it works on dev but not with this PR, makes no sense at all.

johnleider:
Using the above steps provided by @KaelWD reproduces the issue.

KaelWD:
Might be webpack/webpack#1643? Weird that it works with current dev though.

ishitatsuyuki:
Just did a rebase and I'm still investigating, but I think this is really hard to fix since the problem roots from webpack.

ishitatsuyuki:
I'm thinking about postponing this until Babel 7 is out. How do you think? Feel free to close if you are ok with it.

johnleider:
Closing due to inactivity.

PR #3115, opened 2018-01-27
feat (v-date-picker): reactive prop

jacekkarczmarczyk:
## Description
Adds **reactive** prop to `v-date-picker`

## Motivation and Context
Picker now is lazy by default, which means that the `input` event is not emitted when month/year (for date picker) or year (for month picker) is not clicked. Setting **reactive** prop to `true` restores the previous behaviour

## How Has This Been Tested?
jest

## Markup:
```vue
<template>
  <v-app id="app" dark>
    <v-content>
      <v-container grid-list-xl>
        <v-layout>
          <v-flex xs4>
            <v-dialog ref="dialog" full-width v-model="modal" lazy width="330px">
              <v-text-field label="Date" readonly slot="activator" v-model="date" clearable></v-text-field>
              <v-date-picker v-model="date" @input="modal = false"></v-date-picker>
            </v-dialog>
          </v-flex>
          <v-flex xs4>
            <div class="headline">reactive: true</div>
            <v-date-picker reactive v-model="date"></v-date-picker>
          </v-flex>
          <v-flex xs4>
            <div class="headline">reactive: false</div>
            <v-date-picker v-model="date"></v-date-picker>
          </v-flex>
        </v-layout>
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data: () => ({
    modal: false,
    date: null
  })
};
</script>
```

## Types of changes

## Checklist:


KaelWD:
Does this effectively replace `@change`? Might as well remove that if so.

jacekkarczmarczyk:
When picker is not lazy we still may need `@change` so that model updates when user clicks year/month but dialog closes when he clicks the date
We could however remove `@change` from lazy picker, but wouldn't it be nice if we could execute some action on date click regardless the value of `lazy`? Just for consistency

PR #3119, opened 2018-01-27
fix: (VBreadcrumbs) use base name for v-breadcrumbs-item

noldors:
## Description
Change component checking from component tag to component name.

## Motivation and Context
If we would register component locally like this:

```js
import VBreadcrumbs from 'vuetify/es5/components/VBreadcrumbs/VBreadcrumbs'
import VBreadcrumbsItem from 'vuetify/es5/components/VBreadcrumbs/VBreadcrumbsItem'

export default {
  name: 'my-component',
  components: {
    VBreadcrumbs,
    'custom-named-component': VBreadcrumbsItem
  }
}
```
The component would be named as 'custom-named-component' and no delimiters would be applied.

## How Has This Been Tested?

Used existing tests.

## Markup:
```js
if (!elm.componentOptions ||
  elm.componentOptions.Ctor.extendOptions.name !== 'v-breadcrumbs-item' ||
  i === length - 1
) return
```

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VBreadcrumbs/VBreadcrumbs.js
+++ b/src/components/VBreadcrumbs/VBreadcrumbs.js
@@ -41,7 +41,7 @@ export default {
     /**
      * Add dividers between
      * v-breadcrumbs-item
-     * 
+     *
      * @return {array}
      */
     genChildren () {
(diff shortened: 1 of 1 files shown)


johnleider:
If we are going to make this change, it should be consisted across the framework. Here is the other location.

https://github.com/vuetifyjs/vuetify/blob/dev/src/components/VSelect/VSelect.js#L123

KaelWD:
Is there a difference between `Ctor.options` and `Ctor.extendOptions`? `v-tabs` does something similar but uses `options` instead.

Issue #3080, opened 2018-01-23
[Feature Request] allow providing user icons in checkbox/radio


jacekkarczmarczyk:
### New Functionality
It will allow providing user icons in checkbox/radio (will help get rid of material icons in favour of other icon lib)

Currently icons are hardcoded:
```js
    icon () {
      if (this.inputIndeterminate) {
        return 'indeterminate_check_box'
      } else if (this.isActive) {
        return 'check_box'
      } else {
        return 'check_box_outline_blank'
      }
    }
```

```js
    icon () {
      return this.isActive ? 'radio_button_checked' : 'radio_button_unchecked'
    }
```

### Improvements
Just binding few props instead of extending component or other dirty hacking

### Proposed API
```html
<v-checkbox on-icon="fa-square-filled" off-icon="fa-square-empty" indeterminate-icon="fa-something-else">
```

```html
<v-radio on-icon="fa-cicrle-filled" off-icon="fa-circle-empty">
```

KaelWD:
I think this should be done globally instead of with yet another prop.

jacekkarczmarczyk:
That could be the next step - provide a mechanism for setting globally default values of props in other components too (not only icons), but with keeping the possibility of overriding defaults in some special cases. That's why props are imho necessary (also for consistency with other components that have similar prop like carousel, pickers, pagination etc)

jacekkarczmarczyk:
Same goes for `data-iterable` and `expansion-panel-content`.
Additionally (a bit off-topic but still icons thing) - pickers icon props could be renamed to `nextIcon` and `prevIcon` instead of `prependIcon` and `appendIcon` proposed by @johnleider (for constistency with for example `v-pagination`)

jacekkarczmarczyk:
data-iterable, v-pagination and others are done in https://github.com/vuetifyjs/vuetify/pull/3152, checkboxes and radio will be handler in form inputs refactoring

PR #3148, opened 2018-01-31
paramaterize sort icon

thelonecabbage:
Adds the ability to set as a prop the v-icon for data table sort. 
`
<v-data-table  sort-icon="arrow_drop_down" .../>
`
## Description
- Added property "sortIcon"
- use this.sortIcon in genHeaderSortingData

## Motivation and Context
There is currently no way to override the default icon for sort in Data Tables

## How Has This Been Tested?
Ran locally, testing in company storybook

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDataTable/mixins/head.js
+++ b/src/components/VDataTable/mixins/head.js
@@ -1,6 +1,12 @@
 import { consoleWarn } from '../../../util/console'
 
 export default {
+  props: {
+    sortIcon: {
+      type: String,
+      default: 'arrow_upward'
+    }
+  },
   methods: {
     genTHead () {
       if (this.hideHeaders) return // Exit Early since no headers are needed.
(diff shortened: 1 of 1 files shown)


johnleider:
I like this change, but I think we should ultimately move some of this logic into a sort of "config" for vuetify components. You can currently use a custom icon, but you have to use slots. @nekosaur What are your thoughts on how we should proceed forward with this?

nekosaur:
We've had some prior talks about how to decouple material icons from the framework, and I think the general consensus was some sort of config object on $vuetify where you can specify the default internal icons to be used. While still supporting per-component changes through either prop and/or slots.

jacekkarczmarczyk:
see also https://github.com/vuetifyjs/vuetify/issues/3080#issuecomment-359706142 (and issue #3080 itself)

jacekkarczmarczyk:
Some icons are not displayed properly (but that's not a problem with this PR)
image

PR #3188, opened 2018-02-05
Vertical v-tab

IamAnotherOne:
Hi,
Its good idea add vertical tab such: https://codepen.io/anon/pen/WMbGVx
Best regards


KaelWD:
Not like that you don't.

AGPDev:
WTF?

PR #3228, opened 2018-02-08
Added clearable-icon prop on v-text-field and v-select

davidgaroro:
## Description

This is my first PR I do on github.

I've decided to use [Material Design Icons](https://materialdesignicons.com/) instead of [Material Icons](https://material.io/icons/). 

So I deleted Material Icons from index.html: `|Material+Icons'` from `<link href...` and added `materialdesignicons.min.css`

Then when I added a text field with the "clearable" property. Instead of leaving the "X" icon, the text "CLEAR" appeared. This happens because by default `<v-text-field>` and `<v-select>` uses the "clear" icon from Material Icon.

## Motivation and Context

Don't depend on Material Icons and be able to use Material Design Icons.

## How Has This Been Tested?

I tested it locally on my computer.

## Types of changes

## Checklist:

### DIFF
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -118,7 +118,7 @@ export default {
     },
     genIcon (type, defaultCallback = null) {
       const shouldClear = type === 'append' && this.clearable && this.isDirty
-      const icon = shouldClear ? (this.clearableIcon || 'clear') : this[`${type}Icon`]
+      const icon = shouldClear ? this.clearableIcon : this[`${type}Icon`]
       const callback = shouldClear
         ? this.clearableCallback
         : (this[`${type}IconCb`] || defaultCallback)
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -118,7 +118,7 @@ export default {
     },
     genIcon (type, defaultCallback = null) {
       const shouldClear = type === 'append' && this.clearable && this.isDirty
-      const icon = shouldClear ? 'clear' : this[`${type}Icon`]
+      const icon = shouldClear ? (this.clearableIcon || 'clear') : this[`${type}Icon`]
       const callback = shouldClear
         ? this.clearableCallback
         : (this[`${type}IconCb`] || defaultCallback)
--- a/src/components/VSelect/mixins/select-props.js
+++ b/src/components/VSelect/mixins/select-props.js
@@ -15,7 +15,10 @@ export default {
     cacheItems: Boolean,
     chips: Boolean,
     clearable: Boolean,
-    clearableIcon: String,
+    clearableIcon: {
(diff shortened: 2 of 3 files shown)


src/mixins/input.js
```diff
@@ -118,7 +118,7 @@ export default {
     },
     genIcon (type, defaultCallback = null) {
       const shouldClear = type === 'append' && this.clearable && this.isDirty
-      const icon = shouldClear ? 'clear' : this[`${type}Icon`]
+      const icon = shouldClear ? (this.clearableIcon || 'clear') : this[`${type}Icon`]
```
jacekkarczmarczyk: Make 'clear' as a default value of `clearableIcon` prop

jacekkarczmarczyk: Good catch btw

davidgaroro: I did 'clear' as default and just use:
const icon = shouldClear ? this.clearableIcon : this[`${type}Icon`]
But I don't like to duplicate two times 'clear' at v-text and v-select props :P

jacekkarczmarczyk: The thing is that we're planning to add a possibility of setting global defaults, so instead of adding `clearableProp` to every `v-text-field` or `v-select` you could do it just one time in the config. Of course you could make your own components that wrap those two, but the behaviour should be consistent across all components.
Another option is to make a `clearable` mixin if you want to keep it DRY, but I don't think it's worth it

davidgaroro: I agree with you. I do not like redundancy. But in these two cases vuetify should be more flexible and not depend so much on Material Icons. At the moment it is the most viable option. If you want a "clearable" text field, you have to add that property to each textfield equally, unless you, as you said, make your own component.

johnleider:
Thank you for your PR. With v1.0 coming out soon, the team will quickly be shifting to our v1.1 goals. The primary target for that release is a form-components refactor. Due to this, I am currently not accepting any changes that touch those components as it is going to change quite a bit. We are already aware of the issues with not wanting to use Material Icons and being able to setup these particular defaults and it will be addresses.

I understand this is a very small change, however, I would much rather you come talk with the team and be part of the conversation after v1.0. Thank you for your contribution.

jonataswalker:
PR closed but it should be addressed since Vuetify allows the use of another icon pack.

I've made my app entirely based MDI.

@davidgaroro May you handle this since we are in 1.0?

johnleider:
This is something supported with the new config level iconFont option in v1.1

davidgaroro:
@jonataswalker I also use MDI in the app, that's why I realized that little detail at the time :)
@johnleider Awesome work with Vuetify <3

PR #3247, opened 2018-02-10
refactor: improve install function

KaelWD:
- ~~**BREAKING CHANGE:** must `Vue.use(Vuetify)` when using CDN package~~
- ~~You can now set the theme easily when using a CDN~~
- No longer includes the entire package.json in the dist
- Vue version is checked in the Vuetify component, now works with a-la-carte too
- $vuetify is a proper vue instance instead of the hacky defineReactive
- Vuetify.version is now available to CDN users
- Replaced semver with a basic custom function, reduces gzipped size from 62KB to 56KB

## How Has This Been Tested?
- vuetify dev environment with Playground.vue
- vuetifyjs.com repo
- ~~basic html file with markup from below~~

## Types of changes

- [x] ~~Bug fix (non-breaking change which fixes an issue)~~
- [x] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~

## Checklist:

- [x] ~~My change requires a change to the documentation.~~

## ~~Markup~~
```html
<link rel="stylesheet" href="../dist/vuetify.min.css">
<script src="https://unpkg.com/vue"></script>
<script src="../dist/vuetify.min.js"></script>
<div id="app">
  <v-app>
    <v-content>
      <v-text-field v-model="primary" label="primary"></v-text-field>
      <v-btn color="primary">button</v-btn>
    </v-content>
  </v-app>
</div>
<script>
  Vue.use(Vuetify, {
    theme: { primary: '#BADA55' }
  })
  new Vue({
    el: '#app',
    computed: {
      primary: {
        get () { return this.$vuetify.theme.primary },
        set (val) { this.$vuetify.theme.primary = val }
      }
    }
  })
</script>
```

### DIFF
--- a/src/components/Vuetify/index.js
+++ b/src/components/Vuetify/index.js
@@ -1,6 +1,8 @@
 import application from './mixins/application'
 import theme from './mixins/theme'
 import options from './mixins/options'
+import Semver from 'semver'
+import { consoleWarn } from '@util/console'
 import goTo from './util/goTo'
 
 const Vuetify = {
@@ -9,17 +11,20 @@ const Vuetify = {
 
     this.installed = true
 
-    const $vuetify = {}
-    Vue.util.defineReactive($vuetify, 'inspire', {
-      breakpoint: {},
-      application,
-      dark: false,
-      theme: theme(opts.theme),
-      options: options(opts.options),
-      goTo
-    })
+    process.env.NODE_ENV !== 'test' && checkVueVersion(Vue)
 
-    Vue.prototype.$vuetify = $vuetify.inspire
+    Vue.prototype.$vuetify = new Vue({
+      data: {
+        application,
+        breakpoint: {},
+        dark: false,
+        options: options(opts.options),
+        theme: theme(opts.theme)
+      },
+      methods: {
+        goTo
+      }
+    })
 
     if (opts.transitions) {
       Object.values(opts.transitions).forEach(transition => {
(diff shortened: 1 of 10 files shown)


src/components/Vuetify/index.js
```diff
@@ -9,18 +11,21 @@ const Vuetify = {
...
 
-    const $vuetify = {}
-    Vue.util.defineReactive($vuetify, 'inspire', {
-      breakpoint: {},
-      application,
-      dark: false,
-      theme: theme(opts.theme),
-      options: options(opts.options),
-      goTo
+    process.env.NODE_ENV !== 'test' && checkVueVersion(Vue)
```
nekosaur: Why shouldn't we check vue version in test env?

KaelWD: Because the tests don't go through webpack, so each test dumps "Vuetify requires Vue version undefined" to the console.

ekoeryanto:
nuxt development broke when updating to 1.0.2

ram-you:
@ekoeryanto I confirm the error with the message:
 `TypeError: Cannot read property 'split' of undefined
`

beatgrabe:
You can temporally resolve the issue by adding this to your `webpack` config:

```javascript
{
  …
  plugins: [
    new webpack.DefinePlugin({
      'process.env.REQUIRED_VUE': '"<your vue version>"'
      // e.g. 'process.env.REQUIRED_VUE': '"2.5.13"'
    })
  ]
}
```

ram-you:
Thank you @beatgrabe . Worked for me.

JoseGoncalves:
With webpack and an a-la-carte setup the same error happens with 1.0.2: vuetifyjs/a-la-carte#6
While defining the `REQUIRED_VUE` environment variable would solve the issue, is this the recommended way to go?

beatgrabe:
The related error is now fixed in [v1.0.3](https://github.com/vuetifyjs/vuetify/releases/tag/v1.0.3)

PR #3290, opened 2018-02-14
Update applicationable mixin to check uid before removing state

benperiton:
## Description
When making a change to a component that has ```applicationable``` mixin, state will get reset because according to https://github.com/vuejs/vue/issues/6518 the vdom always creates before removing nodes, so ```removeApplication()``` is always called after the new node is inserted.

## Motivation and Context
Testing Vuetify for a new project I noticed it when trying one of the default layouts, mentioned it in Discord, but it was bugging me when trying to test a layout to see if I could switch to this framework.

## How Has This Been Tested?
vuetify dev environment with Playground.vue

## Types of changes

## Checklist:

- [x] I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes.

### DIFF
--- a/src/mixins/applicationable.js
+++ b/src/mixins/applicationable.js
@@ -27,6 +27,8 @@ export default function applicationable (value, events = []) {
     },
 
     created () {
+      this.$vuetify.application['activeuid'] = this._uid
+
       for (let i = 0, length = events.length; i < length; i++) {
         this.$watch(events[i], this.callUpdate)
       }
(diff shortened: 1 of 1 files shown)


KaelWD:
Thank you for your contribution, but this should have been fixed by #3295.

benperiton:
I don't think they are the same issue, my testing using https://gist.github.com/benperiton/067508ff767bdfd5cab87e197e31e75b doesn't contain a footer.

it is more to do with this.$vuetify.application.top/left/right/bottom being reset when a component is created - the old is removed after, so it reset itself.

benperiton:
Having said that, actually it looks like it has been fixed by @johnleider in https://github.com/vuetifyjs/vuetify/commit/f01aa46dd6ce7335e843349b031183ab0b739238

johnleider:
That change will be on hold for a bit as we have a smaller change that fixes some of the layout issues users have been having.

PR #3408, opened 2018-02-27
feat(VDialog): Forward 'after-leave' transition event

### LINKED ISSUE #467: Dialog transition when navigating away
When navigating to a specific route I want to show a dialog with a transition. This works fine when setting the dialog's `v-model` to `true` when the component is `mount()`ed. But, I also want to show the transition when navigating away from this route. How would I do this? Can I somehow programmatically close the dialog and `$router.push` afterwards? Or is there a better way?

Fiddle: http://jsfiddle.net/h4kpvk4v/11/

roparz:
## Description

Forward `after-leave` event from `transition` component.

## Motivation and Context

It allows to make an action after the `VDialog` closing transition is ended.
For exemple, if you have a route-based `VDialog` (a dialog directly bound to a route component), you can go back to the previous route only after the dialog is closed.

Should fix #467

## How Has This Been Tested?

Added `@after-leave` attribute in my `VDialog`, the event is triggered on close.

No new test, no update.

```
  <v-dialog
    :value="true"
    @after-leave="onClose"
  >test</v-dialog>
```

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDialog/VDialog.js
+++ b/src/components/VDialog/VDialog.js
@@ -176,9 +176,7 @@ export default {
         css: this.transitionCss
       },
       on: {
-        afterLeave: () => {
-          this.$emit('after-leave')
-        }
+        afterLeave: () => { this.$emit('after-leave') }
       }
     }, [h('div', data,
       this.showLazyContent(this.$slots.default)
--- a/src/components/VDialog/VDialog.js
+++ b/src/components/VDialog/VDialog.js
@@ -59,6 +59,10 @@ export default {
     transition: {
       type: [String, Boolean],
       default: 'dialog-transition'
+    },
+    transitionCss: {
+      type: Boolean,
+      default: true
     }
   },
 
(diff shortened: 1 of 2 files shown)


johnleider:
I like this change, but it could be used for quite a few things. `v-menu`, `v-toolbar`, `v-navigation-drawer`, `v-bottom-navigation` and probably a dozen other components could technically use this functionality. The question is do we have a one off implementation like this, add it to all components that could support it, or not support it at all.

I'm undecided atm.

roparz:
All Vuetify components that use the `transition` component could implement that. And it's not only the `after-leave` event but all others (https://vuejs.org/v2/guide/transitions.html#JavaScript-Hooks).

Maybe we could use a mixin, a the [`extends`](https://vuejs.org/v2/api/#extends) keyword?

johnleider:
Transitions in general need to be looked at. I'm not a fan of 1 off event emits. As stated above, there are quite a few areas that use transitions, and if we add it to `v-dialog`, we need to add it to them as well. I believe that this could be accomplished from a refactor of the transition generation. As of now, there is no plans to implement this on a one off basis.

If you want to discuss this further, ping me in the community, https://community.vuetifyjs.com

PR #3556, opened 2018-03-13
fix(rippleable): added explicit click event

### LINKED ISSUE #3554: [Bug Report] Clicking v-switch label doesn't fire registered @click handler
### Versions and Environment
**Vuetify:** 1.0.6
**Vue:** 2.5.15
**Browsers:** Chrome 65.0.3325.146
**OS:** Linux x86_64

### Steps to reproduce
Create a v-switch with a label and a custom @click handler function. Clicking the button (the colored circle) will fire your custom handler. Clicking the label element (text to the right) will perform default toggling behavior and ignore the custom click handler.

In the attached Codepen, clicking the button toggles the visibility of the message below it, whereas clicking the label toggles the button.

### Expected Behavior
When using a custom click handler, it should be hit whether I click the button or the label.

### Actual Behavior
Clicking the switch element hits the custom handler as expected, but clicking the label does not. Instead, clicking the label performs default behavior (toggles the v-model) and does not hit the custom handler at all.

### Reproduction Link
[https://codepen.io/anon/pen/JLYWyZ](https://codepen.io/anon/pen/JLYWyZ)

### Other comments
I'm using Vuex, so best practices state that I shouldn't be manipulating store variables directly. A custom click handler is the easiest way to do this, as well as some other things I'm trying to accomplish. Workarounds that come to mind are much more labor-intensive and produce ugly/repetitive code.

I'm also unable to modify the click event on the label element to align with the functionality of the button. This approach would be repetitive and ugly anyway.

johnleider:
## Description
Rippleable mixin did not explicitly emit a click event, and bound click events would overwrite the default

## Motivation and Context
fixes #3554

## How Has This Been Tested?
jest

## Markup:
``` vue
<template>
  <div id="app">
    <v-app id="inspire">
      <v-container fluid>
        <v-switch :label="`Switch 1: ${switch1.toString()}`"
                  v-model="switch1"
                  @click="toggleMessage">
        </v-switch>
        <p v-if="showMessage">I'm showing!</p>
      </v-container>
    </v-app>
  </div>
</template>

<script>
export default {
  data () {
    return {
      switch1: false,
      showMessage: false,
    }
  },
  methods: {
    toggleMessage() {
      this.$data.showMessage = !this.$data.showMessage
    }
  }
}
</script>
```

## Types of changes

## Checklist:

### DIFF
--- a/test/unit/mixins/rippleable.spec.js
+++ b/test/unit/mixins/rippleable.spec.js
@@ -1,0 +1,37 @@
+import { test } from '@/test'
+import Rippleable from '@/mixins/rippleable'
+
+const Mock = {
+  mixins: [Rippleable],
+
+  data: () => ({
+    rippleClasses: null
+  }),
+
+  render (h) {
+    return this.genRipple()
+  }
+}
+
+test('menuable.js', ({ mount }) => {
+  it('should react to click', () => {
+    const wrapper = mount(Mock)
+
+    const click = jest.fn()
+
+    wrapper.setMethods({ toggle: click })
+    wrapper.trigger('click')
+
+    expect(click).toHaveBeenCalled()
+  })
+
+  it('should match snapshot', () => {
+    const wrapper = mount(Mock)
+
+    expect(wrapper.html()).toMatchSnapshot()
+
+    wrapper.setData({ rippleClasses: 'foo' })
+
+    expect(wrapper.html()).toMatchSnapshot()
+  })
+})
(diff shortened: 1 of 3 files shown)


test/unit/mixins/rippleable.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    const click = jest.fn()
+
+    wrapper.setMethods({ toggle: click })
+    wrapper.trigger('click')
+
+    expect(click).toHaveBeenCalled()
+  })
+
+  it('should match screenshot', () => {
```
KaelWD: > screenshot

Issue #3418, opened 2018-02-28
[Bug Report] Tabs with router not marked active


janein:
### Versions and Environment
**Vuetify:** 1.0.3
**Vue:** 2.5.13
**Browsers:** Chrome 64.0.3282.140
**OS:** Linux x86_64

### Steps to reproduce
Create Tabs which are based on a router. The tabs should point to the same route, but with different params in the url. (i.e. '/categories/1/', '/categories/2/')

### Expected Behavior
Active tabs should be marked with the highlight color and the sliding bar below it.

### Actual Behavior
Active tabs are only marked with the highlight color. Sliding bar stays at the first tab matching the route (but not the params).

### Reproduction Link
[https://codesandbox.io/s/4861v3lxmx](https://codesandbox.io/s/4861v3lxmx)

### Other comments
Worked fine in older versions. As the highlighting works fine on the highlight color, it should be solveable :)

janein:
works in 1.0.2, is broken in 1.0.3

janein:
managed to find the cause of the error in these changes:
https://github.com/vuetifyjs/vuetify/compare/v1.0.2...v1.0.3#diff-d429bd0dd45860c5f938bef84a1514a4

As I'm not aware why the checks to `name` have been added, I feel not secure to submit a pull-request. Maybe someone has a solution?

Arteben:
I also came across this error in version 1.0.7.
https://codesandbox.io/s/v6rnpoo2k3

janein:
Thanks a lot! Works like a charm! :)

Arteben:
Thank very much)

PR #3572, opened 2018-03-14
Fix highlight status of tabs

janein:
First check for the complete path before falling back to the name.check.
Fixes Bug #3418

## Description
We should check for the full path including all params before we check on the route's name.

## Motivation and Context
Fixes the bug issued in #3418 

## How Has This Been Tested?
Tested in on a quite huge app I#M working on. As it's a very little change there should not be more tests required.

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x ] The PR title is no longer than 64 characters.
- [ x] The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).

### DIFF
--- a/src/components/VTabs/VTab.js
+++ b/src/components/VTabs/VTab.js
@@ -47,9 +47,9 @@ export default {
 
       if (typeof to === 'string') return to.replace('#', '')
       if (to === Object(to) &&
-        (to.hasOwnProperty('name') ||
-          to.hasOwnProperty('path'))
-      ) return to.name || to.path
+        (to.hasOwnProperty('path') ||
+          to.hasOwnProperty('name'))
+      ) return to.path || to.name
 
       return this
     }


jacekkarczmarczyk:
Can you add test for this fix?

janein:
@jacekkarczmarczyk what should I test for?
It just fixes a bug in the tabs - all exisitngs tests should be enough.

jacekkarczmarczyk:
https://github.com/vuetifyjs/vuetify/blob/dev/test/unit/components/VTabs/VTab.spec.js#L191

Add here the a case where `to` is an object with specified `path` and `name` and check that `wrapper.vm.action` is equal to `path`. It should fail without your patch and pass with it

janein:
@johnleider ok thanks a lot! that's why I didn't want to submit a PR in the first place. I just did it because I really need this fixed as I found out, that I can't properly downgrade to 1.0.2 with the new Vue-Version.

Issue #2871, opened 2017-12-26
[Feature Request] feature and/or documentation for adding custom icon to use in <v-icon>


DrSensor:
### New Functionality
Some of the enterprise projects have their own icon set, its necessary to use that icon in button or add a badge to that icon. Currently, vuetify can do this but only with the predefined icon (e.g mdl and awesome icon).

### Improvements
Rather than adding the custom icon using SVG or PNG, the best practice is adding it using web font (developer can convert the SVG using some loader like [icons-loader](https://github.com/mikevercoelen/icons-loader)). I propose this feature can be set like setting theme so something like:
```javascript
import Icon from '../assets/logo.svg'
import IconCollection from '../assets/CompanySet.font'

Vue.use(Vuetify, {
  icon: {
    prefix: IconCollection,
    icon_name: Icon
  }
})
```
then use it like
```html
<v-icon>icon_name</v-icon>
<v-icon>prefix_company_logo</v-icon>
```

### Bugs or Edge Cases it Helps Avoid
Make sure the documentation is clear that it can only accept `@font-face` style with its rule.

manico:
This is a good idea as we are currently hard-coding font awesome in library.
In my company we also have are own set of icons with prefix `ni`.

Options for icon should be expanded to accept array of icons with different prefixes as some library have multiple like font awesome #2665

johnleider:
Go for it @manico . As it stands, `v-icon` is a bit sloppy. We have a PR for adding svg support, but admittedly, I have not had a chance to really test any alternate implementations. If you happen to pass by that as well.

garcianavalon:
Hi!
I would also be interested in this feature. Any way I can help getting it implemented?

Dynamic-Gravity:
The current fontawesome 5 icons have vue components and are not currently compatible with `v-icon`.

So at this moment in time its either-or.

nekosaur:
font awesome 5 is supported. just use their css.

appurist:
I believe the PR #3591 above resolves this issue and much more.  Not only can you replace all of the internal icons at startup/initialization time, you can also dynamically change them at any time later.  See [the code example](https://codesandbox.io/s/x2w01515wo) where I demonstrate at least two things:
* all internal default MD icons are replaced with MDI icons universally at initialization time (it starts up in MDI mode)
* you can dynamically switch after startup by replacing each known internal icon with a new one.

[...]

Issue #2784, opened 2017-12-13
[Feature Request] Checkbox: Alternative icons


lgullestrup:
### New Functionality
Allow you to select a pair of icons to switch between, instead of using the standard box and tick for checkbox (and other) selection controls.

eg. Favorites usually toggle between an outlined, and a filled star.

### Improvements
allow normal use of checkbox instead of redeveloping a custom peace to use different icons as selection control. 

### Bugs or Edge Cases it Helps Avoid
none that i am aware of.

Akaryatrh:
>Bugs or Edge Cases it Helps Avoid

It would help people who don't use `material design` icons library. Currently, it's not possible to only use `font-awesome` icons library while checkbox icons is tied to `check_box_outline_blank` icon. It's even worst in case of `material-design-icons` library while finally we need to load duplicated icons.

mascit13:
I also ran into this problem. We also only want to use font-awesome. Set fa-icon for checked and not-checked state would be very nice.

appurist:
I've been working on this case and #2871 in a branch that I hope the team will find useful.  It allows easy swapping of all Vuetify internal icons to MD, MDI or FA (font-awesome), as well as the ability to replace specific icons.  See [code example here](https://codesandbox.io/s/x2w01515wo). Note this is not yet part of Vuetify and must be reviewed first, but I intend to submit it today.  Note that it does not support the SVG version of font-awesome, but does for the CSS webfont version, and I'll be looking into that as well, because I need the SVG over JS version.

appurist:
I believe the PR #3591 above resolves this issue and much more.  Not only can you replace the checkbox icons at startup time, you can also dynamically change them at any time later, and you can also do this for any (or all) of the icons referenced internally by Vuetify.  You can see checkbox toggling of Font Awesome checkboxes in [the code example](https://codesandbox.io/s/x2w01515wo) I provided.

For the purposes of this Issue, you could replace the checkbox icons with code like this:
```
Vue.use(Vuetify, {
  icons: {
    checked: 'fas fa-check-square'
    unchecked: 'far fa-square'
  }
})
```

To easily replace **all** internal (MD) icon names with icons from the FA icon sets:
```
Vue.use(Vuetify, { iconfont: "fa" })
```
This will cause Vuetify to use FA (Font-Awesome) icons.

To dynamically change these two icons dynamically later on:
```
this.$vuetify.icons.checked = 'fas fa-check-square'
[...]

PR #3591, opened 2018-03-16
Feature/custom icons improvements for #2784 and #2871

appurist:
## Description
This submission provides a simplified means to replace all internal icon names with their equivalents with either [MD (Material Design) icons](https://material.io/icons/), [MDI (materialdesignicons.com) icons](https://materialdesignicons.com/) or [FA (Font Awesome) icons](https://fontawesome.com/icons?d=gallery). Previously, Vuetify supported only MD icons.

These icons can be configured via the `options` parameter on the `Vue.use` call for Vuetify.  If no choice is specified, it defaults to MD (the same as previous releases).  Irrespective of which choice is made, it is also possible to override one or more individual icons, or all of these, dynamically after initialization via options.

## Motivation and Context
Some deployments have restrictions or requirements that dictate certain icon font needs.  My particular situation is a government enterprise environment that does not allow arbitrary font downloads to managed desktops.  My solution will be to use Font Awesome icons as SVG icons via JavaScript (rather than a font), however Vuetify does not currently allow these to be used for internal icon names, such as those used by `v-checkbox`, the close icon in `v-chip`, etc.

These changes may allow either or both issues #2784 and #2871 to be closed (TBD).

## How Has This Been Tested?
The existing tests have been run, and in fact this exposed a bug (pattern):
```
Test Suites: 8 failed, 1 skipped, 76 passed, 84 of 85 total
Tests:       33 failed, 8 skipped, 659 passed, 700 total
Snapshots:   35 failed, 223 passed, 258 total
```
The previous (left chevron) icon was referenced as 'icon$prev' but was listed in the icons table as 'previous'. (Resolved by changing the icons tables to define it as 'prev'.)

After this, 2 failures remained, and these were caused by the test for checked `v-checkbox` not including the `icon--checkbox` class (wrong test). This was also resolved.

Now all tests pass:
```
Test Suites: 1 skipped, 84 passed, 84 of 85 total
Tests:       8 skipped, 692 passed, 700 total
Snapshots:   258 passed, 258 total
```

Beyond these automated tests, additional visual developer testing has been performed, but no new tests cases added.

## Markup:
It is now possible to easily replace all internal (MD) icon names with icons from the MDI and FA icon sets:
* `Vue.use(Vuetify, { iconfont: "mdi" })` will cause Vuetify to use MDI (materialdesignicons.com) icons.
* `Vue.use(Vuetify, { iconfont: "fa" })` will cause Vuetify to use FA (Font-Awesome) icons.

It is also possible to override specific icons. The following uses the _default_ (MD) icons everywhere, except that marked `v-checkbox` icon will be displayed with the MD bug icon, and an unchecked `v-checkbox` will show the MDI anchor icon:
```javascript
Vue.use(Vuetify, {
  icons: {
    "checked": "bug_report"
    "unchecked": "mdi-anchor",
  }
})
```

It is also possible to combine these to specify an entire alternative icon set and then also override specific icons. The following uses the Font-Awesome (FA) icons everywhere, except that marked `v-checkbox` icon will be displayed with the MD bug icon, and an unchecked `v-checkbox` will show the MDI anchor icon:
```javascript
Vue.use(Vuetify, {
  iconfont: "fa",
  icons: {
    "checked": "bug_report",
    "unchecked": "mdi-anchor"
  }
})
```

A [code sandbox](https://codesandbox.io/s/x2w01515wo) exists demonstrating some of the features (see the `v-select` options at the bottom).

## Caveats / Known Problems
As with previous `v-icon` support, icon fonts based on SVG+JS (such as the SVG/JS implmentation of Font Awesome) bypass the virtual DOM and thus Vue is unaware, which means it does not show dynamic changes to SVG icons.  A [second code sandbox](https://codesandbox.io/s/o14vnjq66) uses the SVG/JS implementation of Font Awesome icons and demonstrates this issue when toggling the checkboxes.  However, it does work fine for icons that aren't changed, such as the close icon in the `v-chip`.

## Types of changes

## Checklist:
These changes do not _require_ a change to the documentation but do require additional documentation to cover the new features. I will submit an update to the docs but have not done so yet.


appurist:
Note that the SVG/JS font-awesome icons _are_ already supported by this PR, but not dynamic updates to them.  For the built-in Vuetify components, I think this means only `v-checkbox` has a problem. I'm investigating ways to make that one work with a bit of additional code, but that will probably follow in a separate incremental update, unless I can resolve that problem quickly.

johnleider:
Looks great. The only thing I'd comment on is the naming convention, 'icon$'. I'd like to brainstorm on this and ensure it will cover current and future (to the best of our ability) use cases. Also the potential of other config defaults.

@vuetifyjs/dev-team please take a look

appurist:
@johnleider Yes, I was trying to think of some way to signal VIcon.js that the icon name was an 'internal indirect' reference. The simplest seemed to be a "starts with" signal akin to the '-' delimiter in custom icons. Except the main difference is that this one would be (intended to be) entirely internal.

I inserted "intended to be" in that last statement because it will prevent rendering of externally (actual) icons that start with the same trigger in the name. So technically it wouldn't be able to support external icons that start with "icon$...".

There were two goals in this naming convention: 1) find a way to identify that the string constant is an icon name, and 2) find something unlikely to be used in an external icon name.  I thought 'icon$cancel' (for example) achieved both of those with relatively good safety, but there may be better choices for this for sure.

I think most icons use '-' and/or '_' as word separators, and I'm not sure what is allowed in a class name in JavaScript, but since this is just a string constant and is filtered before being passed to anything else, I think the safest prefix might be something that isn't even legal in a class name or HTML text content.  I figured since $ was often used to signal template placeholders and other special ids or expressions, that maybe it was a reasonable choice.  

Although this might make it more awkward to specify one from _outside_ Vuetify and that brings up an important point. I realized that it was possible (and potentially extremely useful) _for external code to also use these identifiers_. It's actually a really cool feature.  For example, is possible for a user to specify` <v-icon>icon$close</v-icon>` to make use of _whatever the current close icon is set to_.  In other words, the level of indirection used internally by Vuetify to allow the icons to be redefined without requiring code changes... this could also be used _externally_ by Vuetify users.

And this last point makes the choice of prefix syntax fairly important, and part of the external API, something I hadn't considered when initially choosing 'icon$' as the prefix.  I think it deserves more consideration and feedback from others.  This is something @johnleider and @vuetifyjs/dev-team should discuss and come to a decision on.  At any rate, the unique nature of the prefix makes it easy to find and replace, although at this point I don't see much of a problem with 'icon$'.  Perhaps 'icon*' or "icon!" might be better.

appurist:
I will put in a PR for documentation changes once I think the feature changes have stabilized, but one thing that has not been documented here yet is the indirect reference names to the icons used internally.  This can also be used by end users to provide an indirect reference to an icon.  Here is the list of internal icon names (e.g. this.$vuetify.icons._iconname_) and their previously-hard-coded MD icon names:
```
cancel: 'cancel'       // VAlert (X with circle)
close: 'close'         // just the X (without a circle)
delete: 'cancel'       // VChip close
success: 'check_circle',    // VAlert
info: 'info',          // VAlert
warning: 'priority_high',   // VAlert
error: 'warning',      // VAlert, VStepperStep
prev: 'chevron_left',  // VCarousel, VDatePicker, VPagination, tabs-props, data-iterable
next: 'chevron_right', // VCarousel, VDatePicker, VPagination, tabs-props, data-iterable
checked: 'check_box',                     // VCheckbox
unchecked: 'check_box_outline_blank',     // VCheckbox
indeterminate: 'indeterminate_check_box', // VCheckbox
dot: 'fiber_manual_record',               // VCarousel
sort: 'arrow_upward',                     // VDataTable
expand: 'keyboard_arrow_down',    // VExpansionPanelContent
append: 'keyboard_arrow_down',    // VListGroup
subgroup: 'arrow_drop_down',      // VListGroup, select-props
rchecked: 'radio_button_checked',         // VRadio
runchecked: 'radio_button_unchecked',     // VRadio
edit: 'edit'          // VStepperStep
```

These allow _indirect references_ to the corresponding `this.$vuetify.icons` member, recognized by a '$vuetify.icons' prefix, so you can assign '$vuetify.icons.checked' for example in Vuetify source code (as a string constant), or as the icon name in HTML code, in order to indirectly specify the checked checkbox, without restricting code to a specific icon set.

src/components/Vuetify/mixins/icons.js
```diff
@@ -0,0 +1,110 @@
+// Maps internal Vuetify icon names to actual Material Design icon names.
+const ICONS_MATERIAL = {
```
jacekkarczmarczyk: about the naming:
 - `rchecked` etc - imho no need to create such not-so-easy-to-guess abbreviations, i'd suggest `radionOn`, `radioOff`, `checkboxOn`, `checkboxOff`, `checkboxIndeterminate`
- `dot` - the name should say what it is, not how it looks like, so maybe `delimiter`?
- `append` - the name should say what it is, now where it is located - maybe use `expand`?
- `subgroup`  as a name of the `v-select` icon - maybe `dropdown`? i think `subgroup` for `v-list-group` is ok. So there would be 3 "arrow down" icons: `subgroup`, `dropdown` and `expand`

appurist: I new when I added the missing `rchecked` etc that it was unlikely to survive that way. ;)

In general, when I first added each icon, I wasn't sure to what extent it was currently used. So I tried to preserve the existing naming, at least where it wasn't very iconfont-specific, or simplify/shorten it where it was.  

I'm definitely with you on most of the above and in general. Now that I've produced the list above, the intent is a lot clearer and it's a good chance to review all that naming.

I think that if we're okay with two-word camel-case names, something like contextAction (e.g. radioOn) is great.  And On/Off is nice and short and clear, better than Checked.  So those first 5 I'll do today without needing any further comment from anyone else, I think that would be widely accepted.

The `dot` case isn't really how it looks, and it's not a delimiter.  It's the placeholder blob on the bottom of the [carousel](https://vuetifyjs.com/en/components/carousels) representing an image or item in the list of carousel items. But since it was just a placeholder, circular and meant to be filled in, I just used `dot` initially.  Maybe `carouselItem` or just `placeholder` or `blob` or `placeholderBlob` if it gets used for something else (not currently).  Something like `carouselPlaceholder` is kinda long for my taste.  Yours too, or okay?  I'm kinda leaning towards `placeholder` here now, although I don't really dislike `dot` or `blob` for his one. 😀 Comments?

Yes, the `append` icon (in VListGroups) is an _expand_, and appears to perform the same operation as the icon in expansion panels, and should probably removed entirely and replaced with the existing `expand` icon (tied together). 

Completely agreed on `subgroup`, I think the suggestion is to split the two use cases into `subgroup` and `dropdown`, and I love that idea and I can do that today as well.

jacekkarczmarczyk: About the `delimiter` - i don't like the name too, but the prop name is `delimiterIcon` so i think the icon name should be in line with the prop name (somehow I feel that this idea is valid with this particular icon, but not with the `append`...)

appurist: In that case I'm also totally in favor of `delimiter`, and that solves that one nicely. Thanks for the update/info on that one.  I'll have the commit with all the renames done soon for further review.

appurist:
Okay, $vuetify.icons definition choices and naming tweaks committed.

appurist:
So if I've done this as discussed, I think the new tweaked `$vuetify.icons` definition list should be:
```
cancel: 'cancel'       // VAlert (X with circle)
close: 'close'         // just the X (without a circle)
delete: 'cancel'       // VChip close
success: 'check_circle',    // VAlert
info: 'info',               // VAlert
warning: 'priority_high',   // VAlert
error: 'warning',      // VAlert, VStepperStep
prev: 'chevron_left',  // VCarousel, VDatePicker, VPagination, tabs-props, data-iterable
next: 'chevron_right', // VCarousel, VDatePicker, VPagination, tabs-props, data-iterable
checkboxOn: 'check_box',                // VCheckbox
checkboxOff: 'check_box_outline_blank', // VCheckbox
checkboxIndeterminate: 'indeterminate_check_box', // VCheckbox
delimiter: 'fiber_manual_record', // VCarousel
sort: 'arrow_upward',             // VDataTable
expand: 'keyboard_arrow_down',    // VExpansionPanelContent, VListGroup
subgroup: 'arrow_drop_down',      // VListGroup
dropdown: 'arrow_drop_down',      // select-props
radioOn: 'radio_button_checked',        // VRadio
radioOff: 'radio_button_unchecked',     // VRadio
edit: 'edit'            // VStepperStep
```

appurist:
Updated example codesandbox [here](https://codesandbox.io/s/0p1vvn6wpv).

jacekkarczmarczyk:
It's not in the scope of this PR, not even good repo, but anyway - after this PR is merged there could be an option in templates to select the icon set. It could automatically add the css to index.html (is it always as simple as that?) and set the icon set in `Vue.use(Vuetify, ...)`

appurist:
@jacekkarczmarczyk I forgot to comment on your reply, but I think that's a wonderful idea.

In practice though, it's not as simple as adding the css, etc, but that's probably the 90% case.  For icons there are a few options, first of course which icon set, then whether it will come from the common CDN, another web location, or be self-hosted, and finally which variant (we use the SVG/JS version rather than the CSS version, because our government IE11 network config does not permit downloadable fonts). 

But it would still be nice feature for other sites, and even in our case, we could just edit the index.html after the fact for our special needs; I think the rest would all be the same and would have been set up for us.

D10100111001:
I believe allowing icon overrides at the component instance level (props) would be a great idea as well.

PR #3609, opened 2018-03-18
fix(goTo) Always respect progressPercentage. Fixes #3608

### LINKED ISSUE #3608: [Bug Report] $vuetify.goTo never completes and never releases scroll
### Versions and Environment
**Vuetify:** 1.0.8
**Vue:** 2.5.16
**Browsers:** Chrome 64.0.3282.140
**OS:** Linux x86_64

### Steps to reproduce
Use browser's zoom feature, then programatically scroll to an off-screen element

### Expected Behavior
The scroll animation stops after the set duration.

### Actual Behavior
goTo keeps requesting animation frames as it will never reach the target. Thus, locking the scrolling behaviour permanently.

This never triggers:
```javascript
    if (Math.round(window.pageYOffset) === targetLocation) return
```

### Reproduction Link
[https://codepen.io/anon/pen/MVbyNp](https://codepen.io/anon/pen/MVbyNp)

### Other comments
I tested on my machine, and a possible solution would be to besides the check above, a check on `progressPercentage` could be done as well. This works for my reproduction link.

I can provide a PR if needed.

moustacheful:
## Description

When using `goTo` the `step` function will check `progressPercentage`. If it reaches 1 it means the animation should be finished, according to the given time.

## Motivation and Context

This fixes https://github.com/vuetifyjs/vuetify/issues/3608, where a different zoom setting could result in a target that will never be reached, thus locking the scroll permanently.

## How Has This Been Tested?

Visually.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/Vuetify/util/goTo.js
+++ b/src/components/Vuetify/util/goTo.js
@@ -74,7 +74,12 @@ export default function goTo (target, options) {
     let targetPosition = Math.floor(startLocation + distanceToScroll * easingFunction(progressPercentage))
 
     window.scrollTo(0, targetPosition)
-    if (Math.round(window.pageYOffset) === targetLocation) return
+
+    if (
+      Math.round(window.pageYOffset) === targetLocation ||
+      progressPercentage === 1
+    ) return
+
     window.requestAnimationFrame(step)
   }
 


src/components/Vuetify/util/goTo.js
```diff
@@ -74,7 +74,12 @@ export default function goTo (target, options) {
     let targetPosition = Math.floor(startLocation + distanceToScroll * easingFunction(progressPercentage))
 
     window.scrollTo(0, targetPosition)
-    if (Math.round(window.pageYOffset) === targetLocation) return
+
+    if (
+      Math.round(window.pageYOffset) === targetLocation ||
+      progressPercentage === 1
```
jacekkarczmarczyk: First condition could be removed (it will behave wrong for easing function like http://easings.net/en#easeInOutElastic - i'm aware that it's not in the scope of this issue) and `===` could be changed to `>=`

moustacheful: I wanted to introduce as little changes to the original code as possible, but what you're saying makes sense. If it's no issue with you guys I can make that modification in this same PR.

Regarding the `>=` though that should never trigger because of the `Math.min` a bit above that part. The maximum value of `progressPercentage` is effectively 1.

jacekkarczmarczyk: You're right, i missed the `min` part

PR #3614, opened 2018-03-19
refactor(VInput): convert createElement calls to JSX

KaelWD:
# Merge with rebase please

## Types of changes


johnleider:
It looks clean. We would need to come up with a style guide for it though.

johnleider:
I think that while this has its benefits, the overall goal can be achieved similar to what @jacekkarczmarczyk did with pickers. These can be abstracted into smaller components. Introducing JSX to one component would mean the others would need to be converted as well to maintain the code style, and I don't foresee us having time to do this anytime soon.

PR #3624, opened 2018-03-20
Temporary fixes #2111, workaround Safari z-index menu/dialog

### LINKED ISSUE #2111: v-select does not show the select items when in fullscreen v-dialog on iPhone

aldrinmartoq:
Hello, this is just a workaround to fix a nasty bug in Safari. When a v-select is put inside a full screen v-dialog, the z-index is below the current dialog, so the menu is not displayed.

## Description
I've debugged the following files (with console.log), and found that there is a bug in Safari or getZIndex() function that fails to calculate the zIndex correctly. This bug is present in Mac OS X and iOS versions of Safari.

## Motivation and Context
This PR fixes bug #2111.

We need this temporary fix because we need to run our apps in production ;-) and the original bug was reported on Oct/2017, 4 months ago.

It can gives us time to fix the real problem:
1. fix the getZIndex() function so it works always well in Safari
2. maybe there is something we can do to avoid Safari gives the wrong z-index. Seems something related to `position: fixed`, but I'm not sure.

## How Has This Been Tested?
1. Just debugged with console.log, and checked the problem was present in 1.0.8 and then it was gone with my fix.
2. Published our app and:
   - manually checked in Safari, iOS Simulator and a real iPhone; all of them with latests versions.
   - manually checked in Chrome, and Android
   - manually checked in Firefox 59

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDialog/VDialog.js
+++ b/src/components/VDialog/VDialog.js
@@ -149,7 +149,12 @@ export default {
     if (!this.fullscreen) {
       data.style = {
         maxWidth: this.maxWidth === 'none' ? undefined : (isNaN(this.maxWidth) ? this.maxWidth : `${this.maxWidth}px`),
-        width: this.width === 'auto' ? undefined : (isNaN(this.width) ? this.width : `${this.width}px`)
+        width: this.width === 'auto' ? undefined : (isNaN(this.width) ? this.width : `${this.width}px`),
+        zIndex: this.activeZIndex // temp workaround for #2111, set same zIndex of parent div
+      }
+    } else {
+      data.style = {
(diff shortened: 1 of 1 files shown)


johnleider:
closing in favor of #3624

PR #3642, opened 2018-03-21
Set active class if isActive

smiller171:
By allowing `isActive` to set the `activeClass` The initial active tab can be set with `v-model`

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VTabs/VTab.js
+++ b/src/components/VTabs/VTab.js
@@ -39,7 +39,7 @@ export default {
       return {
         'tabs__item': true,
         'tabs__item--disabled': this.disabled,
-        [this.activeClass]: !this.to && this.isActive
+        [this.activeClass]: this.isActive
       }
     },
     action () {


smiller171:
Should also change how you determine the active link to move the slider appropriately when using the back button https://github.com/vuetifyjs/vuetify/blob/f85456684bd7ef4e91ea34d297d6d1f3be7a98ad/src/components/VTabs/VTabs.js#L139

KaelWD:
That behaviour is intentional. If you are using the `to` prop you probably want to be syncing the tabs with vue-router, so you should be using the router to change the initial path instead. If you have some reasonable rationale for this change, please update the PR filling out all the required details.

smiller171:
@KaelWD I'm using Nuxt generate and want `/` to default to `/about/`. I tried redirecting from `/` to `/about/`, but when I do that my theme doesn't load. Seems to be something wrong with the Nuxt Layout not parsing correctly when doing a redirect. Instead I've set my index file to export about so it renders the same content, but that means if the user is on `/` I want the `about` tab to be active.

I've hacked around this right now to achieve the same result, but first-party support for being able to manually set the active tab instead of being completely dependent on the URL path would be helpful.

johnleider:

If you have any additional questions, please direct them to the official [Discord](https://chat.vuetifyjs.com) server.

PR #3680, opened 2018-03-27
Fix for double emitted 'update:returnValue' event in returnable mixin

mr-july:
The `returnable` mixin emits `update:returnValue` event twice on save.
The first time inside `save` method, the second time directly after
this when the value of `isActive` flag is changed to `false`
(in this case the watcher emits the event), so it is enough to just
set the `isActive` flag to false inside `save` method and the
`update:returnValue` event will be called "automatically" by watcher.


KaelWD:
Please update your PR to follow the [provided template](https://raw.githubusercontent.com/vuetifyjs/vuetify/dev/.github/PULL_REQUEST_TEMPLATE.md).

You should also rebase it onto master if it is a bug fix: 
```
git rebase --onto master HEAD~
git push --force-with-lease
```

image

PR #3682, opened 2018-03-27
 fix: `update:returnValue` event emitted twice in returnable mixin

mr-july:
The `returnable` mixin emits `update:returnValue` event twice on save.
The first time inside `save` method, the second time directly after
this when the value of `isActive` flag is changed to `false`
(in this case the watcher emits the event), so it is enough to just
set the `isActive` flag to false inside `save` method and the
`update:returnValue` event will be called "automatically" by watcher.

### DIFF
--- a/src/components/VTabs/VTab.js
+++ b/src/components/VTabs/VTab.js
@@ -45,8 +45,14 @@ export default {
     action () {
       let to = this.to || this.href
 
-      if (this.$router && this.to) {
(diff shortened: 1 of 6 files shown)


jacekkarczmarczyk:
Please remove unrelated commits and use the PR tempate (https://github.com/vuetifyjs/vuetify/blob/dev/.github/PULL_REQUEST_TEMPLATE.md) for the description

PR #3777, opened 2018-04-05
Don't offset last tab when scrolling

### LINKED ISSUE #3294: [Bug Report] Tabs with grow option move left when selecting second tab (on smaller screen)
### Versions and Environment
**Vuetify:** 1.0.0
**Vue:** 2.5.13
**Browsers:** Chrome 64.0.3282.119
**OS:** Windows 10

### Steps to reproduce
1. Open the pen
2. Select the second tab
3. Resize the window to make it less wide
4. Make window smaller until tab bar shifts left

### Expected Behavior
The tab bar should stay in place

### Actual Behavior
The tab bar receives a CSS style: "transform: translateX(...);" and moves left

### Reproduction Link
[https://codepen.io/anon/pen/eVGxRN](https://codepen.io/anon/pen/eVGxRN)

### Other comments
In the pen this issue is only triggered by resizing the window, when running a collapsible sidebar menu it is triggered on page load once the tab is switched.

smiller171:
fixes #3294

## Description
Check if active tab is last in list and don't add an offset when scrolling if it is.

## Motivation and Context
When using `grow` sometimes tabs would get scrolled (and offset) incorrectly.
This change also improves look and feel of normal scrolling tabs

## How Has This Been Tested?

visually

## Markup:

```vue
<template>
  <v-app>
    <v-toolbar
      dark
      dense
    >
      <v-tabs
        grow
      >
        <v-tabs-slider></v-tabs-slider>
        <v-tab
          v-for="item in tabs"
          :key="item.key"
        >
          {{ item.label }}
        </v-tab>
        <v-tab
          v-for="(item, index) in tabs"
          :key="index + 3"
        >
          {{ item.label }}
        </v-tab>
        <v-tab
          v-for="(item, index) in tabs"
          :key="index + 6"
        >
          {{ item.label }}
        </v-tab>
      </v-tabs>
    </v-toolbar>
  </v-app>
</template>

<script>
export default {
  data() {
    return {
      tabs: [
        {
          key: 'about',
          label: 'About',
          target: '/about/'
        },
        {
          key: 'resume',
          label: 'Resume',
          target: '/resume/'
        },
        {
          key: 'contact',
          label: 'Contact',
          target: '/contact/'
        }
      ]
    }
  }
}
</script>
```
## Types of changes

## Checklist:

### DIFF
--- a/src/components/VTabs/VTabs.js
+++ b/src/components/VTabs/VTabs.js
@@ -188,7 +188,10 @@ export default {
       const wrapperWidth = this.$refs.wrapper.clientWidth
       const totalWidth = wrapperWidth + this.scrollOffset
       const itemOffset = clientWidth + offsetLeft
-      const additionalOffset = clientWidth * 0.3
+      let additionalOffset = clientWidth * 0.3
+      if (this.activeIndex === this.tabs.length - 1) {
+        additionalOffset = 0 // don't add an offset if selecting the last tab
+      }
 
       /* instanbul ignore else */
       if (offsetLeft < this.scrollOffset) {


smiller171:
I don't agree with codecov here. Ideally there should be a special test case for the last tab.

johnleider:
This doesn't actually resolve the underlying issue. The problem is scrollOffset is not being properly reset when **isOverflowed** changes, which happens when going from mobile to desktop.

johnleider:
My current proposition for this fix: #3781. After running through, this ends up being more of a feature request. If I'm wrong please correct me.

smiller171:
@johnleider I think you're correct, though I believe this solves the presentation of the underlying issue as a side effect. That said, I would think that scrolling past the end in any situation is still a bug, as it sends a signal to the user that there are more options off to the right when there are not, and it's inconsistent with the behavior of the overscroll on the left side since the left side will not overscroll beyond the first tab.

johnleider:
It's only a bug if it wasn't intentional by us. The offset is intentional. Your logic in regards to it being a worse user experience, that may be, but this should still be a feature request to `dev` branch.

KaelWD:
Does this still apply, and if so can you explain what it's actually about? The original issue appears to be fixed now.

smiller171:
@KaelWD I'll have to build a new test case. Something has changed that radically altered how my codepen renders.

smiller171:
@KaelWD No, this isn't fixed. I'll get a codepen together in a minute, but the short version is that if you have 6 tabs and only room on screen for 4 tabs, when tab 5 is selected you expect the bar to scroll so that you can see part of tab6, so you know something is there, but you would expect that when tab 6 is selected the right edge of tab 6 should line up with the right edge of the tab bar so that it's clear you're at the end of the list of tabs. Instead the behavior seen is that there is a gap between the right edge of tab 6 and the right edge of the bar, making it appear to the user as if there is another interactive tab when there is not.

smiller171:
@KaelWD Here's an example pen. Make the view narrow enough to overflow the tabs and select tab 10 to see what I mean

https://codepen.io/smiller171/pen/MXrjEq

KaelWD:
That's what the spec says should happen, it's actually incorrect because the left side is supposed to have that offset too.

smiller171:
Could you point to the spec? This behavior increases user friction and as far as I can tell it deviates from every other material design implementation, including those by Google.

KaelWD:
I guess it doesn't say anything about the right-hand side, just that the left should align with the keyline. I don't really mind either way so I'm ok to merge this if the bossman is alright with it.

PR #3810, opened 2018-04-08
fix toolbar spacing

Hkh12:
## Description
I changed the padding and margin of buttons and toolbar title to let toolbar have more space and a better look:
image

## Motivation and Context
The previous toolbar spacing did not match the Material guidelines.

## How Has This Been Tested?
Visually.

## Types of changes


KaelWD:
While this *is* something that needs to be changed at some point, it goes further than just toolbars. You may have noticed that the keylines are **all** reversed between mobile/desktop compared to what material spec says.

PR #3832, opened 2018-04-11
feat: goTo returns a Promise

### LINKED ISSUE #3748: [Feature Request] $vuetify.goTo should return a promise
### Problem to solve
Sometimes we need to `then` and `catch` things after completed/aborted scroll (which happens after sometime due to animation duration).

### Proposed solution
`$vuetify.goTo` function should return a `Promise` which should be resolved when the scroll is successful and rejected when the scroll is aborted (which I guess doesn't happen now, but can be kept as a future hook).

Change in this API won't effect any existing code using `goTo`.

jacekkarczmarczyk:
## Description
`$vuetify.goTo()` returns now a Promise. 

## Motivation and Context
fixes #3748 

## How Has This Been Tested?
updated jest tests, added new

## Markup:
```vue
<template>
  <v-app>
    <v-content>
      <v-btn @click="goto()">Scroll me</v-btn>
      <div style="height: 2000px"></div>
      <div id="target">Target</div>
    </v-content>
  </v-app>
</template>

<script>
export default {
  methods: {
    goto () {
      this.$vuetify.goTo('#target').then(alert)
    }
  }
}
</script>
```

## Types of changes

## Checklist:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -89,7 +89,7 @@ declare interface VuetifyObject {
   dark: boolean
   theme: VuetifyTheme | false
   options: VuetifyOptions
-  goTo: (target: string | number | HTMLElement | Vue, options?: VuetifyGoToOptions) => void
+  goTo: (target: string | number | HTMLElement | Vue, options?: VuetifyGoToOptions) => Promise
 }
 
 declare module 'vue/types/vue' {
--- a/index.d.ts
+++ b/index.d.ts
@@ -89,7 +89,7 @@ declare interface VuetifyObject {
   dark: boolean
   theme: VuetifyTheme | false
   options: VuetifyOptions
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -89,7 +89,7 @@ declare interface VuetifyObject {
   dark: boolean
   theme: VuetifyTheme | false
   options: VuetifyOptions
-  goTo: (target: string | number | HTMLElement | Vue, options?: VuetifyGoToOptions) => void
+  goTo: (target: string | number | HTMLElement | Vue, options?: VuetifyGoToOptions) => Promise
```
KaelWD: `Promise` needs a type argument, in this case it's the same as `target` so I'd make `goTo` generic as well:

```ts
goTo: <T extends string | number | HTMLElement | Vue>(target: T, options?: VuetifyGoToOptions) => Promise<T>
```

PR #3837, opened 2018-04-12
refactor: convert components to typescript

### LINKED ISSUE #3943: [Bug Report] No typings for importing "A La Carte" components
### Versions and Environment
**Vuetify:** 1.0.17
**Vue:** 2.5.16
**Browsers:** Chrome 66.0.3359.117
**OS:** Windows 10

### Steps to reproduce
Run 

```
$ npm install -g vue-cli
$ vue init vuetifyjs/a-la-carte my-project
$ cd my-project
$ npm install
$ npm run dev
```

Then you add typescript support to `my-project` (change `main.js` to `main.ts` etc...) by following the official  vue page on typescript or something like https://github.com/Microsoft/TypeScript-Vue-Starter

Attempting to run this (i.e. `npm run dev`) will result in errors such as:

```
      TS2305: Module '"...../node_modules/vuetify/index"' has no exported member 'VBtn'.

```
which pertaining to the following import lines in `main.ts`

```typesccript
import {
  Vuetify,
  VApp,
  VNavigationDrawer,
  VFooter,
  VList,
  VBtn,
  VIcon,
  VGrid,
  VToolbar,
  transitions
} from 'vuetify'
```

### Expected Behavior
no errors. should build and run fine.

### Actual Behavior
has errors doesn't build or run

### Reproduction Link
[https://github.com/vuetifyjs/vuetify/blob/dev/index.d.ts](https://github.com/vuetifyjs/vuetify/blob/dev/index.d.ts)

### Other comments
Please update https://github.com/vuetifyjs/vuetify/blob/dev/index.d.ts 
You will see over there, that there are actually no references to exported `A La Carte` components etc..

Note that ***Even though this pertains to  A La Carte, it must actually be fixed in the main vuetify repo*** is why this issue has been opened here.

KaelWD:
## Description

This should be ready to go, I've got tests and builds working with no issues as far as I can tell, and it's actually smaller than the current dev build by ~8KB minified. 

So far I've converted VIcon, VBtn, VbtnToggle, and a few mixins and helpers. Everything else can be done incrementally over time. 

<details>
<summary>Possible component patterns:</summary>

```ts
// Regular options object
export default Vue.extend({
  ...
})

// Or with mixins
export default mixins(Foo, Bar).extend({
  ...
})
```

```ts
// Class with decorators, makes props a bit messy
@Component({
  props: {
    foo: String,
    bar: Boolean
  }
})
export default class FooBar extends Vue {
  foo: string
  bar: boolean
}

// Or
@Component
export default class FooBar extends Vue {
  @Prop() foo: string
  @Prop() bar: boolean
}
```

```ts
// Both?
const options = Vue.extend({
  props: {
    foo: String,
    bar: Boolean
  }
})

@Component
export default class FooBar extends options {
  ... // No need to declare props twice
}
```
</details>

&nbsp;
**Things to note:**
 - Components are now vue constructors, not plain objects. This means you have to use `SomeComponent.options` to access properties on the original object. 
 - To convert a component to TS, you must also convert everything that it imports. Probably best to start with mixins and smaller components so you don't have to do too much at once. 
 - Functional components currently don't inherit prop types from their mixins, which is why I've had to declare them manually in VIcon. 

Fixes #3943

## Types of changes

## TODO:
- [ ] The rest of the components
- [x] Check if sourcemaps are still good, considering it goes through two transpilers now
- [ ] Provide/inject with more than just register/unregister
- [ ] $refs declarations
- [x] Tests
- [x] Prod build
- [x] es5 build
- [x] custom linting rules


jsaak:
It may be just me, but it seems this work is pretty useless.
If someone wants typescript then there is angular.

jvdwijngaard:
I don't agree.

Vue is compatible with TypeScript as well, but Vuetify was not entirely ready for that.
The use of TypeScript has nothing to do with Angular. Thát's actually why I love Vue, they give you a choice in what you want to use, instead of forcing users to use TypeScript.

This helps using the Vuetify framework in TypeScript enabled Vue projects and still have the possibility to import only components you need/use (a-la-carte), instead of having to import the entire set of components and corresponding styling, keeping the output files as small as possible.

Ofcourse this is just a start and like Kael said, other components need to be converted, but I'm really happy this is happening.

jsaak:
As long as using typescript remains optional I do agree.
But reading the notes:

> Components are now vue constructors, not plain objects. This means you have to use
> SomeComponent.options to access properties on the original object.

>Functional components currently don't inherit prop types from their mixins, which is why I've had to >declare them manually in VIcon.

I had the feeling that it would be NOT optional.

KaelWD:
This makes 0 difference to you unless you are working on vuetify's itself.

PR #3845, opened 2018-04-13
Refactored input controls

sorinsarca:
## Description

Contains the changes from #3013 (only for checkbox, switch, radio and select).


src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
+  if (last < 0) return obj === undefined ? fallback : obj
+
+  for (let i = 0; i < last; i++) {
+    if (obj === null || obj === undefined) {
+      return fallback
+    }
+    obj = obj[path[i]]
+  }
+
+  if (obj === null || obj === undefined) return fallback
```
KaelWD: `== null` is the same as `=== null || === undefined`

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
 }
 
-export function getObjectValueByPath (obj, path) {
+export function getNestedValue (obj, path, fallback) {
+  const last = path.length - 1
+
+  if (last < 0) return obj === undefined ? fallback : obj
+
+  for (let i = 0; i < last; i++) {
+    if (obj === null || obj === undefined) {
```
KaelWD: Again here

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
+    return false
+  }
+
+  return props.every(p => looseEqual(a[p], b[p]))
+}
+
+export function getObjectValueByPath (obj, path, fallback) {
   // credit: http://stackoverflow.com/questions/6491463/accessing-nested-javascript-objects-with-string-key#comment55278413_6491621
-  if (!path || path.constructor !== String) return
+  if (!path || path.constructor !== String) return fallback
```
KaelWD: Any reason you're using `constructor` rather than `typeof`?

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
-    } else {
-      return
-    }
-  }
-  return obj
+  return getNestedValue(obj, path.split('.'), fallback)
+}
+
+export function getPropertyFromItem (item, property, fallback) {
+  if (property === null || property === undefined) return item === undefined ? fallback : item
```
KaelWD: Split into two ifs, without the ternary

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
-      obj = obj[k]
-    } else {
-      return
-    }
-  }
-  return obj
+  return getNestedValue(obj, path.split('.'), fallback)
+}
+
+export function getPropertyFromItem (item, property, fallback) {
```
KaelWD: Maybe add a comment explaining what this is for. Why does it take a function? The function signature should probably be added to the type definition too.

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
+
+  if (property.constructor === String) return getObjectValueByPath(item, property, fallback)
+
+  if (Array.isArray(property)) return getNestedValue(item, property, fallback)
+
+  if (typeof property !== 'function') return fallback
+
+  const value = property(item, fallback)
+
+  return typeof value === "undefined" ? fallback : value
```
KaelWD: Use single quotes.

src/util/helpers.js
sorinsarca: class X extends String {}
Anyway, it was like that before my commits. So I really don't know the reason.

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
+    }
+    obj = obj[path[i]]
+  }
+
+  if (obj === null || obj === undefined) return fallback
+
+  return obj[path[last]] === undefined ? fallback : obj[path[last]]
+}
+
+export function looseEqual (a, b) {
```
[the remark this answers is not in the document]
sorinsarca: Can be called like that

src/util/helpers.js
KaelWD: Ah right, didn't notice that in the diff.

src/mixins/selectable.js
```diff
@@ -29,12 +30,16 @@ export default {
     id: String,
     inputValue: null,
     falseValue: null,
+    trueValue: null,
+    multiple: {
+      type: Boolean,
+      default: null
```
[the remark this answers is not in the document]
sorinsarca: Yes, setting multiple to false and using array as a single value

src/util/helpers.js
sorinsarca: Sometimes only a path is not enough to get the desired property. The function receives as arg the item.

src/components/VSelect/VSelectList.js
```diff
@@ -144,23 +144,16 @@ export default {
...
-      return getObjectValueByPath(item, this.itemAvatar)
+      return Boolean(getPropertyFromItem(item, this.itemAvatar, false))
     },
     getDisabled (item) {
-      return getObjectValueByPath(item, this.itemDisabled)
+      return Boolean(getPropertyFromItem(item, this.itemDisabled, false))
     },
     getText (item) {
-      return this.getPropertyFromItem(item, this.itemText)
+      return (getPropertyFromItem(item, this.itemText, item) || '').toString();
```
KaelWD: Maybe `String()...`? depends how we want to handle falsy values I guess. 

Remove the semicolon too please.

src/util/helpers.js
```diff
@@ -99,21 +99,63 @@ export function addOnceEventListener (el, event, cb) {
...
-    } else {
-      return
-    }
-  }
-  return obj
+  return getNestedValue(obj, path.split('.'), fallback)
+}
+
+export function getPropertyFromItem (item, property, fallback) {
+  if (property == null) return item === undefined ? fallback : item
```
KaelWD: I meant 
```js
if (property === undefined) return fallback
if (property === null) return item
```

johnleider:
Merging this in as I need to progress with aspects that this touches and I don't want to keep breaking it.

PR #3902, opened 2018-04-22
refactor/input-controls-slider

### LINKED ISSUE #1686: Extend Slider to allow for Ranges and Custom Labels
Allow one OR two selectors on the slider to create a range selector versus a value selector, allow custom labels along selection axis, and embedded labels in thumbs when dragging for  UX purposes. Adding this ability would greatly extend the uses of the slider 

**What will it allow you to do that you can't do today?**
Select a range of values vs. a single value, and extend the semantic meaning of the selection.  This can be used to select a time range, day, date, month range, or any 'range' of values the developer chooses, and creates a more compact and useful UI component that imparts more meaning to the user.   

**How will it make current work-arounds straightforward?**
To accomplish this currently would require two sliders for basic range selection, which would use up more screen space, and be less intuitive to the user. 

**Suggested API extensions** 
- Range true/false ... true =  select a range or single value, false = single value only 
- Embedded Label Function ... if user provided function present slider would pass value, function would return label,  
- Discrete Labels true/false ... true then label the axis instead of embedding in the handle, only viable for discrete sliders, use values, or if user labeling function provided use return values from that.  

**Example**
image

### LINKED ISSUE #1969: [Bug] v-slider does not properly show error colors
### Steps to reproduce
Apply an error state to `v-slider`

### Versions
0.15.7

### What is expected ?
`v-slider` to change color for error state

The label changes color, track and thumb do not

### Reproduction Link
https://codepen.io/johnjleider/pen/WZEQVZ

### LINKED ISSUE #2591: [Feature Request] Add "change" event to v-slider
### New Functionality
Allow updating the user interface or doing any other sort of action only when the user has finished changing the value of a slider. This will enable similar behavior to v-text-field or plain `<input type=range>` which have separate input and change events. 

See also the definition of the change event: "The change event is fired for `<input>, <select>, and <textarea>` elements when a change to the element's value is committed by the user. Unlike the input event, the change event is not necessarily fired for each change to an element's value."

https://developer.mozilla.org/en-US/docs/Web/Events/change

### Improvements
The only way to come up with a similar behavior is to either debounce the input event or somehow tap into the raw DOM events (if it is even possible).

N/a

### LINKED ISSUE #2708: [Feature Request] v-slider: option to always show labels and thumb
### New Functionality
To see the labels and thumb of the slider, user has to touch on it. It would be better to always show labels and thumb (the current selected value) in some cases, at least to show the min, max and current value, for example, a scale-type question in a survey app, user can scale from 0 to 10, it will be much more clear for user if he sees directly min, max and the value he previously selected.

### Improvements
None

None

### LINKED ISSUE #3116: [Feature Request] VSlider has no events for isActive and relase
### New Functionality
I've simply added an event in methods.onMouseDown
```javascript
// on mousedown
this.isActive = true
this.$emit('active')

// onMouseUp
this.isActive = false
this.$emit('release')
```
so that i can listen to this event and programitically show the time or image thumbnail above the slider

### Improvements
NA

NA

### LINKED ISSUE #3567: [Bug Report] Slider is activated by drag not started on the slider
### Versions and Environment
**Vuetify:** 1.0.7
**Vue:** 2.5.13
**Browsers:** Firefox 58.0
**OS:** Mac OS 10.12

### Steps to reproduce
- Press and hold the left mouse button somewhere outside of the slider.
- Move the mouse over the slider.
- Release the mouse button.

### Expected Behavior
The slider should not move.

### Actual Behavior
The slider moves to where the mouse button was released.

### Reproduction Link
[https://codepen.io/anon/pen/zWreYR?editors=1010](https://codepen.io/anon/pen/zWreYR?editors=1010)

### Other comments
This is a minor issue but I noticed native range controls behave as expected (see code example). Seems to affect multiple browsers.

nekosaur:
## Description
Updates to `v-slider` fixing a number of bugs and adding functionality.

### TODO
- Add some tests

### API changes
- Added `range` prop to enable ranged slider.
- `ticks` is now a string, accepted values `''`, `'hover'` and `'always'`. Empty string is same as hover to keep backwards functionality
- `thumb-label` is now a string, accepted values `''`, `'hover'` and `'always'`. Empty string is same as hover to keep backwards functionality
- Added `thumb-label` scoped slot with props `{ index, value }` so that users can customize label.
- Added events `start` and `end` when moving slider using mouse.
- Added event `change` when new value has been 'committed'. Differs from `input` event in that it does not always emit every time value changes. Will emit when:
  - Clicking on slider
  - After moving slider with mouse
  - When using keyboard to move slider

## Motivation and Context
closes #1969
closes #3116
closes #2591
closes #3567
closes #1686
closes #2708

## How Has This Been Tested?
manually using markup

## Markup:
```vue
<template>
  <boilerplate>
    <v-btn @click="$vuetify.rtl = !$vuetify.rtl">rtl</v-btn>
    <h3>Normal</h3>
    <v-slider append-icon="home" v-model="slider" :label="`${slider}`"></v-slider>
    <h3>No v-model</h3>
    <v-slider append-icon="home"></v-slider>
    <h3>No v-model range</h3>
    <v-slider append-icon="home" range></v-slider>
    <h3>Validation</h3>
    <v-slider append-icon="home" v-model="slider" :label="`${slider}`" :rules="[v => v < 25 || 'Value above 25!']"></v-slider>
    <h3>Events</h3>
    <v-slider append-icon="home" v-model="slider" :label="`${slider}`" @start="start" @end="end" @change="change"></v-slider>
    <h3>Non-standard min/max</h3>
    <v-slider append-icon="home" v-model="slider4" :label="`${slider4}`" :min="-25" :max="25"></v-slider>
    <h3>Non-standard min/max range</h3>
    <v-slider append-icon="home" v-model="slider5" :label="`${slider5[0]} to ${slider5[1]}`" :min="-25" :max="25" range></v-slider>
    <h3>Range</h3>
    <v-slider append-icon="home" v-model="slider3" :label="`${slider3[0]} to ${slider3[1]}`" range></v-slider>
    <h3>Ticks hover</h3>
    <v-slider append-icon="home" v-model="slider2" :label="`${slider2}`" ticks :step="5"></v-slider>
    <h3>Ticks always on</h3>
    <v-slider append-icon="home" v-model="slider2" :label="`${slider2}`" ticks="always" :step="5"></v-slider>
    <h3>Thumb label hover</h3>
    <v-slider append-icon="home" v-model="slider2" :label="`${slider2}`" thumb-label></v-slider>
    <h3>Thumb label always on</h3>
    <v-slider append-icon="home" v-model="slider2" :label="`${slider2}`" thumb-label="always"></v-slider>
    <h3>Thumb label slot always on with range</h3>
    <v-slider append-icon="home" v-model="slider3" :label="`${slider3[0]} to ${slider3[1]}`" range thumb-label="always">
      <span slot="thumb-label" slot-scope="{value}">${{value}}</span>
    </v-slider>
  </boilerplate>
</template>

<script>
  export default {
    data: () => ({
      slider: 10,
      slider2: 0,
      slider3: [0, 30],
      slider4: 0,
      slider5: [-5, 5],
    }),
    methods: {
      start (e) {
        console.log('start', e)
      },
      end (e) {
        console.log('end', e)
      },
      change (e) {
        console.log('change', e)
      }
    }
  }
</script>
```

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VSlider/VSlider.js
+++ b/src/components/VSlider/VSlider.js
@@ -189,7 +189,7 @@ export default {
           tabindex: this.disabled ? -1 : this.tabindex
         },
         on: Object.assign({}, {
-          mouseup: this.onSliderMove,
+          click: this.onSliderMove,
           keydown: this.onKeyDown,
           keyup: this.onKeyUp
         }, this.$listeners),
@@ -307,7 +307,6 @@ export default {
       this.app.removeEventListener('touchmove', this.onMouseMove, options)
       this.app.removeEventListener('mousemove', this.onMouseMove, options)
 
-      this.$emit('change', this.inputValue)
       this.$emit('end', this.inputValue)
     },
     onMouseMove (e) {
--- a/src/components/VSlider/VSlider.js
+++ b/src/components/VSlider/VSlider.js
@@ -12,7 +12,7 @@ import VInput from '../VInput'
 import ClickOutside from '../../directives/click-outside'
 
 // Utilities
-import { addOnceEventListener, createRange, keyCodes } from '../../util/helpers'
+import { addOnceEventListener, createRange, keyCodes, deepEqual } from '../../util/helpers'
 import { consoleWarn } from '../../util/console'
 
 export default {
@@ -23,11 +23,12 @@ export default {
   directives: { ClickOutside },
 
   data: vm => ({
+    activeThumb: 0,
     app: {},
     defaultColor: 'primary',
     isActive: false,
     keyPressed: 0,
-    lazyValue: vm.value || 0
+    lazyValue: typeof vm.value !== 'undefined' && (Array.isArray(vm.value) ? vm.value : [vm.value]) || (vm.range ? [0, 0] : [0])
   }),
 
   props: {
@@ -41,6 +42,7 @@ export default {
       type: [Number, String],
       default: 100
     },
+    range: Boolean,
     step: {
       type: [Number, String],
       default: 1
@@ -55,7 +57,7 @@ export default {
       type: String,
       default: null
     },
-    value: [Number, String]
+    value: [Array, Number, String]
   },
 
   computed: {
@@ -75,7 +77,7 @@ export default {
       return this.disabled ? null : (this.trackColor || null)
     },
     computedThumbColor () {
-      if (this.disabled || !this.inputWidth) return null
+      if (this.disabled || !this.isDirty) return null
       return this.validationState || this.thumbColor || this.color || this.defaultColor
     },
     inputValue: {
@@ -84,29 +86,29 @@ export default {
       },
       set (val) {
         const { min, max } = this
-        val = Math.min(Math.max(val, min), max)
 
         // Round value to ensure the
         // entire slider range can
         // be selected with step
-        const value = this.roundValue(val)
+        let value = val.map(v => this.roundValue(Math.min(Math.max(v, min), max)))
+
+        // Switch values if range and wrong order
+        if (this.range && (value[0] > value[1] || value[1] < value[0])) {
+          if (this.activeThumb !== null) this.activeThumb = this.activeThumb === 1 ? 0 : 1
+          value = [value[1], value[0]]
+        }
+
         this.lazyValue = value
 
-        this.$emit('input', value)
+        if (!deepEqual(value, this.value)) this.$emit('input', this.range ? value : value[0])
         this.validate()
       }
     },
     stepNumeric () {
       return this.step > 0 ? parseFloat(this.step) : 1
     },
-    thumbStyles () {
-      return {
-        transition: this.keyPressed >= 2 ? 'none' : '',
-        left: `${this.$vuetify.rtl ? 100 - this.inputWidth : this.inputWidth}%`
-      }
-    },
     trackPadding () {
-      if (this.thumbLabel && this.isActive) return 0
+      if (this.thumbLabel || this.isActive) return 0
 
       return 6 + (this.isActive && !this.disabled ? 3 : 0)
     },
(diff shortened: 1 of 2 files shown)


johnleider:
I've abstracted the **range** prop to a new `v-range-slider` component. I still need to update tests and there is one bug with moving a slider past another.

src/components/VSlider/VSlider.js
```diff
@@ -41,106 +46,120 @@ export default {
...
+    range: Boolean,
     step: {
       type: [Number, String],
       default: 1
     },
-    ticks: Boolean,
+    ticks: {
+      type: [Boolean, String],
+      default: null,
+      validator: v => !v || ['always', 'hover'].indexOf(v) > -1
```
KaelWD: This isn't backwards-compatible.

src/components/VSlider/VSlider.js
```diff
@@ -41,106 +46,120 @@ export default {
...
+        'v-input--slider--ticks': this.showTicks,
         'v-input--slider--inverse-label': this.inverseLabel
       }
     },
+    showTicks () {
+      return !this.disabled && this.stepNumeric && !!this.ticks
+    },
+    showThumbLabel () {
+      return !this.disabled &&
+        (!!this.thumbLabel || this.thumbLabel === '')
```
KaelWD: `thumbLabel` won't be `''` unless explicitly set: `:thumbLabel="''"`

src/components/VSlider/VSlider.js
```diff
@@ -166,32 +185,47 @@ export default {
...
       return this.$createElement(VLabel, data, this.$slots.label || this.label)
     },
+    genListeners () {
+      return Object.assign({}, {
+        blur: this.onBlur,
+        click: this.onSliderClick,
+        focus: this.onFocus,
+        keydown: this.onKeyDown,
+        keyup: this.onKeyUp
+      }, this.$listeners)
```
KaelWD: You want to combine listeners, not override them. Each of these `onXyz` methods needs to re-emit the event.

PR #3952, opened 2018-04-27
feat(theme): allow a nonce to be specified for the stylesheet

### LINKED ISSUE #3950: [Bug Report] Vuetify requires unsafe-inline CSP
### Versions and Environment
**Vuetify:** 1.0.17
**Vue:** 2.5.13
**Browsers:** Safari 11.2, Google Chrome
**OS:** Mac OS 10.13.4

### Steps to reproduce
1) Use Vuetify on a website that uses Content Security Policy headers

### Expected Behavior
Website should load without issues

### Actual Behavior
Loading theme specific styles crashes on CSP, at `document.head.appendChild(style);` with error 

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' https: https://fonts.googleapis.com". Either the 'unsafe-inline' keyword, a hash ('sha256-AiuDn+GCYI1wpuA310/prXDmyv992zTtsIWKiuXkWCA='), or a nonce ('nonce-...') is required to enable inline execution.
```

### Reproduction Link
[https://codepen.io/anon/pen/zjooqW](https://codepen.io/anon/pen/zjooqW)

KaelWD:
## Motivation and Context

resolves #3950 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script

## Markup:

```js
Vue.use(Vuetify, {
  options: {
    cspNonce: '2726c7f26c'
  }
})
```
Resulting HTML:
```html
<style type="text/css" id="vuetify-theme-stylesheet" nonce="2726c7f26c">
  ...
```

## Types of changes

## Checklist:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -130,6 +130,8 @@ declare interface VuetifyOptions {
   themeVariations: string[]
   minifyTheme: ((css: string) => string) | null
   themeCache: VuetifyThemeCache | null
+  /** @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script */
+  cspNonce: string | null
 }
 
 declare type VuetifyGoToEasing =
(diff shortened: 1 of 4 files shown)


KaelWD:
Not really possible, it's only able to be set during bootstrapping. It's also a very trivial feature that won't really benefit from a unit test. I can tell jest to ignore the missed branch though if you like.

KaelWD:
nvm figured it out

PR #4003, opened 2018-05-07
Too loose selector for fill-height

mitar:
## Description

When you use `fill-height` on a container, currently any descendant layout gets `height: 100%`. This is not really what one would expect. You want just the next layout to be expanded, not all future ones.

For example, I want a full-height column where then I stack short rows. Without this fix all rows get expanded in height, without this change I can decide what I want to do with them: align them to start, end, or even expand. But it should not be default.

## How Has This Been Tested?

See this example: https://codepen.io/anon/pen/MGOybq?&editors=100

Here only `layout column` should have `height: 100%`, while `layout row` under it should not.

I changed this in my fork and run it locally to verify it fixes the problem.

## Types of changes

## Checklist:


src/stylus/components/_grid.styl
```diff
@@ -20,9 +20,12 @@
     align-items: center
     display: flex
 
-    .layout
+    > .layout
       height: 100%
       flex: 1 1 auto
+    
+    .layout.fill-height
+      height: 100%
```
KaelWD: Anything `.fill-height` already has 100%.

mitar: Yes, with this you tell which layout element you want to have fill-height. So first you do:

```
<container fill-height>
  <layout column>
    <layout row />
    <layout row />
  </layout>
</container>
```

So this is an example I am working with. A column with two rows. The above makes each of row 50% height because flex aligns them so. So if I want to make only one take the whole available space, you can do:

```
<container fill-height>
  <layout column>
    <layout row />
    <layout row fill-height />
  </layout>
</container>
```

But the issue, from my experimenting, is that this does not work because `.fill-height` is very non-specific selector and it does not take precedence over other stuff. I forgot now but I think something else was preventing it to come into effect, so this is why I added this rule.

One solution could be also to make `.fill-height` rule have `!important`.

KaelWD: Just checked this, there's nothing else overriding the height on `.layout` as far as I can tell. This rule should be removed.

mitar:
Thanks!

PR #4038, opened 2018-05-11
(wip) Add CSS variables

yordis:
## Description
Use CSS variables for the theme. My intention is to keep maintain the branch using CSS variables so when we are ready to switch it will some work done.

Also, worth to point out, I am just trying to remove the needs of the creating different themes, no necessarily embrace CSS variables good practices all over the place (baby steps).

Right now, keeping this branch open helps me to fix merge conflicts.

## Motivation and Context

- Small bundle size
- Faster rendering/painting
- Easier theme

## How Has This Been Tested?

I definitely will need your support on this, regression is done by human eyes catching all the issues; unless you have some advance tooling setup.

## Types of changes

### DIFF
--- a/src/stylus/components/_small-dialog.styl
+++ b/src/stylus/components/_small-dialog.styl
@@ -1,17 +1,4 @@
 @import '../bootstrap'
-@import '../theme'
-
-v-small-dialog($material)
-  &__content
-    background: $material.cards
-
-  &__actions
-    background: $material.cards
-
-  a
-    color: rgba($material.fg-color, $material.primary-text-percent)
-
-theme(v-small-dialog, "v-small-dialog")
 
 .v-small-dialog
   display: block
@@ -20,9 +7,11 @@ theme(v-small-dialog, "v-small-dialog")
 
   &__content
     padding: 0 24px
+    background: var(--vuetify-cards-color)
 
   &__actions
     text-align: right
+    background: var(--vuetify-cards-color)
 
   a
     display: flex
@@ -30,6 +19,7 @@ theme(v-small-dialog, "v-small-dialog")
 
     height: 100%
     text-decoration: none
+    color: var(--vuetify-small-dialog-link-color)
 
     > *
       width: 100%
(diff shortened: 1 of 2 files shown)


jacekkarczmarczyk:
Adding vars first to the theming mechanism would be a useful enhancement instead of breaking change

yordis:
@jacekkarczmarczyk that is my target, just theming and then go crazy with all the good features of CSS variables, but I want to minimize the breaking changes.

jacekkarczmarczyk:
I mean this https://github.com/vuetifyjs/vuetify/blob/dev/src/util/theme.js

yordis:
@jacekkarczmarczyk **at some point the usage of CSS variables is beyond the theming**, also I am confused with that file I don't see that much theming going on (I am new into Vuetify so I don't know the full source code)

The only one left is IE11 and at this point I would kill any support, specially that big piece of Vuetify users are using for admin tools primarily (but I don't care about it right now, I just want to keep this branch update it until everyone decide to move on with IE since all green browsers support CSS variables)

KaelWD:
I've considered doing that but never got around to it. 

Use case:
```vue
<template>
  ...
</template>

<style>
.my-class {
  color: var(--vuetify-primary); /* dynamic, will update if I change this.$vuetify.theme */
}
</style>
```

If you decide to this, it also has to work with SSR without flashing a default color or bloating the stylesheet too much.

KaelWD:
Another option would be to still use stylus variables to store everything, then render all variable usage from a mixin. This would allow a fallback to stylus if someone has to support IE.

yordis:
Folks, my intention is not use CSS variables for theming only by embrace all good the good practices which no preprocessor can do.

This is just the first step and to be honest I wouldn’t put any effort to support IE. I don’t expect you to take this PR in the next couple of months but it is out of my scope to care about IE and all the issues related.

CSS variables are not just for theming, and even just for that, you couldn’t do it in some use cases because you need the DOM structure

johnleider:
Due to the amount of changes coming before 2.0, I'd say this is an effort that should happen later on. Due to the amount of conflicts and current status, closing for now.

Issue #3924, opened 2018-04-24
[Bug Report] Dialog transitions can't be disabled


aldarund:
### Versions and Environment
**Vuetify:** 1.0.17
**Vue:** 2.5.16
**Browsers:** Chrome 66.0.3359.117
**OS:** Windows 10

### Steps to reproduce
Create a dialog with transition false , destroy it, observe that content continues to be displayed after the component is destroyed.

### Expected Behavior
Dialog show/hide immediately

### Actual Behavior
The dialog remains visible for a few hundred millsecs after being destroyed.

### Reproduction Link
[https://codepen.io/anon/pen/ZyVBgB?editors=1010](https://codepen.io/anon/pen/ZyVBgB?editors=1010)

KaelWD:
Regression of #960, which[ wasn't actually fixed. ](https://codepen.io/anon/pen/zjBqre?editors=1010)

aldarund:
The PR fixes dialog transition disable. But overlay still have transition, and it look and act pretty awkward as for me with that pr. Is it desired behavior that  while dialog appears immediately overlay still fade in and fadeout slowly ?

johnleider:
This was fixed in #4041

If you have any additional questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

PR #4041, opened 2018-05-11
Fixes bug #3924 where dialog transitions can't be disabled.

adriancable:
## Description
Fixes bug #3924 where dialog transitions can't be disabled.

## Motivation and Context
v-dialog includes the 'transition' tag even when transition is set to false. This causes a delay of a few hundred milliseconds when the v-dialog is opened or closed, leading to the feeling of an unresponsive UI. Details are in issue #3924.

## How Has This Been Tested?
Tested in my own vuetify project, to confirm that v-dialog now works properly both with and without transitions.

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDialog/VDialog.js
+++ b/src/components/VDialog/VDialog.js
@@ -193,14 +193,19 @@ export default {
       }, [this.$slots.activator]))
     }
 
-    const dialog = h('transition', {
-      props: {
-        name: this.transition || '', // If false, show nothing
-        origin: this.origin
-      }
-    }, [h('div', data,
-      this.showLazyContent(this.$slots.default)
-    )])
+    var dialog
+    if (this.transition) {
+      dialog = h('transition', {
+        props: {
+          name: this.transition,
+          origin: this.origin
+        }
+      }, [h('div', data, this.showLazyContent(this.$slots.default))])
+    } else {
+      dialog = h('div', data,
+        this.showLazyContent(this.$slots.default)
+      )
+    }
 
     children.push(h('div', {
       'class': this.contentClasses,


src/components/VDialog/VDialog.js
```diff
@@ -193,14 +193,19 @@ export default {
...
 
-    const dialog = h('transition', {
-      props: {
-        name: this.transition || '', // If false, show nothing
-        origin: this.origin
-      }
-    }, [h('div', data,
-      this.showLazyContent(this.$slots.default)
-    )])
+    var dialog
```
johnleider: use let instead

src/components/VDialog/VDialog.js
```diff
@@ -193,14 +193,19 @@ export default {
...
+    if (this.transition) {
+      dialog = h('transition', {
```
johnleider: The only thing that really changes here is the tag and props, let's see if we can simplify this a bit.

src/components/VDialog/VDialog.js
adriancable: Done!

src/components/VDialog/VDialog.js
adriancable: Done!

I looked at unit tests, but since the change fixes a UI latency issue it's difficult to functional test.

src/components/VDialog/VDialog.js
```diff
@@ -193,14 +193,18 @@ export default {
...
-    const dialog = h('transition', {
-      props: {
-        name: this.transition || '', // If false, show nothing
-        origin: this.origin
-      }
-    }, [h('div', data,
-      this.showLazyContent(this.$slots.default)
-    )])
+    let dialog
+    let baseDialog = h('div', data, this.showLazyContent(this.$slots.default))
```
johnleider: What about:

```
let dialog = h('div', data, this.showLazyContent(this.$slots.default))

if (this.transition) {
  dialog = h('transition', {
    props: {
      name: this.transition,
      origin: this.origin
    }
  }, [dialog])
}
```

adriancable: Done, and tested. Thanks for the suggestion.

Issue #4062, opened 2018-05-15
[Bug Report] Slider bringing up keyboard and cursor after tap on mobile


Helmi:
### Versions and Environment
**Vuetify:** 1.1.0-alpha.4
**Vue:** 2.5.16
**Browsers:** Safari, Google Chrome
**OS:** Android, iOS

### Steps to reproduce
Tap on the slider from a mobile device (happens on Android and iOS)

### Expected Behavior
Slider moves to the position of the tap.

### Actual Behavior
Slider moves but also the keyboard comes up and a cursor shows up on the left side of the slider (likely due to the input element underneath)

### Reproduction Link
[https://codepen.io/Helmi74/pen/JvaKyL](https://codepen.io/Helmi74/pen/JvaKyL)

KaelWD:
Fixed already: 641426c2b24374d5f35d45585327acd2032070d7

PR #4098, opened 2018-05-19
fix(VSelect): show input placeholder

KaelWD:
## Description

reverts 37c9e539 and part of 641426c2

v-select now can't be tabbed to, it needs tabindex and styles on the root element like in 1.0

## Markup
```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete placeholder="placeholder (already works on dev)"></v-autocomplete>
      <v-select placeholder="can't tab to this any more"></v-select>
    </v-container>
  </v-app>
</template>
```

## Types of changes


KaelWD:
Yeah, it does cause new problems with tabbing though. You're welcome to finish it if you want.

johnleider:
I would if I knew what it was fixing, playground?

KaelWD:
Thought I'd put one already. Updated.

johnleider:
So testing on current dev, the placeholder does not show (as expected by the style that needs changing), but I am able to tab to it because it is readonly. With this PR I am unable to tab to it and by default is disabling the select input which is not correct.

KaelWD:
:man_facepalming: Fuck I'm retarded, no idea why I changed that.

KaelWD:
Oh right, I think it was that IE11 cursor thing. I'm on my shitbox linux laptop rn so I can't check.

KaelWD:
I think it should be ok still - #4062

Issue #3967, opened 2018-04-29
[Bug Report] v-data-table duplicates icons in rows when sorting


creynir:
### Versions and Environment
**Vuetify:** 1.0.17
**Vue:** 2.5.11
**Browsers:** Chrome 66.0.3359.117
**OS:** Mac OS 10.13.4

### Steps to reproduce
1)  Try to sort table with icons in rows

### Expected Behavior
Rows are sorted included icons 

### Actual Behavior
Icons don't change the placement, there is duplication issue also

### Reproduction Link
[https://codepen.io/creynir/pen/rvjrWq](https://codepen.io/creynir/pen/rvjrWq)

### Other comments
I tried many ways to bind icons to the rows, but the problem with wrong placement stays.

KaelWD:
Font awesome 5's js doesn't work with Vue: https://github.com/vuetifyjs/vuetify/issues/3505#issuecomment-371560727

creynir:
Thank you, i tried to add something like a key, but didn't thought about <tr> tag, it seems like a good solution.

jacekkarczmarczyk:
You can also add key to 'td' tag

PR #4294, opened 2018-06-13
Fixes #3841 VDataTable: duplicate key

### LINKED ISSUE #3841: [Bug Report] Datatable component fails with two identical header texts
### Versions and Environment
**Vuetify:** 1.0.14
**Vue:** 2.5.16
**Browsers:** Firefox 59.0
**OS:** Windows 10

### Steps to reproduce
1. Open codepen https://codepen.io/anon/pen/GxaoRV
2. Check console. It says ""[Vue warn]: Duplicate keys detected: ''. This may cause an update error.
found in ---> <VDataTable>"
3. Change list array by adding the missing "text" value
4. voilá, it works.

### Expected Behavior
No error.

### Actual Behavior
Console error, although table indeed renders.

### Reproduction Link
[https://codepen.io/anon/pen/GxaoRV](https://codepen.io/anon/pen/GxaoRV)

### Other comments
Sometimes you just don't want to add header text to all columns. I assume that for whatever reason the text value from the header was used as 'item-key'. The error exists if two headers have any identical text values not just in the case of empty strings.

dmecke:
## Description
As in #3841 reported, using the same header text in a VDataTable causes vue to output an console error "duplicate key". This pull request keeps the original behaviour, but allows to pass a `key` property to the headers object to be used as the internal key instead. This allows using the same header text while using different keys.

## Motivation and Context

Fixes #3841

## How Has This Been Tested?

visually tested via the dev tools console

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDataTable/mixins/head.js
+++ b/src/components/VDataTable/mixins/head.js
@@ -58,7 +58,7 @@ export default {
     genHeaderData (header, children) {
       const classes = ['column']
       const data = {
-        key: header[this.headerText],
+        key: header['key'] ? header['key'] : header[this.headerText],
         attrs: {
           role: 'columnheader',
           scope: 'col',


src/components/VDataTable/mixins/head.js
```diff
@@ -58,7 +58,7 @@ export default {
     genHeaderData (header, children) {
       const classes = ['column']
       const data = {
-        key: header[this.headerText],
+        key: header['key'] ? header['key'] : header[this.headerText],
```
jacekkarczmarczyk: I'd rather see it as just `key: header[this.headerKey]` with new `headerKey` prop defaulted to `"text"`

KaelWD: Shouldn't it use `value`?

jacekkarczmarczyk: `value` could be duplicated too (if someone wants to display same column twice, for example as a first one and last one)

dmecke: Yes, that's exactly my use case. I am not that deep in the internals of vuetify yet, but as far as I understand, `key` simply needs to be something unique here. Generating it somehow would of course be preferable to passing somehting manually. Can we maybe access the column index somehow to append it to the `key`?

dmecke: Maybe pass the index down from https://github.com/vuetifyjs/vuetify/blob/23888743866cf7c74568f3810d937bc99fbe8b10/src/components/VDataTable/mixins/head.js#L29?

jacekkarczmarczyk: Index could be a solution as well, you just need to pass it in `genTHead()` and maybe set the default value of `headerKey` to be `null` (that means index should be used)
Something like
`const row = this.headers.map((o, i) => this.genHeader(o, this.headerKey ? o[this.headerKey] : i))`

dmecke: Do we need the `headerKey` prop at all then? I mean, could there be an use case where someone would want to pass a key explicitly? Again, I am not really into the internals and think that the `key` is only there for being unique. So if this is guaranteed, there would be no need to override it.

jacekkarczmarczyk: You may want to fully rerender your header cell when the headers are dynamically updated. Sometimes it can produce a bug if it uses index as a key, see for example https://github.com/vuetifyjs/vuetify/issues/3967

dmecke:
Now the coumn index is used as the key. It can be overridden by using the new `headerKey` prop.

jacekkarczmarczyk:
Can you also provide the Playground.vue (minimal, probably just 2 headers and no items should be enough)

dmecke:
I've added the Playground.vue now.

jacekkarczmarczyk:
Looks fine for me. We're on feature freeze and it's kinda feature, but quite minor and new prop is required to fix the existing bug so imho it still can go to 1.1. @nekosaur?

Last one thing @dmecke - this should be rebased to `dev` branch

dmecke:
I've rebased to `dev` now.

src/components/VDataTable/mixins/head.js
```diff
@@ -26,7 +26,7 @@ export default {
 
         children = [this.hasTag(row, 'th') ? this.genTR(row) : row, this.genTProgress()]
       } else {
-        const row = this.headers.map(o => this.genHeader(o))
+        const row = this.headers.map((o, i) => this.genHeader(o, this.headerKey ? o[this.headerKey] : i))
```
KaelWD: Should we give a warning if `headerKey` is set but now defined in the object? Use the `consoleError` helper.

dmecke: Shall I add it?

jacekkarczmarczyk: We don't add it here https://github.com/vuetifyjs/vuetify/blob/dev/src/components/VDataTable/mixins/body.js#L47, i don't think it's necessary

PR #4301, opened 2018-06-14
fix(translatable): removed normalizedHeight

jacekkarczmarczyk:
## Motivation and Context
translatable::normalizedHeight was a mess, returning string 'NNNpx' with `jumbotron` prop and then adding it or subsctracting from numbers, also `.replace(/(^[0-9]*$)/, '$1')` does nothing

## Types of changes


johnleider:
Poor parallax. I think it was the second component I created for the framework. It's old as dirt.

PR #4318, opened 2018-06-16
fix: removed unnecessary prefix from lang keys

nekosaur:
## Description
When using our internal translation engine there is no need to use a prefix. It only leads to verbose code, e.g. `{{ $vuetify.t('$vuetify.lang.component.key') }}`. Now it's `{{ $vuetify.t('component.key') }}`. Prefixing the keys should only be relevant if we perhaps want to integrate with vue-i18n down the line? But then it should be user configurable.

## Motivation and Context
See above

## How Has This Been Tested?
manually

## Markup:
n/a

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VDataTable/VDataTable.js
+++ b/src/components/VDataTable/VDataTable.js
@@ -45,7 +45,7 @@ export default {
     hideHeaders: Boolean,
     rowsPerPageText: {
       type: String,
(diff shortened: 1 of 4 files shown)


wcjr:
I'm seeing the prefix is still required in `1.2.6`, correct?

PR #4351, opened 2018-06-20
refactor: convert theme.js to typescript

jacekkarczmarczyk:
## Motivation and Context
learning typescript

## How Has This Been Tested?
yarn lint

## Types of changes


src/util/theme.ts
```diff
@@ -1,14 +1,15 @@
 import { colorToInt, intToHex } from './colorUtils'
 import * as sRGB from './color/transformSRGB'
 import * as LAB from './color/transformCIELAB'
+import { VuetifyTheme, RGB, VuetifyParsedTheme } from 'types'
 
 /**
- * @param {object} theme
- * @returns {object}
+ * @param {VuetifyTheme} theme
+ * @returns {VuetifyParsedTheme}
```
KaelWD: docblock can be removed now

types/index.d.ts
```diff
@@ -114,7 +114,11 @@ export interface VuetifyBreakpoint {
   xsOnly: boolean
 }
 
+export type RGB = number
+export type XYZ = [number, number, number]
+export type LAB = [number, number, number]
 export type VuetifyThemeItem = string | number
+export type VuetifyParsedThemeItem = RGB
```
KaelWD: These are all internal, should be somewhere in utils instead. `VuetifyParsedTheme` is compatible with `VuetifyTheme` and doesn't need to exist.

types/index.d.ts
```diff
@@ -114,7 +114,10 @@ export interface VuetifyBreakpoint {
   xsOnly: boolean
 }
 
-export type VuetifyThemeItem = string | number
+export type RGB = number
+export type XYZ = [number, number, number]
+export type LAB = [number, number, number]
```
KaelWD: That goes for all these too

src/util/theme.ts
```diff
@@ -20,77 +21,69 @@ export function parse (theme) {
...
 .${name}--text input,
 .${name}--text textarea {
-  caret-color: ${value} !important;
+  caret-color: ${rgb} !important;
 }`
 }
 
 /**
  * Generate the CSS for a variant color (.primary.darken-2)
  *
```
KaelWD: Remove blank line

src/util/theme.ts
```diff
@@ -1,14 +1,15 @@
 import { colorToInt, intToHex } from './colorUtils'
 import * as sRGB from './color/transformSRGB'
 import * as LAB from './color/transformCIELAB'
+import { VuetifyTheme, RGB } from 'types'
 
 /**
- * @param {object} theme
- * @returns {object}
+ * @param {VuetifyTheme} theme
+ * @returns {VuetifyTheme}
```
KaelWD: remove

types/index.d.ts
```diff
@@ -114,7 +115,7 @@ export interface VuetifyBreakpoint {
   xsOnly: boolean
 }
 
-export type VuetifyThemeItem = string | number
+export type VuetifyThemeItem = string | RGB
```
KaelWD: Leave this as it was, that's just complicating stuff for no reason

src/util/colorUtils.ts
```diff
@@ -1,9 +1,13 @@
 import { consoleWarn } from './console'
 
+export type RGB = number
+export type XYZ = [number, number, number]
+export type LAB = [number, number, number]
```
KaelWD: Do non-distinct types help here? Typescript isn't nominally typed, so calling `number` something else doesn't change anything.

src/util/theme.ts
```diff
@@ -1,14 +1,11 @@
...
 
-/**
- * @param {object} theme
- * @returns {object}
- */
-export function parse (theme) {
+export function parse (theme: VuetifyTheme): VuetifyTheme {
   const colors = Object.keys(theme)
-  const parsedTheme = {}
+  const parsedTheme: any = {}
```
KaelWD: Try to avoid `any`, use the actual type (`VuetifyTheme`) if possible.

src/util/colorUtils.ts
KaelWD: It's only ever used from a namespace, so it's obvious what color format is returned

`LAB.fromXYZ(...`

KaelWD: Although I guess vscode does show the actual type, so it isn't too bad. I'd definitely remove it from the public type file however.

PR #4425, opened 2018-06-29
feat: add russian locale

Djaler:
Just a russian locale


jacekkarczmarczyk:
@vuetifyjs/dev-team should we name files ru.js, en.js, es.js or ru_RU.js, en_US.js, en_GB.js, es_MX.js etc?

KaelWD:
Simple names are fine, especially for languages with only slight regional variation.

PR #4532, opened 2018-07-07
refactor(theme): remove from v-app, convert to TS

KaelWD:
## Types of changes


src/components/Vuetify/services/ThemeService.ts
```diff
@@ -46,10 +70,10 @@ export default {
 
       return css
     },
-    vueMeta () {
-      if (this.$vuetify.theme === false) return
+    vueMeta (): object | undefined {
+      if (this.$vuetify.theme as any === false) return
```
KaelWD: We can probably do this in `Vuetify/index.ts` instead, because it isn't supposed to be reactive

```
if (options.theme !== false) {
  new ThemeService()
}
```

KaelWD: Done in 45adc011338d8eaaab99ea691faf45523506d581

PR #4560, opened 2018-07-10
refactor(v-bottom-nav): convert to typescript

jacekkarczmarczyk:
## How Has This Been Tested?
lint

## Types of changes


src/components/VBottomNav/VBottomNav.ts
```diff
@@ -6,40 +6,45 @@ import Applicationable from '../../mixins/applicationable'
...
 import Colorable from '../../mixins/colorable'
 
+// Util
+import mixins from '../../util/mixins'
+
+// Types
+import { VNode } from 'vue'
+import { PropValidator } from 'vue/types/options'
+
 /* @vue/component */
```
KaelWD: This needs to be directly above the `.extend({`

PR #4598, opened 2018-07-12
fix(types): prevent type leak in Toggleable, add $el to VBtn

KaelWD:
## Types of changes


src/components/VBadge/VBadge.ts
```diff
@@ -49,7 +49,7 @@ export default mixins(
     const badge = this.$slots.badge ? [h('span', {
       staticClass: 'v-badge__badge',
       'class': this.addBackgroundColorClassChecks(),
-      attrs: this.attrs,
+      attrs: this.$attrs,
```
KaelWD: Found this after I fixed the `Record<string, any>` leak.

PR #4613, opened 2018-07-14
Localize numbers in pagination

alireza4050:
## Description
Localize presentation of numbers based on current locale in data-iterable mixin.

## Motivation and Context
To localize numbers in v-data-table and v-data-iterator pagination. 

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <v-app dark>
    <v-content>
      <v-container fill-height>
        <v-layout column
                  justify-space-between>
          <v-data-table :headers="headers"
                        :items="desserts"
                        class="elevation-1">
            <template slot="items"
                      slot-scope="props">
              <td>{{ props.item.name }}</td>
              <td class="text-xs-right">{{ props.item.calories }}</td>
              <td class="text-xs-right">{{ props.item.fat }}</td>
              <td class="text-xs-right">{{ props.item.carbs }}</td>
              <td class="text-xs-right">{{ props.item.protein }}</td>
              <td class="text-xs-right">{{ props.item.iron }}</td>
            </template>
          </v-data-table>
          <v-spacer></v-spacer>
          <v-data-iterator :items="desserts"
                           content-tag="v-layout"
                           row
                           wrap>
            <v-flex slot="item"
                    slot-scope="props"
                    xs12
                    sm6
                    md4
                    lg3>
              <v-card>
                <v-card-title>
                  <h4>{{ props.item.name }}</h4>
                </v-card-title>
                <v-divider></v-divider>
                <v-list dense>
                  <v-list-tile>
                    <v-list-tile-content>Calories:</v-list-tile-content>
                    <v-list-tile-content class="align-end">{{ props.item.calories }}</v-list-tile-content>
                  </v-list-tile>
                  <v-list-tile>
                    <v-list-tile-content>Protein:</v-list-tile-content>
                    <v-list-tile-content class="align-end">{{ props.item.protein }}</v-list-tile-content>
                  </v-list-tile>
                </v-list>
              </v-card>
            </v-flex>
          </v-data-iterator>
          <v-switch :label="`Language: ${this.$vuetify.lang.current}`"
                    v-model="langSwitch"></v-switch>
        </v-layout>
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
export default {
  mounted() {
    this.langSwitch = true;
  },
  computed: {
    langSwitch: {
      get() {
        return this.$vuetify.lang.current === 'fa-IR';
      },
      set(val) {
        this.$vuetify.lang.current = val ? 'fa-IR' : 'en';
      }
    }
  },
  data() {
    return {
      _radioBtn: 0,
      headers: [
        {
          text: 'Dessert (100g serving)',
          align: 'left',
          sortable: false,
          value: 'name'
        },
        { text: 'Calories', value: 'calories' },
        { text: 'Protein (g)', value: 'protein' },
      ],
      desserts: [
        {
          value: false,
          name: 'Frozen Yogurt',
          calories: 159,
          protein: 4.0,
        },
        {
          value: false,
          name: 'Ice cream sandwich',
          calories: 237,
          protein: 4.3,
        }
      ]
    }
  }
}
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/src/mixins/data-iterable.js
+++ b/src/mixins/data-iterable.js
@@ -168,7 +168,7 @@ export default {
           ? Object.assign({}, item, {
             text: this.$vuetify.t(item.text)
           })
-          : item
+          : { value: item, text: Number(item).toLocaleString(this.$vuetify.lang.current) }
       })
     },
     hasPagination () {
@@ -454,7 +454,8 @@ export default {
             pageStop: stop,
             itemsLength: this.itemsLength
           })
-          : this.$vuetify.t('$vuetify.dataIterator.pageText', this.pageStart + 1, stop, this.itemsLength)
+          : this.$vuetify.t('$vuetify.dataIterator.pageText',
+            ...([this.pageStart + 1, stop, this.itemsLength].map(n => Number(n).toLocaleString(this.$vuetify.lang.current))))
       }
 
       return this.$createElement('div', {


src/mixins/data-iterable.js
```diff
@@ -45,9 +45,9 @@ export default {
       type: Array,
       default () {
         return [
-          5,
-          10,
-          25,
+          { value: 5, text: 5 },
+          { value: 10, text: 10 },
+          { value: 25, text: 25 },
```
KaelWD: Why do these need to be objects?

src/components/Vuetify/mixins/lang.ts
```diff
@@ -29,6 +29,9 @@ export default function lang (config: Options['lang'] = {}): VuetifyLanguage {
     locales: Object.assign({ en }, config.locales),
     current: config.current || 'en',
     t (key, ...params) {
+      if (typeof key === 'number') {
```
KaelWD: Should be added to the public type definitions, then you can remove `as number` because it'll be narrowed.

src/mixins/data-iterable.js
alireza4050: > Why do these need to be objects?

In the `computedRowsPerPageItems`method, if it's an object the `text` will be passed to the `t` method and since the text is a number instead of a string, it get's localized in the 'if' statement I added.

src/components/Vuetify/mixins/lang.ts
alireza4050: > Should be added to the public type definitions

Will do &#128077;

src/mixins/data-iterable.js
KaelWD: Wouldn't it make more sense to pass it through `t()` on the other side of that ternary too?

alireza4050: No, beacause then the values could be non-english digit charcters and it won't work in the rest of the code where the actual numbers are needed. Coercion of string to number only happens for strings of ascii digits.
We could however convert it to an object inside the computed method instead of changing the default items to objects

alireza4050:
Also we could leave the `t` method alone and just call `toLocaleString` in the four places needed. But the `t` method translating numbers might be a usefull feature in itself

alireza4050:
Removed changes to `t` method

PR #4675, opened 2018-07-20
Allow overriding the default breakpoint object pixel thresholds

cb109:
That will enable Users to globally adjust these values from javascript to fix certain browser inconsistencies or clashes with other frameworks across resolutions e.g. for safari.

## Description
`$vuetify.breakpoint` now allows to override the pixel values used to calculate which breakpoint ranges we are on by exposing two new attributes:
- `$vuetify.breakpoint.thresholds`
   - `xs: 600`
   - `sm: 960`
   - `md: 1280`
   - `lg: 1920`
- `$vuetify.breakpoint.scrollbarWidth: 16`

## Motivation and Context
A user approached me having difficulties in a project that mixes bootstrap with vuetify. They seem to have different thresholds and they also differ depending on the browser (he was testing in Safari). He was able to override `$grid-breakpoints` stylus helpers, but the breakpoint's thresholds were hardcoded so far. These are now exposed so they can be adjusted if needed, e.g. in the main.js or App.vue based on the current browser.

## How Has This Been Tested?

I have extended the following unit test file to make sure the overrides don't break existing functionality and are recognized when calculating the current breakpoints: `vuetify/test/unit/components/Vuetify/Vuetify.breakpoint.spec.js`

Run like:

    $ jest test/unit/components/Vuetify/Vuetify.breakpoint.spec.js

I have also updated the type/interface definition, please see the diff.

## Types of changes
## Checklist:
- [x] My change requires a change to the documentation. -> This page is affected: https://vuetifyjs.com/en/layout/breakpoints#breakpoint-object
- [x] I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes. -> https://github.com/vuetifyjs/vuetifyjs.com/pull/539

### DIFF
--- a/src/components/Vuetify/index.ts
+++ b/src/components/Vuetify/index.ts
@@ -34,7 +34,7 @@ const Vuetify: VuetifyPlugin = {
 
     Vue.prototype.$vuetify = new Vue({
       mixins: [
-        breakpoint
+        breakpoint(opts.breakpoint)
       ],
       data: {
         application,
--- a/src/components/Vuetify/mixins/breakpoint.ts
+++ b/src/components/Vuetify/mixins/breakpoint.ts
@@ -22,16 +22,16 @@ export default Vue.extend({
       md: 1280,
       lg: 1920
     },
-    scrollbarOffset: 16
+    scrollbarWidth: 16
   }),
 
   computed: {
     breakpoint (): VuetifyBreakpoint {
       const xs = this.clientWidth < this.thresholds.xs
       const sm = this.clientWidth < this.thresholds.sm && !xs
-      const md = this.clientWidth < (this.thresholds.md - this.scrollbarOffset) && !(sm || xs)
-      const lg = this.clientWidth < (this.thresholds.lg - this.scrollbarOffset) && !(md || sm || xs)
-      const xl = this.clientWidth >= (this.thresholds.lg - this.scrollbarOffset)
+      const md = this.clientWidth < (this.thresholds.md - this.scrollbarWidth) && !(sm || xs)
+      const lg = this.clientWidth < (this.thresholds.lg - this.scrollbarWidth) && !(md || sm || xs)
+      const xl = this.clientWidth >= (this.thresholds.lg - this.scrollbarWidth)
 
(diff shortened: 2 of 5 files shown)


cb109:
One question: `scrollbarOffset`: I named it like this without actually knowing for certain, if those 16px are subtracted because of a scrollbar or something else. If that naming is wrong please suggest an alternative name, thanks.

johnleider:
If we did this I would want it to be configurable from the Vuetify bootstrap. That means this is going to have to turn into a factory function.

cb109:
So something like this?

```javascript
Vue.use(Vuetify, {
  breakpoint: {
    thresholds: {
      xs: 360
    }
  }
})
```

- [ ] Will look into that and update

SmootherSpike:
As in the related issue, please consider to centralize the breakpoints, that you do not have to maintain them in _variables.styl and breakpoints.ts. Such an example would look like:

`json("../../config/screenSizes.json") $grid-breakpoints := { xs: 0 sm: unit(sm-lower, px)....` and 
`import screenSizes from '../../../config/screenSizes.json' ...
breakpoint (): VuetifyBreakpoint {
      const xs = this.clientWidth < this.screenSizes.xs.upper...`
This would make this issue much cleaner than it is right now.

jacekkarczmarczyk:
That would rather go to this PR: https://github.com/vuetifyjs/vuetify/pull/4705

cb109:
Both this PR and the sidekick docs-PR updated. Mixin is installed from a factory function now that we pass the optional threshold overrides during app bootstrap.

I tried to keep it simple as there is a lot of movement right now around the breakpoint object with the other PRs. @SmootherSpike I agree, but would see this as a next step. First step is to make it configurable at all, imho.

johnleider:
This Pull Request is being closed due to unresolved merge conflicts. Once corrected we will re-open.

Thank you for your contribution and interest in improving Vuetify!

cb109:
I've opened a separate PR instead of reopening this one as requested by @jacekkarczmarczyk, please see #5666 from here on

PR #4705, opened 2018-07-24
Optimize breakpoints mixin

qm3ster:
## Description
Instead of being a mixin to `$vuetify` instance, exporting a single `breakpoint` computed property `breakpoint` is now its own instance of Vue, with the various `xs` and `mdAndDown` as individual
properties. Since these are implemented as instance `data` and not `computed`, they don't bother
their dependents on resize if the value of the actual dependency is unchanged.
The performance of actual matching breakpoints is also separately improved.

## Motivation and Context
Less useless execution on resize, perhaps now we will be able to change the debounce value
from 200ms to 100ms or less.

## How Has This Been Tested?
The actual mixin's synthetic unit test was slightly tweaked.
Other tests that rely on breakpoints passed without modification.
I didn't test visually (yet?).

## Types of changes

## Checklist:

- [ ] The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes). **Dunno, can we shove this into `master`?**
- [x] My code follows the code style of this project. *switch to Prettier already*


src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -111,8 +83,16 @@ export default Vue.extend({
       this.resizeTimeout = window.setTimeout(this.setDimensions, 200)
     },
     setDimensions (): void {
-      this.clientHeight = getClientHeight()
-      this.clientWidth = getClientWidth()
+      const width = getClientWidth()
```
jacekkarczmarczyk: this method (after small tweaking) could be used in `data()` to avoid code duplication

src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -1,90 +1,62 @@
...
 import { VuetifyBreakpoint } from 'types'
 
+const breakpoints = {
+  xs: 600,
+  sm: 960,
+  md: 1280 - 16,
+  lg: 1920 - 16,
+  xl: Infinity
+}
+const keys = Object.keys(breakpoints) as (keyof typeof breakpoints)[]
```
KaelWD: Use the `keys` function from util/helpers

src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -1,90 +1,62 @@
...
+  const result: Record<
+  string,
+  keyof Omit<VuetifyBreakpoint, 'width' | 'height'> | boolean
+  > = { name }
+  keys.forEach((key: keyof typeof breakpoints, i) => {
+    result[key] = result[`${key}Only`] = i === rowIndex
+    result[`${key}AndDown`] = i >= rowIndex
+    result[`${key}AndUp`] = i <= rowIndex
+  })
+  return result as Static
```
KaelWD: No idea what you've done with the types here, `result` should be defined as `Record<string, Static | boolean>` but then you're asserting it to `Static` instead.

src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -1,90 +1,62 @@
...
+const keys = Object.keys(breakpoints) as (keyof typeof breakpoints)[]
+const values = Object.values(breakpoints)
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+type Static = Omit<VuetifyBreakpoint, 'width' | 'height'>
+const breakpointExplainations: Static[] = keys.map((name, rowIndex) => {
+  const result: Record<
+  string,
+  keyof Omit<VuetifyBreakpoint, 'width' | 'height'> | boolean
+  > = { name }
+  keys.forEach((key: keyof typeof breakpoints, i) => {
```
KaelWD: `key` should be inferred if `keys` has the correct type.

src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -1,90 +1,62 @@
...
-
-        // For custom breakpoint logic.
-        width: this.clientWidth,
-        height: this.clientHeight
-      }
+  data () {
+    const width = getClientWidth()
+    const height = getClientHeight()
+    const cl = classifyBreakpoint(width)
+    return Object.assign(
```
KaelWD: Use spread

src/components/Vuetify/mixins/breakpoint.ts
```diff
@@ -121,10 +101,7 @@ export default Vue.extend({
 // https://stackoverflow.com/questions/1248081
 function getClientWidth () {
   if (typeof document === 'undefined') return 0 // SSR
-  return Math.max(
-    document.documentElement.clientWidth,
-    window.innerWidth || 0
-  )
+  return Math.max(document.documentElement.clientWidth, window.innerWidth || 0)
```
KaelWD: revert this

qm3ster: Donezo

src/components/Vuetify/mixins/breakpoint.ts
qm3ster: I cannot. I need to assign to `this`

src/components/Vuetify/mixins/breakpoint.ts
qm3ster: Donezo. At some point it didn't used to get inferred, it does now.

src/components/Vuetify/mixins/breakpoint.ts
qm3ster: Nuh-uh, that's the `return` of the function being mapped.
It returns a single row of the lookup table, containing all the static results (hence the name) for the viewport range. This includes one string, `name`, and the rest booleans.
A dense integer-indexed array of these, `Static[]` is the product of the whole mapping and the lookup table we want.
Finally, the reason I had to assert is this object is created in a loop, furthermore, through string concatenation of keys. Hence, it doesn't know of any specific key except `name`, and manages to lose even that.

To be completely honest, it actually produces more than a `Static`, by including the nonsense of `xsAndDown` and `xlAndUp` and always-`true` of `xlAndDown` and `xsAndUp`, but I am not sure whether putting a condition to not create them is necessary. Please advise.

Would be this:
```ts
const max = keys.length - 1
keys.forEach((key, i) => {
  result[key] = result[`${key}Only`] = i === rowIndex
  if (i === 0 || i === max) return
  result[`${key}AndDown`] = i >= rowIndex
  result[`${key}AndUp`] = i <= rowIndex
})
```

src/components/Vuetify/mixins/breakpoint.ts
qm3ster: Why don't you just override the definition from `lib.es5.d`?
Just doing an
```ts
// index.d.ts
interface ObjectConstructor {
  keys<O extends object>(o: O): (keyof O)[];
  keys<O>(o: O): [];
}
```
works for me I think?

src/components/Vuetify/mixins/breakpoint.ts
qm3ster: I think I should unify getClientWidth and getClientHeight to return a tuple, they are always run together. But I am not sure if destructuring the tuple is actually worse for performance than two function calls :man_shrugging:
Maybe it should return an object so that the names `height` and `width` could be put into those two contexts by destructuring? :woman_shrugging:

qm3ster: I could return an object of
```ts
clientHeight: height,
clientWidth: width,
width,
height,
cl,
...breakpointExplainations[cl]
```
but then I'd have to merge that in with 
```ts
{resizeTimeout: undefined as number | undefined}
```
and in the second case, simply assign it to
```ts
Object.assign(this, extractedFunction())
```
So there would be two destructurings/calls to Object.assign per method (`data` and `setDimensions`) Since both of those already have a call like that, I used those.

The only reason a second Object.assign is used in the `cl` watcher is because many resizes might not end up changing the breakpoint, meaning we can skip assigning all of those numerous keys (and Vue individually comparing them all)

src/components/Vuetify/mixins/breakpoint.ts
johnleider: No you don't,

```
{
      clientHeight: height,
      clientWidth: width,
      width,
      height,
      cl,
      resizeTimeout: undefined as number | undefined,
      ...breakpointExplainations[cl]
    }
```

src/components/Vuetify/index.ts
```diff
@@ -37,6 +34,7 @@ const Vuetify: VuetifyPlugin = {
         t: lang.t.bind(lang)
       }
     })
+    Vue.prototype.$vuetify.breakpoint=new Vue(breakpoint)
```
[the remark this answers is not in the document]
johnleider: Currently everything lives within the above declared statement. You are enabling breakpoints by explicitly adding it as opposed to simply being a data value. I understand your reasoning for it, to not have the watchers applied from including it as an explicit value, but I don't know if it is worth it.

qm3ster:
@johnleider would you like me to do
```ts
Vue.prototype.$vuetify = new Vue({
  data: {
    dark: false,
    icons: icons(opts.iconfont, opts.icons),
    lang,
    rtl: opts.rtl,
    theme: theme(opts.theme)
  }
})

Object.assign(Vue.prototype.$vuetify, {
  application:new Vue({data:application}),
  breakpoint: new Vue(breakpoint),
  options: options(opts.options),
  goTo,
  t: lang.t.bind(lang)
})
```
or can we merge this as-is?

qm3ster:
@johnleider should I rebase this or do you hate it and don't want it at all?

johnleider:
This is something we can look at more closely after 1.2 releases tomorrow.

johnleider:
I'm going to close this for now as we will be moving to a different approach for v2.0 in how we manage these services. We can discuss this further in the community.

PR #4774, opened 2018-08-01
feat(theme): css vars

### LINKED ISSUE #827: Feature request: Native css variables
As far as I understand, at the given moment, all the stylys settings ([colors](https://github.com/vuetifyjs/vuetify/blob/master/src/stylus/settings/_colors.styl), [theme](https://github.com/vuetifyjs/vuetify/blob/master/src/stylus/settings/_theme.styl), etc.) are compiled into css as values.

This causes some inconvenience. For example, to change the primaty color of a theme, I have to integrate the sources into my project, redefine the `$theme` variable and compile the result.

How much easier it would be if the colors of the theme were set using the **native CSS variable**. In this case, redefining the color palette becomes much easier and more flexible. And the original code of the vuetifyjs is reduced.
```css
:root {
  --primaryColor: #f06d06;
}
```

I understand that the use of native variables may not give all the possibilities as a stylus, but I am convinced that this is something to which one should strive.

I think you should take the parameters of the theme into native variables. And in the future expand other [settings](https://github.com/vuetifyjs/vuetify/tree/master/src/stylus/settings).

jacekkarczmarczyk:
## Description
turns on css variables for theme

## Motivation and Context
resolves #827

## How Has This Been Tested?
playground, unit

## Markup:

<details>

```vue
<template>
  <v-app id="inspire">
    <v-content>
      <v-container>
        <v-toolbar color="primary"></v-toolbar>
        <v-slider min="0" max="255" label="Red" color="red" v-model="red" />
        <v-slider min="0" max="255" label="Green" color="green" v-model="green" />
        <v-slider min="0" max="255" label="Blue" color="blue" v-model="blue" />
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
export default {
  data: () => ({
    red: 0,
    green: 0,
    blue: 0,
  }),

  watch: {
    red: 'update',
    green: 'update',
    blue: 'update',
  },

  methods: {
    update () {
      document.documentElement.style.setProperty('--v-primary-base', `rgb(${this.red}, ${this.green}, ${this.blue})`)
    }
  },

  mounted () {
    this.$vuetify.options.useCssVariables = true
    this.update()
  }
}
</script>
```
</details>

## Types of changes

## Checklist:


src/util/theme.ts
```diff
@@ -81,31 +81,45 @@ const genVariantColor = (name: string, variant: string, value: string): string =
...
+  let variablesCss = ''
+  let css = ''
+
+  const aColor = cssVar ? genColorVariable('primary') : theme.primary.base
+  css += `a { color: ${aColor}; }`
 
   for (let i = 0; i < colors.length; ++i) {
     const name = colors[i]
     const value = theme[name]
     if (typeof value === 'object') {
```
Phlow2001: invert and continue to reduce nesting

src/util/theme.ts
```diff
@@ -81,31 +81,45 @@ const genVariantColor = (name: string, variant: string, value: string): string =
...
-      css += genBaseColor(name, value.base)
+      css += genBaseColor(name, cssVar ? genColorVariable(name) : value.base)
+      cssVar && (variablesCss += `  ${genColorVariableName(name)}: ${value.base};\n`)
 
       const variants = Object.keys(value)
       for (let i = 0; i < variants.length; ++i) {
         const variant = variants[i]
         const variantValue = value[variant]
         if (variant !== 'base') {
```
Phlow2001: invert and continue to reduce nesting

yordis:
I would drop support to IE at this point. IE will no die in the near future and we should move on with it, but that is just my opinion.

Xowap:
Thanks I need that feature :+1: 

Two small suggestions:

### Also include text color

I'd like to also propose to include this:

```
diff --git a/src/stylus/components/_app.styl b/src/stylus/components/_app.styl
index 8633c37..59fb23a 100644
--- a/src/stylus/components/_app.styl
+++ b/src/stylus/components/_app.styl
@@ -20,6 +20,9 @@
 application($material)
   background: $material.background
   color: $material.text.primary
+  --v-text-color-primary: $material.text.primary
+  --v-text-color-secondary: $material.text.secondary
+  --v-text-color-disabled: $material.text.disabled
 
   a
     cursor: pointer
```

This would allow to get the text colors available from custom properties too. Here's a demo use case below:

```html
<template>
  <v-app>
    <div>
      This is primary text
      <a href="https://vuejs.org/" class="secondary-text">
        yet this is secondary
      </a>
      (and a link (well, it's a demo))
    </div>
  </v-app>
</template>

<style>
  a.secondary-text,
  a.secondary-text:visited {
    color: var(--v-text-color-secondary);
  }

  a.secondary-text:hover {
    color: var(--v-text-color-primary);
  }
</style>
```

### Why `:root`

The custom properties seem to be applied to `:root`, yet wouldn't it make more sense if they were applied to the application directly?

KaelWD:
`$material.text` is already available if you use stylus, so you can already do that without this PR. `:root` is because v-app will be removed at some point, or at least made optional.

Xowap:
Thanks for the clarifications!

If I don't use Stylus, it's a bit harder to extract the value(s) properly. This is an opportunity to export Stylus variables to every CSS stack, independently of your build system. I find the possibility rather exciting!

PR #4863, opened 2018-08-14
Add appear prop to VSnackbar to enable transition on load

hackel:
## Description
Adds an `appear` prop which is passed along to the transition component, allowing the snackbar transition to run on page load.  Ref:
https://vuejs.org/v2/guide/transitions.html#Transitions-on-Initial-Render

## Motivation and Context
I wanted the snackbar slide animation for a snackbar that is displayed on page load.

## How Has This Been Tested?
I couldn't think of a good way to test this functionality in a unit test.  Perhaps it could check that the appropriate CSS classes are set?  Wasn't sure that was necessary in this case, since it's just passing the prop to Vue.  It is working fine in my dev environment.

## Markup:

<details>

```vue
<v-snackbar appear/>
```

</details>

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VSnackbar/VSnackbar.ts
+++ b/src/components/VSnackbar/VSnackbar.ts
@@ -16,6 +16,10 @@ export default mixins(
   name: 'v-snackbar',
 
   props: {
+    appear: {
+      type: Boolean,
+      default: false,
+    },
     autoHeight: Boolean,
     multiLine: Boolean,
     // TODO: change this to closeDelay to match other API in delayable.js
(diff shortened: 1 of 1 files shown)


KaelWD:
What's wrong with this? https://codepen.io/anon/pen/OwKYVX?editors=1010

hackel:
@KaelWD Aha, I didn't think of that!  I always just set the value directly.  It is a good workaround, but it does require using two separate variables (if using a prop), which seems a bit messier to me.  Definitely an edge-case, though, so I understand if it doesn't seem useful enough to include.

johnleider:

PR #4895, opened 2018-08-17
feat(VEditDialog): add open on hover option

### LINKED ISSUE #4896: [Feature Request] VEditDialog: Add hover option
A table row might serve as a route, therefore, a click on the row triggers the router. The `VEditDialog` is activated by a click. To still provide an edit dialog, it should be possible to show the dialog while hovering.

I made a [merge request](https://github.com/vuetifyjs/vuetify/pull/4895).

ndabAP:
## Description
Adds an option to show the `VEditDialog` on hover. This makes it possible to activate the dialog through hovering. 

## Motivation and Context
Useful when a table row servers as a route.

Resolves #4896

## How Has This Been Tested?
None.

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-content>
      <v-container fluid fill-height>
        <v-layout align-center justify-center>
          <v-flex xs8 text-xs-center>
            <v-edit-dialog open-on-hover>
              Test
              <v-text-field
                slot="input"
                label="Edit"
                single-line
                counter
              ></v-text-field>
            </v-edit-dialog>
          </v-flex>
        </v-layout>
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
    })
  }
</script>

```
</details>

## Types of changes

## Checklist:

- [x] I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes.

### DIFF
--- a/src/components/VDataTable/VEditDialog.js
+++ b/src/components/VDataTable/VEditDialog.js
@@ -29,6 +29,10 @@ export default {
     transition: {
       type: String,
       default: 'slide-x-reverse-transition'
+    },
+    openOnHover: {
+      type: Boolean,
+      default: false
     }
   },
 
@@ -108,7 +112,8 @@ export default {
         value: this.isActive,
         closeOnClick: !this.persistent,
         closeOnContentClick: false,
-        lazy: this.lazy
+        lazy: this.lazy,
+        openOnHover: this.openOnHover
       },
       on: {
         input: val => (this.isActive = val)


jacekkarczmarczyk:
What about doing it in a similar way as https://github.com/vuetifyjs/vuetify/pull/4795?

ndabAP:
@jacekkarczmarczyk I don't see the exact relation but `VEditDialog` serves perfectly as an inline edit modal. I was able to imitate the behavior with `VMenu` but only with help of a class (which may change) and a lot more code.

jacekkarczmarczyk:
I meant to deprecate all menu related props and introduce ` :menu-props` prop where you could set `openOnHover` to what you want and also another menu props. This would be a bigger change than yours but more flexible and in line with menitoned PR
Thoughts @vuetifyjs/dev-team ?

ndabAP:
@jacekkarczmarczyk I'm sorry, I still don't get it. What will happen with `VEditDialog` after this deprecation? Should we build this by ourselves then?

jacekkarczmarczyk:
No, see `v-select`:
- before: `<v-select auto content-class="foo" ....>`
- now: `<v-select :menu-props="{ auto: true, contentClass: 'foo', someOtherMenuProp: 'bar' }">`

VEditDialog: 
- before: `<v-edit-dialog transition="foo">`
- after: `<v-edit-dialog :menu-props="{ transition: 'foo', openOnHover: true, ... }">`

jacekkarczmarczyk:
#4795 is only for select/autocomplete/overflow-btn and doesn't affect v-edit-dialog. I just suggested that instead of adding proxy to `openOnHover` prop to edit dialog you could just here `:menu-props` prop that would allow not only setting `openOnHover` but also other menu related props

KaelWD:
v-edit-dialog is basically just a menu with some pre-defined content, completely unlike v-select. It should be refactored to extend VMenu, `menu-props` would probably be overkill.

johnleider:

ndabAP:
@johnleider you are welcome. Thanks a lot for this great product.

What's your proposed approach of dealing with this? I got a table where every cell is a route. Cells should be editable as well.

johnleider:
You can use `v-hover`

PR #4930, opened 2018-08-21
fix(VSlider) add more aria attributes to VSlider

### LINKED ISSUE #4806: [Feature Request] V-slider a11y:  readonly is always set to true and aria-readonly is present which isn't allowed by specs
### Problem to solve
https://codepen.io/lol2x/pen/oMMmdx
Only the second slider have readonly prop set to true, however the HTML generated is always with readonly="readonly". 
What is more aria-readonly attribute should only be used for elements that are not allowed to have a readonly attribute in HTML5 - according to spec https://www.w3.org/TR/html-aria/

### Proposed solution
The presence of HTML readonly prop should be based on readonly prop from Vuetify
`readonly: String(this.readonly)`
and https://github.com/vuetifyjs/vuetify/blob/dev/src/components/VSlider/VSlider.js#L253 should be removed
I can create PR if you want this changes.

lol2x:
## Description

Aria-label was passed to root element without ...this.$attrs which causes errors in Google Insights
Also adds aria attributes -min -max -now

## Motivation and Context

fixes #4806

## How Has This Been Tested?

unit, playground

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-content>
      <v-container>
        <v-slider aria-label="just aria label"></v-slider>
        <v-slider label="this is label with aria label" aria-label="foo"></v-slider>
        <v-slider label="this is just label"></v-slider>
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
    })
  }
</script>
```
</details>

## Types of changes

## Checklist:

- [ ] I have created a PR in the documentation with the necessary changes.


src/components/VSlider/VSlider.js
```diff
@@ -31,6 +31,7 @@ export default {
   props: {
     alwaysDirty: Boolean,
     inverseLabel: Boolean,
+    hideLabel: Boolean,
```
KaelWD: This seems unnecessary, if you just add `...this.$attrs` to the bottom of the attrs object then the user can apply `aria-label` themselves.

KaelWD: - `label="something"` -> `<label>something</label>` + `aria-label="something"`
- `aria-label="something"` -> `aria-label="something"` + no `<label>`
- `label="something" aria-label="something else` -> `<label>something</label>` + `aria-label="something else"`

lol2x: will do

lol2x:
fixed according to your suggestions, now this PR doesn't require change to docs, also fixed markup

Issue #4784, opened 2018-08-02
[Bug Report] V-tab achor hash doesn't update url


amritk:
### Versions and Environment
Vuetify: 1.1.9
Vue: 2.5.16
Browsers: Chrome 68.0.3440.75
OS: Ubuntu 18.04

### Steps to reproduce
Use an href="#something" with the v-tab

### Expected Behavior
It changes the tab as well as updates the url with the new hash

### Actual Behavior
It changes the tab but does not update the url, nor does it load the correct tab if you manually type in the hash in the url.

### Reproduction Link
https://codepen.io/amritk/pen/djeqBX

---
### Additional Comments:
Can't really see the url changing in the codepen, not sure how to showcase that, but if you copy this code locally you should see that it doesn't change the url.

amritk:
I tracked it down to this line in VTab.js line 75
```
if (this.href && this.href.indexOf('#') > -1) e.preventDefault();
```
If I comment it out, it works perfectly. Why is it preventing default on a hash fragment url?

amritk:
If it was to allow people to use a '#' by itself, can we not instead do 
```
if (this.href && this.href === '#') e.preventDefault();
```
So this would work correctly:
```
<tab href="#teams">
```

johnleider:
Closed via #5124

johnleider:
I'm reverting this because it causes a bad user experience when using hrefs. After releasing we realized what its true purpose was.

amritk:
Maybe we should update the docs to tell users to use router-link instead of an anchor for hash fragment tab routing?

PR #5124, opened 2018-09-19
fix(v-tab): anchor hash routing

amritk:
## Description

Just a minor fix for the href attribute on v-tab when using hash urls. Originally it prevents default, but I believe that the intention was to only prevent default when the url is exactly '#'. In the case of using hash urls for the tab, this will not work. 

## Motivation and Context

I had this issue open for a while to see why this code was here in the first place, after no feedback decided to fix it as I am going live with a feature that is using this soon.
https://github.com/vuetifyjs/vuetify/issues/4784

## How Has This Been Tested?

I have tested that it works visually in my project, not quite sure how to write a test for this one.

## Markup:

<details>

```vue
<template lang="pug">
.main
    v-tabs.shadow(v-model="activeTab" slider-color="primary" fixed-tabs)
        v-tab(v-for="tab in tabs" :key="tab.name" :href="`#${tab.name}`")
            | {{ $t(tab.title) }}
        v-tab-item(id="teams")
            | These are the teams
        v-tab-item(id="players")
            | These are the players
        v-tab-item(id="matches")
            | These are the matches
</template>

<script lang="ts">
import { Component, Vue } from 'vue-property-decorator'
import * as VTabs from 'vuetify/es5/components/VTabs'

@Component({
    components: {
        ...VTabs
    }
})
export default class TeamFinderList extends Vue {
    activeTab: string = null
    tabs = [
        {
            name: 'teams',
            title: 'Teams'
        },
        {
            name: 'players',
            title: 'Players'
        },
        {
            name: 'matches',
            title: 'Matches'
        }
    ]

    created () {
        // Load the correct tab in the beinning
        this.activeTab = this.$route.hash.substr(1)
    }
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/src/components/VTabs/VTab.js
+++ b/src/components/VTabs/VTab.js
@@ -88,7 +88,7 @@ export default {
       // actual link, do not
       // prevent default
       if (this.href &&
-        this.href.indexOf('#') > -1
+        this.href === '#'
       ) e.preventDefault()
 
       this.$emit('click', e)


amritk:
sounds good

egeersoz:
Is this still slated for 1.3.0? A critical part of my app requires switching tabs using the URL.

KaelWD:
This doesn't fix anything: you still can't link directly to a tab by its hash, and clicking a tab scrolls the page to the top of the tab item (https://vuetifyjs.com/en/components/alerts#props)

amritk:
@kaelwd I had a working example. Let me try out the fix tomorrow and I'll see if it still works.

amritk:
@KaelWD This fix was to set the hash fragment in the url, which it correctly does now. To set it initially, on mounted you can do something like this:
```
this.activeTab = this.$route.hash.substr(1)
```
Not sure if we want to include this as part of the tabs component.

amritk:
@KaelWD actually that is probably why this code was in here in the first place, to prevent the scrolling in the docs. Maybe we should keep the routing but prevent the scroll.

amritk:
I just realized if you use router-link instead of an achor it works much better, including linking directly to the tab
```
<template lang="pug">
    v-tabs(v-model="activeTab" slider-color="primary" fixed-tabs)
        v-tab(v-for="tab in tabs" :key="tab" :to="`#${tab}`")
            | {{ tab }}
        v-tab-item(:id="tab" v-for="tab in tabs" :key="tab")
            | This is the {{ tab }} feed
</template>

<script lang="ts">
import { Component, Vue } from 'vue-property-decorator'
import * as VTabs from 'vuetify/es5/components/VTabs'

@Component({
    components: {
        ...VTabs
    }
})
export default class Amrit extends Vue {
    activeTab: string = null
    tabs = ['teams', 'players', 'matches']
}
</script>
```

KaelWD:
Correct. You could've also set the hash yourself with a watcher on `activeTab`, seeing as you're already doing `this.activeTab = this.$route.hash.substr(1)` in userland.

amritk:
Yes that is what I was doing while I was waiting for the pull request. But the tabs component should be handling that.

PR #5241, opened 2018-10-08
fix(v-data-table): support nested values for item key

Ignigena:
## Description
Currently, the data table supported nested values with dot syntax for the `headers.value` and `pagination.sortBy` configuration options among others. When implementing the expandable table, the `itemKey` property is required. However, this property doesn't supported nested values and when trying to specify one, an error is thrown: `[Vue warn]: <transition-group> children must be keyed: <div>`

## Motivation and Context

This change just brings parity to the other properties which supported accessing nested values with dot syntax. It uses the same `getObjectValueByPath` utility function which is used on the other properties.

Since this is not documented, and the error message is not very descriptive, it feels like a bug. The only workaround currently is to bind to a list with a key that is not nested in an object. But with complex data this can become difficult or result in extra computed properties and mapping of the data which could be solved with this change.

## How Has This Been Tested?

Added a new test using the existing `dataTableNestedTestData` fixture. This compares the `key` value used in the resulting table with the `itemKey` path property.

## Markup:

<details>

```vue
<template>
  <v-data-table
    :headers="headers"
    :items="items"
    hide-actions
    item-key="nested.value.id"
  >
    <template slot="items" slot-scope="props">
      <tr @click="props.expanded = !props.expanded">
        <td>{{ props.item.nested.value.id }}</td>
        <td>{{ props.item.nested.value.name }}</td>
      </tr>
    </template>
    <template slot="expand" slot-scope="props">
      <v-card flat>
        <v-card-text>Peek-a-boo!</v-card-text>
      </v-card>
    </template>
  </v-data-table>
</template>

<script>
  export default {
    data () {
      return {
        headers: [
          { text: 'ID', value: 'nested.value.id' },
          { text: 'Name', value: 'nested.value.name' },
        ],
        items: [
          { nested: { value: { id: 0, name: 'foo' } } },
          { nested: { value: { id: 1, name: 'bar' } } },
          { nested: { value: { id: 2, name: 'baz' } } }
        ]
      }
    }
  }
</script>

```
</details>

## Types of changes

## Checklist:


Ignigena:
@nekosaur Rebased against latest changes in `master`

Issue #5477, opened 2018-10-29
[Feature Request] [VDataTable] Make actions-prepend items align to the left by default


mrodal:
### Problem to solve
off looking default position for items on actions-prepend slot:
https://codepen.io/mrodal/pen/mzgQoQ
![image](https://user-images.githubusercontent.com/20667544/47673871-9b2f2100-db94-11e8-96f8-41d2fac48633.png)

### Proposed solution
Add flex:1 to the div enclosing the items to make them align to the left by default
![image](https://user-images.githubusercontent.com/20667544/47673882-a41ff280-db94-11e8-92a3-d9df9f2d6146.png)

xuanduyrn:
@johnleider I updated the project's vuetify to> 2.0 @latest version, and my v-datatable previously used slot = "actions-prepend". Now it's gone. How do I know what this slot has changed to?

mrodal:
looks like this slot is gone. Could you confirm @KaelWD @MajesticPotatoe? 
Then this issue no longer applies. Also I couldnt find a reference to this slot being gone on the migration guide

PR #5478, opened 2018-10-29
Align actions-prepend slot items to the left on v-data-table 

mrodal:
## Description

Added flex:1 to prepend-actions div on v-data-table and data-iterator

## Motivation and Context

Fixes https://github.com/vuetifyjs/vuetify/issues/5477

## How Has This Been Tested?

visually

## Types of changes

## Checklist:

- [ ] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).

### DIFF
--- a/packages/vuetify/src/mixins/data-iterable.js
+++ b/packages/vuetify/src/mixins/data-iterable.js
@@ -474,7 +474,9 @@ export default {
       return [this.$createElement('div', {
         'class': this.actionsClasses
       }, [
-        this.$slots['actions-prepend'] ? this.$createElement('div', {}, this.$slots['actions-prepend']) : null,
+        this.$slots['actions-prepend'] ? this.$createElement('div', {
+          'class': 'flex'
+        }, this.$slots['actions-prepend']) : null,
         this.rowsPerPageItems.length > 1 ? this.genSelect() : null,
         rangeControls,
         this.$slots['actions-append'] ? this.$createElement('div', {}, this.$slots['actions-append']) : null


packages/vuetify/src/stylus/components/_data-iterator.styl
```diff
@@ -52,3 +52,6 @@ theme(v-dataiterator, "v-data-iterator")
 
       .v-select__selection--comma
         font-size: 12px
+
+  &__prepend-slot
+    flex: 1
```
johnleider: This has the potential to cause issues on IE11

nekosaur:
I'm fine with adding a class that could be used to target some custom css. I would probably leave what that css is up to whomever uses the slot. Another solution could also be to skip the wrapping div entirely and just render the slot.

mrodal:
Yeah, I thought about just adding a class and let the developer do whatever he wants with it, but I also thought that it was nice to display the items better that it is now, that doesnt look right.

Skipping the div maybe makes the aligning of the items harder?

nekosaur:
Skipping the div, you could do `<div slot="prepend-slot" class="d-flex">...</div>`

mrodal:
Tru dat, I like it

mrodal:
Ok, now I just add the flex class to the slot div. I didnt know about that class 😁

johnleider:
Waiting for @nekosaur 's final approval.

nekosaur:
If you want to keep the wrapping div, then please add back the `__prepend-slot` class so that it can be easily targeted with css selectors.

KaelWD:
We can't use .flex, it won't work unless the user has loaded it themselves. This should be its own table-specific class and css.

packages/vuetify/src/mixins/data-iterable.js
```diff
@@ -474,7 +474,9 @@ export default {
       return [this.$createElement('div', {
         'class': this.actionsClasses
       }, [
-        this.$slots['actions-prepend'] ? this.$createElement('div', {}, this.$slots['actions-prepend']) : null,
+        this.$slots['actions-prepend'] ? this.$createElement('div', {
+          'class': 'flex'
```
KaelWD: We can't use .flex, it won't work unless the user has loaded it themselves. This should be its own table-specific class and css.

johnleider:
Closing due to inactivity.

PR #4726, opened 2018-07-26
fix(v-checkbox, v-radio): fix adding 'checked' attribute to input element

ochikoji:
## Description

I fixed v-checkbox and v-radio components to apply for non-javascript form page by adding 'checked' attribute on them when they are actived.

## Motivation and Context

I'm using this framework with Laravel to make website to be rich appearance.
When I'm creating form page on laravel-blade-template, I realized v-checkbox and v-radio are not applied for non-javascript html form page.
So I fixed them working.

Sorry for my poor english. and I'm a biginner.
## How Has This Been Tested?

Sorry, I can't understand how to test yet..
But, I checked them on my own web site.

## Markup:

<details>

```vue
// Paste your FULL Playground.vue here
```
</details>

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have created a PR in the [documentation](https://github.com/vuetifyjs/vuetifyjs.com) with the necessary changes.

PR #5505, opened 2018-11-01
Fix for non-spa checkbox, radio and switch values

ivanwitzke:
## Description

Updated the v-switch, v-checkbox and v-radio to set the 'checked' attribute on the hidden input they create with the values.
Also, updated the 'Selectable' mixin to allow the value of the input to be fixed to this value if set on the component. (But I'm not sure if this could mean breaking changes or not... It still respects the `input-value` property if `value` is not set)

## Motivation and Context

When using vuetify with non-spa pages, these inputs were not being sent correctly on form submit

It is a follow up on this PR #4726

## How Has This Been Tested?

Tested using the dev/Playground.vue

I have not created any other test.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-layout>

        <v-checkbox
          v-model="checkboxModel"
          :true-value="1"
          :false-value="0"
          :value="1"
          label="Checkbox"
          name="is_checked"
        />

        <v-radio-group
          v-model="radioValue"
          name="radio_id"
        >
          <v-radio
            v-for="n in 3"
            :key="n"
            :label="`Radio ${n}`"
            :value="n"
          />
        </v-radio-group>

        <v-switch
          v-model="switchValue"
          true-value="on"
          false-value="off"
          value="on"
          label="Switch"
          name="switch_on"
        />

        <v-switch
          v-model="switchValue"
          true-value="off"
          false-value="on"
          value="off"
          label="Switch Off"
          name="switch_off"
        />
      </v-layout>
      <pre>checkboxModel: {{ checkboxModel }}</pre>
      <pre>radioValue: {{ radioValue }}</pre>
      <pre>switchValue: {{ switchValue }}</pre>
    </v-container>

  </v-app>
</template>

<script>
export default {
  data: () => ({
    switchValue: 'on',
    checkboxModel: 1,
    radioValue: 2
  })
}
</script>

```
</details>

## Types of changes

## Checklist:


KaelWD:
Can you explain what the issue is here? That playground seems fine to me already on master, and if you wrap the whole lot in a `<form action="/" method="GET">` you'll see the parameters are updated correctly as `checked` is already being applied in selectable:https://github.com/vuetifyjs/vuetify/blob/02b9e7972104129bbc3f26ddea4e879f43d8ea7a/packages/vuetify/src/mixins/selectable.js#L98-L100

That ternary you added in the mixin will always be true - it's only used for these three components, and VSwitch uses a checkbox.

ivanwitzke:
The problem I'm having and tried to solve with this PR is that when using these components on non-spa page, the `value` that is sent on the request is always the value of the `input-value`, which is also the value that is compared on the `selectable` mixin.
So, for example, if the current `input-value = 0` (false), the value that will be sent will be 0. Laravel will understand this as false (even though the checkbox/switch was checked).
Also, the 'checked' property currently is not added to the input. Looks like `domProps` is not working properly.

ivanwitzke:
Rebased to master...

KaelWD:
Thanks for clarifying, I'll take another look at it.

johnleider:
Could this be simplified by adding that to https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/selectable.js#L87 ?

ivanwitzke:
Correct.. the `checked` attr is there, however, the behavior must be broken, because it never gets added to the input when the checkbox is activated (or radio or switch) as you can see in the image below.

image

ivanwitzke:
Well, after some more testing, I now understand that even though the `checked` attr is not added to the input, it will be 'virtually' added, so the behavior is the same, making the changes to the components irrelevant.
However, the `value` problem still exists, and [this](https://github.com/ivanwitzke/vuetify/blob/fix_non-spa_checkbox_values/packages/vuetify/src/mixins/selectable.js#L96) changes solves my problem

packages/vuetify/src/mixins/selectable.js
```diff
@@ -96,6 +96,7 @@ export default {
           value: this.inputValue
```
KaelWD: This should be removed now. It is incorrect and being overridden by the domProp

PR #5593, opened 2018-11-12
feat(v-dialog): adds persistent callback

### LINKED ISSUE #5533: [Bug Report] V-Dialog closes before it can be caught
### Versions and Environment
**Vuetify:** 1.2.8
**Vue:** 2.5.16
**OS:** Mac OS 10.13.6

### Steps to reproduce
* Click the Open button.
* Click outside the dialog (this will open the "Are you sure dialog")

### Expected Behavior
The original dialog should remain open until the user says they are sure they want to close the original dialog

### Actual Behavior
The original dialog closes as soon as the user clicks the overlay

### Reproduction Link
<a href="https://codepen.io/jonperry/pen/wQKBQq" target="_blank">https://codepen.io/jonperry/pen/wQKBQq</a>

### Other comments
I believe this happens because the Toggleable mixin used by Vuetify provides a variable called isActive which is updated when the value prop is updated and emits 'input' when isActive is updated. This leaves the two variables with inconsistent values when isActive is updated by the v-dialog component and that 'input' event is caught by the calling component as in the codepen example (isActive = false, value = true).

It would seem that isActive should be a computed as follows:
```javascript
isActive: {
    get () {
        return this.value;
    },
    set (val) {
        this.$emit('input', val);
    }
}
```

rkomiyama:
## Description

Adds a prop called `persistentCb` to `v-dialog` which allows callbacks to happen when a user clicks outside of a persistent `v-dialog`.

## Motivation and Context

Fixes #5533 

## How Has This Been Tested?

None

## Markup:

<details>

```vue
<template>
  <div id="app">
    <v-app>
      <v-content>
        <v-container grid-list-xl>
          <v-layout row justify-center>
            <v-dialog
              v-model="open"
              persistent
              :persistent-cb="callback"
              width="300"
            >
              <v-btn slot="activator"> Open </v-btn>
              <v-card>
                My Dialog Content
              </v-card>
            </v-dialog>
            <v-dialog
              v-model="areYouSureDialog"
              persistent
              width="300"
            >
              <v-card>
                <v-flex>
                  Are you sure?
                </v-flex>
                <v-flex>
                  <v-btn @click="closeBoth"> Yes </v-btn>
                  <v-btn @click="areYouSureDialog = false"> No </v-btn>
                </v-flex>
              </v-card>
            </v-dialog>
          </v-layout>
        </v-container>
      </v-content>
    </v-app>
  </div>
</template>

<script>
export default {
  name: 'app',
  components: {
  },
  methods: {
    callback() {
      this.areYouSureDialog = true;
    },
    closeBoth() {
      this.open = false;
      this.areYouSureDialog = false;
    }
  },
  data: () => ({
    open: false,
    areYouSureDialog: false
  })
}
</script>

<style lang="scss" scoped>
#app {
  font-family: 'Avenir', Helvetica, Arial, sans-serif;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
  text-align: center;
  color: #2c3e50;
  margin-top: 60px;
}
</style>

```
</details>

## Types of changes

## Checklist:


jacekkarczmarczyk:
I can see 2 options here:
- it would work only for persistent dialog, in this case personally i'd prefer an event (`click:outside` for example) instead of callback prop
- it would work both for persistent and not persistent dialog, in this case callback would return the bool/undefined value, true would close the dialog, false would leave it open, undefined would perform default action (nothing for persistent, close for non-persistent)

@vuetifyjs/core-team ?

johnleider:
I'm an absolute fan of `click:outside`. I'm not sure I completely understand bullet 2 though.

jacekkarczmarczyk:
`click:outside` is perfect but only for persistent dialogs

Example of bullet 2:

`<v-dialog :cb="check"></v-dialog>`

```js
check () {
  if (this.formIsEdited()) return true;
  showAlert('Form is edited, please click "Save"');
  return false;
}
```

johnleider:
I would probably approach this differently. `formIsEdited` would be a computed property and bound to `:persistent`. I see what you are saying though. It could still work for non persistent dialogs, it would just be the same as adding a watcher to a bound model.

jonperryxlm:
I don't know of the complexities associated with this, but my 2 cents regardless...

While a click:outside event would work perfectly fine for my use case (as would the callback option), would it not be more appropriate for the dialog component to emit an input event of false when the overlay is clicked? 

Let's say the parent component has a variable "open" which it passes to v-dialog's value prop...

The dialog is already opened with its value prop, then the parent component can either bind a "check" function to the input event, before setting the "open" variable (same as the click:outside option), continue to use v-model blissfully unaware of the new functionality or (as I would) make the "open" variable passed to v-model prop a computed and put the check logic in its setter.

The reason I suggest this as an alternative is because currently the dialog component is determining whether it should be open or closed based off whether the user has clicked the overlay. This should be controlled by the component that opens the dialog. Plus it would work for non-persistent dialogs (and probably **shouldn't** work for persistent dialogs - as the name suggests).


Thanks for looking into this.

jacekkarczmarczyk:
> would it not be more appropriate for the dialog component to emit an input event of false when the overlay is clicked?

It does emit the `@input` event, because that's how the `v-model` works: https://codepen.io/anon/pen/yQMWjN?editors=1010

However it's doesn't work for non-persistent dialogs, because they don't close on clicking the overlay so the model doesn't change so the event is not being fired. And that's why we need something different than `@input`

Does it answer your question?

jonperryxlm:
Hi @jacekkarczmarczyk, thanks for the response.

My suggestion was overly simple. The bug that I reported was to do with v-dialog not correctly using v-model. It has a data attribute called isActive that controls the visibility of the dialog, clicking the overlay sets isActive to false in non-persistent dialogs and emits the input event.

When the parent component handles that event and decides not to update the boolean it passes to v-model, isActive is already set to false and v-dialog's value prop remains as true. This is because of the way that the Toggleable mixin works. So I would have to set the v-model Boolean to false (to get the 2 properties back in sync), then back to true to re-open the dialog.

Like I said, I'd be happy to use a persistent dialog and handle a separate click:outside event (providing the dialog doesn't close then re-open - as in the codepen you posted).

Thanks again.

rkomiyama:
## Revised Markup:

<details>

```vue
<template>
  <div id="app">
    <v-app>
      <v-content>
        <v-container grid-list-xl>
          <v-layout row justify-center>
            <v-dialog
              v-model="open"
              persistent
              @click:outside="callback"
              width="300"
            >
              <v-btn slot="activator"> Open </v-btn>
              <v-card>
                My Dialog Content
              </v-card>
            </v-dialog>
            <v-dialog
              v-model="areYouSureDialog"
              persistent
              width="300"
            >
              <v-card>
                <v-flex>
                  Are you sure?
                </v-flex>
                <v-flex>
                  <v-btn @click="closeBoth"> Yes </v-btn>
                  <v-btn @click="areYouSureDialog = false"> No </v-btn>
                </v-flex>
              </v-card>
            </v-dialog>
          </v-layout>
        </v-container>
      </v-content>
    </v-app>
  </div>
</template>

<script>
export default {
  name: 'app',
  components: {
  },
  methods: {
		callback() {
			this.areYouSureDialog = true;
		},
    closeBoth() {
      this.open = false;
      this.areYouSureDialog = false;
    }
  },
  data: () => ({
    open: false,
    areYouSureDialog: false
  })
}
</script>

<style lang="scss" scoped>
#app {
  font-family: 'Avenir', Helvetica, Arial, sans-serif;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
  text-align: center;
  color: #2c3e50;
  margin-top: 60px;
}
</style>
```
</details>

packages/vuetify/src/components/VDialog/VDialog.js
```diff
@@ -133,11 +133,11 @@ export default {
       // If we made it here, the click is outside
       // and is active. If persistent, and the
       // click is on the overlay, animate
+      this.$emit('click:outside')
       if (this.persistent) {
-        if (!this.noClickAnimation &&
-          this.overlay === e.target
-        ) this.animateClick()
-
+        if (!this.noClickAnimation && this.overlay === e.target) {
```
johnleider: revert these style changes

packages/vuetify/src/components/VDialog/VDialog.js
```diff
@@ -133,11 +133,11 @@ export default {
       // If we made it here, the click is outside
       // and is active. If persistent, and the
       // click is on the overlay, animate
+      this.$emit('click:outside')
```
johnleider:
```suggestion
      this.$emit('click:outside')

```

jonperryxlm:
Hi all,

I noticed this has been closed. Does this mean there is no chance of getting the "click:outside" event? I think this is still a fairly standard use case to want to prompt the user that they have unsaved changes before they leave a dialog with a form contained within it.

Is there an alternative solution that you would suggest?

Thanks,
Jon

rkomiyama:
Hey @jonperryxlm,

I've created a new PR (https://github.com/vuetifyjs/vuetify/pull/6719) which simply adds one line to emit a `click:outside` event.

PR #5646, opened 2018-11-17
fix(VAutocomplete): hide-selected may not show menu

### LINKED ISSUE #4580: [Bug Report] When "v-autocomplete" has one or more items selected, filtering breaks after x amount of characters are entered.
### Versions and Environment
Vuetify: 1.1.4
Vue: 2.5.16
Browsers: Chrome, Firefox, etc.
OS: macOS High Sierra 

### Expected Behavior

1. Try and type "California" (or any other state) in full.
2. The list should be filtered accordingly, and you should be able to enter as many characters as desired.

### Actual Behavior

1. Try and type "California" (or any other state) in full.
2. After the third letter of the state is entered, the list goes away, as if there aren't any results.

### Reproduction Link

https://codepen.io/anon/pen/WKbzNe?&editors=101

---

### Additional Comments:
It works as expected for the initial search when there are no items selected. However, once one is selected, the issue starts to appear.

The issue appears to be caused by the combination of the `hide-no-data` and `hide-selected` properties. If you remove either of them, it works as expected.

dima74:
## Description

Modify `displayedItemsCount` method in VAutocomplete to work correctly with `hide-selected` prop. Previous code (`this.filteredItems.length - this.selectedItems.length`) didn't work if there exists selected item which doesn't match search. Now explicit check is used (thus helpers methods are added — `hasItem` and `selectedValues`).

## Motivation and Context

Fix #4580

## How Has This Been Tested?

unit & visually

## Markup:

<details>

```vue
<template>
  <div>
    <div>Type <code>Cali</code> in VAutocomplete. Expect to see menu with one item (California)</div>
    <v-autocomplete
      v-model="model"
      :items="states"
      chips
      multiple
      hide-no-data
      hide-selected
    />
  </div>
</template>

<script>
  export default {
    data: () => ({
      model: ['Alabama'],
      states: [
        'Alabama', 'Alaska', 'American Samoa', 'Arizona',
        'Arkansas', 'California', 'Colorado', 'Connecticut',
        'Delaware', 'District of Columbia', 'Federated States of Micronesia',
        'Florida', 'Georgia', 'Guam', 'Hawaii', 'Idaho',
        'Illinois', 'Indiana', 'Iowa', 'Kansas', 'Kentucky',
        'Louisiana', 'Maine', 'Marshall Islands', 'Maryland',
        'Massachusetts', 'Michigan', 'Minnesota', 'Mississippi',
        'Missouri', 'Montana', 'Nebraska', 'Nevada',
        'New Hampshire', 'New Jersey', 'New Mexico', 'New York',
        'North Carolina', 'North Dakota', 'Northern Mariana Islands', 'Ohio',
        'Oklahoma', 'Oregon', 'Palau', 'Pennsylvania', 'Puerto Rico',
        'Rhode Island', 'South Carolina', 'South Dakota', 'Tennessee',
        'Texas', 'Utah', 'Vermont', 'Virgin Island', 'Virginia',
        'Washington', 'West Virginia', 'Wisconsin', 'Wyoming'
      ]
    })
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.js
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.js
@@ -44,7 +44,7 @@ export default {
     menuCanShow () {
       if (!this.isFocused) return false
 
-      return (this.displayedItemsCount > 0) ||
+      return this.hasDisplayedItems ||
         (!!this.$slots['no-data'] && !this.hideNoData)
     }
   },
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.js
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.js
@@ -69,10 +69,13 @@ export default {
     computedItems () {
       return this.filteredItems
     },
-    displayedItemsCount () {
+    selectedValues () {
+      return this.selectedItems.map(item => this.getValue(item))
+    },
+    hasDisplayedItems () {
       return this.hideSelected
-        ? this.filteredItems.length - this.selectedItems.length
-        : this.filteredItems.length
+        ? this.filteredItems.some(item => !this.hasItem(item))
+        : this.filteredItems.length > 0
     },
     /**
      * The range of the current input text
@@ -116,7 +119,7 @@ export default {
     menuCanShow () {
       if (!this.isFocused) return false
 
-      return (this.displayedItemsCount > 0) || !this.hideNoData
+      return this.hasDisplayedItems || !this.hideNoData
     },
     $_menuProps () {
       const props = VSelect.computed.$_menuProps.call(this)
@@ -362,6 +365,9 @@ export default {
       )) {
         this.setSearch()
       }
+    },
+    hasItem (item) {
+      return this.selectedValues.indexOf(this.getValue(item)) > -1
     }
   }
 }
--- a/packages/vuetify/test/unit/components/VAutocomplete/VAutocomplete.spec.js
+++ b/packages/vuetify/test/unit/components/VAutocomplete/VAutocomplete.spec.js
@@ -219,7 +219,7 @@ test('VAutocomplete.js', ({ mount, shallow, compileToFunctions }) => {
   it('should not display menu when tab focused', async () => {
     const wrapper = shallow(VAutocomplete, {
       propsData: {
-        items: [1 ,2],
+        items: [1, 2],
         value: 1
       }
     })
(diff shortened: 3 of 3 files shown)


dima74:
@johnleider
Sorry, I didn't understand the perfomance issue completely. Aren't there a few items in `selectedItems` array usually (like <10)? Because `selectedValues` computed property iterates only over `selectedItems` array, and iteration over `filteredItems` using `some` method will explore up to `selectedItems.length + 1` items from `filteredItems` array in worst case.

Also I want to point out that same operations have to be perfomed in `VSelectList`. Actually code for `selectedValues` was copied [from there](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VSelect/VSelectList.js#L70-L72).

johnleider:
It's more about this https://github.com/vuetifyjs/vuetify/pull/5646/files#diff-9009177e046b7e8ee6410bada06c0904R77

dima74:
In line

    this.filteredItems.some(item => !this.hasItem(item))

calback will be executed less than `selectedItems.length + 1` times, won't it?
(because among first `selectedItems.length + 1` items of `filteredItems` array there necessarily will be one, that will not belongs to selectedItems, so callback for this item will return false, and after first time when callback return false, method `some` will return false immediately). Maybe I miss something...

johnleider:
`filteredItems`'s length isn't determined by `selectedItems`. So if the user had 3k results it would trigger for all of them.

dima74:
It can't be 1000th record. It is guaranteed that match happens no further than on `selectedItems.length + 1`-th item. https://codepen.io/dima74/pen/zMPKJp?editors=0011

johnleider:
You are 100% right. For some reason when I first reviewed my eyes did not catch the `!` in the filter.

packages/vuetify/src/components/VCombobox/VCombobox.js
```diff
@@ -44,7 +44,7 @@ export default {
     menuCanShow () {
```
johnleider: Since changes are already being made here, we can (and probably should) move this logic to `v-autocomplete` and remove its re-declaration here.

dima74: Could you, please, explain it more. (Because behaviour of `v-autocomplete` and `v-combobox` supposed to be different when `hideNoData` is `false`, there are no items to display, and there is no `no-data` slot. In `v-autocomplete` menu with standard message appears (`No data available`), and in `v-combobox` no menu appears)

PR #5695, opened 2018-11-21
fix(VTimeline): remove arrows from nested cards

### LINKED ISSUE #5437: [Bug Report] Arrows on nested cards in the timeline component
### Versions and Environment
**Vuetify:** 1.3.3
**Vue:** 2.5.17
**Browsers:** Chrome 70.0.3538.77
**OS:** Windows 10

### Steps to reproduce
See the nested cards in the sample

### Expected Behavior
No arrows on the nested card

### Actual Behavior
Arrows on the nested card

### Reproduction Link
[https://codepen.io/anon/pen/Jmxdzr?&editors=101](https://codepen.io/anon/pen/Jmxdzr?&editors=101)

dima74:
## Description

Move key styles for card arrow from `.v-timeline-item` to `.v-timeline-item__body` with `>` selector, so now these styles are applied only to top-level cards (direct children of `.v-timeline-item__body`)

## Motivation and Context

Fixes #5437

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
<template>
  <v-timeline>
    <v-timeline-item
      color="red lighten-2"
      large
    >
      <span slot="opposite">Foo</span>
      <v-card class="elevation-2">
        <v-card-title class="headline">Title</v-card-title>
        <v-card-text>
          Card
          <v-card class="elevation-2">
            <v-card-text>
              Nested card
            </v-card-text>
          </v-card>
        </v-card-text>
      </v-card>
    </v-timeline-item>
  </v-timeline>
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/stylus/components/_timeline.styl
+++ b/packages/vuetify/src/stylus/components/_timeline.styl
@@ -22,7 +22,7 @@ timeline-dots($dot-size, $inner-dot-size, $inner-dot-margin)
 
   .v-timeline-item__inner-dot
     height: $inner-dot-size
-    margin:$inner-dot-margin
+    margin: $inner-dot-margin
     width: $inner-dot-size
 
 .v-timeline-item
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
The question is why is the arrow added to `.v-card` instead of `.v-timeline-item__body` (or `<v-card class="with-arrow">`)? Probably it would require an additional prop (either for `v-treeview` or `v-card`) in case you don't want to add the arrow for every item, but still i think it would better than using a context for changing the `v-card` style

dima74:
@jacekkarczmarczyk sorry for late response. Could you, please, explain why your proposed solution is better then using a context for changing styles? For me it looks like that using context is the simplest and most reliable solution, which works even with cards in opposite slot ([code](https://github.com/vuetifyjs/vuetify/blob/bb3ef969b0bbfda631219822c315d572bdaff6fe/packages/vuetify/src/stylus/components/_timeline.styl#L44-L46), [codepen](https://codepen.io/dima74/pen/xQBZQg), [appropriate pr](https://github.com/vuetifyjs/vuetify/pull/5400))

jacekkarczmarczyk:
1. When you're using context you need to make sure that when you're changing the main style of the component it won't affect all possible contexts. It happened before many times
2. Your solution works only for v-cards with elevation > 0, what if I want to use another component or have a flat card?
3. Adding a separate classes that are responsible for displaying arrows allows to use them not only in timeline but also in other contexts
4. Some pros/cons of different solutions to this problem are described [here](http://simurai.com/blog/2015/05/11/nesting-components)
5. This is just my personal feelings about this, I'm not saying that this is the only way, but I hope this few points explain my reasons

dima74:
Yes, thank you, now it is more clear for me. Closing this PR because I'm not sure I can properly implement another solution

PR #5702, opened 2018-11-22
docs: load fonts asynchronously

peluprvi:
## Description

Async load Roboto fonts on docs.

## Motivation and Context

It should improve some performance by not waiting for font load before display/paint some text.
On the other hand, texts will be displayed with secondary font (sans-serif) and icons will display its name (example: "MENU") until the font is loaded.
This setup can be done/chosen for each font (currently in use: Roboto, Roboto Mono, Material Icons) to balace the performance improvement and load/FOUC expectations.

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
// Paste your FULL Playground.vue here
```
</details>

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [ ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #5799, opened 2018-12-03
docs: load fonts asynchronously

peluprvi:
## Description

Async load Roboto fonts on docs.

Replaces https://github.com/vuetifyjs/vuetify/pull/5702

## Motivation and Context

It should improve some performance by not waiting for font load before display/paint some text.
On the other hand, texts will be displayed with secondary font (sans-serif) and icons will display its name (example: "MENU") until the font is loaded.
This setup can be done/chosen for each font (currently in use: Roboto, Roboto Mono, Material Icons) to balace the performance improvement and load/FOUC expectations.

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


MajesticPotatoe:
"copy" icon in all code examples are reverting to top-left instead of centered. Seems to be related to loading the 'Material Icons' family vui WebFontLoader in entry-client.js

packages/vuetifyjs.com/src/App.vue
```diff
@@ -45,3 +45,9 @@
     }
   }
 </script>
+
+<style>
+  .wf-loading .material-icons {
+    display: none;
```
peluprvi: @MajesticPotatoe maybe change this line to `opacity` can fix the example copy icon.
```suggestion
    opacity: 0;
```

KaelWD: It's probably vuetifyjs/vuetifyjs.com#353

`display: none` is needed so the icon text doesn't displace other elements.

MajesticPotatoe: Tried the opacity, no go. Seems that `display: inline-flex;` on `.v-icon` is getting overridden somewhere or is being ignored.
adding 
```suggestion
.v-icon {
    display: inline-flex;
}
```
into App.vue does seem to fix it though (but this seems hacky)

KaelWD: Yeah that's vuetifyjs/vuetifyjs.com#353, `.material-icons` adds `display: inline-block`

MajesticPotatoe: AH! gotcha, sorry, read it too fast. Seems just need to move `import 'vuetify/dist/vuetify.css';` below the WebFontLoader in `entry-client.js`

KaelWD: es6 imports are hoisted

PR #5831, opened 2018-12-06
Expose Vuetify components so that they can be dynamically used

christophehenry:
Exports components to let them be imported in code and [dynamically and asynchronously](https://vuejs.org/v2/guide/components-dynamic-async.html) be used.

### DIFF
--- a/packages/vuetify/src/index.ts
+++ b/packages/vuetify/src/index.ts
@@ -21,3 +21,4 @@ if (typeof window !== 'undefined' && window.Vue) {
 }
 
 export default Vuetify
+export * from './components'


KaelWD:
There's no point to this, all the components have already been registered globally. If you import from `vuetify/lib` then you can already do this. 

https://vuetifyjs.com/en/guides/a-la-carte

christophehenry:
> There's no point to this, all the components have already been registered globally. If you import from `vuetify/lib` then everything's already exposed. `vuetify-loader` will also dynamically import components if your routes are asynchronously loaded already.
> 
> https://vuetifyjs.com/en/guides/a-la-carte

Oh, nice! Sorry for the noise.

packages/vuetify/src/index.ts
```diff
@@ -21,3 +21,4 @@ if (typeof window !== 'undefined' && window.Vue) {
 }
 
 export default Vuetify
+export * from './components'
```
elegantwater: I add this code but It does not work, still can not be dynamically used

PR #6086, opened 2019-01-08
feat(VSelect): replicate html select hotkeys

### LINKED ISSUE #4853: [Feature Request] Replicate HTML Select hotkeys in Vuetify <v-select>
### Problem to solve
In a vanilla html select input, once the input is focused, the different options can be quick-selected by typing the first letter of an option. This not only highlights the value, but also  **selects** that value.

This may seem trivial, but it is a standardized feature in select inputs, and is an important feature for semi-advanced users who are accustomed to faster data input methods.

This pen shows the issue:
https://codepen.io/smcolburn/pen/zLbXOy?&editors=101

To reproduce:
1. Click on the vuetify `<v-select>` input.
2. Press `F`. Nothing happens.
3. Click on the HTML `<select>` input.
3. Press `D`. `Dog` is automatically highlighted. If the user tabs away, `Dog` is saved as the input value.

### Proposed solution
`<v-select>` should emulate standard HTML inputs in this regard. If a user begins to type with the input focused, it should automatically select the closest matched value to their input based on the `itemValue` prop, string match, or possibly a filter function. 

**THIS IS NOT THE SAME AS V-AUTOCOMPLETE** 
Note that is not the same as `v-autocomplete` or the basic filtering functionality therein. The autocomplete suggests options based on the users input, and filters possible values based on that input.

This is recreating the standard HTML select functionality. No options should be filtered out, and the closest matching option should be automatically selected from the list of options.

dima74:
## Description

Possible implementation for quick select item in `v-select` by typing first few letters of item

## Motivation and Context

Fixes #4853

## How Has This Been Tested?

visually & unit

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-layout align-center justify-center>
      <v-flex xs4 text-xs-center>

        <v-text-field solo />
        <v-select
          solo
          :items="items"
          label="Standard"
        />
        <a href="https://codepen.io/dima74/pen/yGXabW" target="_blank">codepen with native select</a>

      </v-flex>
    </v-layout>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
      items: ['Aaa', 'Ooo', 'Faa', 'Foo'],
    }),
  };
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.js
+++ b/packages/vuetify/src/components/VSelect/VSelect.js
@@ -571,7 +571,7 @@ export default VTextField.extend({
       if (this.multiple) return
 
       const KEYBOARD_LOOKUP_THRESHOLD = 1000 // milliseconds
-      const now = Date.now()
+      const now = performance.now()
       if (now - this.keyboardLookupLastTime > KEYBOARD_LOOKUP_THRESHOLD) {
         this.keyboardLookupPrefix = ''
       }
--- a/packages/vuetify/src/components/VSelect/VSelect.js
+++ b/packages/vuetify/src/components/VSelect/VSelect.js
@@ -108,7 +108,9 @@ export default VTextField.extend({
       ? vm.value
       : vm.multiple ? [] : undefined,
     selectedIndex: -1,
-    selectedItems: []
+    selectedItems: [],
+    keyboardLookupPrefix: '',
+    keyboardLookupLastTime: 0
   }),
 
   computed: {
@@ -387,6 +389,7 @@ export default VTextField.extend({
       input.data.domProps.value = null
       input.data.attrs.readonly = true
       input.data.attrs['aria-readonly'] = String(this.readonly)
+      input.data.on.keypress = this.onKeyPress
 
       return input
     },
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/components/VSelect/VSelect.js
```diff
@@ -387,6 +389,7 @@ export default VTextField.extend({
       input.data.domProps.value = null
       input.data.attrs.readonly = true
       input.data.attrs['aria-readonly'] = String(this.readonly)
+      input.data.on.keypress = this.onKeyPress
```
KaelWD: keypress is deprecated, move to the keydown handler

packages/vuetify/src/components/VSelect/VSelect.js
```diff
@@ -564,6 +567,22 @@ export default VTextField.extend({
         this.isMenuActive = false
       }
     },
+    onKeyPress (e) {
+      if (this.multiple) return
+
+      const KEYBOARD_LOOKUP_THRESHOLD = 1000 // milliseconds
+      const now = Date.now()
```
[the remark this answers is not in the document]
dima74: Yes, thank you

packages/vuetify/src/components/VSelect/VSelect.js
dima74: ok, i will change it soon

dima74: @KaelWD so I tried to use keydown handler, but I couldn't find good way to detect if pressed key is printable (https://stackoverflow.com/q/43877434/). Looks like proper solution would be to use [`beforeinput`](https://developer.mozilla.org/en-US/docs/Web/Events/beforeinput) event, but it has poor browser support ([not implemented in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=970802)). Maybe it is good idea to keep `keypress`...

KaelWD: That's a bit silly, guess it'll have to do then.

PR #6114, opened 2019-01-11
feat: allow do something with vnode when image error

bluelovers:
## Description

something about if `src` is fail loaded, we can set it back to `lazy-src` or do other with vnode

## How Has This Been Tested?

i can't build this repo at local by some unknow reason
so i didn't test
but i try find vnode by chrome and set `vnode.__vue__.image.src = "xxxx" ` is work, so i think this change should be ok too

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

- [ ] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).

### DIFF
--- a/packages/vuetify/src/components/VImg/VImg.ts
+++ b/packages/vuetify/src/components/VImg/VImg.ts
@@ -130,7 +130,7 @@ export default VResponsive.extend({
     onLoad () {
       this.getSrc()
       this.isLoading = false
-      this.$emit('load', this.src)
+      this.$emit('load', this.src, this)
     },
     onError () {
       consoleError(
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
What is this for? You can access vnode using refs

KaelWD:
Please explain what this is for and how it's supposed to be used.

bluelovers:
i wanna do something like this

```html
<v-img
    src="a_image_url_maybe_not_exists1"
    lazy-src="a_image_url_maybe_not_exists2"
    height="150px"
    align-center="true"
    align-content-center="true"
    v-on:error="imageError"
    v-on:load="imageLoaded"
    slot="activator"

>
</v-img>
```

```ts
...

	imageError(url, vnode)
	{
		vnode.image.src = 'another_image_url_exists'
	}

...
```

KaelWD:
Bind `:src` dynamically, you should never change props that way.

PR #6123, opened 2019-01-12
refactor(util): convert to TS

sh7dm:
## Description
Convert all JS utils to TS

## Motivation and Context
Framework TS conversion

## How Has This Been Tested?
`jest`

## Types of changes

## Checklist:


packages/vuetify/src/util/mask.ts
```diff
@@ -11,32 +11,32 @@ export const defaultDelimiters = /[-!$%^&*()_+|~=`{}[\]:";'<>?,./\\ ]/
...
-export const isMaskDelimiter = char => char && defaultDelimiters.test(char)
+export const isMaskDelimiter = (char: string): boolean => char ? defaultDelimiters.test(char) : false
 
 /**
  * Mask keys
  *
  * @type {Object}
  */
-const allowedMasks = {
+const allowedMasks: any = {
```
jacekkarczmarczyk:
```ts
interface MaskItem {
  test: (char: string) => boolean;
  convert?: (char: string) => string;
}
const allowedMasks: Record<string, MaskItem>;
```
(or something like this, didn't test it)

packages/vuetify/src/util/rebuildFunctionalSlots.ts
```diff
@@ -4,7 +4,7 @@
  * @param {function} h
  * @returns {array}
  */
-export default function rebuildFunctionalSlots (slots, h) {
+export default function rebuildFunctionalSlots (slots: object[], h: Function) {
```
jacekkarczmarczyk: `h: CreateElement`, slots also should be properly typed, it's not just array of any objects

packages/vuetify/src/util/mask.ts
```diff
@@ -60,7 +68,7 @@ const isMask = char => allowedMasks.hasOwnProperty(char)
  *
  * @return {String}
  */
-const convert = (mask, char) => {
+const convert = (mask: MaskType, char: string): string => {
   return allowedMasks[mask].convert ? allowedMasks[mask].convert(char) : char
```
jacekkarczmarczyk: if `convert` is always in MaskItem then this check is not necessary anymore

packages/vuetify/src/util/mask.ts
```diff
@@ -89,7 +97,7 @@ const maskValidates = (mask, char) => {
  *
  * @return {String}
  */
-export const maskText = (text, masked, dontFillMaskBlanks) => {
+export const maskText = (text: string | any[], masked: string | any[], dontFillMaskBlanks: boolean): string => {
```
jacekkarczmarczyk: shouldn't it be `string | string[]`?
also maybe it would be a good idea to add maskable.js as well, just as a additional verification if everything is ok

packages/vuetify/src/util/rebuildFunctionalSlots.ts
```diff
@@ -1,10 +1,12 @@
+import { CreateElement, VNodeChildren } from "vue";
+
 /**
  *
  * @param {object} slots
  * @param {function} h
  * @returns {array}
  */
-export default function rebuildFunctionalSlots (slots, h) {
+export default function rebuildFunctionalSlots (slots: Record<string, VNodeChildren>, h: CreateElement) {
```
jacekkarczmarczyk: in vue.d.ts $slots is defined as `readonly $slots: { [key: string]: VNode[] | undefined };`, so maybe it should be the same here

packages/vuetify/src/util/mask.ts
```diff
@@ -11,35 +17,37 @@ export const defaultDelimiters = /[-!$%^&*()_+|~=`{}[\]:";'<>?,./\\ ]/
  *
  * @return {Boolean}
```
KaelWD: Remove docblocks

packages/vuetify/src/util/mask.ts
```diff
@@ -11,35 +17,37 @@ export const defaultDelimiters = /[-!$%^&*()_+|~=`{}[\]:";'<>?,./\\ ]/
...
-export const isMaskDelimiter = char => char && defaultDelimiters.test(char)
+export const isMaskDelimiter = (char: string): boolean => char ? defaultDelimiters.test(char) : false
 
 /**
  * Mask keys
  *
  * @type {Object}
  */
-const allowedMasks = {
+const allowedMasks: Record<MaskType, MaskItem> = {
```
KaelWD: `Dictionary<MaskItem>`, it's ok to add more masks here

packages/vuetify/src/util/mask.ts
```diff
@@ -11,35 +17,37 @@ export const defaultDelimiters = /[-!$%^&*()_+|~=`{}[\]:";'<>?,./\\ ]/
...
   '#': {
-    test: char => char.match(/[0-9]/)
+    test: (char: string): boolean => Boolean(char.match(/[0-9]/)),
```
KaelWD: annotations shouldn't be needed as they're already defined by `MaskItem`

packages/vuetify/src/util/mask.ts
```diff
@@ -11,35 +17,37 @@ export const defaultDelimiters = /[-!$%^&*()_+|~=`{}[\]:";'<>?,./\\ ]/
...
+    convert: (char: string) => char
```
KaelWD: `convert` can be made optional in the interface

sh7dm: If I try this, I have some problems

PR #6350, opened 2019-02-04
fix(VDatePicker): Unwanted form submission on header click.

iPaat:
## Description

If a DatePicker is used within a form, a click on the month in the DatePicker header will cause an unwanted submit of the form. 

My change will correct this behavior.

## Motivation and Context

I would like to create a form where the user can select a date.

## How Has This Been Tested?

I tested with the existing tests.

## Markup:
https://jsfiddle.net/jkd4rbo8/21/

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePickerHeader.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePickerHeader.ts
@@ -112,7 +112,7 @@ export default mixins(
       const header = this.$createElement('div', this.setTextColor(color, {
         key: String(this.value)
       }), [this.$createElement('button', {
-        domProp: {
+        domProps: {
             type: 'button'
         },
         on: {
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VDatePicker/VDatePickerHeader.ts
```diff
@@ -112,6 +112,9 @@ export default mixins(
       const header = this.$createElement('div', this.setTextColor(color, {
         key: String(this.value)
       }), [this.$createElement('button', {
+        domProps: {
```
KaelWD: attrs

johnleider: I think he means you should use attrs instead of domProps.

iPaat: Oh, my bad. Of course I changed it.

PR #6445, opened 2019-02-10
fix(v-ripple): skip mouse events if triggered by touch

### LINKED ISSUE #6418: [Bug Report] Ripple is broken since 1.4.7
### Versions and Environment
**Vuetify:** 1.5.0
**Last working version:** 1.4.6
**Vue:** 2.6.3
**Browsers:** Chrome 71.0.3578.98
**OS:** Windows 10

### Steps to reproduce
Create a v-btn or other component using v-ripple.

### Expected Behavior
Ripple effect

### Actual Behavior
No ripple effect

### Reproduction Link
<a href="https://codepen.io/anon/pen/JxMQdZ" target="_blank">https://codepen.io/anon/pen/JxMQdZ</a>

KaelWD:
## Description

Always bind touch events, and skip the fake mouse events if ripple was triggered by a touch

## Motivation and Context

fixes #6418

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-card ripple>
        <v-card-text>
          ripple
        </v-card-text>
      </v-card>
    </v-container>
  </v-app>
</template>
```
</details>

## Types of changes

### DIFF
--- a/packages/vuetify/src/directives/ripple.ts
+++ b/packages/vuetify/src/directives/ripple.ts
@@ -130,7 +130,10 @@ function isRippleEnabled (value: any): value is true {
 function rippleShow (e: MouseEvent | TouchEvent) {
   const value: RippleOptions = {}
   const element = e.currentTarget as HTMLElement
-  if (!element) return
+  if (!element || element._ripple!.touched) return
+  if (isTouchEvent(e)) {
+    element._ripple!.touched = true
+  }
   value.center = element._ripple!.centered
   if (element._ripple!.class) {
     value.class = element._ripple!.class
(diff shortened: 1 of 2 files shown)


cdubois:
This is triggering an exception for me in most of my cases when a v-btn immediately hides a parent element with a v-if, with the code added to ripple.ts, starting at line 145:
`const element = e.currentTarget as HTMLElement | null
	  if (!element) return
	
	  window.setTimeout(() => { element._ripple!.touched = false })
	  ripple.hide(element)`
I think it should return on line 146 if element or element._ripple are undefined.  My code throws as element._ripple is not defined in the setTimeout.  This of course could be indicative of another issue, but that patch gets my code around the exception.  I'm using my WebPack app in Chrome 72, Vue 2.6.6, and Vuetify 1.5.1.

MajesticPotatoe:
@cdubois fixed this a lil while ago 0cd5671

BuhoSabio:
IE 11 issue...

If using the Vuetify 1.5.1 js file, need to update function line: 20629

window.setTimeout(function () {
        **if (typeof (element._ripple) === 'undefined') {return true;}**  
        element._ripple.touched = false;
    });

PR #6513, opened 2019-02-15
Removes unnecessary outline for circle pagination buttons.

iliyaZelenko:
This will remove the unnecessary border: 

For rectangular buttons, this does not interfere, but I think it is not necessary for circle ones.


jacekkarczmarczyk:
@iliyaZelenko please use the PR template in description, don't forget to include the Playground.vue for testing
image

sh7dm:
@iliyaZelenko please use `commitizen` for good commit names

iliyaZelenko:
@jacekkarczmarczyk @sh7dm In this case, we could use the pseudo-class `:focus`. We can remove the `outline`, but for the `focus` to make a beautiful round border.

Issue #6530, opened 2019-02-17
[Feature Request] Support for flex-grow: 0 and flex-shrink: 0


justinhelmer:
### Problem to solve
Currently, the flex CSS for `.layout`, `.flex`, etc. assume `flex-grow: 1` and `flex-shrink: 1`. Although this works for a lot of cases, there are situations where you may only want _some_ children of a flex container to grow/shrink, and others not to.

For example, consider a layout where you want to push a footer to the bottom of the page (but not `fixed`/`sticky`). If you have a wrapper with two children (i.e. `.content` and `.footer`) and they _both_ have `flex-grow: 1`, the `.footer` can grow even though you want it to remain a fixed height. The opposite is also true: the `flex-shrink: 1` results in the content of the footer being shrunk.

The `.shrink` option sets `flex-grow: 0`, but also sets `flex-shrink: 1`, which we wouldn't want in this case (don't want the footer wrapper to shrink smaller than the content).

In this case, it would be nice to be able to set `flex-grow: 0` and `flex-shrink: 0` for _just_ the `.footer`, and allow the `.content` to continue to grow to fill the remaining height.

### Proposed solution
Support for specifying the `.flex-grow` and `.flex-shrink` values, i.e.

```html
<v-layout grow-0 shrink-0>
```

Other solutions (such as [MDBootstrap](https://mdbootstrap.com/docs/vue/utilities/flexbox)) support this capability, and I think it would be a great value-add.

Currently we are adding an additional class to support this behavior and would like to be able to remove it.

sh7dm:
Good for beginner, anyone wants to contribute?

justinhelmer:
sure, I took a stab at it: https://github.com/vuetifyjs/vuetify/pull/6533

sh7dm:
Fixed in #6538

PR #6533, opened 2019-02-18
feat(6530): support for no grow/shrink, and better documentation

justinhelmer:
## Description
Adds ability to reset the default behavior of flex components that grow/shrink (`container`, `layout`, `flex`, `d-flex`).

After diving deeper on https://github.com/vuetifyjs/vuetify/issues/6530, I made the following realizations:
- Following the existing `.grow` and `.shrink` patterns would not be ideal, because they use the `!important` tag and could be applied to anything (not just our flex components)
- There is not a need to separate `grow-0` and `shrink-0`. The current `grow` and `shrink` attributes essentially serve to "reset" the _other_ property, since the default behavior without _any_ additional attributes is to grow and shrink.
- The documentation is unclear and needed some love. It is not clear _why_ one would use `grow` or `shrink`, given the fact that the default behavior already sets `flex-grow: 1` and `flex-shrink: 1`. Therefore, I provided better explanation and examples:

capture
> Please note that the documentation only contains the `en` translation. However I noticed the majority of components don't have a translation, so I didn't worry too much about it.

## Motivation and Context
Resolves https://github.com/vuetifyjs/vuetify/issues/6530

## How Has This Been Tested?
Visually. I actually could not get the docs running locally. It was looking to install a package that did not exist:

```
npm ERR! code E404
npm ERR! 404 Not Found: @vuetify/api-generator@^1.5.1
```

However, I have `vuetify` installed in a project and I tested the changes there.

## Markup:

<details>

```vue
<template>
  <v-container fluid>
    <v-layout row>
      <v-flex
        pa-1
      >
        <v-card
          dark
          color="green darken-3"
        >
          <v-card-text>I'm a Flex that will grow and shrink (default behavior)</v-card-text>
        </v-card>
      </v-flex>
      <v-flex
        pa-1
      >
        <v-card
          dark
          color="green darken-1"
        >
          <v-card-text>I'm a Flex that will grow and shrink (default behavior)</v-card-text>
        </v-card>
      </v-flex>
    </v-layout>
    <v-layout row>
      <v-flex
        pa-1
      >
        <v-card
          dark
          color="green darken-1"
        >
          <v-card-text>I'm a Flex that will grow and shrink (default behavior)</v-card-text>
        </v-card>
      </v-flex>
      <v-flex
        grow
        pa-1
      >
        <v-card
          dark
          color="green lighten-1"
        >
[...]

PR #6538, opened 2019-02-18
feat(VGrid): support for no grow/shrink, and better documentation

justinhelmer:
Reference: PR https://github.com/vuetifyjs/vuetify/pull/6533
Resolves https://github.com/vuetifyjs/vuetify/issues/6530

### DIFF
--- a/packages/docs/src/examples/grid/growShrink.vue
+++ b/packages/docs/src/examples/grid/growShrink.vue
@@ -2,64 +2,60 @@
   <v-container fluid>
     <v-layout row>
       <v-flex
-        grow
         pa-1
       >
         <v-card
           dark
           color="green darken-3"
         >
-          <v-card-text>#1 Im a Grow Flex</v-card-text>
+          <v-card-text>I'm a Flex that will grow and shrink (default behavior)</v-card-text>
         </v-card>
       </v-flex>
       <v-flex
-        shrink
         pa-1
       >
         <v-card
           dark
           color="green darken-1"
         >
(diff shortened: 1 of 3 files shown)


sh7dm:
Don't worry about this:
default

justinhelmer:
Wow you are fast :-)

One question - since this is in `next`, any ETA on when the package will be published for all to use? Otherwise we need to keep having custom classes in our code.

sh7dm:
@justinhelmer what version do you use?

justinhelmer:
@sh7dm - The latest published artifact:

```
npm view vuetify version
1.5.1
```

My assumption is that the `next` branch will be for `2.x` and as such, is not yet ready. Perhaps you are going to suggest we jump onto `2.0.0-alpha.X`?

sh7dm:
`next` changes will be released in `v2.0.0-alpha.5` on Tuesday. `master` and `dev` are feature-frozen now

MajesticPotatoe:
@justinhelmer would never suggest using an alpha/beta in production, considering stability is questionable and we are not accepting bug reports for 2.0 at this time. However you would be correct that if you would like to use these changes you would need to use use the alpha build. You are also welcome to patch these changes locally and use your own custom build knowing they will be there when upgrading to 2.0.

justinhelmer:
Thanks, @sh7dm and @MajesticPotatoe.

Please let me know if you need anything else from me to get this merged into `next`, and become part of the `2.0.0-alpha.5` version being released tomorrow.

Cheers.

PR #6543, opened 2019-02-19
docs(VCard): use text defined in data() function

begueradj:
## Description
In this example, I believe you wanted to use the `card_text` returned by `data()` function.  That is what I did.

## Motivation and Context
It may be confusing for beginners (as it happened to me in the start) as to why you defined `card_text` but you used different text for the description.  If you prefer not using `card_text`, as it is the case with the present situation, then please remove totally `data()` function because it is a good practice  to code  only what we need to use.

## Markup:
This is what I did, and I believe it makes sens in this case:
```
<div> {{ card_text }} </div>
```
And the output looks nice.
<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/examples/cards/usage.vue
+++ b/packages/docs/src/examples/cards/usage.vue
@@ -10,7 +10,7 @@
         <v-card-title primary-title>
           <div>
             <h3 class="headline mb-0">Kangaroo Valley Safari</h3>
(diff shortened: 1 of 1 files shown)


sh7dm:
@begueradj thanks!

Issue #6626, opened 2019-02-27
[Bug Report] v-dialog with tabs hides the tabs slider until you click any of the tabs


mrodal:
### Versions and Environment
**Vuetify:** 1.5.3
**Vue:** 2.6.7
**Browsers:** Chrome 72.0.3626.109
**OS:** Windows 10

### Steps to reproduce
* Create a v-dialog with tabs inside
* Open the dialog

### Expected Behavior
The tabs slider showing on the active tab.

### Actual Behavior
The tabs slider is hidden until you click on any tab other than the initial.

### Reproduction Link
<a href="https://codepen.io/mrodal/pen/PLqWOP" target="_blank">https://codepen.io/mrodal/pen/PLqWOP</a>

### Other comments
This is a duplicate of https://github.com/vuetifyjs/vuetify/issues/1978, that the royal sweet potato suggested me to do since you guys usually don't respond on closed issues.

I copy my comment on that issue, which is a possible solution:
>Have you considered something like [this](http://marcj.github.io/css-element-queries/) instead of the window resize event for example? I made a quick test adding that on the resize directive and it fixes this.

>Im not sure how it might affect the rest of the codebase, but they even claim that it performs better than the window resize event that is used currently on the resize directive, so it might improve the whole vuetify performance!

As you can see in the original bug, there's a lot of duplicates, and I think its an annoying bug that shouldn't exist on a library as good as this. I'm not too familiar with the codebase to say if it breaks things or not, but I think its worth considering, and if you want I can make a PR since its a simple change.

X1machinemaker1X:
Happens to me as well

cn-xufei:
Tab pagination controls not appear in expansion panel.
https://codepen.io/cn-xufei/pen/EzJXQQ

johnleider:
This is part of a larger issue with components that have internal elements that rely on a DOM element's dimensions. When a component like `v-tabs` is inside of a `v-menu`, or a `v-dialog` or a `v-expansion-panel`, it is unable to determine its slider position because its parent is hidden. This is also something we experience with inputs that use **autofocus**.

It is a larger issue that needs a more framework wide solution and the only workaround for now is to either use the **lazy** prop, or add a **$ref** to the `v-tabs` component and manually call `this.$refs.tabs.callSlider()` when the model of the `v-expansion-panel` changes.

johnleider:
[...]

Issue #1978, opened 2017-10-01
[Bug] v-dialog with tabs hides the tabs slider until you click any of the tabs


anthonysan95:
### Steps to reproduce
* Create a v-dialog with tabs inside
* Open the dialog

### Versions
0.15.7

### What is expected ?
The tabs slider showing on the active tab.

### What is actually happening ?
The tabs slider is hidden until you click on any tab.

### Reproduction Link
https://codepen.io/anon/pen/wrqeav

johnleider:
This is unfortunately part of how `v-tabs` works, it needs to know its dimensions to calculate the bar position. You can manually do this, or you can just add the **lazy** prop to your `v-dialog`.

https://codepen.io/johnjleider/pen/gGxGBZ

BennyAlex:
@johnleider How do I know when the dialog is ready when using lazy? I have soem properties inside the dialog and I get errors now. Without lazy it works fine

jacekkarczmarczyk:
The issue list of this repo is **exclusively** for bug reports and feature requests. For general questions, please join [the Discord chat room](https://community.vuetifyjs.com). You can also check [reddit](https://www.reddit.com/r/vuetifyjs/) or [stackoverflow](https://stackoverflow.com/questions/tagged/vuetify.js). Do not forget to create a reproduction that shows your problem

mrodal:
This is an annoying bug. 

Have you considered something like [this](http://marcj.github.io/css-element-queries/) instead of the window resize event for example? I made a quick test adding that on the resize directive and it fixes this. 

Im not sure how it might affect the rest of the codebase, but they even claim that it performs better than the window resize event that is used currently on the resize directive.

simonmaass:
@mrodal i agree

simonmaass:
any news here if this will be refactored?

thomas-topway-it:
hello, I confirm that using "lazy" attribute it works,
here is our app https://app.mappd.io

MTakashiMori:
Hello, if anyone have this issue outside a dialog, just wrap with "v-lazy" component. As it's described in docs, they will render when needed.

yoji0806:
This may not be your case, 
but in my case, binding <v-tabs> and <v-tabs-items> solved this issue.
(using the same v-mode attribute which is defined 'null' in Vue object's data bracket )

PR #6627, opened 2019-02-27
fix(resize): use ResizeSensor

mrodal:
## Description

Change the resize directive to use ResizeSensor instead of the window resize event

## Motivation and Context

It fixes this long lasting and with many duplicates issue https://github.com/vuetifyjs/vuetify/issues/1978 that I opened again here https://github.com/vuetifyjs/vuetify/issues/6626

It fixes the issue because the current window resize implementation requires the window to be resized for the onresize event to be triggered, and with this alternative, the onresize is fired when the _element_ is resized. For this reason, I think this might help with other possible issues as well.

Here are the docs to css-element-queries: http://marcj.github.io/css-element-queries/. 

## How Has This Been Tested?

Visually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

- [X] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).


mrodal:
😁 sorry, I thought I had selected next by mistake

mrodal:
@MajesticPotatoe did you leave comments on the files? I cant see them

MajesticPotatoe:
> @MajesticPotatoe did you leave comments on the files? I cant see them

I did not.

johnleider:
This will have to be done differently. We have a no external dependencies rules that I do not intend on breaking 😉

mrodal:
hmm what about copying the ResizeSensor code into vuetify? 😂 
Its this file:
https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js

DRoet:
I don't think copying in upstream source code is a valid solution to the no-dependencies rule, but maybe it's possible to allow people to create adapters for stuff like this? so we can always go back to the default behaviour if none is provided?

johnleider:
Closing due to inactivity.

mrodal:
@johnleider I proposed taking the ResizeSensor code into vuetify, which isnt much code.. isnt that ok?

PR #6770, opened 2019-03-20
fix(component): fix "item-key" in v-data-table

zdsfwy:
## Description

make sure "item-key" also applied when users try to control the whole item row
## Motivation and Context

while "item-key" is declared in v-data-table, the key is only automatically applied when "td" is used in the "items" scoped slot. When "tr" (or even "div") is used to control the whole row, the key is not applied to the row as "item-key" itself implies.

This should be regarded as a bug, as the code behavior and the spec is inconsistent. This sometimes will result in problems which usually happen when "key" is missing but desired(strange UI). 

In the following "markup" section, I provided a simplified demo from my program. The problem I encounter is if I delete one row which shouldn't have a leading icon and then add it back via the autocomplete list, the newly added item will somehow have the icon appended. If I explicitly add ":key=xxx" on the "tr", the problem is gone.

*Sorry that the playground I provided cannot reproduce my problem, it only demonstrates my operation. The real component where I encounter this problem has a more complicated structure and behavior with "expand" scoped slot.*

My solution is to first judge whether component users are trying to control the whole row, if yes and if the key is not set, the code will set the key based on "item-key".

## How Has This Been Tested?

tested manually, and tested with unit test
## Markup:

```vue
<template>
  <v-app>
    <v-data-table
      :headers="headers"
      :items="itemsShownInTable"
      item-key="name"
      class="elevation-1"
    >
      <template v-slot:items="props">
        <tr>
          <td>
            <div style="display:flex; align-items:center;">
              <v-checkbox hide-details v-if="props.item.calories < 300"
                class="my-3"
                color="#339999">
              </v-checkbox>
              <v-tooltip right v-else>
                  <v-icon  v-on="on" color='#f4b142'>error</v-icon>
                  <v-checkbox hide-details 
                    class="my-3" disabled
                    color="#339999">
                  </v-checkbox>
                </template>
                some hint here
              </v-tooltip>
              {{ props.item.name }}
            </div>
          </td>
          <td class="text-xs-left">{{ props.item.calories }}</td>
          <td class="text-xs-left">{{ props.item.fat }}</td>
          <td class="text-xs-left">{{ props.item.carbs }}</td>
          <td class="text-xs-left">{{ props.item.protein }}</td>
          <td class="text-xs-left">{{ props.item.iron }}</td>
          <td class="text-xs-left">
            <v-icon small @click="removeItem(props.item)" >delete</v-icon>
          </td>
        </tr>
      </template>
    </v-data-table>
    <v-autocomplete
      v-bind:value="selected"
      v-on:input="onSelect"
      :items="deletedFromTable"
      hide-no-data
      item-text="name"
      chips
      return-object
      dense
    >
    </v-autocomplete>
  </v-app>
</template>
<script>
  export default {
    data: () => ({
        selected: [],
        deletedFromTable: [],
        headers: [
          {
            text: 'Dessert (100g serving)',
            align: 'left',
            sortable: false,
            value: 'name'
          },
          { text: 'Calories', value: 'calories' },
          { text: 'Fat (g)', value: 'fat' },
          { text: 'Carbs (g)', value: 'carbs' },
          { text: 'Protein (g)', value: 'protein' },
          { text: 'Iron (%)', value: 'iron' },
          { text: '', value: '' }
        ],
        desserts: [
          {
            name: 'Eclair',
            calories: 262,
            fat: 16.0,
            carbs: 23,
            protein: 6.0,
            iron: '7%'
          },
          {
            name: 'Cupcake',
            calories: 305,
            fat: 3.7,
            carbs: 67,
            protein: 4.3,
            iron: '8%'
          },
          {
            name: 'Gingerbread',
            calories: 356,
            fat: 16.0,
            carbs: 49,
            protein: 3.9,
            iron: '16%'
          }
        ]
      }),
    computed: {
      itemsShownInTable: function() {
        return this.desserts.filter(function(d) {
          return this.deletedFromTable.indexOf(d) === -1
        }, this)
      }
    },
    methods: {
      removeItem: function(item) {
        this.deletedFromTable.push(item)
      },
      onSelect: function($event) {
        let idx = this.deletedFromTable.indexOf($event)
        if (idx !== -1) {
          this.$nextTick(function() {
            this.selected = []
            this.deletedFromTable.splice(idx, 1)
          })
        }
      }
    }
  }
</script>
```

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDataTable/mixins/body.js
+++ b/packages/vuetify/src/components/VDataTable/mixins/body.js
@@ -33,6 +33,9 @@ export default {
 
       return this.genTR([transition], { class: 'v-datatable__expand-row' })
     },
+    getItemKeyOrDefault (props, defalutVal) {
+      return this.itemKey ? getObjectValueByPath(props.item, this.itemKey) : defalutVal
+    },
     genFilteredItems () {
       if (!this.$scopedSlots.items) {
         return null
@@ -42,15 +45,19 @@ export default {
       for (let index = 0, len = this.filteredItems.length; index < len; ++index) {
         const item = this.filteredItems[index]
         const props = this.createProps(item, index)
-        const row = this.$scopedSlots.items(props)
+        let row = this.$scopedSlots.items(props)
 
-        rows.push(this.hasTag(row, 'td')
-          ? this.genTR(row, {
-            key: this.itemKey ? getObjectValueByPath(props.item, this.itemKey) : index,
+        if (this.hasTag(row, 'td')) {
+          row = this.genTR(row, {
+            key: this.getItemKeyOrDefault(props, index),
             attrs: { active: this.isSelected(item) }
           })
-          : row)
-
+        } else if (row.length === 1 && !row[0].key) {
+          row[0].key = this.getItemKeyOrDefault(props, index)
+        } else {
+          // reach cases of bad usage
+        }
+        rows.push(row)
         if (this.$scopedSlots.expand) {
           const expandRow = this.genExpandedRow(props)
           rows.push(expandRow)
--- a/packages/vuetify/test/unit/components/VDataTable/VDataTable.spec.js
+++ b/packages/vuetify/test/unit/components/VDataTable/VDataTable.spec.js
@@ -456,4 +456,24 @@ test('VDataTable.vue', ({ mount, compileToFunctions }) => {
 
     expect('Unable to locate target [data-app]').toHaveBeenTipped()
   })
+
+  it('item-key should also be applied when users try to control the whole row in items slot', async () => {
+    const data = dataTableNestedTestData()
+    const component = Vue.component('test', {
+      render (h) {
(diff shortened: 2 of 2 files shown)


nekosaur:
Thank you for your interest in improving vuetify. However since the data-table component has been completely reworked for the upcoming 2.0 release, and since there exists a simple workaround (simply adding the key yourself in the slot template), I will be closing this.

zdsfwy:
Hi @nekosaur, looking forward to the 2.0 release.  
Do you think we instead need to modify the Vuetify doc a bit, to remind users about the need of adding ":key" manually? It would be not necessary though if Vuetify 2.0 is coming quite soon.

nekosaur:
Yes. A note somewhere in the docs could be a good idea.

zdsfwy:
To anyone who concerns, I create a new PR to change the doc correspondingly. #6772

PR #6837, opened 2019-03-26
added Hebrew locale file

MordiSacks:
Added Hebrew locale file.


bdsoha:
If you are looking for translation help, let me know.
אשמח לעזור!

PR #6905, opened 2019-04-03
added check for componentOptions existing on slot fixes #6901

### LINKED ISSUE #6901: [Bug Report] v-speed-dial  error when using slots
### Versions and Environment
**Vuetify:** 1.5.2
**Last working version:** 1.5.1
**Vue:** 2.6.10
**Browsers:** Chrome 73.0.3683.86
**OS:** Windows 10

### Steps to reproduce
codepen

### Expected Behavior
No errors

### Actual Behavior
i'm receiving this error:
`webpack-internal:///./node_modules/vue/dist/vue.runtime.esm.js:1872 TypeError: Cannot read property 'Ctor' of undefined`

This happens when i use a slot in a v-speed-dial

it happens in the render function of the v-speed-dial perhaps because i am using slots inside the speeddail
```
if (this.isActive) {
            var btnCount_1 = 0;
            children = (this.$slots.default || []).map(function (b, i) {
                if (b.tag && b.componentOptions.Ctor.options.name === 'v-btn') {  // <<<<<--------------------- here
                    btnCount_1++;
```

and i see in this commit that this loop/map is added:
https://github.com/vuetifyjs/vuetify/commit/7e0a521e29f12c2e62b5ce0525452e00fa401f81#diff-c9647319ceb66515a767fac3e1419f3b

### Reproduction Link
<a href="https://codepen.io/anon/pen/BENVJK?editors=1111" target="_blank">https://codepen.io/anon/pen/BENVJK?editors=1111</a>

ricardovanlaarhoven:
## Description
Adding a `typeof !== 'undefined'` check so only slots with `componentOptions` will be adding an animation.

## Motivation and Context
it fixes #6901 

## How Has This Been Tested?
I've tested this in my project with the same code as the issues codepen: https://codepen.io/anon/pen/BENVJK?editors=1111

and with the playground as documented below

## Markup:

<details>
Playground.vue

```
<template>
  <v-app>
    <child>
      <div slot="speedDialAfter">
        test
      </div>
    </child>
  </v-app>
</template>

<script>
  import Child from './Child.vue';
  export default {
    components: {Child},
    data: () => ({
    })
  }
</script>


```
Child.vue

```
<template>
  <div>
    <v-fab-transition>
      <v-speed-dial
          v-model="fab"
          :top="true"
          :right="true"
          :open-on-hover="true"
          direction="bottom"
          transition="slide-y-reverse-transition"
      >
        <v-btn slot="activator" color="accent" dark fab hover v-model="fab">
          <v-icon>view_headline</v-icon>
          <v-icon>close</v-icon>
        </v-btn>
        <v-tooltip left>
          <v-btn
              fab
              dark
              small
              color="red"
              slot="activator"
          >
            <v-icon>delete</v-icon>
          </v-btn>
          <span>delete</span>
        </v-tooltip>
        <slot name="speedDialAfter"></slot>
      </v-speed-dial>
    </v-fab-transition>
  </div>
</template>

<script>
  export default {
    name: 'child',
    data: () => ({
      fab: null
    })
  }
</script>

```
</details>

## Types of changes

## Checklist:

No documentation needed

### DIFF
--- a/packages/vuetify/src/components/VSpeedDial/VSpeedDial.js
+++ b/packages/vuetify/src/components/VSpeedDial/VSpeedDial.js
@@ -65,7 +65,7 @@ export default {
     if (this.isActive) {
       let btnCount = 0
       children = (this.$slots.default || []).map((b, i) => {
-        if (b.tag && b.componentOptions.Ctor.options.name === 'v-btn') {
+        if (b.tag && typeof b.componentOptions !== 'undefined' && b.componentOptions.Ctor.options.name === 'v-btn') {
           btnCount++
           return h('div', {
             style: {


packages/vuetify/src/components/VSpeedDial/VSpeedDial.js
```diff
@@ -65,7 +65,7 @@ export default {
     if (this.isActive) {
       let btnCount = 0
       children = (this.$slots.default || []).map((b, i) => {
-        if (b.tag && b.componentOptions.Ctor.options.name === 'v-btn') {
+        if (b.tag && typeof b.componentOptions !== 'undefined' && b.componentOptions.Ctor.options.name === 'v-btn') {
```
johnleider:
```suggestion
        if (b.tag && typeof b.componentOptions && b.componentOptions.Ctor.options.name === 'v-btn') {
```

KaelWD: Should probably be any non-empty vnode, dunno why it singles out buttons.

ricardovanlaarhoven: @johnleider is this typescript specific?
in javascript `typeof something` results in a string `'undefined'`
And then it would always results in true because it's a string.

Anyway the code you suggested results in the same error.

ricardovanlaarhoven: @KaelWD Could you tell me how to do that? Your solution sounds better, if i understand it correctly you mean that we change the if to check if there is any vuetify component (starts with `v-`)

Perhaps we should ask @kuromoka and @sh7dm if there was any particular reason for this

johnleider: My suggested change was merely for style purposes. As far as @KaelWD 's suggestion. This is used in a few places, maybe it's time to normalize the functionality into a helper of sorts.

ricardovanlaarhoven: Hi John,

But when i apply your auggestion and test it, it breaks.

KaelWD: > check if there is any vuetify component (starts with v-)

any element, not just vuetify components

ricardovanlaarhoven: How could i get this PR accepted?
there is a lot of discussion and good suggestions, but nothing specific.

John's comment breaks code and i don't get what is better about it since only `!== 'undefined'` is removed.

Kael's comment is a really good suggestion, but i don't know how to implement that.

AdrianSkierniewski: I'm facing same issue and code change above fixes it

johnleider: My suggestion was just to change the compare logic. If it doesn't work, so be it.

As far as getting this accepted, I'd say that this works and let's get it in.

PR #7019, opened 2019-04-18
Feat(VIcon): remove the need for svg- prefix

DRoet:
## Description

Currently when using mdiSvg users are suppose prefix the icon with `svg-{{icon}}`

## Motivation and Context

Ain't nobody got time for prefixing their icons, all of the `@mdi/js` icons have the same structure: `M ... Z`, which is an relative easy check to perform.

## How Has This Been Tested?

unit | visually

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-icon>{{ icons.mdiAccount }}</v-icon>
  </v-container>
</template>

<script>
import { mdiAccount } from '@mdi/js'

export default {
  data: () => ({
    icons:{
      mdiAccount: mdiAccount
    }
  })
}
</script>
```
</details>

## Types of changes

## Checklist:


DRoet:
Pretty sure I have to update snapshots? wasn't sure how

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -182,7 +176,7 @@ const VIcon = mixins(
     const icon = this.getIcon()
 
     if (typeof icon === 'string') {
-      if (icon.indexOf('svg-') === 0) {
+      if (icon[0] === 'M' && icon[icon.length - 1] === 'Z') {
```
DRoet: Any reason we don't just check for `this.$vuetify.icon.iconfont` here? there is an service for the iconfont but I haven't found any usage for it in the codebase? what is currently the point of setting the `iconfont` in the Vuetify constructor when it doesn't get used anywhere ?

jacekkarczmarczyk:
`yarn test -u`

DRoet:
> yarn test -u

@jacekkarczmarczyk cheers, altho I realized it is not needed for this change, since the snapshot render is suppose to be the same

packages/vuetify/src/components/VIcon/VIcon.ts
KaelWD: https://github.com/vuetifyjs/vuetify/blob/2d52dac4ebd02efaddc5289a0d29b09a6a1bd003/packages/vuetify/src/services/icons/index.ts#L14-L21

It just sets what the other components use by default, you can use any type of icon no matter what it's set to.

KaelWD: `startsWith`/`endsWith`

KaelWD: svg path's aren't guaranteed to have this structure, so it'll only work for mdi

DRoet: > svg paths also aren't guaranteed to have this structure, so it'll only work for mdi

true, but the current `svg-` prefix also only works for `@mdi/js` 🤷‍♂️

DRoet:
On a side note, any chance I could backport the `@mdi/js` stuff into `master` or will it get rejected? I'd love to use `mdi` instead of `md`, but the complete `.css` bundle is quite large... (189kb gzipped) so it would be cool to be able to treeshake in `1.5.x` or `1.6.x`

johnleider:
There are currently no plans for a 1.6.

DRoet:
So yes it would get rejected for 1.5.x?

sh7dm:
@DRoet yea, it'll be in v2.x.x, not in v1.5

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -182,7 +176,7 @@ const VIcon = mixins(
     const icon = this.getIcon()
 
     if (typeof icon === 'string') {
-      if (icon.indexOf('svg-') === 0) {
+      if (icon.startsWith('M') && icon.endsWith('Z')) {
```
sh7dm: it can end with `Z`, but it can not to do that. I don't know about `M`

DRoet: See https://github.com/vuetifyjs/vuetify/pull/7019#discussion_r276595311

sh7dm: :ok_hand:

packages/vuetify/src/components/VIcon/VIcon.ts
peluprvi: The previously `svg-` prefix was because we shouldn't assume that every icon in every css icon library doesn't starts with `M` and doesn't ends with `Z`.
It was working with any other svg path if you prefix it with `svg-`, not only `@mdi/js`.

packages/vuetify/src/components/VIcon/VIcon.ts
peluprvi: Are there any css icon library that starts the icon name with `M` and ends with `Z`?
We are not accepting `@mdi/js` only. It is supposed to work with any custom svg path.

packages/vuetify/src/components/VIcon/VIcon.ts
DRoet: While I get that and agree, is there another icon library that exports its icons as svg, and specifically a svg `d` path-only and with a 24px viewport? I just noticed that the current implementation already heavily targets `@mdi/js` (even through the documentation)

Fontawesome for example already has it's own implementation of an SVG renderer.

sh7dm:
@peluprvi I need your approval for this to be merged

johnleider:
Hold up on the merging, we need to verify that this is compatible with other svg path libraries.

DRoet:
I would love to know which others it was suppose to work for since both implementations currently heavily target `@mdi/js` 😅

nekosaur:
I haven't looked at the code too closely, but in what way was the old prefixed solution hardcoded or targeted to `@mdi/js` specifically?

DRoet:
> I haven't looked at the code too closely, but in what way was the old prefixed solution hardcoded or targeted to `@mdi/js` specifically? (other than through documentation)

it heavily relies on the export that `@mdi/js` provides:

1) expects a single `<path d>` value, you can't just randomly insert an full svg
2) has a hardcoded "0 0 24px 24px" viewport, so even u manage to fit in an icon, if it is bigger than 24x24 it won't fix, I have actually tried this with fontawesome:

`import { library } from '@fortawesome/fontawesome-svg-core'`
`<v-icon>{{ library.icon[0] }}</v-icon> // the "M .... Z path structure is located in here"`

but the fontawesome icons are 512x512 by default so it wouldn't fit.

...so the current implementation doesn't seem generic enough either

nekosaur:
mm. those are some decent points. maybe change prefix to `mdi-` to signal that it's not a generic svg solution?

DRoet:
that's also a possibility, my point was that if you are only going to target `mdi` you might aswell remove the prefix all together to make it less cumbersome to use 🤷‍♂️ 

we could always extend the renderer in the future if other icon-fonts need support, like what was done for `fontawesome`

packages/vuetify/src/components/VIcon/VIcon.ts
peluprvi: @DRoet You are right, I'm just trying to see more possibilities and without break the previously solutions. I already talked with the @vuetifyjs/core-team about the focused aproach on `@mdi/js` and about the setting's name `mdi-svg`. The choice for the naming on settings was because we copied the paths strings from `@mdi/js` library so we have to refer them.

Both solutions were designed for a specific case, like you said, but they allow its usage by coping the inner path.

Another thing we can add there is a way to allow `<v-icon><svg ...>...</svg></v-icon>` itself.

Libraries, apps, icon packs, and others that uses or exports icons and images as svg or path:
https://www.flaticon.com/ (that one has some that doesn't end with `Z` or `z`)
https://icomoon.io/app/
https://simpleicons.org/
https://linearicons.com/free

Also, please notice that a lot of svg paths ends with `z` instead of `Z`.

packages/vuetify/src/components/VIcon/VIcon.ts
peluprvi: Libraries, apps, icon packs, and others that uses or exports icons and images as svg or path:
https://www.flaticon.com/ (that one has some paths that don't end with `Z` or `z`)
https://icomoon.io/app/
https://simpleicons.org/
https://linearicons.com/free

Please notice that a lot of svg paths ends with `z` instead of `Z`.

DRoet:
@peluprvi I'll change the detection to include `M ... z` aswell, however the hardcoded viewport still seems to be an issue with (some) different fonts/svg:
<https://codepen.io/anon/pen/VNVwOw>

DRoet:
> I think it will be better to test if icon is an SVG path using the following method:
> 
> 1. Trim string
> 2. check it with `^[mzlhvcsqta]\s*[-+.0-9][^mlhvzcsqta]+/i.test(str) && /[\dz]$/i` regexp

I'll check this out/test it when I get time, or feel free to add it ofcourse @sh7dm

peluprvi:
> [...] however the hardcoded viewport still seems to be an issue with (some) different fonts/svg:
> https://codepen.io/anon/pen/VNVwOw

Maybe add a way to define `viewBox` as a prop or add a way to allow a fully `<svg>` inner `v-icon`.

nekosaur:
Not sure if this is a premature optimization, but regex aren't the fastest thing around.

DRoet:
@sh7dm regex LGTM
@nekosaur still seems fast enough in my tests with a bunch of icons, so I think it's OK
@peluprvi yeah something like that could work, however I feel like it's a bit out of scope with this PR, we could always add this later

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -182,7 +176,7 @@ const VIcon = mixins(
     const icon = this.getIcon()
 
     if (typeof icon === 'string') {
-      if (icon.indexOf('svg-') === 0) {
+      if (/^[mzlhvcsqta]\s*[-+.0-9][^mlhvzcsqta]+/i.test(icon) && /[\dz]$/i.test(icon) && icon.length > 4) {
```
jacekkarczmarczyk: This line deserves some comment with explanation

Could be also extracted to a seperate methods and made more readable

DRoet: > Could be also extracted to a seperate method

done

PR #7131, opened 2019-05-03
Adding test for applitools

megsachdev:
Hello!
At times maintaining separate visual test cases becomes a tiring task so using your existing examples setup we tried coming up with something that can ease this process. So I chose to use Cypress as the browser test harness, which Applitools has an SDK for. I started off with writing tests for some of the components so it's sort of a "smoke test" but you can add more if you'd like.

Visual regression involves comparing the pixels visually like a human would, ideally ignoring differences that a human couldn't perceive e.g. 1 pixel changes, differences in how browsers render images, drop shadows, etc. to avoid false positives; but you can still do a strict 1:1 pixel match if that's your preference.

Applitools thankfully provides an SDK that made it trivial to write tests using Cypress. So other than the things I included what you'll need to do is create your own Applitools account and then set up your GitHub CI integration if you want it to run automatically on Pull Requests; optional, but probably a good idea.

Every test we write will include selectors for elements. To save a lot of headaches, we should write selectors that are resilient to changes.

Oftentimes we see users run into problems targeting their elements because:
Your application may use dynamic classes or ID’s that change
Your selectors break from development changes to CSS styles or JS behavior

Luckily, it is very easy to avoid both of these problems:
Don’t target elements based on CSS attributes such as: id, class, tag
Don’t target elements that may change their textContent
Add data-* attributes to make it easy to target elements

That's what I have done here, I added data attributes to the existing components to make the code neat and consistent.

To run Applitools:

Create Applitools account
Go to the menu in the top right corner and click the profile icon on the far right.
55111935-b006d680-5098-11e9-9b32-26ccbb3ff591
Click on "My API Key" and copy your key
Go to your terminal and paste
`export APPLITOOLS_API_KEY=Your_API_Key_Here`
Go to your directory and run:
`yarn cypress:visual`
In your Applitools dashboard you should see the results of your tests
PIXLR

### DIFF
--- a/packages/docs/.gitignore
+++ b/packages/docs/.gitignore
@@ -9,3 +9,6 @@ npm-debug.log
 yarn-error.log
 *.swp
 .node-version
+# cypress
+cypress/screenshots
+cypress/videos
--- a/packages/docs/cypress.json
+++ b/packages/docs/cypress.json
@@ -1,0 +1,3 @@
+{
+    "baseUrl": "http://localhost:8095/en/components"
+}
--- a/packages/docs/cypress/integration/visual-testing/bottomsheet-test.spec.js
+++ b/packages/docs/cypress/integration/visual-testing/bottomsheet-test.spec.js
@@ -1,0 +1,36 @@
+import {  environment } from "../../../utils";
+describe("Visual Regression Testing", () => {
+  before(() => {
+    cy.eyesOpen({
+      appName: "Vuetify",
+      testName: "bottomsheet page",
+      browser: Object.assign(environment, {width: 800})
+    });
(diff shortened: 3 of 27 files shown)


KaelWD:
You should talk to john about this, I believe he already has someone working on e2e tests. We also have a kitchen sink on `next` instead of relying on documentation examples.

johnleider:
A few things to note:

- We have a [package](https://github.com/vuetifyjs/vuetify/tree/next/packages/kitchen) already setup for this on the `next` branch
- `master` is on lock down until v2 and any changes should be directed to `next`

megsachdev:
Closing in favor of #7152

PR #7175, opened 2019-05-10
fix(VTextField): clearable should respect readonly=true

One1L:
This fix is made because of issue #7111 

## Description
Extend condition inside genClearIcon method

## Motivation and Context
https://github.com/vuetifyjs/vuetify/issues/7111

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [x] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #7187, opened 2019-05-12
revert: fix(VTextField): clearable should respect readonly=true

sh7dm:
Reverts vuetifyjs/vuetify#7175


One1L:
@sh7dm why revert?

DRoet:
> @sh7dm why revert?

welp, we probably should've provided why. It was because it broke the picker examples. for now we think it's better if the dev explicitely removes the clearable prop if he doesn't want the readonly field to be clearable.

unless there is a better way of doing this

sh7dm:
@One1L ask @jacekkarczmarczyk

PR #7198, opened 2019-05-13
feat(VIcon): Accessible icon button

### LINKED ISSUE #4770: [Enhancement] Accessible icon button
### Versions and Environment

Vuetify: 1.1.9
Vue: -
Browsers: Firefox
OS: Linux

### Steps to reproduce

1. Go to https://vuetifyjs.com/en/components/icons#example-clickable
2. Inspect the clickable button

### Actual Behavior

    <i aria-hidden="true" class="v-icon v-icon--link 
          mdi 
          mdi-chevron-right" style="font-size: 36px;"></i>

This icon serves as a button it should not have `aria-hidden="true"`.

### Expected Behavior

The icon should probably have `role="button"`.

A workaround is to explicitly override them:

    <v-icon aria-hidden="false" role="button" @click="foo">add</v-icon>

### Reproduction Link

https://codepen.io/anon/pen/oMqxXe

### Additional Comments:

Additionally, the documentation should probably recommend setting `aria-label` and/or `title`:

    <v-icon "@click="foo" aria-label="Add">add</v-icon>

swiecaJS:
When VIcon has a click handler, aria-hidden attribute is removed. In addition, role attribute is set
to `button`.

Resolves #4770

## How Has This Been Tested?

I've extended unit tests and updated snapshots

## Markup:

```
    it('should be an accessible link', () => {
      const clickHandler = jest.fn()
      const context = functionalContext({ on: { click: clickHandler } }, '$vuetify.icons.testIcon')
      const wrapper = mount(VIcon, context)
      wrapper.trigger('click')

      expect(wrapper.element.classList).toContain('v-icon--link')
      expect(wrapper.element.getAttribute('aria-hidden')).toBeFalsy()
      expect(wrapper.element.getAttribute('role')).toBe('button')
      expect(clickHandler).toBeCalled()
    })
```

<details>

```vue

```
</details>

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] The PR title is no longer than 64 characters.
- [x ] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).
- [x ] My code follows the code style of this project.


packages/docs/src/data/api/supporters.json
```diff
@@ -145,6 +145,13 @@
...
       "type": "Patron",
       "pledge": "Gold ($50)"
+    },
+    {
+      "name": "Loomio",
+      "href": "https://www.loomio.org/",
+      "logo": "backers/loomio.svg",
+      "type": "Patron",
+      "pledge": "Gold ($50)"
     }
```
DRoet: this should not be included

DRoet:
please target the `next` branch, `dev` has been frozen for a while

swiecaJS:
I was also considering placing `console.warn()` when `<v-icon>` has click handler, and there is no `aria-label` attribute. WDYT?

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -202,9 +204,7 @@ export default Vue.extend({
 
     // Support usage of v-text and v-html
     if (data.domProps) {
-      iconName = data.domProps.textContent ||
-        data.domProps.innerHTML ||
-        iconName
+      iconName = data.domProps.textContent || data.domProps.innerHTML || iconName
```
DRoet: nit: eventho the linter doesn't complain, I feel like this can be reverted since it doesn't actually change anything

packages/vuetify/src/components/VIcon/__tests__/VIcon.spec.ts
```diff
@@ -206,9 +206,13 @@ describe('VIcon', () => {
...
-        return h('div', {
-          class: 'test-component'
-        }, this.name)
+        return h(
+          'div',
+          {
+            class: 'test-component'
+          },
+          this.name
+        )
```
DRoet: nit: same here ^^

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -73,17 +73,19 @@ const VIcon = mixins(
...
           'v-icon--left': this.left,
-          'v-icon--link': this.$listeners.click || this.$listeners['!click'],
+          'v-icon--link': hasClickListener,
           'v-icon--right': this.right,
           'v-icon--dense': this.dense
         },
         attrs: {
-          'aria-hidden': true,
+          'aria-hidden': !hasClickListener,
+          role: hasClickListener ? 'button' : null,
```
sh7dm: maybe `aria-role`?

swiecaJS: I think it should be `role`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

Also, I noticed, that in such case there should be `tabindex` - to allow icon to be selectable

PR #6814, opened 2019-03-23
feat(VSlider): add decreasing prop

sureshvv:
Implementation is very similar to global Vue rtl setting but specific setting is local to the particular v-slider.

## Motivation and Context

Needed just one particular slider to have a decreasing mode - max on the left and min on the right

## How Has This Been Tested?

Manual testing only.

## Markup:

<details>

```vue
// Paste your FULL Playground.vue here
```
</details>

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any feature but make things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [ ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #7215, opened 2019-05-15
feat(VSlider): reverse property for VSlider allows for max on left and min on right

sureshvv:
## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
```diff
@@ -146,7 +147,7 @@ export default VSlider.extend({
         ref: 'track'
       }, children)
     },
-    genChildren () {
```
sh7dm: no need for strict types on methods

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
```diff
@@ -83,7 +84,7 @@ export default VSlider.extend({
   },
 
   methods: {
-    getTrackStyle (startLength: number, endLength: number, startPadding = 0, endPadding = 0) {
+    getTrackStyle (startLength: number, endLength: number, startPadding = 0, endPadding = 0): Partial<CSSStyleDeclaration> {
```
sh7dm: No need for `: Partial<CSSStyleDeclaration>`

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
sureshvv: Copied from VSlider.ts. Should I leave it there?

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
sureshvv: Copied from VSlider.ts. Should I leave it there?

sh7dm: no

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
sh7dm: no

jacekkarczmarczyk:
Why not just check if max < min instead of adding a new prop?

DRoet:
> Why not just check if max < min instead of adding a new prop?

yeah that would be pretty neat, altho it would kind of defeat the purpose of "max" and "min" as concepts. maybe the props should've been called `start` and `end` 🤷‍♂

sureshvv:
> > Why not just check if max < min instead of adding a new prop?
> 
> yeah that would be pretty neat, altho it would kind of defeat the purpose of "max" and "min" as concepts. maybe the props should've been called `start` and `end`

or left and right would be even better :D

sureshvv:
Screenshot from 2019-03-23 11-35-09

This is to give you an idea of why it may be useful. Or if you know how this can be done without a new property.

jacekkarczmarczyk:
@sureshvv Where would you display `left` value in RTL? On the left (like in LTR) or on the right?

sureshvv:
> @sureshvv Where would you display `left` value in RTL? On the left (like in LTR) or on the right?

The beauty of left and right is they don't change with rtl. Left is left is left.

jacekkarczmarczyk:
And the cons are that 
- you'll need another props for vertical slider (there's no left/right)
- i'm not sure about this, so correct me if i'm wrong, but i guess that the default layout in RTL mode is just the mirror image of LTR, so in multilanguage site if you want to switch from LTR to RTL you'll need to switch left and right values

sureshvv:
> And the cons are that
> 
> * you'll need another props for vertical slider (there's no left/right)
> * i'm not sure about this, so correct me if i'm wrong, but i guess that the default layout in RTL mode is just the mirror image of LTR, so in multilanguage site if you want to switch from LTR to RTL you'll need to switch left and right values

1. yep.. may be top and bottom for vertical sliders

2. yep.. we are punting on rtl and let the individual programmer worry about it - but it may be all right for most users. not really sure about that though :)

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
sureshvv: fixed

packages/vuetify/src/components/VRangeSlider/VRangeSlider.ts
sureshvv: fixed

sureshvv:
@sh7dm Can we merge?

DRoet:
Have we settled on the extra prop in the discussion above?

sureshvv:
> Have we settled on the extra prop in the discussion above?

You can check out the discussion on #6814 which was the previous version of this.

johnleider:

Thank you for your contribution and interest in improving Vuetify!

sureshvv:
This is ready to be merged. Please merge.

jacekkarczmarczyk:
There's still no consensus on the props names

sureshvv:
> There's still no consensus on the props names

Let us vote and settle on one please.

nekosaur:
My vote is for whatever we've already used before. Which I think is `reverse`?

sureshvv:
My vote is also for "reverse".

KaelWD:
I'm ok with reverse. The min/max thing sounds confusing, and left/right doesn't make sense for rtl or vertical.

jacekkarczmarczyk:
The problem with `reverse` is that for text fields it aligns label and input texts, it could be used for placing label on the opposite side also in v-slider and selection controls (there's an issue for this afaik).

Definitely I agree that min/max and left/right are not the best, but there is still start/end option, could be also from-value/to-value or just from/to or maybe event something better

jacekkarczmarczyk:
And another thing - with this playground:
```vue
<template>
  <v-app>
    <v-content>
      <v-container>
        <v-slider min="1" max="10" :tick-labels="[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]" />
        <v-slider min="1" max="10" :tick-labels="[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]" reverse />
        <v-slider min="1" max="10" :tick-labels="[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]" vertical />
        <v-slider min="1" max="10" :tick-labels="[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]" vertical reverse />
      </v-container>
    </v-content>
  </v-app>
</template>
```

i got this:

image


So first of all labels are wrong in reverse mode, second - I'd expect that the initial value will be the left side value, so for normal slider it'd be 5, for reverse - 15, third - keyboard navigation with up/down keys is reversed in reverse mode (`up` moves the slider indicator down)

sureshvv:
@jacekkarczmarczyk... your playground works correctly for me now.

sureshvv:
Screenshot from 2019-06-01 00-22-54

jacekkarczmarczyk:
Great, ty, still there are 2 other issues in my comment

sureshvv:
> Great, ty, still there are 2 other issues in my comment

Fixed 1. Can you check the keyboard input again? Seems quite intuitive for me. Let me know if you would like the keys reversed too.

jacekkarczmarczyk:
For me it's not intuitive for me when you press down arrow and slider goes up, don't know about team's opinion. Also i don't recall using vertical slider with keyboard at all, so won't force anything

There also a new issue:
image

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules/
 .env
+*~
```
sh7dm: What's that

sureshvv:
@jacekkarczmarczyk can you explain new issue?

jacekkarczmarczyk:
colors of the track and ticks are inverted in reverse mode

sureshvv:
> colors of the track and ticks are inverted in reverse mode

No idea how that happened :) will check.

jacekkarczmarczyk:
1. `pageDown`/`pageUp` are not handled properly - in vertical mode page UP should move UP, not sure what should be proper behaviour for horizontal
2. `home`/`end` - imho in all cases (vertical/horizontal, and normal/reverse) `home` should set the initial value (5 for normal, 15 for reversed), `end` - the opposite
2. RTL switches top/bottom in vertical slider

I'm starting to think that allowing `min` to be greater than `max` would be a better solution, also in terms of implementation...

packages/vuetify/src/components/VSlider/VSlider.ts
```diff
@@ -549,7 +553,8 @@ export default mixins<options &
       if ([left, right, down, up].includes(e.keyCode)) {
         this.keyPressed += 1
 
-        const increase = this.$vuetify.rtl ? [left, up] : [right, up]
+        const increase = this.decreasing ? this.vertical ? [right, down] : [left, up]
```
jacekkarczmarczyk: that's a bit unreadable, also i wonder if we should handle left/right for vertical slider?

sureshvv: will try to make it more readable

sureshvv:
> 1. `pageDown`/`pageUp` are not handled properly - in vertical mode page UP should move UP, not sure what should be proper behaviour for horizontal
> 2. `home`/`end` - imho in all cases (vertical/horizontal, and normal/reverse) `home` should set the initial value (5 for normal, 15 for reversed), `end` - the opposite

Will check.

> 3. RTL switches top/bottom in vertical slider
 

How can I test RTL in the Playground?

> I'm starting to think that allowing `min` to be greater than `max` would be a better solution, also in terms of implementation...

I think it will boil down to current implementation anyway at the end.

.gitignore
sureshvv: Backup files created by "vi" editor. Prett standard in .gitignores where I come from :D

jacekkarczmarczyk:
`<v-checkbox v-model=$vuetify.rtl />`

nekosaur:
ping @sureshvv

johnleider:
This issue is being closed due to:

- [ ] Reproduction environment not provided (inactivity for 14 days)
- [x] Requested information not provided (inactivity for 14 days)
- [ ] Violation of the [Code of Conduct](https://github.com/vuetifyjs/vuetify/blob/dev/CODE_OF_CONDUCT.md)

PR #7220, opened 2019-05-15
fix(VTabs): do not hardcode v-tabs text color to white when in dark mode

nekosaur:
## Description
Checks if text color should be set to white or not when in dark mode instead of hardcoding it to always white

## Motivation and Context
This way `color` is applied in dark mode as well

## How Has This Been Tested?
playground

## Markup:

<details>

```vue
<template>
  <div>
    <v-switch v-model="$vuetify.theme.dark" label="dark"></v-switch>
    <v-tabs color="purple">
      <v-tab>Item One</v-tab>
      <v-tab>Item Two</v-tab>
      <v-tab>Item Three</v-tab>
    </v-tabs>

    <v-tabs>
      <v-tab>Item One</v-tab>
      <v-tab>Item Two</v-tab>
      <v-tab>Item Three</v-tab>
    </v-tabs>
  </div>
</template>
```
</details>

## Types of changes

## Checklist:


DRoet:
FYI it seems like you've pushed ur playground 😅

packages/vuetify/dev/Playground.calendar.vue
```diff
@@ -0,0 +1,332 @@
+<template>
```
sh7dm: why adding this? mis-added?

PR #7229, opened 2019-05-16
(fix) ensure required typescript folders are included during packaging

cadriel:
## Description
When a new Vuetify Alpha is released - the types folder is missing two sibling folders called `services` and `presets`. During development you likely would not notice this - but when consuming the npm package in your own project - you receive numerous typescript errors about missing referenced files.

This commit should ensure the two folders are now included in the published package, removing the errors at compile time.

## Motivation and Context
We've been testing the Alpha locally, but find it hard to develop because the errors and warnings produced without these two folders overwhelm any actual errors you might have.

## How Has This Been Tested?
I've run an `npm pack` locally to ensure the generated package includes the two missing folders.

## Types of changes

### DIFF
--- a/packages/vuetify/.npmignore
+++ b/packages/vuetify/.npmignore
@@ -10,4 +10,6 @@
 !/es5/**/*
 
 # Keep types
+!/types/services/*.d.ts
+!/types/presets/*.d.ts
 !/types/*.d.ts


jacekkarczmarczyk:
That would add also `test` folder, which i'm not sure whether it's required or not (for testing? and if so do we need tsconfig.json as well?)

cadriel:
This change only affects the packaged, production release of the library. The test folder should not be required for the packaged release. I've tested this locally, and not having the testing folder there doesn't cause issues.

The `tsconfig` is already configured correctly - and should not require changes.

May I request another release of the alpha after this gets merged? It'd help us out a bunch!

Thank you :)

chanlito:
@cadriel I'm running into this issue as well while trying to test the v2 of vuetify hope this gets published soon.

PR #7261, opened 2019-05-21
Include header 'class' and 'align' properties

flahol:
## Description
As the title explains just add header.class and header.align

## Motivation and Context
Make it easier


nekosaur:
Closing due to inactivity

PR #7443, opened 2019-06-10
fix: prevent duplicated theme stylesheet id

kevinmarrec:
## Description
When using SSR, `vue-theme-stylesheet` is already in DOM, but there's missing check that make client add an empty `style` tag with a duplicated id. Having two same `id` in DOM is not HTML valid.

Assuming comment here https://github.com/vuetifyjs/vuetify/blob/next/packages/vuetify/src/services/theme/index.ts#L138, I think it's just a check that have been forgotten but already thought.

## How Has This Been Tested?
I checked locally with an SSR project that there was no duplicatd tag with same id anymore.

## Types of changes

### DIFF
--- a/packages/vuetify/src/services/theme/index.ts
+++ b/packages/vuetify/src/services/theme/index.ts
@@ -156,7 +156,7 @@ export class Theme extends Service {
   // Generate the style element
   // if applicable
   private genStyleElement (): void {
-    if (typeof document === 'undefined') return
+    if (typeof document === 'undefined' || document.getElementById('vuetify-theme-stylesheet')) return
 
     /* istanbul ignore next */
     const options = this.options || {}
--- a/packages/vuetify/src/services/theme/index.ts
+++ b/packages/vuetify/src/services/theme/index.ts
@@ -70,7 +70,7 @@ export class Theme extends Service {
   // When setting css, check for element
   // and apply new values
   set css (val: string) {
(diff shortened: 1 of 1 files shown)


jousepo:
Hi!!!!
After download the last beta.5 tag using nuxt vuetify loader, i've noticed that the stylesheet is still duplicated in DOM:

image

This could be due the merge of both commits because the second one removes the change of the other as you can see:

First commit screenshot:
image

Second commit screenshot:
image

I don't know if this is the correct fix but the problem persists in the last version.

Thx!!!

jousepo:
Finally i've found the problem. The problem resides in the **initNuxt** method from _index.ts_ that injects the style into head tag but it's applied after the call of **genStyleElement** inside **checkOrCreateStyleElement**

`
this.styleEl = document.getElementById('vuetify-theme-stylesheet') as HTMLStyleElement
`

**this.styleEl** is undefined at this point 

```
 private initNuxt (root: Vue) {
    const options = this.options || {}

    root.$children.push(new Vue({
      head: {
        style: [
          {
            cssText: this.generatedStyles,
            type: 'text/css',
            id: 'vuetify-theme-stylesheet',
            nonce: options.cspNonce,
          },
        ],
      },
    } as any))
  }
```

Hope this info will be useful. I can't help you more because i'm new in vue and nuxt and all of this its new for me.

PR #7482, opened 2019-06-12
docs: add layout menu

peluprvi:
## Description

Add a `Layout` menu item and move to it the submenus about `Text Aligment`, `Aspect Ratio`, `Colors`, `Content`, `Display`, `Elevation`, `Grid`, `Grid Lists`, `Pre Made Layouts`, `Spacing`, `Transitions`, `Typography`.

## Motivation and Context

To have a clear menu that have separated groups for frameworks settings (where dev have to change some options or create setting files) and for layout (for style helpers or other related components/elements/etc).

## How Has This Been Tested?

visually

## Types of changes

## Checklist:


jacekkarczmarczyk:
I'd change here few things:

1. merge heper classes into styles section and move it to the root level
2. too much clicks need imho to get to Framework Options/Services/Whatever, services could be moved level up
3. I've never understood why grids and aspect ratios are not just in the components section, layout is the place where grid terminology and philosophy could be described, but components should be in components section
4. alacarte and predefined layouted as things that are configured at the beginning of the project could be moved just below the quick start

So this is something I'm thinking of:
- Gettings started
  - Quick start
  - A la carte
  - Pre-made layouts
  - Why Vuetify
  - ...
- Services
  - Application
  - ...
- Components
  - ...
  - Aspect ratios
  - Grid
  - ...
- Directives
  -  ...
- Styles & transitions
  - Colors
  - Content
  - Display
  - Elevation
  - ...
  - Transitions
- Permium themes


I think that makes the root level more clear and easier to find where to look for things that you want to know

nekosaur:
Agreed that meet the team etc should not be in getting started, but probably get their own group instead. Do we have a better word than Services for things like app, icons, etc? Isn't that only an internal code term. Perhaps Options, or something.

jacekkarczmarczyk:
Another improvement that could be made (not necessarily in this PR) is decoupling pages' names from their location in the navigation, for example `meet-the-team` page should be just `meet-the-team` without any prefix like `getting-started/meat-the-team`. That would made navigation refactoring much easier - no need to move json files from one folder to another, update 301 redirections or internal links etc (except just one time for this refactor ofc). Sound a big task though

@nekosaur breakpoints, icons, i18n, themes - sound like "Customizing". Application doesn't fit in here though, but we could have an `VApp` component page where this service could be mentioned. It could also mention the necessity of having `VApp` component, now I always have a problem with finding this information if I want to show it to somebody. And the last one - goTo - maybe it should go to "Styles & transitions" or "Styles & animations" or "Styles, motion & transitions"? "Customizing" group should also have the "Modifying SASS variables", the current location (Framework options/Services/Theme) is completely unintuitive
That would made services scattered in different places in navigation, but actually their locations should be based on their functionality instead of internal implementation

I've updated the structure proposal to reflect this comment

MajesticPotatoe:
closed in favor of #7594

PR #7557, opened 2019-06-18
fix(types): give explicit type to empty arrays

kevinmarrec:
## Description
Fix types of some places where empty arrays with implicit type would resolve to `never[]` type instead of `any[]` and cause TypeScript errors.

## Motivation and Context
Running a simple TypeScript project with Vuetify was displaying 5 errors regarding Vuetify types, see below the screenshots.

<details>
<summary>Screenshots of errors</summary>

</details>

Weirdly, in editor, we can't see these errors in Vuetify source code but only inside a TS project using Vuetify under `node_modules/vuetify/src/util/{file}.ts` (it's probably due to some `tsconfig.json` options).

In TypeScript, empty arrays that have no explicit types are resolving to the type `never[]`, so pushing anything (even of `any` type) in to this array leads to a type error `"Argument of type 'any' is not assignable to parameter of type 'never'."`

It can be easily fixed by explicitly defining the array type as `any[]`.

Sources :
https://github.com/Microsoft/TypeScript/issues/9976 
https://github.com/Microsoft/TypeScript/issues/9976#issuecomment-237872781

## How Has This Been Tested?
Locally in the project I had the 5 errors (seen on screenshots). These changes fixed all of them.

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -349,7 +349,7 @@ export const camelize = (str: string): string => {
  * Returns the set difference of B and A, i.e. the set of elements in B but not in A
  */
 export function arrayDiff (a: any[], b: any[]): any[] {
-  const diff = []
+  const diff: any[] = []
   for (let i = 0; i < b.length; i++) {
     if (a.indexOf(b[i]) < 0) diff.push(b[i])
   }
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/util/helpers.ts
```diff
@@ -486,7 +486,7 @@ export function padEnd (str: string, length: number, char = '0') {
 }
 
 export function chunk (str: string, size = 1) {
-  const chunked = []
+  const chunked: any[] = []
```
KaelWD:
```suggestion
  const chunked: string[] = []
```

PR #7672, opened 2019-06-25
feat(VIcon): shorter iconName prefix

kevinmarrec:
## Description
Make `$x` icon names resolves `$vuetify.icons.values.x`, while keeping `$vuetify.icons.x` working too (unless we decide to not support it anymore).
See Markup.

## Motivation and Context
I was feeling that `$vuetify.icons.x` was too verbose to tell `VIcon` to use internal/custom icons.

A simple use case that can easily showcase how it can be too verbose is when dynamically setting `preprend-icon` or `append-icon` props of a `VTextField` (could apply on few other components like `VCombobox`, `VAutocomplete` or `VListGroup`) :

```vue
<template>
  <v-text-field
    :append-icon="show ? '$vuetify.icons.visibility' : '$vuetify.icons.visibilityOff'"
    :type="show ? 'text' : 'password'"
    @click:append='show = !show'
  ></v-text-field>
</template>

<script>
export default {
  data () {
    return {
      show: false
    }
  }
}
</script>
```

It may not look too verbose with only one text field, and could also be simplified by `'$vuetify.icons.' + show ? 'visibility' : 'visibilityOff'` but when you have a component template (or render function) with multiple `$vuetify.icons.x` statement, it felt to me enough redundant to think about a shorter way to tell Vuetify to remap names to internal/custom icons : `$x`.

Taking only the `$` from `$vuetify.icons` make it still easily understandable and that it represents a internal mapped variable (or user's one through Vuetify icons options).

## How Has This Been Tested?
Jest

## Markup

```vue
<template>
  <div class="pa-5">
    <v-icon>$cancel</v-icon>
    <v-icon>$vuetify.icons.cancel</v-icon>
  </div>
</template>
```

## Types of changes

## Checklist:

- [x] I've added relevant changes to the documentation


packages/vuetify/src/components/VAlert/VAlert.ts
```diff
@@ -148,10 +148,10 @@ export default mixins(
       if (typeof this.icon === 'string' && this.icon) return this.icon
 
       switch (this.type) {
```
kevinmarrec: This `switch` and more globally the `computedIcon` function can probably can be simplified/shortened.
Maybe :
```ts
computedIcon (): string | boolean {
  return this.icon || (['info', 'error', 'sucess', 'warning'].includes(this.type) && `$${this.type}`)
}
```

The thing I don't know is : do users use `<v-alert type="info" icon="false">Message</v-alert>` behavior ?

packages/vuetify/src/util/helpers.ts
```diff
@@ -322,17 +322,15 @@ export const keyCodes = Object.freeze({
...
 // to the current name or component for that icon.
 export function remapInternalIcon (vm: Vue, iconName: string): VuetifyIcon {
-  if (!iconName.startsWith(ICONS_PREFIX)) {
+  if (!iconName.startsWith('$')) {
     return iconName
   }
 
   // Get the target icon name
-  const iconPath = `$vuetify.icons.values.${iconName.split('.').pop()}`
+  const iconPath = `$vuetify.icons.values.${iconName.split('$').pop()!.split('.').pop()}`
```
KaelWD: Deprecate `$vuetify.icons.`?

kevinmarrec: My changes doesn't make it deprecated, both will work.
Not sure if depecrating it is relevant unless everyone upvotes it
`$vuetify.icons.` is more verbose but some people probably will prefer it over just `$`

kevinmarrec: Still with my changes even `$whatever.you.want.name` will resolve to `$vuetify.icons.values.name`

PR #7707, opened 2019-06-30
docs: Add accessibility (a11y) section to docs pages

morphatic:
## Description

This is the first PR in a 3-PR bundle (see #7708 and #7709) looking to add an "Accessibility Considerations" section to the standard UI Components documentation pages. This PR makes the actual template changes necessary to display the section. The second PR is an example of what this would look like in practice on an actual docs page, `VDivider`. The third PR updates the actual `VDivider` component to implement the a11y features recommended for components of this type.

## Motivation and Context

This does not (to my knowledge) address any open (or closed) issues. The motivations are:

1. One of the explicit goals of Vuetify 2.0 is to implement and highlight a focus on a11y issues. Adding a standard a11y section to every docs page communicates that we're serious about this.
2. Many developers are not aware of, don't have much experience with, and/or don't know how to learn about, much less implement, a11y in their apps.
3. While in many cases a11y is "baked in" to Vuetify 2.0, i.e. relevant attributes are automatically added to components by the framework, in other cases, e.g. `VDialog`, developers will have to make a conscious effort to take advantage of Vuetify's a11y features. In the case of `VDialog`, `v-bind="attrs"` must be explicitly set on the activator for the dialog in order for it to be active. Adding a section to every docs page will make it easier for developers to know/remember this and make these changes to their code.
4. Going through the process of adding these sections to the docs pages will give us another opportunity to audit and publicly document our compliance with WAI-ARIA guidelines.

## How Has This Been Tested?
Visually. Here's a screenshot:

Preview of Accessibility Considerations section for Vuetify 2.0 Docs

## Types of changes

## Checklist:

- [x] The PR is submitted to the correct branch (`next` ).
- [x] I've added relevant changes to the documentation (see accompanying PRs)


johnleider:
Thoughts on just naming it Accessibility for brevity?

morphatic:
Agnostic. Simpler seems better.

johnleider:
Also just noticed that divider doesn't actually have the correct role, fixing that now.

morphatic:
Already [had a PR ready for that](https://github.com/morphatic/vuetify/commit/24bc4187b78215fa3756bf8175f8390e8c962548) 😎

johnleider:
Nice!

morphatic:
I updated the PR to change the section heading to just "Accessibility."

Also, I forgot to mention that I opted to put the a11y section at the end of the page, rather than elsewhere, but I think that decision could be debated.

On the one hand, putting the a11y section higher up on the page, say after the Usage section but before the API section, might emphasize the degree to which it is, or should be, taken seriously. Also, it might encourage more people to learn more about it and make it a priority in their own apps. At the very least, people would be forced to scroll by it on the way to the info they were seeking. Putting it at the end makes it less likely to be seen.

On the other hand, the primary reason people visit the docs is to answer immediate implementation questions. Since most devs will visit the docs dozens or hundreds of times in the course of a project, it might be a disservice, or annoy people, if we force them to scroll past this section every time to get to the API or Examples.

I think it's probably smarter to keep it at the end where it is, but I don't feel super strongly about that. Kael had a great suggestion that we can update the eslint plugin to give people a clue as to when they can/should be addressing a11y issues. If we did that, we wouldn't need to make it quite as prominent in the docs.

MajesticPotatoe:
Having it at the end is fine. There's always the right nav bar to scroll to it.

packages/docs/src/components/doc/Accessibility.vue
```diff
@@ -0,0 +1,35 @@
...
+        default: undefined,
+      },
+    },
+
+    computed: {
+      ...mapGetters('documentation', [
+        'namespace',
+        'page',
+      ]),
+      internalValue () {
```
johnleider: This doesn't appear to be used.

morphatic: This whole component or just some of it? You won't find a direct reference to the component because it [gets loaded dynamically from here](https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/components/core/Page.vue#L24). The component type is [specified in the `Divider.json` flie](https://github.com/morphatic/vuetify/blob/v-divider-docs-updates/packages/docs/src/lang/en/components/Dividers.json#L29).

morphatic: This component is basically a modified and simplified copy of [the `Usage` component](https://github.com/morphatic/vuetify/blob/v-divider-docs-updates/packages/docs/src/lang/en/components/Dividers.json#L29). TBH, I didn't fully understand the plumbing of this component, but it worked and it seemed to be basically parallel with other section-definition components.

johnleider: That's because `doc/Usage.vue` uses the internalValue for `<doc-example>`

image

morphatic: Ha! Yeah, missed that. Good catch! I updated the PR to remove it.

PR #7929, opened 2019-07-23
chore: deploy docs with a docker container

KaelWD:
This reverts commit 1e28f5a2addf4bcf3403337b8f7f2d66591104a8.

Zeit claims to support node 10.15, but only installs up to 10.7

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM mhart/alpine-node:12
+
+COPY . .
+
+EXPOSE 8095
+
+RUN yarn
+RUN yarn build
+CMD ["yarn", "start"]
--- a/now.json
+++ b/now.json
@@ -5,5 +5,8 @@
     "autoAlias": false,
     "silent": true
   },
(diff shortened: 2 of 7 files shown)


DRoet:
Idk how well Zeit uses the docker cache, but usually its better to first copy over the package.json / .lock file and install the dependencies and then copy over the rest of the source files, instead of a wildcard copy.

something like:
```dockerfile
FROM mhart/alpine-node:12

COPY package.json yarn.lock ./

RUN yarn
RUN yarn build

COPY . .

EXPOSE 8095

CMD ["yarn", "start"]

```

This way it won't have to install all the dependencies if it is just the source files that are changing.

KaelWD:
ah k, it wasn't any slower than the default node deploy so I didn't bother seeing if it could be improved. 

This would only work for the initial `yarn install` stage though, and would require a bit of maintenance to make sure all the packages and postinstall hooks are copied correctly what with the whole monorepo thing

PR #7979, opened 2019-07-24
chore: throw if vuetify is not bootstrapped properly

jacekkarczmarczyk:
image

I've tried `consoleError` as well but imho the message wasn't visible enough

## How Has This Been Tested?
replace `vuetify` with `v: vuetify` in dev/index.js

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VApp/VApp.ts
+++ b/packages/vuetify/src/components/VApp/VApp.ts
@@ -34,6 +34,12 @@ export default mixins(
     },
   },
 
+  beforeCreate () {
+    if (!this.$vuetify || (this.$vuetify === this.$root)) {
+      throw new Error('Vuetify is not properly initialized, see https://vuetifyjs.com/getting-started/quick-start#bootstrapping-the-vuetify-object')
+    }
+  },
+
   render (h) {
     const wrapper = h('div', { staticClass: 'v-application--wrap' }, this.$slots.default)
 


packages/vuetify/src/components/VApp/VApp.ts
```diff
@@ -34,6 +34,12 @@ export default mixins(
     },
   },
 
+  beforeCreate () {
+    if (process.env.NODE_ENV === 'development' && (!this.$vuetify || this.$vuetify === this.$root)) {
```
KaelWD: > `this.$vuetify === this.$root`

When does that happen?

jacekkarczmarczyk: When vuetify is not bootstrapped properly. 

*dev/index.js*

```js
import Vue from 'vue'
import App from './App'
import Vuetify from 'vuetify'
// import * as locales from '../src/locale'
import router from './router'
import '@mdi/font/css/materialdesignicons.css'

Vue.config.performance = true

Vue.use(Vuetify)

// const vuetify = new Vuetify({
//   lang: {
//     locales,
//   },
// })

const vm = new Vue({
  data: () => ({ isLoaded: document.readyState === 'complete' }),
  router,
  render (h) {
    return this.isLoaded ? h(App) : undefined
  },
}).$mount('#app')

// Prevent layout jump while waiting for styles
vm.isLoaded || window.addEventListener('load', () => {
  vm.isLoaded = true
})
```

*VApp.ts*
```ts
  render (h) {
    console.log(this.$vuetify === this.$root)
```

image

Don't ask me why. Although maybe it shouldn't happen and there's another issue with it

KaelWD: Yeah that looks like a bug. This check could be in src/install.ts:49 with the logic there fixed to return undefined

PR #7984, opened 2019-07-24
chore: restrict vuetify and other libs version to current major

jacekkarczmarczyk:
## Description
replaces `npm/vuetify/dist` with `npm/vuetify@^2/dist` to avoid breaking apps after releasing next major

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).
- [x] My code follows the code style of this project.
- [ ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)
- [ ] I've added new examples to the kitchen (applies to new features and breaking changes in core library)

PR #8002, opened 2019-07-25
chore: restrict vuetify and other libs version to current major

jacekkarczmarczyk:
see https://github.com/vuetifyjs/vuetify/pull/7984

This one is for `stable` branch

### DIFF
--- a/packages/docs/src/components/doc/Codepen.vue
+++ b/packages/docs/src/components/doc/Codepen.vue
@@ -23,7 +23,7 @@
 
   const jsResources = [
     'https://cdn.jsdelivr.net/npm/babel-polyfill/dist/polyfill.min.js',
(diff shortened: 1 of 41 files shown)


packages/docs/src/components/doc/Codepen.vue
```diff
@@ -1,135 +1,135 @@
...
-    },
-
-    methods: {
-      submit () {
-        this.$el.submit()
-      }
-    }
-  }
-</script>
+<template lang="pug">
```
KaelWD: Did you change line endings or something?

jacekkarczmarczyk: not deliberately but yes, it was crlf before

btw
image

PR #8024, opened 2019-07-25
fix(VBtnToggle ): fix borders between buttons

antfu:
## Description
Fix VButtonsGroup borders between buttons (from two pixel wide to one pixel wide)

## How Has This Been Tested?

None

## Compares

Before
image

After
image

## Types of changes

## Checklist:

- [x] I've added new examples to the kitchen (applies to new features and breaking changes in core library)

### DIFF
--- a/packages/vuetify/src/components/VBtnToggle/VBtnToggle.sass
+++ b/packages/vuetify/src/components/VBtnToggle/VBtnToggle.sass
@@ -52,6 +52,9 @@
     &:after
       display: none
 
+    &:not(:first-child)
+      margin-left: -1px
+
 // Modifiers
 .v-btn-toggle--rounded
   border-radius: $btn-toggle-round-radius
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VBtnToggle/VBtnToggle.sass
```diff
@@ -52,6 +52,9 @@
     &:after
       display: none
 
+    &:not(:first-child)
+      margin-left: -1px
```
johnleider:
```suggestion
      border-left-width: 0
```

jacekkarczmarczyk: border-width doesn't solve the problem
image

antfu: @johnleider After some tests, I think border-width is the better approach. Due to the `border-width: thin` is not determined.

Change has updated.

PR #8041, opened 2019-07-26
feat(VStepper): center v-stepper__label in 'alt-labels' mode

nionoku:
# Centring v-stepper__label in 'alt-labels' mode

## Description
I propose to leave only the beginning of the title, the end to hide behind three points and show the full title when you hover on the label.

## Motivation and Context
Long headers are difficult to fit into 1 line (especially in Russian, hard to make a short headline in Russian). 

## How Has This Been Tested?
Visually, look at the result before:
https://i.imgur.com/AeToGHV.png
After:
https://i.imgur.com/4CaX1FN.png
And after with show title hint:
https://i.imgur.com/PS3JV3R.png

## Markup:
<details>

```vue
<template>
  <v-stepper v-model="stepper.step" alt-labels>
    <v-stepper-header>
      <template v-for="(item, index) in stepper.headers">
        <v-stepper-step
          :key="index + 1"
          :complete="stepper.step > index + 1"
          :step="index + 1"
        >{{ item }}</v-stepper-step>
        <v-divider 
          v-if="index < stepper.headers.length - 1"
          :key="stepper.headers.length + index + 1"/>
      </template>
    </v-stepper-header>
    <v-stepper-items>
      <v-stepper-content step="1">
        it\'s work
      </v-stepper-content>
      <v-stepper-content step="2">
        fantastic!
      </v-stepper-content>
    </v-stepper-items>
  </v-stepper>
</template>

<script lang="ts">
import Vue from 'vue'
export default Vue.extend({
  data() {
    return {
      stepper: {
        step: 0,
        headers: [
          'it\'s work adspawjfjaw pjfp jawpfj pawjf piawjf pawf jpaw',
          'fantastic! awfiawjfpafjawjfpawj opawjf pawjfp ajwpf jawpf jp'
        ]
      }
    }
  }
})
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VStepper/VStepper.sass
```diff
@@ -192,6 +192,13 @@
         margin: $stepper-alt-labels-header-divider
         align-self: flex-start
 
+    .v-stepper__label
+      max-width: 10vw
```
johnleider: Why this value?

packages/vuetify/src/components/VStepper/VStepperStep.ts
```diff
@@ -104,6 +104,7 @@ export default baseMixins.extend<options>().extend({
     genLabel () {
       return this.$createElement('div', {
         staticClass: 'v-stepper__label',
+        attrs: { title: this.$slots.default && this.$slots.default.length > 0 ? this.$slots.default[0].text : null },
```
johnleider: We can pass attrs here, but no explicit title declaration. We should instead opt for a11y support. @phiter any feedback?

phiter: I think it is valid to use title since it works for visually enabled people to check the full contents on hover.

`aria-label` would only work for screen readers, but we don't need that since even though the text is visually hidden, it is readable by the screen reader.

packages/vuetify/src/components/VStepper/VStepper.sass
nionoku: To avoid collision of labels. When using the value `10vw`, saved the correct tab size

jacekkarczmarczyk: I don't know if this is the correct solution, but if it is the value should be probably extracted to the sass var

nionoku: @jacekkarczmarczyk 
I understand that this is not a 100% correct solution, but I did not come up with how to do it more correctly =)
I extract the value as a variable and leave the decision to reviewers

packages/vuetify/src/components/VStepper/VStepperStep.ts
johnleider: I suppose the larger concern is assuming title without also passing in the props. I'd opt to pass the title attribute to it for now and let that be part of another PR since the scope of this is to fix the label.

phiter: On that case I agree. The user should be able to pass the title. This change will always add a title to the labels even if they're small enough not to be collapsed.

johnleider:
At this time the team is not considering any feature requests for the v2.x Major. If you are interested in working with the team on the v3 implementation, please reach out to us in the [Discord community](https://community.vuetifyjs.com).

PR #8143, opened 2019-07-29
feat(VCard): add hover effect when focusing

cawa-93:
## Description
When navigating using the keyboard, the v-card:focus has no visual confirmation. It is necessary to add a similar visual elevation as when hovering the cursor

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-card hover to="/">
      <v-img :aspect-ratio="225/314" src="https://fakeimg.pl/225x314/"></v-img>
    </v-card>
  </v-container>
</template>
```
</details>

Demo with problem: https://codepen.io/cawa-93/pen/gVBEmP?editors=1000
Demo with suggested changes: https://codepen.io/cawa-93/pen/xvyBdm?editors=1100

## Types of changes

## Checklist:


nekosaur:
Missing playground markup @cawa-93

cawa-93:
@nekosaur I don't really understand what you mean.  I do not change the standard of behavior in any way and do not add new behavior.  I just suggest complementing your existing behavior.  To take the example, the focus and focus are identical.  I think cards should be done the same way.

jacekkarczmarczyk:
He just asked you to provide a sample code where we could compare the behaviour before and after PR, see the `Markup` section in other PRs, for example <https://github.com/vuetifyjs/vuetify/pull/8409>

cawa-93:
Oh. Ok. I will make this later

cawa-93:
@nekosaur I added playground markup and codepen demos

johnleider:
Why only apply when using the hover prop? The same issue would apply for regular cards as well.

cawa-93:
I think the focus and the hover are the same. And they should behave the same way (with some very rare exceptions). If there is no hover effect on the card, then this effect is not required when focusing.

johnleider:
Remove the **hover** prop and try to focus those cards. The first 2 aren't noticed at all.

Issue #7102, opened 2019-04-29
[Bug Report] VStepper & VTabs transitions between steps are sometimes broken.


MatthewAry:
### Versions and Environment
**Vuetify:** 1.5.13+ & 2.0.0-beta.9
**Vue:** 2.6.10
**Browsers:** Google Chrome
**OS:** Windows, Mac OSX

### Steps to reproduce
Go to Codepen.
Click the continue button and observe transition.
Click the continue button and observe the transition again.

### Expected Behavior
The previous step should exit left while the next step should enter right using a sliding transition.

### Actual Behavior
The previous step exits left while the next step appears. This issue can be seen on the first step, however the second step seems to not exhibit the transition issue.

### Reproduction Link
<a href="https://codepen.io/MatthewAry/pen/rbbzJz" target="_blank">https://codepen.io/MatthewAry/pen/rbbzJz</a>

### Other comments
I have been able to reproduce this on MacOS and Windows on Google Chrome 74. Assuming that none of my Google Chrome plugins are the cause, this issue should be reproducible.

Video:
[https://drive.google.com/open?id=1Mr8vgQXi-kWx38vbhAOzeunIMBZLcwQM](https://drive.google.com/open?id=1Mr8vgQXi-kWx38vbhAOzeunIMBZLcwQM)

EDIT:
This same behavior has been observed with vtabs

EDIT: Similar issues are observable on Vuetify 2 Beta: [Video](https://drive.google.com/open?id=1GOKF1EoVjoHxeUF7YUqo-RB9E979dF8G)

sean-kennedy:
We noticed this yesterday after my Chrome updated to 74. We tested on a computer running Chrome 73 where it worked as expected then after only updating Chrome to 74 it started the same behaviour as above.

CesarGomezTissini:
We noticed this sometime ago too but never was able to reproduce it outside our project. We take decision to avoid autofocus when go to next step. At least in Safari works.

MatthewAry:
I have observed this behavior with V-Tabs as well.

CesarGomezTissini:
> I have observed this behavior with `v-tabs` as well.

I noticed now this issue is related to `window` itself.

CesarGomezTissini:
We noticed now this issue happened much more in Safari / Chrome in iOS. When i say "much more", I mean "easier to reproduce".

MatthewAry:
Likely related to #7088

MatthewAry:
[...]

Issue #6962, opened 2019-04-10
[Bug Report] Tab height stuck for tab-item with lazy, transition and async content


jongeyer:
### Versions and Environment
**Vuetify:** 1.5.10
**Last working version:** 1.2.10
**Vue:** 2.6.10
**Browsers:** Chrome 71.0.3578.80
**OS:** Windows 10

### Steps to reproduce
Create tab-item with combination of `lazy` and `transition` props, where tab's content is loaded asynchronously.

### Expected Behavior
Tab height expands to fit content

### Actual Behavior
Tab height remains fixed with an inline style, hiding async content

### Reproduction Link
<a href="https://codesandbox.io/s/m45l18nw3y" target="_blank">https://codesandbox.io/s/m45l18nw3y</a>

### Other comments
See codesandbox link... problem only occurs when all three conditions are met (lazy, transition, async content).

Workaround was to add style below, but not sure if this will create problems inside a v-dialog

    .v-tabs .v-window__container {
      height: auto !important;
    }

KaelWD:
This might be a bug in vue, onAfterEnter isn't called on the first two tabs for some reason.

jongeyer:
Looks like this might be related to: https://github.com/vuetifyjs/vuetify/issues/5363

Hopefully fixed with: https://github.com/vuetifyjs/vuetify/pull/6924

KaelWD:
Your `custom-transition` doesn't actually have any css, so vue can't know when it ends. If you just want to disable the transition you should use `false`: https://codesandbox.io/s/r0kxr6x3nn

jongeyer:
Right, so in my application I'm using `fade-transition` and that's how I discovered the problem. 

I created a new codesandbox (https://codesandbox.io/s/o403om35ky) and did some more testing.
Apparently the problem is not present for all of the built-in transitions only `fade-transition` and `expand-transition`.

Haven't yet tested a real `custom-transition` -- I'll try to see if that works or not.
And of course, without the transition, there's no problem.

katjang:
this seems to be related to a problem i'm having, some transitions dont emit events. https://codepen.io/anon/pen/vMJBVw

EpicFailv2:
My issue: 
![image](https://user-images.githubusercontent.com/32414344/58333089-5bd46480-7e45-11e9-997d-d5bec6bf9e24.png)

Issue consistently repeatable when rapidly switching between tabs; when second transition is called while first one is still going.

[...]

PR #8205, opened 2019-07-31
fix(#6962): Fix race conditions in VWindowItem.ts during transition.

chouhanyang:
From time to time, the transition state of the VWindow container is
not finished and the height will be fixed. This change fixed this
problem.

FINDINGS:

- If transitionend fired before nextTick(), then done() is never
  going to be invoked.
- onEnter() itself may be re-entered during transiton. Then the
  state is messed up.
- onBeforeLeave() may be called after cancel, then the height
  is stuck.
- onBeforeEnter() may be re-entered after first call. Calling
  this multiple times also wrongly reset transition state.

SOLUTION:

- Put correct guards on each event handler. Check of isActive and
  wasCancelled accordingly for each event to prevent re-entrant.
- Put a timer in onEnter() to ensure done() is always invoked and
  deactivate() after timeout if anything goes wrong.

TESTS:

- Add log instrumentations for observation.
- Test quickly switching tabs and see container always deactivate.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

- [x] I've added new examples to the kitchen (applies to new features and breaking changes in core library)

### DIFF
--- a/packages/vuetify/src/components/VItemGroup/VItemGroup.ts
+++ b/packages/vuetify/src/components/VItemGroup/VItemGroup.ts
@@ -17,7 +17,6 @@ export type GroupableInstance = InstanceType<typeof Groupable> & {
   id?: string
   to?: any
   value?: any
-  beforeChange?: (v: boolean) => void // Callback before setting isActive
  }
 
 export const BaseItemGroup = mixins(
@@ -169,12 +168,7 @@ export const BaseItemGroup = mixins(
     updateItem (item: GroupableInstance, index: number) {
       const value = this.getValue(item, index)
 
-      var val = this.toggleMethod(value)
-      // Trigger callback only if value changed.
-      if (val !== item.isActive && item.beforeChange) {
-        item.beforeChange(val)
-      }
-      item.isActive = val
+      item.isActive = this.toggleMethod(value)
     },
     updateItemsState () {
       if (this.mandatory &&
--- a/packages/vuetify/src/components/VItemGroup/VItemGroup.ts
+++ b/packages/vuetify/src/components/VItemGroup/VItemGroup.ts
@@ -17,6 +17,7 @@ export type GroupableInstance = InstanceType<typeof Groupable> & {
   id?: string
   to?: any
   value?: any
+  beforeChange?: (v: boolean) => void // Callback before setting isActive
  }
 
 export const BaseItemGroup = mixins(
@@ -168,7 +169,12 @@ export const BaseItemGroup = mixins(
     updateItem (item: GroupableInstance, index: number) {
       const value = this.getValue(item, index)
 
-      item.isActive = this.toggleMethod(value)
+      var val = this.toggleMethod(value)
+      // Trigger callback only if value changed.
+      if (val !== item.isActive && item.beforeChange) {
+        item.beforeChange(val)
+      }
+      item.isActive = val
     },
     updateItemsState () {
       if (this.mandatory &&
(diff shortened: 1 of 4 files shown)


chouhanyang:
Hmm... I think I have fixed all broken tests. However, the "radioOn" breakage was caused by previous commits. How can I solve this? Any pointer?

chouhanyang:
Never mind, all checks passed. Please review my code. Thank you!

chouhanyang:
Weird. Travis CI job passed. But github shows pending. I'll force update again to see if I can trigger the check to run again.

johnleider:
This breaks the tab content height transition.

**Expected:**
expected

**Actual:**
actual

chouhanyang:
> This breaks the tab content height transition.
> 

Hi John,
Thanks for pointing that out. Now I understand what the business of setting clientHeight is. I have investigated more and have tried several different approach to fix this issue:

1. I tried to put back onBeforeLeave(), but this is triggered on different item. And it may be fired before or after onBeforeEnter(), which depends on if we move forward the tabs or backward. This is bad because in the case of switch tabs continuously, onBeforeLeave() may fire very late, which cause the height to stuck.

2. I also tried to set initial height within onBeforeEnter() where we kick off the transition process. I use the current height of the parent element as initial height. However, it works only if we walk backward of the tabs. When walking forward the tabs, parent element collapse to 0px height. The root cause may be similar to 1. Where I tried to fix the order where we set isActive in VItemGroup.ts, but no use. Vue just fired with its own order.

3. Finally, I have to add an beforeChange() from VItemGroup.ts. So in this callback, we can make sure no transition events are fired. Then we can obtain current height correctly and start transition height inside onBeforeEnter(). 

This change is a bit more involved than I originally though. Appreciate if you can take another look and let me know if this callback is something that fits the coding standard.

johnleider:
For the most part everything looks good. I did uncover an issue with the playground of `v-tabs` if you could take a look.

tabs

chouhanyang:
John,

Thanks for taking another look. Just to clarify, the problem you mentioned here is that the content breaks out of its container transiently, right?

If so, this is related to another issue #7102 where I find the problem is because the active state terminated early due to numbers of reasons.

But my change could make this issue worse because the change preemptively remove the active state while cancel. I'll see if I can also fix #7102 by allowing multiple transition tabs running at the same time.

chouhanyang:
Good news!

The reason why onAfterEnter() was skipped is simply it was cancelled. With the new change we allow 2 windows to have their own transition states individually. And then, we can use a counter activeWindows as a semiphore to capture entire transition duration with multiple cancellation in between.

- Use activeWindows to deduce if the container should be in active state.
- Compute isActive instead.
- Got rid of transitionEnd() event. 
- Let cancel go to the same path as after enter/leave path.
- Got rid of the done() callback related code and the timer in previous change.
- Got rid of all async calls requestAnimationFrame() since it's no longer necessary.

So basically as long as we handle the following 4 paths correctly. We don't need any hacks.

1. Normal enter: onBeforeEnter() .... onAfterEnter().
2. Cancel enter: onBeforeEnter() ... onEnterCancelled().
3. Normal leave: onBeforeLeave() ... onAfterLeave().
4. Cancel leave: onBeforeLeave() ... onLeaveCancelled().

Every handler is now synchronous. I'm pretty happy with the new change since it is a lot simpler than previous attempts. I'm doing more tests in the mean time. Let me know what you think.

chouhanyang:
Since now we handle both enter and leave paths. We can detect the first onBeforeXXX() event easily. So we can get rid of the beforeChange() event and the trigger from VItemGroup.ts. It's much cleaner now.

chouhanyang:
@johnleider 

Thanks John. I think you can close issue #7102 as well. The root causes are the same.

PR #8293, opened 2019-08-04
feat(VTextField): hide affixes when neither focused nor dirty

### LINKED ISSUE #7872: [Feature Request] Have VTextField prefix and suffix text only appear when the input is focused or dirty
### Problem to solve
The md2 spec indicates that prefix and suffix text should only display when the text-field input is dirty or has the focus. This is not the current VTextField behavior. This may be a bug rather than a feature request, but it could be reasoned either way.

### Proposed solution
Change the behavior of the VTextField having prefix or suffix text so that the text is only visible as indicated in the md2 spec. The codepen has my suggested css already included so the effect can be observed for the various text-field styles.

Alternatively, add a prop to VTextField so that the user can choose to display prefix/suffix text on focus/dirty or display them persistently like in the current implementation.

https://material.io/design/components/text-fields.html

prefix

suffix

sh7dm:
## Description
Hide prefix/suffix when text field isn't focused or dirty.
- [x] Add a prop to force display affixes
- [x] Add animations
- [x] Fix VSelect etc

## Motivation and Context
Fixes #7872

## How Has This Been Tested?
Markup
- [x] `jest`

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-text-field prefix="$" suffix="%"></v-text-field>
  </v-container>
</template>

<script>
export default {
  data: () => ({
  }),
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -73,6 +73,14 @@ export default baseMixins.extend<options>().extend({
     chips: Boolean,
     clearable: Boolean,
     deletableChips: Boolean,
+    displayPrefix: {
+      type: Boolean,
+      default: true,
+    },
+    displaySuffix: {
+      type: Boolean,
+      default: true,
+    },
     dense: Boolean,
     eager: Boolean,
     hideSelected: Boolean,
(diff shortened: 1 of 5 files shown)


johnleider:
Pulse check.

185driver:
I wonder if this is close to being completed. It seems so, but maybe not.

johnleider:
This won't make it into v2.x branch unless this is poised as a fix. Otherwise it will have to wait for v3 and should be closed.. I'll leave that up to you @sh7dm .

PR #8491, opened 2019-08-13
fix(VDialog): inherit listeners

### LINKED ISSUE #3358: [Bug Report] v-dialog only captures keydown events
### Versions and Environment
**Vuetify:** 1.0.3
**Vue:** 2.5.13
**Browsers:** Chrome 64.0.3282.167
**OS:** Mac OS 10.12.4

### Steps to reproduce
Create v-dialog and use @keyup.enter or any other keyup event on that dialog.

### Expected Behavior
It should handle the keyup events

### Actual Behavior
it's only handling the down events. 

### Reproduction Link
[https://codepen.io/ianstallings/pen/aqGerE?editors=1010](https://codepen.io/ianstallings/pen/aqGerE?editors=1010)

### Other comments
In VDialog.js it's lines 117-125. I'm honestly not sure if this is intentional or not. I didn't want to issue a patch/PR without asking first.

avengerweb:
fix #3358

## Description
We don`t inherit any listeners on VDialog, only emit keydown event.

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <v-layout row justify-center>

    <v-dialog v-model="dialog" @keydown.enter="dialog = false">
        <v-btn v-on="on">Open Dialog</v-btn>
      </template>

      <v-card >

        <v-card-title>Use Google's location service?</v-card-title>

        <v-card-text>Let Google help apps determine location. This means sending anonymous location data to Google, even when no apps are running.</v-card-text>
        <v-card-actions>

          <v-btn class="green--text darken-1" @click.native="dialog = false">Disagree</v-btn>

          <v-btn class="green--text darken-1" @click.native="dialog = false">Agree</v-btn>
        </v-card-actions>
      </v-card>
    </v-dialog>
  </v-layout>
</template>

<script>
export default {
  data: () => ({
    dialog: false,
  })
}
</script>

```
</details>

## Types of changes

## Checklist:


husayt:
@johnleider This is causing problems if I am using `:value/@input` instead of `v-model`
Before this used to work just fine, as I was only receiving `false` on esc/clickOutside.
```
 <VDialog
            :value="dialogType !== EDialogType.NONE"
            @input="dialogType = EDialogType.NONE"
          >
```

But now `@input` receives also InputEvents from texteditors inside. The reason it is not broken for most is because they would use `v-model` which  would interpret InputEvent as true and thus Dialog will not be closed. BUt in my case keydown inside an editor in dialog closes dialog. So currently this breaks dialogs all over for me.

atwhiteley:
This update has also broken dialogs for us with text fields in them, even with v-model use. We have watch functions on the v-model that opens/closes dialogs, and those watch functions previously would only fire when the dialog opens or closes. Now they are fired on every single key stroke, which breaks every dialog with text fields in them.

johnleider:
This is being reverted due to regressions caused with inputs inside dialogs.

PR #8790, opened 2019-08-29
feat(VTextField): add customizable counter value

### LINKED ISSUE #8788: [Feature Request] Ability to provide a custom count function for the counter in the v-text-field
### Problem to solve
We have some `v-text-field` that holds text with variables like `this is my text with a ${variable}`.
Then we have multiple values that can placed into this variable and we choose the top length. So there is in example a value `hello world example`. It is 19 characters long + the 23 characters from the input text. So now we allow only a maximum of 30 characters, so the counter should show `42 / 30` in red.

### Proposed solution
This could be solved with a new optional `prop` that takes the current `input` of the `v-text-field` and returns a `number`. We could name it `counter-length` or `length` or `input-length` or something similar.

So we use `<v-text-field v-model="line" :counter="30" :counter-length="myCustomCounter">`
```ts
myCustomCounter(value: string): number {
  // implementation
}
```

---

It would also be nice if the `:counter-length` could be an `async function`, because we calculate the length via backend / api.

Shinigami92:
Please see #8802


Shinigami92:
To allow Promise also, we could change

src/components/VTextField/\_\_tests\_\_/VTextField.spec.ts
```ts
it('should use counter length function', async () => {
  // ...

  wrapper.setProps({
    counterLength: async (value?: string): Promise<number> => {
      return await Promise.resolve((value || '').replace(/foo/g, '').length)
    },
  })

  expect(counter.element.innerHTML).toBe('8 / 50')
}
```

and

src/components/VCounter/VCounter.ts line 19
```diff
- type: [Number, String],
+ type: [Number, String, Promise],
```

But unfortunately I do not know how to unwrap the maybe-promise in the render call.
Maybe someone can help me?

packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -55,6 +55,7 @@ export default baseMixins.extend<options>().extend({
       default: '$vuetify.icons.clear',
     },
     counter: [Boolean, Number, String],
```
myleslee: Simply add `Function` to the list instead of creating a new prop?
`counter: [Boolean, Number, String, Function],`

Shinigami92: Attention, the `counter` property in `VTextField` is the value of the `maxLength` not the value itself!
I want to name the `counterLength` `counterValue`, but there is already the `counterValue` method and I didn't want to conflict with it.
If someone in the vuetify core team would accept to use `counterValue` instead of `counterLength`, I can rewrite the PR if this is wanted

MajesticPotatoe: Would say name it something like `counterText`.

MajesticPotatoe:
note: this should be rebased to `dev`

packages/vuetify/src/components/VTextField/VTextField.ts
jacekkarczmarczyk: `counterText` suggests a bit more than just a length:

`${counterValue} / 30` vs `${counterText}`

Maybe slot would be a good idea? You could add literally anything, including colors, icons etc. More powerful but also more to typing

```html
<v-text-field #counter="{ value, max }">{{ someFun(value) }} / {{ max }}</v-text-field>
```

```html
<v-text-field>
  <template #counter="{ value, max }">
    <v-icon class="red">foo</v-icon> 
    {{ value < 10 : 'too short' : 'too long'}}, max: {{ max }}</span>
  </template>
</v-text-field>
```

Async would need an additional variable in data, but i don't think it's a frequent case.

MajesticPotatoe: ah, I was under the impression it was replacing the whole thing not just the starting value, in which what would be the point of making it a function (you could just bind a string or number). 
Maybe a slot would be more valuable, providing the most flexibility.

Shinigami92:
> note: this should be rebased to `dev`

How can I change the target branch to `vuetifyjs:dev`?

packages/vuetify/src/components/VTextField/VTextField.ts
Shinigami92: I **try** not to make any breaking changes

Shinigami92:
I don't know if I made anything wrong with the 'git rebase dev' command 🤔 
Also I don't know why there are 172 files different to `vuetify:dev`
Can someone help me? Alternative could be for me to close this mess-up branch and open it clean again

MajesticPotatoe:
probably better doing a rebase locally

packages/vuetify/src/components/VTextField/VTextField.ts
jacekkarczmarczyk: It wouldn't be a breaking change

Shinigami92:
> probably better doing a rebase locally

I'm trying, but I think I messed something up 🤷‍♂️
I'm not very well into `rebasing`
So I think it's easier to copy my changes into a new branch, sorry

PR #8802, opened 2019-08-30
feat(VTextField): add customizable counter value

### LINKED ISSUE #8788: [Feature Request] Ability to provide a custom count function for the counter in the v-text-field

Shinigami92:
## Description

added customizable counter value via optional function

_I messed up a previous branch #8790, this is the same but recreated_

## Motivation and Context

closes #8788 

## How Has This Been Tested?

unit test

## Markup:

<details>

```vue
<template>
  <v-text-field
    counter
    :counter-length="(value) => (value || '').replace(/\s/g).length"
  ></v-data-table>
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -55,6 +55,7 @@ export default baseMixins.extend<options>().extend({
       default: '$clear',
     },
     counter: [Boolean, Number, String],
+    counterValue: Function,
     filled: Boolean,
     flat: Boolean,
     fullWidth: Boolean,
@@ -107,7 +108,10 @@ export default baseMixins.extend<options>().extend({
         'v-text-field--shaped': this.shaped,
       }
     },
-    counterValue (): number {
+    computedCounterValue (): number {
+      if (typeof this.counterValue === 'function') {
+        return this.counterValue(this.internalValue)
+      }
       return (this.internalValue || '').toString().length
     },
     internalValue: {
@@ -296,7 +300,7 @@ export default baseMixins.extend<options>().extend({
           dark: this.dark,
           light: this.light,
           max,
-          value: this.counterValue,
+          value: this.computedCounterValue,
         },
       })
     },
--- a/packages/vuetify/src/components/VCombobox/VCombobox.ts
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.ts
@@ -31,7 +31,7 @@ export default VAutocomplete.extend({
   }),
 
   computed: {
-    counterValue (): number {
+    computedCounterValue (): number {
       return this.multiple
         ? this.selectedItems.length
         : (this.internalSearch || '').toString().length
--- a/packages/vuetify/src/components/VFileInput/VFileInput.ts
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.ts
@@ -79,7 +79,7 @@ export default VTextField.extend({
         'v-file-input': true,
       }
     },
-    counterValue (): string {
+    computedCounterValue (): string {
       const fileCount = (this.isMultiple && this.lazyValue)
         ? this.lazyValue.length
         : (this.lazyValue instanceof File) ? 1 : 0
@@ -203,7 +203,7 @@ export default VTextField.extend({
       const length = this.text.length
 
       if (length < 2) return this.text
-      if (this.showSize && !this.counter) return [this.counterValue]
+      if (this.showSize && !this.counter) return [this.computedCounterValue]
       return [this.$vuetify.lang.t(this.counterString, length)]
     },
     genSelections () {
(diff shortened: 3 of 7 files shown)


Shinigami92:
Now this will merge into `vuetify:dev` instead of `vuetify:master`. Sorry for inconvenience.

packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -55,6 +55,7 @@ export default baseMixins.extend<options>().extend({
       default: '$vuetify.icons.clear',
     },
     counter: [Boolean, Number, String],
```
Shinigami92: Continue discussion to `counterText`:
@myleslee, @MajesticPotatoe, @jacekkarczmarczyk 

`${counterValue} / 30` vs `${counterText}``

> Maybe slot would be a good idea? You could add literally anything, including colors, icons etc. More powerful but also more to typing
> 
> ```
> <v-text-field #counter="{ value, max }">{{ someFun(value) }} / {{ max }}</v-text-field>
> ```
> 
> ```
> <v-text-field>
>   <template #counter="{ value, max }">
>     <v-icon class="red">foo</v-icon> 
>     {{ value < 10 : 'too short' : 'too long'}}, max: {{ max }}</span>
>   </template>
> </v-text-field>
> ```
> 
> Async would need an additional variable in data, but i don't think it's a frequent case.

Correct me if I'm wrong, but would a `slot` not mean a breaking change?

Shinigami92: The more I look at it the `slot` solution, the more uncomfortable I feel with it
It could hurt material design guidelines and is also thought to be slightly beyond the feature request 🤔

Shinigami92:
I need an `internalValue` in `VCounter` to support `async` / `Promise`.
I tried something like the following to prepare updatable content

`VCounter.ts`
```ts
export default mixins(Themeable).extend({
  name: 'v-counter',

  functional: true,

  props: {
    value: {
      type: [Number, String],
      default: '',
    },
    max: [Number, String],
  },

  data: () => ({
    content: '',
    isGreater: false,
  }),

  render (h, ctx): VNode {
    const { props } = ctx
    const max = parseInt(props.max, 10)
    const value = parseInt(props.value, 10)
    this.content = max ? `${value} / ${max}` : String(props.value)
//      ~ this is null
    this.isGreater = !!max && (value > max)

    return h('div', {
      staticClass: 'v-counter',
      class: {
        'error--text': this.isGreater,
        ...functionalThemeClasses(ctx),
      },
    }, this.content)
  },
})
```

But this isn't working and I think it's because it is a functional component?! I'm not very into functional components.

I want to create a method `updateContent`, this will call what ever is needed to update the content.
When the content is then updated, the render function will be triggered to rerender the state.

But unfortunately I currently don't know how to do this 🤔

jacekkarczmarczyk:
async calculation of text length seems a very edge case and personally I wouldn't implement it in the framework

Shinigami92:
> async calculation of text length seems a very edge case and personally I wouldn't implement it in the framework

In this case, my changes can be merged after a review.
What do I else need to update? Docs?

Anything else I should change? Code-Style?

packages/vuetify/src/components/VTextField/VTextField.ts
MajesticPotatoe: slot would simply be a feature/enhancement. it wouldn't be breaking unless it actually broke existing functionality. In this case its just providing additional.

Shinigami92:
I renamed the computed function `counterValue` to `computedCounterValue` so that I can name the new property `counterValue` instead of `counterLength`.
Since the `computedCounterValue` is only used internally, I think this will not be a breaking change except to someone who uses vuetify heavily and rely on this.

packages/vuetify/src/components/VTextField/VTextField.ts
Shinigami92: IMO `counter-slot` could be another feature request and be independent of this feature

Shinigami92:
I wanted to update this PR, but I see that many related tests are currently being skipped
https://github.com/vuetifyjs/vuetify/blob/b222c90/packages/vuetify/src/components/VTextField/__tests__/VTextField.spec.ts

I'm also a bit confused why the file in `master` (a870dfa 18 days ago) is more recent than the one in `dev` (b222c90 27 days ago)

So I'll wait and come back later

Shinigami92:
🎉

PR #8838, opened 2019-09-03
fix(VCarousel): Always respect original internal height of VWindow.

chouhanyang:
## Description

Add an intermediate height to overwrite component internal height if internalHeight is undefined.
This change makes VWindow universally respect internalHeight if set without special flags.

## Motivation and Context

Improve VWindow.

## How Has This Been Tested?

Test on v-tabs and v-carousel docs.

## Types of changes

### DIFF
--- a/packages/vuetify/src/components/VCarousel/VCarousel.ts
+++ b/packages/vuetify/src/components/VCarousel/VCarousel.ts
@@ -64,7 +64,6 @@ export default VWindow.extend({
   data () {
     return {
       internalHeight: this.height,
-      noHeightReset: true,
       slideTimeout: undefined as number | undefined,
     }
   },
--- a/packages/vuetify/src/components/VCarousel/VCarouselItem.ts
+++ b/packages/vuetify/src/components/VCarousel/VCarouselItem.ts
@@ -49,9 +49,5 @@ export default baseMixins.extend({
 
       return this.$createElement(tag, data, this.showLazyContent(this.genDefaultSlot()))
     },
-    onBeforeEnter () { /* noop */ },
-    onEnter () { /* noop */ },
-    onAfterEnter () { /* noop */ },
-    onEnterCancelled () { /* noop */ },
   },
 })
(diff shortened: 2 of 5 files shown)


chouhanyang:
@johnleider 

With one extra variable, I think it is better to make VWindow universally restore original internalHeight after transition. This is less hacky and less error prone in the future I believe. Feel free to discard this change if you have any other concerns.

johnleider:
Please post an example of what this resolves.

chouhanyang:
@johnleider This fix is basically what you did in v2.0.12 (https://github.com/vuetifyjs/vuetify/commit/2246337) with some extra benefits and cleanup.

### VCarousel w/o 2246337

https://codepen.io/chouhanyang/pen/NWKXmRE

### With this fix

https://codepen.io/chouhanyang/pen/BaBJgEY

### Descriptions

The main goals and benefits for this change are:

1. Decoupling the business of transition from VCarouselItem.ts. At the end of this file, it disabled all event handlers for enter path, which is now invalid because function names are change. I would like to make VCarouselItem not mess around the transition logic.

```javascript
    onBeforeEnter () { /* noop */ }, // invalid now
    onEnter () { /* noop */ }, // dangerous
    onAfterEnter () { /* noop */ }, // invalid now
    onEnterCancelled () { /* noop */ }, // invalid now
```

2. Remove the extra flag from VCarouseItems.ts. Once internalHeight is fixed, VWindow will respect that without setting another flag. Which is less error prone.
3. Avoid side effect of VWindow transition. After the VWindow transition business is done. Everything should be restored as it was before transition. Having a flag that leaves transition height untouched is difficult to debug in the future.
4. Keep the transition business simple. Since the transition itself is fairly complex. I'd suggest we keep it as simple as possible. So in the future, if we need to trace it for bugs, it will be very simple.

As I mentioned previously, you'll have better judgement if this change is necessary or not. I generally prefer to make sure I clean up the mess I created and make sure the code would be easy to debug in the future. Hope this explains my change. Thanks!

PR #8962, opened 2019-09-14
fix(VPagination.ts): fix init for lazy parent

### LINKED ISSUE #8935: [Bug Report] v-pagination inside v-expansion-panel-content does not respect total-visible prop
### Environment
**Vuetify Version:** 2.0.17
**Last working version:** 1.5.18
**Vue Version:** 2.6.10
**Browsers:** Chrome 76.0.3809.132
**OS:** Mac OS 10.14.6

### Steps to reproduce
See the code pen. Basically, a pagination component outside of the expansion panel renders with the appropriate number of buttons, but a pagination component inside the panel does not (number of buttons seems to equal value of length prop). Note: once you interact with the component total-visible prop value is respected and the number of visible buttons is reduced.

### Expected Behavior
Expect total-visible to be respected on initial render even inside a v-expansion-panel-content.

### Actual Behavior
Total-visible prop is not respected inside expansion panels until the user interacts with the pagination controls.

### Reproduction Link
<a href="https://codepen.io/mcculleydj/pen/QWLrJKN?&editable=true&editors=101#anon-signup" target="_blank">https://codepen.io/mcculleydj/pen/QWLrJKN?&editable=true&editors=101#anon-signup</a>

mcculleydj:
fix #8935

Not sure if you want to keep this comment from 2 years ago:
`// TODO: Change this (f75dee3a, cbdf7caa)`

Using `await $nextTick()` to deal with this feels very brittle. I don't think it is worse than the existing setTimeout and it does address the issue I opened, but this could probably be done differently as part of a larger refactor that deals with any component listening to the parent's width when that parent could be something like v-expansion-panel-content and provide a 0 width value to any initial life cycle hook in the child component.

## Description

- this.$el.parentElement.clientWidth is 0 when onResize is called in the init method
- this fix uses async / await to replace the existing setTimeout and alters the init method so that clientWidth is properly defined
- since setTimeout no longer appears in the component, this PR removes all calls to jest.runAllTimers() in the .spec file

## Motivation and Context
https://github.com/vuetifyjs/vuetify/issues/8935

## How Has This Been Tested?
Placed a v-pagination component inside a v-expansion-panel in Playground.vue. Also, used the code on the v-pagination docs to see if fixing this problem would lead to another.

Updated the existing tests because there is no longer a need to call jest.runAllTimers()

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-expansion-panels>
      <v-expansion-panel>
        <v-expansion-panel-header>My Panel</v-expansion-panel-header>
        <v-expansion-panel-content>
          <v-pagination v-model="page" :length="10" :total-visible="5"></v-pagination>
        </v-expansion-panel-content>
      </v-expansion-panel>
    </v-expansion-panels>
    <v-pagination v-model="page" :length="10" :total-visible="5"></v-pagination>
  </v-container>
</template>

<script>
export default {
  data: () => ({
    page: 1
  })
};
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VPagination/VPagination.ts
```diff
@@ -105,12 +105,11 @@ export default mixins(Colorable, Themeable).extend({
   },
 
   methods: {
-    init () {
+    async init () {
```
johnleider: we don't use async/await

packages/vuetify/src/components/VPagination/VPagination.ts
```diff
@@ -105,12 +105,11 @@ export default mixins(Colorable, Themeable).extend({
...
       this.selected = null
-
-      this.$nextTick(this.onResize)
-      // TODO: Change this (f75dee3a, cbdf7caa)
-      setTimeout(() => (this.selected = this.value), 100)
+      await this.$nextTick()
```
johnleider: we don't use async/await

mcculleydj:
I took another stab at this by calling onResize recursively until a non-zero width is detected. If this doesn't work for you (frankly it makes me nervous) I'm wondering if you can point me to somewhere else in the code that the issue of clientWidth 0 on a parent component is dealt with without the luxury. In my test the method went through 4 "ticks" before detecting a non-zero width.

I reverted the test file, since I have no real reason to touch the existing setTimeout to address my issue.

Also, can you educate me on what `this.selected` is doing and why this set timeout is necessary? I removed it during testing for this PR and as far as I can tell it isn't doing anything. I'm sure there is something I'm not seeing.

packages/vuetify/src/components/VPagination/VPagination.ts
mcculleydj: Ok. I'm pretty new to working on this and open source in general. Is there a reason why, just so I can keep it in mind for the future?

mcculleydj:
Doesn't look like the recursive solution of waiting on width to be non-zero is going to work for the tests, but I'd argue that the tests aren't catching the case where the component is rendered in a parent that transitions into existence the way v-expansion-panel-content does or master would not be passing. Happy to continue working on this, but I'm going to wait for replies to previous questions so I can better understand what the original thinking was behind init().

mcculleydj:
Expected value of -3 is hard coded into the test for maxButtons. I disagree that a property with the name maxButtons should ever have a negative value. The fact that the component and tests are written to support this is leading me to a larger PR than just addressing the expansion panel problem.

I've been thinking about this component a lot and I'm wondering why the minimum number of buttons isn't 3 by design. For this control to be useful you need a previous button, selected page indication, and next button at a minimum. So why shouldn't we start from that base assumption? It'll be responsive to the point where you can no longer fit three buttons in the container, but I'm not sure what else a developer using this product can expect.

My latest _local_ version of this PR addresses the expansion panel issue, by asking if the parent's classList contains 'v-expanion-panel-content__wrapper' and if so obtaining width from the great grand parent component (which is the first component to have a non-zero width during the mount LCH for the embedded v-pagination) given this structure:

```
v-expansion-panels
  v-expansion-panel
    v-expansion-panel-header
    v-expansion-panel-content
      v-pagination
```

This feels brittle and I am not steeped in this enough to know if there are other parent components that will cause this same behavior. I also attempted to get onResize to fire again once the actual parent's width was established, however, the directive is not reactive to the parent's width and only changing the window width will trigger CD.

I can submit an update to this that only handles the expansion panel issue, if you are good with a solution that relies on detecting the parent class. However, if that's the water's edge on this PR, I'd like to open another issue to rework v-pagination under the constraint that at least 3 buttons are required, which simplifies a lot of the logic that currently exists.

mcculleydj:
Definitely open to suggestions. The simple solution is to await $nextTick when the width of the parent would be defined but you rejected that with "we don't use async/await" without any further explanation. Looking for a class on a parent and then getting the width of a great grandparent is brittle and going to look terrible in code, especially if I need to quiet down TS's "could be undefined" warnings. Maybe the resize directive needs to be changed to detect when the width of the parent changes from 0 to whatever it actually is after the expansion panel transition completes. 

Once this problem is fixed I'm not huge fan of `expect(pagination.vm.maxButtons).toBe(-3)`. So, I'd like to keep working on this component.

mcculleydj:
In effort to flatten the first solution I created two new methods:

- checkParent: checks the classList of the parent element against an object which can be added to if `v-expansion-panel-content__wrap` isn't the only class that would trigger this problem. The object maps classes to the number of steps up the tree we need to look to get a working value for clientWidth

- getContainerWidth: given a number steps up the DOM tree to look, obtains a reference to that ancestor and returns its width

johnleider:
I think that this may end up being resolved as part of a greater effort. This PR has some insight as to what we are trying to accomplish and overall I think this may be a better way to approach this from a macro level.

If you agree, please close this issue. Otherwise, let me know how it differs and how the provided approach would not suite this bug.

https://github.com/vuetifyjs/vuetify/pull/9341

mcculleydj:
Supplying parent component width once it becomes available is a better solution. I'll close this. Curious if the PR you've highlighted will work in all the browsers you might wish to support.

PR #9021, opened 2019-09-19
docs(Theme): fix Chinese translation

351768593:


sh7dm:
Sorry, we have to close this. Since we use Crowdin, please apply your fix using it. Click translate here and you'll go into Crowdin localization:
image

Issue #9026, opened 2019-09-19
[Bug Report] File input truncated to fixed amount of characters


peteruithoven:
### Environment
**Vuetify Version:** 2.0.18
**Vue Version:** 2.6.10
**Browsers:** Firefox 69.0
**OS:** elementary OS Juno (built on Ubuntu 18.04)

### Steps to reproduce
1. Add a file input component (`v-file-input`)
2. Set `truncate-length` to `10`
3. Select file with a name that's longer than 10 characters. 
  For example: "Longer filename.js".

### Expected Behavior
Shows filename that's truncated in such a way it's 10 characters. 
For example: "Longe…e.js".

### Actual Behavior
It's truncated to a fixed amount of characters. A "…" with 10 characters on both sizes. 
For example: "Longer fi…ilename.js".
![Screenshot from 2019-09-19 14-36-19](https://user-images.githubusercontent.com/523210/65244458-df981800-daea-11e9-9fcd-71598134249d.png)

### Reproduction Link
<a href="https://codepen.io/peteruithoven/pen/KKPrwVw" target="_blank">https://codepen.io/peteruithoven/pen/KKPrwVw</a>

### Other comments
Relevant code:  
https://github.com/vuetifyjs/vuetify/blob/2d51ae1011812996bc7f2525f25b3b59e8fa12c9/packages/vuetify/src/components/VFileInput/VFileInput.ts#L241-L244

peteruithoven:
@yuzhoumao thanks for improving this! :)

PR #9036, opened 2019-09-19
fix(VFileInput): replace hardcoded string boundary with computed

yuzhoumao:
## Description

instead of hardcoding the number of chars taken from head and tail of user input to be 10, computed the number of chars to keep based on the `truncate-length` field.

## Motivation and Context

undesired input truncation when `truncate-length` is smaller than default value 22

#9026

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VFileInput/VFileInput.ts
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.ts
@@ -240,7 +240,8 @@ export default VTextField.extend({
     },
     truncateText (str: string) {
       if (str.length < Number(this.truncateLength)) return str
-      return `${str.slice(0, 10)}…${str.slice(-10)}`
+      const charsKeepOneSide = Math.floor((this.truncateLength - 1) / 2)
+      return `${str.slice(0, charsKeepOneSide)}…${str.slice(-charsKeepOneSide)}`
     },
   },
 })


packages/vuetify/src/components/VFileInput/__tests__/VFileInput.spec.ts
```diff
@@ -231,4 +231,68 @@ describe('VFileInput.ts', () => {
...
+          {
+            name: 'testFile15Chars',
+            size: 1,
+          },
+        ],
+      },
+    })
+
+    expect(wrapper.find('.v-file-input__text').text()).toBe('test…hars')
+  })
```
sh7dm: integrate all these tests into one, change props using `wrapper.setProps({ ... })`

yuzhoumao: thanks for the review! I agree and have merge the test cases

packages/vuetify/src/components/VFileInput/__tests__/VFileInput.spec.ts
```diff
@@ -231,4 +231,34 @@ describe('VFileInput.ts', () => {
...
+      propsData: {
+        truncateLength: 1,
+        value: fifteenCharFile,
+      },
+    })
+
+    expect(wrapper.find('.v-file-input__text').text()).toBe('…')
+
+    wrapper.setProps({
+      truncateLength: 2,
```
sh7dm: What's about leaving char in the beginning here? I'm not sure if it's a good idea

yuzhoumao: Yeah I also wanted to discuss this. Would you say it's better if we completely hide the filename if user specifies `truncateLength` to be 1?

sh7dm: I don't know

Issue #9045, opened 2019-09-20
[Bug Report] Unable to select same file twice when value cleared


matthieusieben:
### Environment
**Vuetify Version:** 2.0.18
**Vue Version:** 2.6.10
**Browsers:** Chrome 76.0.3809.132
**OS:** Mac OS 10.14.6

### Steps to reproduce
Pick a file using a v-file-input, clear the v-model programmatically (using the button in the codepen), try to pick the exact same file again, the input value will not be set.

### Expected Behavior
The inner "file" input's value should be cleared when the value prop is changed

### Actual Behavior
The inner "file" input element's value is not cleared when the value prop changes

### Reproduction Link
<a href="https://codepen.io/matthieusieben/pen/NWKEEQo" target="_blank">https://codepen.io/matthieusieben/pen/NWKEEQo</a>

### Other comments
I was able to "hack" around the issue using the following *watch* function (in the vuetify's VFileInput.ts source code):

```
    value(val) {
      if (!val) this.$refs.input.value = ''
    },
```

sh7dm:
you can open a PR with your fix

ardabzlk:
I've solved this problem by adding :key property to <v-file-input> tag, and initialized key value with  0, whenever I want to reset file input I increase  :key="inputKey" with 1
My data =>
`  data() {
    return {
      inputKey: 0
 }
}`
My reset function =>
` closeNewPhotoDialog() {
      this.inputKey++;
    }`
My component =>
`<v-file-input  v-model="newPhotoTextField" :key="inputKey">`

johnleider:
We kindly ask users to not comment on closed/resolved issues. If you believe that this issue has not been correctly resolved, please create a new [issue](https://issues.vuetifyjs.com) showing the regression or create a new [discussion](https://discussions.vuetifyjs.com).

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

johnleider:
We kindly ask users to not comment on closed/resolved issues. If you believe that this issue has not been correctly resolved, please create a new [issue](https://issues.vuetifyjs.com) showing the regression or create a new [discussion](https://discussions.vuetifyjs.com).

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

PR #9091, opened 2019-09-25
fix(VFileInput): trigger onInput after value has changed

matthieusieben:
## Description

When the `value` prop of a `VFileInput` is changed programatically, the internal `input(type=file)` element's value must be cleared.

This is due to the fact that the input element won't emit a `change` event is the user selects the same files as the ones already in `$refs.input.files`. But since the `VFileInput`'s value has changed and no longer corresponds to `$refs.input.files`, we *do* want the `change` event to be triggered.

Ideally, we would do something as follows, but the `files` property is read only.
```js
watch: {
  value (v) {
    const value = !v ? [] : Array.isArray(v) ? v : [v]
    this.$refs.input.files = value
  }
}
```

## Motivation and Context

Fixes: #9045

## How Has This Been Tested?

none

## Markup:

<details>

```vue
<template>
      <v-container>
        <v-file-input ref="fileInput" v-model="value">
        </v-file-input>
        <v-btn @click="clear">Clear</v-btn>
      </v-container>
</template>

<script>
export default {
  data: () => ({
    value: null
  }),
  methods: {
    clear() {
      this.value = null
      this.$nextTick(() => {
        console.log('This should be empty:', this.$refs.fileInput.$refs.input.value)
      })
    }
  }
}
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VFileInput/VFileInput.ts
```diff
@@ -141,6 +141,18 @@ export default VTextField.extend({
       },
       immediate: true,
     },
+    value: {
```
sh7dm: You don't need object-style watch here.

matthieusieben: You should add a linting rule for this... Otherwise, how are we supposed to know all the "implicit" rules?

sh7dm: okay, we'll consider it

sh7dm: but for now please fix it, it's not required here.

johnleider: But why? I've never seen anyone use a handler for a watcher that wasn't using immediate or deep. It's not an implicit rule and more of a general code style.

sh7dm: Upvote to @johnleider, object style must only be used in deep/immediate

matthieusieben: I agree the style is better. It actually is written like that because I was testing with `immediate: true` and forgot to cleanup styling afterwards. An eslint rule would have avoided this conversation and make us all gain some time ;-)

packages/vuetify/src/components/VFileInput/VFileInput.ts
```diff
@@ -141,6 +141,17 @@ export default VTextField.extend({
...
+    value (v) {
+      const value = this.isMultiple ? v : v ? [v] : []
+      if (!deepEqual(value, this.$refs.input.files)) {
+        // When the input value is changed programatically, clear the
+        // internal input's value so that the `onInput` handler
+        // can be triggered again if the user re-selects the exact
+        // same file(s). Ideally, `input.files` should be
+        // manipulated directly but that property is readonly.
+        this.$refs.input.value = ''
+      },
```
johnleider:
```suggestion
      }
```

Issue #9120, opened 2019-09-26
[Bug Report] Performance issue when using large date range (0000-01-01 to 9999-12-31) in date range picker


TheInvoker:
### Environment
**Vuetify Version:** 2.1.0-beta.1
**Vue Version:** 2.6.10
**Browsers:** Firefox 69.0
**OS:** Windows 10

### Steps to reproduce
Click the left arrow to go back a month.

### Expected Behavior
Should change instantly with all days selected in the month.

### Actual Behavior
Takes long time and kills cpu to change months.

### Reproduction Link
<a href="https://codepen.io/sneaky666/pen/zYOXpOy?&editable=true&editors=101" target="_blank">https://codepen.io/sneaky666/pen/zYOXpOy?&editable=true&editors=101</a>

### Other comments
There should not be a huge spike in CPU performance and time because switching months or when setting a long date range should only draw at most 7x5=35 circles on the calendar. It feels like it is iterating through the huge date range (10000 years worth of days in this case) and drawing circles around them all in realtime.

KaelWD:
#8891

It's creating a `new Date()` object for each day in the selection, so in this case 3,652,424 times. Really you shouldn't be using this for anything longer than a couple of months so it isn't a big problem. 

https://github.com/vuetifyjs/vuetify/blob/72e279c5/packages/vuetify/src/components/VDatePicker/VDatePicker.ts#L373-L381

TheInvoker:
Why was it designed to loop through all dates though? Shouldn't it lazy load the current month (or visible view) and process that only?

TheInvoker:
This code snippet also seems really inefficient. Firstly I feel like it should be lazy load the visible view only. On top of that it's doing a linear map of the range, then sorting it for some reason when it should already be sorted if going from min to max in the range. Then again it's looping through the range for some reason when it could have processed it in the first linear map.

KaelWD:
I didn't write it. You shouldn't be using this for longer ranges so I guess the guy who did didn't think lazy loading was worth the effort. The map then sort is making sure the array is actually in the format `[from, to]` in case you passed `[to, from]` instead, and the second loop is over the number of days in the range not the range itself.

PR #9128, opened 2019-09-27
Fix/#9120 VDatePicker range performance

YipingRuan:
## Description
Check date selection for range in the current month only.
Currently smooth to support range as big as '0000-01-01', '2019-09-20', should be enough.
## Motivation and Context
https://github.com/vuetifyjs/vuetify/issues/9120

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-date-picker v-model="dates" range width="290" class="mt-4"></v-date-picker>
    v-model value: {{ dates }}
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      dates: ['0000-01-01', '2019-09-20'],
    })
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
+++ b/packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
@@ -50,7 +50,7 @@ export default mixins(
       type: String,
       required: true,
     },
-    value: [String, Array],
+    value: [String, Array] as PropValidator<string | string[]>,
   },
 
   data: () => ({
@@ -101,7 +101,7 @@ export default mixins(
     },
     genButton (value: string, isFloating: boolean, mouseEventType: string, formatter: DatePickerFormatter) {
       const isAllowed = isDateAllowed(value, this.min, this.max, this.allowedDates)
-      const isSelected = value === this.value || (Array.isArray(this.value) && (this.value.indexOf(value) !== -1 || this.isInRange(value)))
+      const isSelected = this.isSelected(value)
       const isCurrent = value === this.current
       const setColor = isSelected ? this.setBackgroundColor : this.setTextColor
       const color = (isSelected || isCurrent) && (this.color || 'accent')
@@ -193,13 +193,17 @@ export default mixins(
         directives: [touchDirective],
       }, [transition])
     },
-    isInRange (value: string): boolean {
-      if (this.range && this.value.length === 2) {
-        const [from, to] = [...this.value as string[]].sort()
-        return from <= value && value <= to
+    isSelected (value: string): boolean {
+      if (Array.isArray(this.value)) {
+        if (this.range && this.value.length === 2) {
+          const [from, to] = [...this.value].sort()
+          return from <= value && value <= to
+        } else {
+          return this.value.indexOf(value) !== -1
+        }
       }
 
-      return false
+      return value === this.value
     },
   },
 })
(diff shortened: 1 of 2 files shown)


nekosaur:
1. What is this code even supposed to do?
2. Why cache manually instead of using a computed prop?

YipingRuan:
> 1. What is this code even supposed to do?
> 2. Why cache manually instead of using a computed prop?

Thanks. Moved the logic to date-picker-table.ts for the displaying month only

packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
```diff
@@ -100,7 +101,7 @@ export default mixins(
     },
     genButton (value: string, isFloating: boolean, mouseEventType: string, formatter: DatePickerFormatter) {
       const isAllowed = isDateAllowed(value, this.min, this.max, this.allowedDates)
-      const isSelected = value === this.value || (Array.isArray(this.value) && this.value.indexOf(value) !== -1)
+      const isSelected = value === this.value || (Array.isArray(this.value) && (this.value.indexOf(value) !== -1 || this.isInRange(value)))
```
nekosaur: refactor all of the logic into a `isSelected` function or something similar

packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
```diff
@@ -192,5 +193,13 @@ export default mixins(
         directives: [touchDirective],
       }, [transition])
     },
+    isInRange (value: string): boolean {
+      if (this.range && this.value.length === 2) {
+        const [from, to] = [...this.value as string[]].sort()
+        return from <= value && value <= to
```
nekosaur: This is all down to personal preferences, but for me
```
value >= from && value <= to
```
is much clearer

packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
```diff
@@ -192,5 +193,13 @@ export default mixins(
         directives: [touchDirective],
       }, [transition])
     },
+    isInRange (value: string): boolean {
+      if (this.range && this.value.length === 2) {
+        const [from, to] = [...this.value as string[]].sort()
```
nekosaur: Add `PropValidator<string | string[]>` to `value` prop and you should be able to remove `as string[]`

packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
KaelWD: I prefer this way, it's like in mathematics from &leq; value &leq; to

packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
YipingRuan: This is super convenient.

TheInvoker:
when will this be released?

johnleider:
Is there a reason this is to dev instead of master?

YipingRuan:
> Is there a reason this is to dev instead of master?

No, bugs should -> master. 
Forgot this new feature is already there.

TheInvoker:
So is this coming in a version after 2.1.2?

PR #9163, opened 2019-10-01
fix(VDialog/VMenu/VTooltip): hide wrapper element when not attached

### LINKED ISSUE #6951: [Bug Report] Dialog/Menu/Tooltip output an extra element and break layout
### Versions and Environment
**Vuetify:** 1.5.9
**Vue:** 2.6.10
**Browsers:** Chrome 73.0.3683.103
**OS:** Linux x86_64

### Steps to reproduce
Add a dialog inside a nav drawer, using a v-list-tile in the activator slot of the dialog.

### Expected Behavior
The dialog activator should have the exact spacing as the other drawer items.

### Actual Behavior
The dialog activator has extra spacing below it.

### Reproduction Link
<a href="https://codepen.io/anon/pen/ROKPZo" target="_blank">https://codepen.io/anon/pen/ROKPZo</a>

### Other comments
Poking around the inspector, it looks like the problem stems from the v-dialog__container div right under the v-list-tile activator. Setting its style to display: none fixes the issue.

KaelWD:
## Description

Adds `display: none` to the root element unless it's attached to itself (because then we'd be hiding the content too)

## Motivation and Context

fixes #6951
The previous commit 97681ce only helps in an inline context

## Types of changes

### DIFF
--- a/packages/vuetify/src/components/VTooltip/VTooltip.ts
+++ b/packages/vuetify/src/components/VTooltip/VTooltip.ts
@@ -110,6 +110,10 @@ export default mixins(Colorable, Delayable, Dependent, Detachable, Menuable, Tog
         'v-tooltip--right': this.right,
         'v-tooltip--bottom': this.bottom,
         'v-tooltip--left': this.left,
+        'v-menu--attached':
+          this.attach === '' ||
+          this.attach === true ||
+          this.attach === 'attach',
       }
     },
     computedTransition (): string {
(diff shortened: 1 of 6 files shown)


packages/vuetify/src/components/VTooltip/VTooltip.ts
```diff
@@ -110,6 +110,10 @@ export default mixins(Colorable, Delayable, Dependent, Detachable, Menuable, Tog
         'v-tooltip--right': this.right,
         'v-tooltip--bottom': this.bottom,
         'v-tooltip--left': this.left,
+        'v-menu--attached':
```
DRoet:
```suggestion
        'v-tooltip--attached':
```
right?

PR #9231, opened 2019-10-04
Use Now 2.0 deployment for docs

styfle:
## Description
Update docs website from Now 1.0 to Now 2.0

## Motivation and Context
Faster builds, faster cold-starts.

## How Has This Been Tested?
Work in progress

## Types of changes

- [x] Docs

## Checklist:

### DIFF
--- a/now.json
+++ b/now.json
@@ -3,7 +3,7 @@
   "name": "vuetifyjs.com",
   "alias": "vuetifyjs.com",
   "builds": [
-    { "src": "packages/docs/src/public/**", "use": "@now/static" },
+    { "src": "packages/docs/src/dist/**", "use": "@now/static" },
     { "src": "packages/docs/server.js", "use": "@now/node" }
   ],
   "routes": [
--- a/now.json
+++ b/now.json
@@ -3,7 +3,7 @@
   "name": "vuetifyjs.com",
   "alias": "vuetifyjs.com",
   "builds": [
-	{ "src": "packages/docs/package.json", "use": "@now/static-build" },
+    { "src": "packages/docs/package.json", "use": "@now/static-build" },
     { "src": "packages/docs/server.js", "use": "@now/node" }
   ],
   "routes": [
--- a/now.json
+++ b/now.json
@@ -3,7 +3,7 @@
   "name": "vuetifyjs.com",
   "alias": "vuetifyjs.com",
   "builds": [
-    { "src": "packages/docs/src/dist/**", "use": "@now/static" },
+	{ "src": "packages/docs/package.json", "use": "@now/static-build" },
     { "src": "packages/docs/server.js", "use": "@now/node" }
   ],
   "routes": [
--- a/now.json
+++ b/now.json
@@ -1,10 +1,17 @@
 {
+  "version": 2,
   "name": "vuetifyjs.com",
   "alias": "vuetifyjs.com",
+  "builds": [
+    { "src": "packages/docs/src/public/**", "use": "@now/static" },
+    { "src": "packages/docs/server.js", "use": "@now/node" }
+  ],
+  "routes": [
+    { "handle": "filesystem" },
+    { "src": ".*", "dest": "/packages/docs/server.js" }
+  ],
   "github": {
     "autoAlias": false,
     "silent": true
-  },
-  "version": 1,
-  "type": "docker"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
     "babel-jest": "^24.8.0",
+    "copy-webpack-plugin": "^5.0.5",
     "cross-spawn": "^6.0.5",
     "eslint": "^6.1.0",
     "eslint-config-standard": "^13.0.1",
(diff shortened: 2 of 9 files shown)


johnleider:
Testing this locally I'm running into this error:

image

This is with a cleared node_modules and reinstalling the now plugin.

johnleider:
Here is the stack trace
image

styfle:
Hi John,

Does it work with canary? `npm i -g now@canary`

johnleider:
@vuetifyjs/core-contributors @vuetifyjs/core-contributors feedback welcome

now.json
```diff
@@ -1,10 +1,17 @@
 {
+  "version": 2,
   "name": "vuetifyjs.com",
   "alias": "vuetifyjs.com",
+  "builds": [
```
KaelWD: Is there any documentation on this option? We need it to run `yarn` from the root directory before building so the whole monorepo thing actually works properly.

KaelWD: I tried making a `server.js` at the root instead with just `module.exports = require('./packages/docs/server')` but that didn't work either.
Adding a `functions` key to the config to use `includeFiles` gives the error `The property `builds` is not allowed in now.json when using Now 2.0 – please remove it.`
Trying to test locally with `now dev -d` so I can see what's going on without waiting for now to start building tries to type-check every ts file in the repo even though it's already been built by webpack, then infinitely loops back to building when it's done.

KaelWD:
> This seems like a babel issue but why is babel used at runtime?

It's using the same babel config as the client, I corrected it but then it can't find lodash instead. If I download the function content there's a lot of packages missing from node_modules which I tried to correct with `includeFiles` but got the error described above.

now.json
styfle: Hi @KaelWD 

This PR was submitted awhile ago while `builds` was still documented. The new `functions` property has mostly replaced it. In this case, the `./packages/docs` website relies on build output from the other directories (I think) so I had to use `builds` instead of moving `./now.json` to `./packages/docs/now.json`. You can't mix `builds` and `functions`.

styfle:
> there's a lot of packages missing from node_modules

@KaelWD There should be some missing because [node-file-trace](https://github.com/zeit/node-file-trace) runs on the input file, in this case `packages/docs/server.js`, and removes any unused dependencies. This is necessary because the maximum Serverless Function size is 50 MB and your repo is currently 671 MB.

It sounds like there might be a bug in `node-file-trace`. Do you see anything that is required at runtime but missing from `node_modules`?

KaelWD:
Worked around the node_modules thing by not using webpack externals, but it's removing pretty much everything from /dist too.

styfle:
@KaelWD Thanks for the update!

The reason why there are only two files included with the function is because those are the only two that were required in the `packages/docs/server.js` entry point:

```js
 const bundle = require('./dist/vue-ssr-server-bundle.json')
 const clientManifest = require('./dist/vue-ssr-client-manifest.json')
```

It seems like you are intending to have the remaining files served statically to the client so I think `@now/static-build` is best here.

KaelWD:
Sorry I missed this reply somehow. That still doesn't work the way we want - all the static requests are going through server.js instead of being served from the file system, and they need to be available under /dist instead of being moved to the root.  
The output view also doesn't match what I see when I download the function:
image
image

KaelWD:
Dist is still empty:
image
And exists, so the output view is lying - dist doesn't get moved to the root: https://vuetifyjscom-ahargm7pp.now.sh/dist/vue-ssr-client-manifest.json

johnleider:
@styfle Is there anything else that we can provide help with regarding this issue? We will need to make some massive changes if your new version cannot support this use-case. Please follow-up as soon as possible, thank you.

styfle:
@johnleider I haven't been able to identify the root cause yet.

It seems that one or more of your dependencies is using a dynamic require which isn't statically analyzable by [node-file-trace](https://github.com/zeit/node-file-trace). I'm guessing its the line `app.use('/dist', serve('./dist', true))`. We could try using [includeFiles](https://zeit.co/docs/runtimes#advanced-usage/technical-details/including-additional-files) but its probably best to have those static files served from the CDN instead of the node server.

johnleider:
@styfle Do you know how much longer before Now v1 goes away?

styfle:
@johnleider Sorry, no date yet. You'll see an announcement with a date once that is decided.

There are a few more fixes coming to `node-file-trace` in the next stable release (already on canary).

I'll take another try at this migration in a couple weeks, thanks!

johnleider:
Does anything here https://zeit.co/blog/advanced-project-settings help us with this migration?

KaelWD:
And why is it a config page instead of something in now.json?

styfle:
[Project Settings](https://zeit.co/blog/advanced-project-settings) is the first step towards zero config monorepos. It links the code on your machine to a ZEIT Project so that the settings can be synced.

For your purposes, the `rootDirectory` property won't work because your `docs` package depends on other packages and `rootDirectory=docs` will ignore the other directories. So in the future we'll have more options to enable that type of behavior for zero config.

[Project Linking](https://zeit.co/docs/now-cli#commands/now/project-linking) allows you to rename a project without breaking deployments or losing your domain name.

styfle:
Closing in favor of #10598 which is using SPA instead of SSR.

This will drastically improve page load times 🚀 

Here's the current PageSpeed score for reference.

<a href="https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fvuetifyjs.com%2Fen%2F">vueitify-score</a>

Here's the Audit from my local dev tools.

image

I'm confident that the SPA approach in #10598 can improve this score.

Issue #8395, opened 2019-08-08
[Bug Report] Chrome's input history suggestions shows for input fields with Vuetify v2


rgb-one:
### Environment
**Vuetify Version:** 2.0.4
**Vue Version:** 2.5.17
**Browsers:** Chrome 76.0.3809.87
**OS:** Linux x86_64

### Steps to reproduce
1. Choose an item from the `v-autocomplete` dropdown list
2. Click 'Submit'
3. Repeat step 1
4. Observe the browser input suggestion overlaying the autocomplete list

### Expected Behavior
When typing content into a `v-autocomplete` or `v-combobox`, No browser input suggestions should be provided.

### Actual Behavior
When typing content into a `v-autocomplete` and `v-combobox`, the Chrome input history dropdown overlays the results provided by these components. For the v-text-field, it provides suggestions as well but this isn't much of an issue.

### Reproduction Link
<a href="https://codesandbox.io/s/vue-template-5b1p7" target="_blank">https://codesandbox.io/s/vue-template-5b1p7</a>

### Other comments
In the reproduction example, if you submit multiple items, only the first one submitted shows up in the browser autocomplete list. In the initially observed scenario however, multiple would show. The more data submitted from that input field, the more browser input suggestions would show.

Also, what has been observed is that these browser suggestions show for `v-text-field`'s as well even if the context is different (i.e. the text field is not related to the data); for example the 'Reproduction Link' in this issue form, it shows some of the suggestions that first showed up in an app I am working on.

Here are some images:
![browser-input-history-dialog-v-combobox](https://user-images.githubusercontent.com/5672238/62804809-f0994680-bab3-11e9-92a3-c6f39cd3ba0d.png)
![browser-input-history-v-autocomplete](https://user-images.githubusercontent.com/5672238/62804810-f131dd00-bab3-11e9-8de8-e6edc403558a.png)

myleslee:
Adding `autocomplete="off"` to  `v-autocomplete` seems to fix the problem.

rgb-one:
@myleslee, indeed. Cool stuff. `autocomplete="off"` by default on `v-autocomplete` and `v-combobox` would be good.

simjes:
I would like to work on this

PR #9246, opened 2019-10-05
fix(VAutocomplete): Disable browser autocomplete, fix#8395

simjes:
## Description
Disable browser autocomplete by default in VAutocomplete

## Motivation and Context
Component should suggest autocompletion items, not the browser. 

## How Has This Been Tested?
Tested it in the docs by recreating the example that is linked in #8395

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-content>
      <v-container>
        <v-form ref="form" @submit.prevent="submit()">
          <v-autocomplete v-model="item" :items="items" placeholder="Type something..." id="item"></v-autocomplete>

          <v-btn type="submit" color="primary" :disabled="isValid">Submit</v-btn>
        </v-form>
      </v-container>
    </v-content>
  </v-app>
</template>

<script>
export default {
  name: "App",
  data() {
    return {
      items: ["One", "Two", "Three"],
      item: ""
    };
  },

  computed: {
    isValid() {
      return !this.item;
    }
  },

  methods: {
    submit() {
      alert(`Item ${this.item}`);
      this.$refs.form.reset();

    }
  }
};
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts
```diff
@@ -281,7 +281,10 @@ export default VSelect.extend({
       const input = VTextField.options.methods.genInput.call(this)
 
       input.data = input.data || {}
-      input.data.attrs = input.data.attrs || {}
+      input.data.attrs = {
```
MajesticPotatoe:
```
input.data.attrs = input.data.attrs || {}
input.data.attrs.autocomplete = 'off'
```

simjes: I just assumed that you should be able to turn browser autocomplete on, but that doesn't make sense at all :D Will update

MajesticPotatoe: autocomplete is just passed through to the native input, the `input.data.attrs || {}` is to safeguard against a null or empty value. would likely throw an error if you spread a null value.

vcastro45:
This issue has been fixed by v2.1.2 but v2.1.3 is has regression.

In v2.1.3, the autocomplete attribute is set to `"disabled"` by default but it is not a valid value.
The doc for the `<input>` HTML tag says the values can be `"on"` or `"off"`: 

https://www.w3schools.com/tags/att_input_autocomplete.asp

MajesticPotatoe:
@vcastro45 `disabled` is in fact not a valid value, however it was chosen to combat against chrome ignoring the value of `off` in certain scenarios while still effectively disabling it for other browsers. This of course is overridden if you set the value yourself maintaining functionality for people who have explicitly set it. If you are experiencing that this is not the case please open a new issue.

vcastro45:
@MajesticPotatoe In what kind of scenario Chrome doesn't disable the input ?
I currently do my tests using Chrome and since I updated vuetify from 2.1.2 to 2.1.3, the browser autocomplete is shown everywhere...

MajesticPotatoe:
> If you are experiencing that this is not the case please open a new issue.
https://issues.vuetifyjs.com/

Issue #9263, opened 2019-10-07
[Bug Report] vue-meta 2.3.0 breaks vuetify theme


MatthewAry:
### Environment
**Vuetify Version:** 2.1.1
**Vue Version:** 2.6.10
**Browsers:** Chrome 77.0.3865.90
**OS:** Mac OS 10.14.6

### Steps to reproduce
Clone project and run it. Observe that the default hello world component has some of its links replaced with buttons.

It's that or you can add vue-meta 2.3.0 to a project with custom theme colors.

### Expected Behavior
The buttons should be green.

### Actual Behavior
The buttons are white.

### Reproduction Link
<a href="https://github.com/MatthewAry/bug-reproduction-vue-meta-breaks-vuetify" target="_blank">https://github.com/MatthewAry/bug-reproduction-vue-meta-breaks-vuetify</a>

### Other comments
Downgrading vue-meta from 2.3.0 to 2.2.2 fixes the problems with the theme. Other observations include that the theme's light and dark settings are not propagating through out.

kevinmarrec:
Invoking @pimlie , hopefully he may have some clues about this issue :)

FPurchess:
Can't confirm that downgrading vue-meta to 2.2.2 fixes the problem:

Tested on Firefox 69.0.1 (64-Bit) / Linux (Ubuntu) and the repository provided by @MatthewAry + this codepen: https://codepen.io/0in1/pen/GRRgmRa

DRoet:
just had a quick look, but I think it might be caused by: https://github.com/nuxt/vue-meta/commit/c344d60

since we are using `root.$options.metaInfo = () => {}` in `initVueMeta`

the removal of 
```js
if (isFunction(data)) {
   data = data.call(component)
}
```

prevents the metaInfo from loading

pimlie:
The docs needs to be updated still, but in `v2.3` i've added (due to all the vuetify issues) better support for adding meta info from 3rd party libs to prevent issues like here. Docs are currently my comments in [this pr](https://github.com/nuxt/vue-meta/pull/452)

DRoet:
@pimlie so something like this? https://github.com/vuetifyjs/vuetify/pull/9268

tvld:
I can confirm that after upgrading to vuetify@2.3.1 today, my theme immediately broke as well (using nuxt). All colors are back to default

PR #9268, opened 2019-10-07
fix(theme): use new vue-meta api

DRoet:
## Description

To make sure everything still works with vue-meta `2.3.0` by using the new API , should fix: https://github.com/vuetifyjs/vuetify/issues/9263

## Motivation and Context

https://github.com/vuetifyjs/vuetify/issues/9263#issuecomment-539168292

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


DRoet:
@kevinmarrec Could you see if everything still works when using nuxt (ssr)?

the < 2.3.0 code below also has an extra check to see if the vuetify-theme-stylesheet already exists so Im not sure if this new api needs that aswell

packages/vuetify/src/services/theme/index.ts
```diff
@@ -179,6 +179,22 @@ export class Theme extends Service {
     this.vueMeta = true
 
     const meta = root.$meta()
+
+    const metaAdd = typeof meta.addApp === 'function'
+    if (metaAdd) {
+      const { set } = root.$meta().addApp('vuetify')
```
pimlie: This should work too I guess:

```suggestion
      const { set } = meta.addApp('vuetify')
```

packages/vuetify/src/services/theme/index.ts
```diff
@@ -179,6 +179,22 @@ export class Theme extends Service {
...
+
+      set({
+        style: [{
+          cssText: this.generatedStyles,
```
pimlie: You could probably also set [`once: true`](https://vue-meta.nuxtjs.org/api/#once), but you probably shouldnt call `.set` multiple times with the same stylsheet then because it will probably be added each time then

If thats an issue you could set a prop on `$root` to check if its already added or not

DRoet: The old code would check if the stylesheet is already there and if so only re-apply the `cssText`:
https://github.com/vuetifyjs/vuetify/blob/06419697cd3ba33ff116397d272ebda902706beb/packages/vuetify/src/services/theme/index.ts#L204
I assume this only happens during SSR, but you are saying we should just skip it instead?

DRoet: > If thats an issue you could set a prop on `$root` to check if its already added or not

is there no `getApp()` available? 😅

pimlie: If that check is done only for SSR it should be possible to add a [`[tagidkeyname]`](https://vue-meta.nuxtjs.org/api/#tagidkeyname) prop so its only added once. Ofc you have to omit `once` in that case.

kevinmarrec:
Please be aware that there is multiple cases to test, it includes being able to change theme  & navigate through the app without issues (SPA + SSR)

DRoet:
@kevinmarrec yeah I'm not able to test this with nuxt anytime soon so PTAL when u can.

packages/vuetify/src/services/theme/index.ts
kevinmarrec: How does it work ? I'm not sure to understand how it fixes thing, it makes double places where we set styles.
Also with the `addApp` I also don't understand how `this.generatedStyles` gonna be reactive if not in a function. Please give me your light 😁

pimlie: My remarks are generic of nature. Havent look at the vuetify specifics

packages/vuetify/src/services/theme/index.ts
```diff
@@ -202,6 +213,24 @@ export class Theme extends Service {
...
+    }
+    set({
+      style: [{
+        cssText: this.generatedStyles,
+        type: 'text/css',
+        id: 'vuetify-theme-stylesheet',
+        nonce: (this.options || {}).cspNonce,
+      }],
+    })
+    this.hasMeta = true
```
DRoet: @pimlie does this seem like an OK way to make things reactive? `applyVueMeta23()` get triggered by a watcher

DRoet:
@kevinmarrec during my testing, I've noticed that the vue-meta (<2.3) related code is not fully reactive. of you override the theme colors on `mounted()` the changes are not shown.

the newer `2.3` code doesn't have this problem. @kevinmarrec could you try and test it out aswell and see if it works in a nuxt environment?

DRoet:
Tested this with nuxt in spa mode which works fine, Im unable to test SSR mode since `yarn link vuetify` caused errors (multiple instances of Vue detected) but that is unrelated to this PR

packages/vuetify/src/services/theme/index.ts
pimlie: Calling `remove` is actually not required here, set always overwrites all the previous meta info. See the [implementation](https://github.com/nuxt/vue-meta/blob/master/src/shared/additional-app.js) in vue-meta.

`remove` is only required to call if you want to remove all meta data without setting new.

DRoet: sweet, that makes it a lot easier

DRoet:
Hmm, the only issue I'm having is that `applyVueMeta23()` is called to early, since `vue-meta` only applies on `mounted()` whilst our theme watcher triggers a lifecycle before that (due to `vm.$watch.immediate`)

packages/vuetify/src/services/theme/index.ts
```diff
@@ -202,6 +214,19 @@ export class Theme extends Service {
     }
   }
 
+  private applyVueMeta23 () {
+    const { set } = this.vueMeta.addApp('vuetify')
```
kevinmarrec: Shouldn't `addApp` need to be called only once ?

pimlie: Both is fine, calling addApp everytime adds only [a little overhead](https://github.com/nuxt/vue-meta/blob/master/src/shared/additional-app.js)

kevinmarrec: Okay, well unless user is making theme update every millisecond it shouldn't affect performance much

packages/vuetify/src/services/theme/index.ts
```diff
@@ -176,10 +182,16 @@ export class Theme extends Service {
   }
 
   private initVueMeta (root: any) {
-    this.vueMeta = true
+    this.vueMeta = root.$meta()
+    if (this.isVueMeta23) {
+      // vue-meta needs to apply after mounted()
+      root.$nextTick(() => {
+        this.applyVueMeta23()
+      })
```
sh7dm:
```suggestion
      root.$nextTick(this.applyVueMeta23)
```

PR #9300, opened 2019-10-09
fix(VTextField): use intersect to recalculate elements width

### LINKED ISSUE #9299: [Bug Report] Incorrect text field size calculation inside expansion panel
### Environment
**Vuetify Version:** 2.1.2
**Vue Version:** 2.6.10
**OS:** Linux x86_64

### Steps to reproduce
1. Open expansion panel
2. Focus second field

### Expected Behavior
Correct notch width in border
Correct label positioning in second field

### Actual Behavior
Incorrect notch width in border
Incorrect label positioning in second field

### Reproduction Link
<a href="https://codepen.io/djaler/pen/oNNXywM" target="_blank">https://codepen.io/djaler/pen/oNNXywM</a>

### Other comments
Elements width calculated while they are hidden, so offsetWidth = 0

Djaler:
fixes #9299

## Description
Use v-intersect directive to recalculate element width after input becomes visible, 

## Motivation and Context
Cause they have zero width on mount in hidden state (inside expansion panel, for example)
https://github.com/vuetifyjs/vuetify/issues/9299

## How Has This Been Tested?
visually

## Markup:
https://codepen.io/djaler/pen/oNNXywM

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -13,6 +13,7 @@ import Loadable from '../../mixins/loadable'
 
 // Directives
 import ripple from '../../directives/ripple'
+import intersect from '../../directives/intersect'
 
 // Utilities
 import { convertToUnit, keyCodes } from '../../util/helpers'
@@ -42,7 +43,7 @@ const dirtyTypes = ['color', 'file', 'time', 'date', 'datetime-local', 'week', '
 export default baseMixins.extend<options>().extend({
   name: 'v-text-field',
 
-  directives: { ripple },
+  directives: { ripple, intersect },
 
   inheritAttrs: false,
 
@@ -84,6 +85,7 @@ export default baseMixins.extend<options>().extend({
     initialValue: null,
     isBooted: false,
     isClearing: false,
+    isVisible: false,
   }),
 
   computed: {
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -448,6 +465,10 @@ export default baseMixins.extend<options>().extend({
 
       VInput.options.methods.onMouseUp.call(this, e)
     },
+    onObserve (entries: IntersectionObserverEntry[]) {
```
johnleider: This function receives 3 arguments, entries, observer and isIntersecting. You can simply use this.

johnleider:
I like this solution. We should fabricate this into a function mixin so we can fix other areas of the framework such as text-field autofocus, tabs slider and probably many others.

packages/vuetify/src/components/VTextField/VTextField.ts
Djaler: Will fix. I just look in v-lazy component (https://github.com/vuetifyjs/vuetify/blob/058b1fe857d3aab1dbdb2143a69e66b9435dc1e9/packages/vuetify/src/components/VLazy/VLazy.ts#L58), so, maybe it can be simplified too

Djaler:
The only negative is that we need an intersection polyfill for whole framework, not the only VLazy and v-intersect

bdeo:
> The only negative is that we need an intersection polyfill for the whole framework, not the only VLazy and v-intersect. But, maybe this mixin can be written in such a way that it simple will do nothing if intersection API is not supported, instead of failing

This is a good point @Djaler

Djaler:
@johnleider should I try to implement this mixin or you'll merge this PR?

johnleider:
If you use v-lazy, v-img or v-intersect at all you have to include intersection observer regardless.

johnleider:
Closing this in favor of #9341

PR #9320, opened 2019-10-11
Typescript Declaration File Fix

donniewa:
## Description
- Updated the lib.d.ts declaration file to add back the module for `vuetify/lib/directives`.
  This matches the same module setup as the ES5 a-la-carte model.  As well as the documentation for loading the directives directly from `vuetify/lib/directives`

- According to the other files I noticed that the directives were expected to be in a directives object when being imported from `vuetify/lib`.
  I've updated the export to wrap the directives only from `vuetify/lib` to be encapsulated in the `directives` namespace.

Since the second point edits the actual entry-lib.ts file and changes the export of the directives, it's possible I've introduced a breaking change, or fixed an unintentional oversight.  I followed the documentation and other examples from the library to make an educated guess at what it should be.

## Motivation and Context
After upgrading vuetify from 2.0.7 to 2.1.1, my Typescript compilation starts failing due to the following:
```Try 'npm install @types/vuetify' if it exists or add a new declaration (.d.ts) file containing "declare module 'vuetify/lib/directives';"```

The Ripple import is failing to compile in Typescript.

I noticed that the Directive declaration was removed in this commit: https://github.com/vuetifyjs/vuetify/commit/64056b45fedb8781c9415812117923b0c83f505b#diff-4c39902ef939fb8c21742bb13af686c8L75

Also after doing some investigation it looks like the directives were under the vuetify/lib, inside a lowercased directives object, https://github.com/vuetifyjs/vuetify/blob/64056b45fedb8781c9415812117923b0c83f505b/packages/vuetify/types/lib.d.ts
but when I try to import them from there I get:
"export 'directives' was not found in 'vuetify/lib' so they were not being exported either

However when looking at the entry-lib.ts: https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/entry-lib.ts
I'm seeing that
export * from './directives'

which would result in being able to import Ripple directly. This fails normal compilation

 FAIL  Failed to compile
18:38 Module '"vuetify/lib"' has no exported member 'Ripple'.

## How Has This Been Tested?
I linked my existing project, all my errors went away with the current changes, my compiler runs properly.

I also built locally via the `yarn build` command, verified that linting passed via `yarn lint` and I ran the provided test suite via `yarn test`

```
Test Suites: 172 passed, 172 total
Tests:       52 skipped, 1317 passed, 1369 total
Snapshots:   556 passed, 556 total
```

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/entry-lib.ts
```diff
@@ -1,4 +1,5 @@
+import * as directives from './directives'
 export * from './components'
-export * from './directives'
+export { directives }
```
KaelWD: Why are you changing exports? It's just an incorrect typescript definition.

donniewa: according to the documentation and the other files it's expected that from vuetify/lib that the directives were supposed to be encapsulated in a 'directives' object.  I was trying to re-align it with what was here: https://github.com/vuetifyjs/vuetify/blob/64056b45fedb8781c9415812117923b0c83f505b/packages/vuetify/types/lib.d.ts 

If that's changed and the expectation is that the directives are individually exported just like the components, then the lib.d.ts is wrong and we'll have to update that as well.

KaelWD: Yes, types and documentation are just documentation. If someone is successfully using this without typescript, then this change would be a breaking one.

donniewa: So which one is correct? the documentation and types or the entry-lib? From everything I've seen this is what's expected, but maybe it was changed and the docs and types are out of date?  I just want the types to align, I'm not opposed to changing either, but it just started to happen in a recent update is all.

KaelWD: The source is correct.

donniewa:
@johnleider I just pushed the latest updates based on comments from @KaelWD.  I did have to change the import of the directives in the types test since the directives aren't encapsulated.

packages/vuetify/src/entry-lib.ts
donniewa: I've reverted this change back to what was in the lib as per the conversation.  Updated the typescript lib to match instead

PR #9341, opened 2019-10-13
fix(VTextField): use intersect to recalculate elements width

### LINKED ISSUE #9299: [Bug Report] Incorrect text field size calculation inside expansion panel

Djaler:
fixes #9299

## Description
Use IntersectionObserver (if it available) to recalculate element width after input becomes visible, 

## Motivation and Context
Cause they have zero width on mount in hidden state (inside expansion panel, for example)
https://github.com/vuetifyjs/vuetify/issues/9299

## How Has This Been Tested?
visually

## Markup:
https://codepen.io/djaler/pen/oNNXywM

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/mixins/intersectable/index.ts
+++ b/packages/vuetify/src/mixins/intersectable/index.ts
@@ -1,0 +1,49 @@
+// Directives
+import Intersect from '../../directives/intersect'
+
+// Utilities
+import { consoleWarn } from '../../util/console'
+
+// Types
+import Vue from 'vue'
+
+export default function intersectable (options: { onVisible: string[] }) {
+  if (typeof window === 'undefined' || !('IntersectionObserver' in window)) {
+    // do nothing because intersection observer is not available
+    return Vue.extend({ name: 'intersectable' })
+  }
+
+  return Vue.extend({
+    name: 'intersectable',
+
+    mounted () {
+      Intersect.inserted(this.$el as HTMLElement, {
+        name: 'intersect',
+        value: {
+          handler: this.onObserve,
+        },
+      })
+    },
+
+    destroyed () {
+      Intersect.unbind(this.$el as HTMLElement)
+    },
+
+    methods: {
+      onObserve (entries: IntersectionObserverEntry[], observer: IntersectionObserver, isIntersecting: boolean) {
+        if (!isIntersecting) return
+
+        for (let i = 0, length = options.onVisible.length; i < length; i++) {
+          const callback = (this as any)[options.onVisible[i]]
+
+          if (typeof callback === 'function') {
+            callback()
+            continue
+          }
+
+          consoleWarn(options.onVisible[i] + ' method is not available on the instance but referenced in intersectable mixin options')
+        }
+      },
+    },
+  })
+}
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -9,6 +9,7 @@ import VCounter from '../VCounter'
 import VLabel from '../VLabel'
 
 // Mixins
+import Intersectable from '../../mixins/intersectable'
 import Loadable from '../../mixins/loadable'
 
 // Directives
@@ -24,7 +25,14 @@ import { VNode } from 'vue/types'
 
 const baseMixins = mixins(
   VInput,
-  Loadable
+  Intersectable({
+    onVisible: [
+      'setLabelWidth',
+      'setPrefixWidth',
+      'setPrependWidth',
+    ],
+  }),
+  Loadable,
 )
 interface options extends InstanceType<typeof baseMixins> {
   $refs: {
--- a/packages/vuetify/src/mixins/intersectable/__tests__/intersectable.spec.ts
+++ b/packages/vuetify/src/mixins/intersectable/__tests__/intersectable.spec.ts
@@ -1,0 +1,39 @@
+import intersectable from '../index'
+
+import {
+  mount,
+  Wrapper,
+} from '@vue/test-utils'
+import { ComponentOptions } from 'vue'
+
+describe('intersectable.ts', () => {
+  let mountFunction: (options?: ComponentOptions<any>) => Wrapper<any>
+
+  beforeEach(() => {
+    mountFunction = (options?: ComponentOptions<any>) => {
+      return mount({
+        render: h => h('div'),
+        ...options,
+      })
+    }
+  })
+
+  it('should call callbacks when element is intersected', () => {
+    const callback = jest.fn()
+
+    const wrapper = mountFunction({
+      mixins: [intersectable({ onVisible: ['callback'] })],
+      methods: { callback },
+    })
+
+    expect(callback).not.toHaveBeenCalled()
+
+    wrapper.vm.onObserve([] as IntersectionObserverEntry[], null as any as IntersectionObserver, false)
+
+    expect(callback).not.toHaveBeenCalled()
+
+    wrapper.vm.onObserve([] as IntersectionObserverEntry[], null as any as IntersectionObserver, true)
+
+    expect(callback).toHaveBeenCalledTimes(1)
+  })
+})


Djaler:
@johnleider this is improved version of #9300

johnleider:
@Djaler Could you message me in the community please, https://community.vuetifyjs.com

Djaler:
@johnleider I updated a code according to what we discussed. Also, @sh7dm, I added a test to mixin

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        this.onObserve(entries)
+      })
+
+      observer.observe(this.$el)
+    },
+    methods: {
+      onObserve (entries: IntersectionObserverEntry[]) {
+        const isVisible = Boolean(entries.find(entry => entry.isIntersecting))
+
+        if (isVisible) {
```
johnleider: reduce block nesting. If !visible can return early

johnleider:
@KaelWD can you look over the mixin for inaccuracies in typescript implementation.

packages/vuetify/src/mixins/intersectable/index.ts
Djaler: @johnleider done

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
+import Vue from 'vue'
+import { consoleWarn } from '../../util/console'
+
+export default function intersectable (options: { onVisible: string[] }) {
+  if (!('IntersectionObserver' in window)) {
```
johnleider: this will break on SSR

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.onObserve(entries)
+      })
+
+      observer.observe(this.$el)
+    },
+    methods: {
+      onObserve (entries: IntersectionObserverEntry[]) {
+        const isVisible = Boolean(entries.find(entry => entry.isIntersecting))
+
+        if (!isVisible) {
```
johnleider: this can be one lined

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+          return
+        }
+
+        for (let i = 0, length = options.onVisible.length; i < length; i++) {
+          const callback = (this as any)[options.onVisible[i]]
```
johnleider: blank line between variable declaration and block logic

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+          if (typeof callback === 'function') {
```
johnleider: opt for early return in this conditional.

```
if (typeof callback === 'function') {
  callback()

  return
}
```

Djaler: But we need to call all callbacks, not just first

packages/vuetify/src/mixins/intersectable/index.ts
Djaler: Because of the `window`?

johnleider: You have to check for the existence of the window

packages/vuetify/src/mixins/intersectable/index.ts
johnleider: sorry, `continue`

Djaler:
@johnleider done

Djaler:
@joeledenberg can you check it?

Djaler:
@johnleider can you check it?

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    mounted () {
+      const observer = new IntersectionObserver((
+        entries: IntersectionObserverEntry[] = [],
+      ) => {
+        this.onObserve(entries)
+      })
+
+      observer.observe(this.$el)
+    },
```
johnleider: add line between mounted and methods

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    })
+  }
+
+  return Vue.extend({
+    name: 'intersectable',
+
+    mounted () {
+      const observer = new IntersectionObserver((
+        entries: IntersectionObserverEntry[] = [],
+      ) => {
```
johnleider: change to `) => this.onObserve(entries)`

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
+import Vue from 'vue'
+import { consoleWarn } from '../../util/console'
+
+export default function intersectable (options: { onVisible: string[] }) {
+  if (typeof window === 'undefined' || !('IntersectionObserver' in window)) {
+    return Vue.extend({
+      name: 'intersectable',
+      // do nothing because intersection observer is not available
```
johnleider: Move to line 6

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
+import Vue from 'vue'
+import { consoleWarn } from '../../util/console'
+
+export default function intersectable (options: { onVisible: string[] }) {
+  if (typeof window === 'undefined' || !('IntersectionObserver' in window)) {
+    return Vue.extend({
+      name: 'intersectable',
```
johnleider: one line this, `return Vue.extend({ name: intersectable })`

Djaler:
How can I add directive from mixin?

Djaler:
@johnleider how can we add directive from mixin?

KaelWD:
> @KaelWD can you look over the mixin for inaccuracies in typescript implementation.

It's alright. Could make the options type safe by being a wrapping function instead of a mixin factory, but that gets pretty complicated.

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+      // do nothing because intersection observer is not available
+    })
+  }
+
+  return Vue.extend({
+    name: 'intersectable',
+
+    mounted () {
+      const observer = new IntersectionObserver((
+        entries: IntersectionObserverEntry[] = [],
```
KaelWD: Don't need this, the type is inferred and correct me if I'm mistaken but isn't `entries` always defined?

packages/vuetify/src/mixins/intersectable/index.ts
```diff
@@ -0,0 +1,43 @@
...
+      name: 'intersectable',
+      // do nothing because intersection observer is not available
+    })
+  }
+
+  return Vue.extend({
+    name: 'intersectable',
+
+    mounted () {
+      const observer = new IntersectionObserver((
```
KaelWD: Do we need to tear this down when the component is destroyed?

Djaler:
> The mixin could have everything else though and the implementing component would just have to make sure to use the directive somewhere.

Do you mean, in implementing component we will have something like 
```
directives: [
          {
            name: 'intersect',
            value: {
              handler: this.onObserve,
            },
          },
        ],
```, where `onObserve` provided by mixin?

KaelWD:
You can't do that, implementing component would have https://github.com/vuetifyjs/vuetify/pull/9300/files#diff-dc0c0fccb5e57dfaca5093b2096562fdR386-R393 in its render function somewhere. See my edit for another idea.

Djaler:
> ```js
> mounted () {
>   Intersect.inserted(this.$el, options)
> }
> destroyed () {
>   Intersect.unbind(this.$el)
> }
> ```

Do you mean to get rid of mixin at all?

KaelWD:
No, the mixin can do that.

Djaler:
Oh, I did not get it that Intersection.inserted is a way to initialize directive. Yeah, this could work

Djaler:
@KaelWD @johnleider done

Djaler:
So what about it, guys?

PR #9535, opened 2019-11-01
fix(VFileInput): should be File object if you don't set the multiple prop

### LINKED ISSUE #9530: [Bug Report] VFileInput multiple="false" changes the value for rules
### Environment
**Vuetify Version:** 2.1.7
**Vue Version:** 2.6.10
**OS:** Mac OS 10.15.1

### Steps to reproduce
See codepen in the reproduction link.
1. Add a file in the first file input and the value parameter in the rules function is a File object. See console.
2. Add a file in the second file input and the File object is wrapped in an array. This should be the same as in step 1.

### Expected Behavior
The value parameter of the rules functions should be a File object if you don't set the multiple property or set multiple to false.

### Actual Behavior
If you don't set the multiple property, the value in the rules function is a File object as it should be. If you set multiple to false the value is an array which contains the File object.

### Reproduction Link
<a href="https://codepen.io/Aukemid/pen/BaawqEM" target="_blank">https://codepen.io/Aukemid/pen/BaawqEM</a>

### Other comments
I believe this is caused by VFileInput.isMultiple which only checks whether the multiple property exists. The following change works in my test cases.

```javascript
isMultiple() {
  // return this.$attrs.hasOwnProperty('multiple');
  return this.$attrs.hasOwnProperty('multiple') && this.$attrs.multiple !== false;
},
```

nkoik:
…prop or set multiple to false

fix #9530

If multiple property added to false, should be return a File object like if you don't provide the prop at all.

## Description

This pull req. fixes the bug described in #9530 . Added extra check for multiple prop if is truthy.

## How Has This Been Tested?

visually
## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VFileInput/VFileInput.ts
```diff
@@ -115,7 +115,7 @@ export default VTextField.extend({
       return this.isDirty
     },
     isMultiple (): boolean {
-      return this.$attrs.hasOwnProperty('multiple')
+      return this.$attrs.hasOwnProperty('multiple') && this.$attrs.multiple
```
nekosaur: Hmm, why are we even checking this through $attrs, and not just `this.multiple`?

nkoik: Yes you are right, maybe confused with an input element.

Aukemid: In #9530 I compared the multiple attribute to false because it was an empty string when adding the attribute without a value, which should result in true. The isMultiple should return the correct boolean in these cases:
```html
<v-file-input></v-file-input> // false
<v-file-input multiple></v-file-input> // true
<v-file-input :multiple="false"></v-file-input> // false
<v-file-input :multiple="true"></v-file-input> // true
```

I don't know enough about the Vuetify core to fix this properly, but I also noticed that "multiple" isn't defined in the props either. Isn't this required for it to be processed as a boolean?

nekosaur: yes. @nkoik add multiple prop definition and use `this.multiple` instead.

packages/docs/src/lang/en/components/FileInputs.json
```diff
@@ -51,7 +51,7 @@
     "chips": "Components.Selects.props.chips",
     "counterSizeString": "The text displayed when using the **counter** and **show-size** props. Can also be customized globally on the [internationalization page](/customization/internationalization).",
     "counterString": "The text displayed when using the **counter** prop. Can also be customized globally on the [internationalization page](/customization/internationalization).",
-    "multiple": "Adds the **multiple** attribute to the input, allowing multiple file selections.",
+    "multiple": "Allowing multiple file selections.",
```
johnleider: Is there a reason for changing this?

packages/vuetify/src/components/VFileInput/VFileInput.ts
```diff
@@ -70,6 +70,7 @@ export default VTextField.extend({
         return typeof val === 'object' || Array.isArray(val)
       },
     } as PropValidator<File | File[]>,
+    multiple: Boolean,
```
johnleider: Defining this here breaks text-field propagating the **multiple** attribute to the input. https://github.com/vuetifyjs/vuetify/blob/0f3fa6a9204688abe23996d2df2be665f52a7e18/packages/vuetify/src/components/VTextField/VTextField.ts#L363

packages/vuetify/src/components/VFileInput/VFileInput.ts
```diff
@@ -115,7 +116,7 @@ export default VTextField.extend({
       return this.isDirty
     },
     isMultiple (): boolean {
-      return this.$attrs.hasOwnProperty('multiple')
+      return this.multiple
```
johnleider: The reporting issue had a one liner to solve this issue, is there a reason why we can't use that instead?

```return this.$attrs.hasOwnProperty('multiple') && this.$attrs.multiple !== false```

packages/vuetify/src/components/VFileInput/VFileInput.ts
nekosaur: Ah, I see. At a glance it is not at all clear why it has to be done with `$attrs` instead of an idiomatic prop. So let's add an explanatory comment in the `isMultiple` computed prop.

johnleider:

PR #9552, opened 2019-11-02
Fix(VInput): Browser Autocomplete

### LINKED ISSUE #9516: [Bug Report] input field in v-autocomplete should be autocomplete="off" instead of disabled
### Environment
**Vuetify Version:** 2.1.7
**Vue Version:** 2.6.10
**OS:** Linux x86_64

### Steps to reproduce
Write many many times e.g. "Comboboxes" and after a while Google Chrome gives you a second "dropdown" with "Comboboxes" this is because the input field is like this:

```
<input id="input-4" type="text" autocomplete="disabled">
```

but should be rather like this:

```
<input id="input-4" type="text" autocomplete="off">
```

### Expected Behavior
no extra dropdown from Chrome

### Actual Behavior
second overlaying dropdown from Chrome

### Reproduction Link
<a href="https://codepen.io/samuelba/pen/eYYGGeN" target="_blank">https://codepen.io/samuelba/pen/eYYGGeN</a>

melbahja:
## Description
Disable browser autofill menu on VSelect, VCombobox, VAutocomplete

## Motivation and Context
fixes #9516

## How Has This Been Tested?
unit | visually

## Types of changes

## Checklist:


packages/docs/src/lang/en/components/Selects.json
```diff
@@ -2,6 +2,7 @@
   "heading": "# Selects",
   "headingText": "Select fields components are used for collecting user provided information from a list of options.",
   "objectAlert": "When using objects for the **items** prop, you must associate **item-text** and **item-value** with existing properties on your objects. These values are defaulted to **text** and **value** and can be changed.",
+  "autocompleteAlert": "Browser autocomplete are set to off by default, and may vary given that browsers such as chrome may ignore it. [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion)",
```
johnleider:
```suggestion
  "autocompleteAlert": "Browser autocomplete is set to off by default and may vary by browser and may be ignored. [MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion)",
```

lilalinux:
At least Chrome takes the attribute autocomplete literally, so "off" is no better than any other fixed value. The recommended fix (according to google) is to use random values (or semantic random values):
https://gist.github.com/niksumeiko/360164708c3b326bd1c8#gistcomment-2956701

melbahja:
@lilalinux
> The problem is, if the user comes back to fill this form again, my-special-phonenumber-field isn't a new thing anymore. Chrome autocompletes. In my case, it does it wrong and modifies some fields it should not. which is boring as people validate a form with modified values and they don't notice....

The problem here is you need a unique value each and every time. and how about other browsers?
I think maybe if we can use selects without `input` HTML tag like MD web:
https://material.io/develop/web/components/input-controls/select-menus/

lilalinux:
I *think* random values would work for other browsers, too. I don't see a problem in creating a random value each time the input is used, why not setting a UUID via JS?

An alternative approach of the linked discussion seems to be dynamically setting the input readonly.

melbahja:
Yes I think `readonly` with `autocomplete="off"` should works as expected.

packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts
```diff
@@ -283,7 +283,8 @@ export default VSelect.extend({
 
       input.data = input.data || {}
       input.data.attrs = input.data.attrs || {}
-      input.data.attrs.autocomplete = input.data.attrs.autocomplete || 'disabled'
+      input.data.attrs.readonly = true
```
jacekkarczmarczyk: That makes autocomplete and combobox readonly

melbahja: I didn't think of that :man_facepalming:

johnleider:
Requesting final comments @vuetifyjs/core-contributors @vuetifyjs/core-team

PR #9621, opened 2019-11-07
fix(VTooltip): make tooltip activator visible on SSR initial render

adrienbaron:
## Description
Remove `display: none;` from `.v-tooltip` class.

## Motivation and Context
Upon initial SSR rendering, the activator for a tooltip is contained inside the root of the VTooltip component like so:
```html
<span class="v-tooltip v-tooltip--bottom">
  <div
    class="v-tooltip__content"
    style="left:0px;max-width:auto;opacity:0;top:12px;z-index:0;display:none;"
  ></div>
  <div>ACTIVATOR</div>
</span>
```

Because `v-tooltip` is `display: none;` the activator is not visible.

When JS kicks in the DOM [gets modified by the `beforeMount` of `Detachable`](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/detachable/index.ts#L61) and the activator gets moved out like so:
```html
<div>ACTIVATOR</div>
<span class="v-tooltip v-tooltip--bottom">
  <div class="v-tooltip__content" style="left:0px;max-width:auto;opacity:0;top:12px;z-index:0;display:none;"></div>
</span>
```

Removing this `display: none;` fix the SSR issue as the activator gets displayed before JS kicks in. Moreover as `tooltip__content` is empty before JS kicks in and also have `display:none;` set, this change shouldn't have any side effect on that front.

## How Has This Been Tested?

I've tested this by overriding this style on my project that uses Nuxt + Vuetify and checking it fixes the SSR issue. I didn't notice any side effects, tooltips still work as expected.
I've also tested in the playground and it also seems to work without issue.

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-tooltip bottom>
        <v-btn v-on="on">Hover me</v-btn>
      </template>
      This is a tooltip
    </v-tooltip>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
    //
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTooltip/VTooltip.sass
+++ b/packages/vuetify/src/components/VTooltip/VTooltip.sass
@@ -1,8 +1,6 @@
 @import ./_variables
 
 .v-tooltip
-  display: none
-
   &--attached
     display: inline
 


KaelWD:
Hmm I forgot about SSR when I did that. I'm going to have to do my detachable rewrite to fix this, having the empty element can break flex layouts which is why it was hidden in the first place.

adrienbaron:
@KaedWD 
Would working on a potential reimplemation of detachable in a codesandbox be useful? Or do you already have a clear idea of how to go about it?

KaelWD:
Yeah I've already got a plan for it, thanks for the offer though.

PR #9636, opened 2019-11-09
fix(VDataIterator): emit two item-selected events in single-select mode

### LINKED ISSUE #8477: [Bug Report] Data Table - @item-selected emit once
### Environment
**Vuetify Version:** 2.0.0
**Vue Version:** 2.6.6
**OS:** Windows 10

### Steps to reproduce
Use a datatable
Put it in single select mode
Show the select using show-select
Detect the item-selected event and console it

Now check one row, after this check the other one without uncheck the first.
Only one event is throw, for the new checked value, but the uncheck value is not throw.

### Expected Behavior
When we check one row then another one, it should throw tho events in a row : the unchecked row with false and the checked row with true

### Actual Behavior
We just need to check one row, in my sample, go in the developer console
"Item : Jack M. value : true"

If we uncheck, there is it
"Item : Jack M. value : false"

But because I'm using the single select, when we check one row, if there is one already checked, this also uncheck the second one. That's a good point and what we expect.

I'm looking at the console log and I just see the 
"Item : John D. value : true"
But there is no 
"Item : Jack M. value : false"

The event didn't throw the false automatic value changed, but only the new true one.

### Reproduction Link
<a href="https://codepen.io/audricch/pen/ymrqZY" target="_blank">https://codepen.io/audricch/pen/ymrqZY</a>

### Other comments
I'm pretty sure it's not a bug, but I also need to understand why it works like this and how to do simply to have what I want.

nekosaur:
## Description
bug where in single-select mode we did not emit a 'deselect' item-select event for previously selected item when selecting a new one.

## Motivation and Context
closes #8477

## How Has This Been Tested?
playground, unit test

## Markup:

<details>

```vue
<template>
  <div>
    <v-data-table :headers="header" :items="people" single-select show-select @item-selected="itemSelected">
      <template v-slot:item.libelle="{ item }">
        <td class="text-xs">{{ item.libelle }}</td>
      </template>
    </v-data-table>
  </div>
</template>

<script>
export default {
  data: () => ({
    header: [
      {
        value: 'name',
        text: 'Name'
      }
    ],
    people: [
      {
        id: 1,
        name: 'John D.'
      },
      {
        id: 2,
        name: 'Jack M.'
      }
    ]
  }),
  methods: {
    itemSelected(item) {
      console.log('Item : ' + item.item.name + ' value : ' + item.value)
    }
  }
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDataIterator/VDataIterator.ts
+++ b/packages/vuetify/src/components/VDataIterator/VDataIterator.ts
@@ -154,6 +154,11 @@ export default Themeable.extend({
       if (value) selection[key] = item
       else delete selection[key]
 
+      if (this.singleSelect && emit) {
+        const keys = Object.keys(this.selection)
+        const old = keys.length && getObjectValueByPath(this.selection[keys[0]], this.itemKey)
+        old && old !== key && this.$emit('item-selected', { item: this.selection[old], value: false })
+      }
       this.selection = selection
       emit && this.$emit('item-selected', { item, value })
     },
--- a/packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
+++ b/packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
@@ -550,4 +550,30 @@ describe('VDataTable.ts', () => {
 
     expect(itemsPerPage).toHaveBeenCalledWith(-1)
   })
+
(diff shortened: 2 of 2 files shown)


johnleider:
This is not happening on master, but on your branch it is:
image

PR #9750, opened 2019-11-21
fix(vdatatable): add display style to <tr> when Mobile

### LINKED ISSUE #8799: v-data-table with tooltip breaking design in Chrome mobile view
### Environment
**Vuetify Version:** 2.0.11
**Vue Version:** 2.6.10
**OS:** Windows 10

### Steps to reproduce
Create a data table with a column that contains a tooltip and at least 2 additional columns afterwards. so for example:
- column 1
- column 2 with tooltip
- column 3
- column 4

If you then resize the viewport then down to the mobile view, then rendering of the columns below the one with the tooltip breaks.

Instead of the expected
- column 1
- column 2 with tooltip
- column 3
- column 4

it actually looks like
- column 1                       |  - column 3
- column 2 with tooltip  |  - column 4

This does not happen when loading the page directly in the mobile view, only when resizing it down from desktop to mobile version.

### Expected Behavior
All columns are arranged under each other and not in two columns.

### Actual Behavior
The columns AFTER the one with tooltip are placed on a second column.

### Reproduction Link
<a href="https://codepen.io/DonDozone/pen/MWgvYvE" target="_blank">https://codepen.io/DonDozone/pen/MWgvYvE</a>

### Other comments
This only happens in Chrome for me, not in Firefox or Edge. I can also reproduce it on my phone with chrome browser

YipingRuan:
fix #8799

## Description
Override user agent stylesheet for chrome <tr>

## Motivation and Context
https://github.com/vuetifyjs/vuetify/issues/8799

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <div id="app">
  <v-app>
    <v-content>
      <v-data-table :headers="headers" :items="items" disable-pagination hide-default-footer>
              <template v-slot:item.tooltip="{ item }">
              <v-tooltip bottom>
                  <span v-on="on">{{item.tooltip}}</span>
                </template>
                <span>This is the tooltip</span>
              </v-tooltip>
            </template>
        </v-data-table>
    </v-content>
  </v-app>
</div>
</template>

<script>
export default {
  data: () => ({
    headers: [
      { text: 'Header 1', value: 'valueA' },
      { text: 'Tooltip', value: 'tooltip' },
      { text: 'Header after', value: 'valueC' },
      { text: 'Header 2 after', value: 'valueD' },
    ],
    items: [{
       valueA: 'asd',
       tooltip: 2,
       valueC: 12345,
       valueD: 12345
    }
    ]
    //
  })
}
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDataTable/VDataTable.ts
+++ b/packages/vuetify/src/components/VDataTable/VDataTable.ts
@@ -514,7 +514,9 @@ export default VDataIterator.extend({
         height: this.height,
         fixedHeader: this.fixedHeader,
         dense: this.dense,
-        isMobile: this.isMobile,
+        tableClasses: {
+          'v-data-table__mobile': this.isMobile,
+        },
       }
 
       // if (this.virtualRows) {
--- a/packages/vuetify/src/components/VDataTable/VDataTable.ts
+++ b/packages/vuetify/src/components/VDataTable/VDataTable.ts
@@ -514,6 +514,7 @@ export default VDataIterator.extend({
         height: this.height,
         fixedHeader: this.fixedHeader,
         dense: this.dense,
+        isMobile: this.isMobile,
       }
 
       // if (this.virtualRows) {
--- a/packages/vuetify/src/components/VDataTable/VDataTable.ts
+++ b/packages/vuetify/src/components/VDataTable/VDataTable.ts
@@ -515,7 +515,7 @@ export default VDataIterator.extend({
         fixedHeader: this.fixedHeader,
         dense: this.dense,
         tableClasses: {
-          'v-data-table__mobile': this.isMobile,
+          'v-data-table--is-mobile': this.isMobile,
         },
       }
 
--- a/packages/vuetify/src/components/VDataTable/VDataTable.ts
+++ b/packages/vuetify/src/components/VDataTable/VDataTable.ts
@@ -514,9 +514,6 @@ export default VDataIterator.extend({
         height: this.height,
         fixedHeader: this.fixedHeader,
         dense: this.dense,
-        tableClasses: {
-          'v-data-table--is-mobile': this.isMobile,
-        },
       }
 
       // if (this.virtualRows) {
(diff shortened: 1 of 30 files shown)


YipingRuan:
Before:
Before
After:
After

packages/vuetify/src/components/VDataTable/VSimpleTable.ts
```diff
@@ -33,7 +34,7 @@ export default mixins(Themeable).extend({
           height: convertToUnit(this.height),
         },
       }, [
-        this.$createElement('table', this.$slots.default),
+        this.$createElement('table', { class: { 'v-data-table__mobile': this.isMobile } }, this.$slots.default),
```
nekosaur: Apply this class from v-data-table since it is a data-table specific class

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -514,6 +514,9 @@ export default VDataIterator.extend({
         height: this.height,
         fixedHeader: this.fixedHeader,
         dense: this.dense,
+        tableClasses: {
+          'v-data-table__mobile': this.isMobile,
```
johnleider:
```suggestion
          'v-data-table--is-mobile': this.isMobile,
```

packages/vuetify/src/components/VDataTable/VSimpleTable.ts
```diff
@@ -12,6 +12,7 @@ export default mixins(Themeable).extend({
     dense: Boolean,
     fixedHeader: Boolean,
     height: [Number, String],
+    tableClasses: Object,
```
johnleider: Should have a default value

```
tableClasses: {
  type: Object,
  default: () => ({}),
}
```

packages/vuetify/src/components/VDataTable/VSimpleTable.ts
```diff
@@ -33,7 +34,7 @@ export default mixins(Themeable).extend({
           height: convertToUnit(this.height),
         },
       }, [
-        this.$createElement('table', this.$slots.default),
+        this.$createElement('table', { class: { ...this.tableClasses } }, this.$slots.default),
```
johnleider: This should be done differently, there is no reason change the snapshots.

packages/vuetify/src/components/VDataTable/VSimpleTable.ts
YipingRuan: Tried move the mobile css to tr at MobileRow

PR #9823, opened 2019-11-29
fix(vbtn): fix VBtn custom loader content overflow

### LINKED ISSUE #9672: [Bug Report] Button size is not changed to accommodate custom loader content
### Environment
**Vuetify Version:** 2.1.9
**Vue Version:** 2.6.10
**Browsers:** Chrome 77.0.3865.120
**OS:** Windows 10

### Steps to reproduce
When using <template v-slot:loader> to define custom loader content for button, if this content is wider than original button text then content will overflow to the outside of button.

### Expected Behavior
Accommodate button size (width) so that custom loader template can fit in.

### Actual Behavior
Button size stays the same and in case loader content is larger than that it will go outside button.

### Reproduction Link
<a href="https://codepen.io/milost1982/pen/WNNgwoq?&editable=true&editors=111" target="_blank">https://codepen.io/milost1982/pen/WNNgwoq?&editable=true&editors=111</a>

### Other comments
Reproduction link is based on the example from the loaders sections in the v-btn page https://vuetifyjs.com/en/components/buttons

yazfield:
## Description
Custom loader content overflows when it grows bigger than it's parent

## Motivation and Context
fixes #9672

Fix VBtn custom loader content overflow issue by allowing custom content to expand it's parent (VBtn)

## How Has This Been Tested?
visually

## Reproduction link:
From the reported issue #9672 https://codepen.io/janvennemann/pen/gOOdpvq

## Types of changes

## Checklist:

## Notes:
I'm not sure if: 
- `.v-btn__content` depends on `opacity` for some animations
- something relies on `.v-btn__loader` to be absolutely positioned
Pointers will be appreciated :)

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.sass
+++ b/packages/vuetify/src/components/VBtn/VBtn.sass
@@ -135,7 +135,7 @@
   height: 100%
   justify-content: center
   left: 0
-  position: absolute
+  position: relative
   top: 0
   width: 100%
 
@@ -241,7 +241,7 @@
   transition: none
 
   .v-btn__content
-    opacity: 0
+    display: none
 
 .v-btn--outlined
   border: $btn-outline-border-width solid currentColor


yazfield:
@jacekkarczmarczyk  thanks I missed that. 
Do you have an idea about how the fix would look like?
I'm thinking of adding a css modifier class `v-btn--custom-loading` and alternate the style changes from the PR there (only apply them if there is a custom loader.

jacekkarczmarczyk:
I'm not sure if it's a bug, loading state shouldn't change the size of the button so imho it should be up to dev to assure that custom loader is not bigger that button content

yazfield:
@jacekkarczmarczyk makes sense to me. could you flag the issue as won't fix?

PR #9912, opened 2019-12-08
fix(VBtn): adjust loading indicator size with the x-small prop

### LINKED ISSUE #9870: [Bug Report] v-btn: Loading indicator too large on small buttons
### Environment
**Vuetify Version:** 2.1.12
**Vue Version:** 2.6.10
**Browsers:** Chrome 78.0.3904.108
**OS:** Windows 7

### Steps to reproduce
1. create v-btn
2. set property x-small to true
3. set property loading to true

### Expected Behavior
The loading indicator should not be bigger than the button.

### Actual Behavior
The loading indicator is bigger than the button.

### Reproduction Link
<a href="https://codepen.io/tinooooo/pen/dyPPGOr" target="_blank">https://codepen.io/tinooooo/pen/dyPPGOr</a>

yoongti:
## Motivation and Context

https://github.com/vuetifyjs/vuetify/issues/9870

## How Has This Been Tested?

Test on Playground

none

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-btn
      x-small
      loading
    ></v-btn>
    <v-btn
      small
      loading
    ></v-btn>
    <v-btn
      x-large
      loading
    ></v-btn>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
    //
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

- [X] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and breaking changes).


jacekkarczmarczyk:
How about 60% width/height for all sizes?
https://codepen.io/jkarczm/pen/JjoXjxG

leeyt54:
Way better, I'll commit shortly using your solution

leeyt54:
@jacekkarczmarczyk I tried using style 60% of height and width on v-progress-circular. The build fails because viewBoxSize and others depends on size. Now using computed to compute size but I still couldn't get the test case to pass. Is there anything that I miss?

jacekkarczmarczyk:
@yoongti you need to refresh the snapshots (`yarn test -u`).
Was the tests the only issue with the percentage version?

yoongti:
@jacekkarczmarczyk I get command "test" not found when I use `yarn test -u` as it is not included in `package.json`. `yarn lerna run test -u` I get unknown argument: u. `yarn lerna run test` works for me but it doesn't refresh.

Another issue I see is that the computedSize is a hardcoded value(60% of the btn-size, for example, 12 for x-small (20), 16.8 for small(28)...). Again, it was set to pass the test case. Perhaps I should use 20 * 0.6 for better understanding for future developers. What do you think?

johnleider:
I almost wonder if we can just target this with css and skip the explicit prop. Then we wouldn't have 2 separate sets of sizing values, one in component and one in sass. Thoughts?

jacekkarczmarczyk:
Targetting with css is an option too, regarding testing - did you run it from the packages/vuetify folder or from the root?

yoongti:
Targeting with css definitely feels more elegant, but would have to set it with !important to override as calculateSize will take prop `size`(default to 32) to set the `height` and `width` of the loading indicator. Would that be okay?

It was my mistake, I was running the command at root. It works when I run it in packages/vuetify, thanks @jacekkarczmarczyk

johnleider:
Still need to remove line 171.

packages/vuetify/src/components/VBtn/VBtn.sass
```diff
@@ -241,6 +241,10 @@
 
   .v-btn__content
     opacity: 0
+  
+  .v-progress-circular 
+    width: 60% !important
```
johnleider: This needs to be done without **important** and should also be in `_variables.scss`

johnleider:
I'm going to close this and work on the changes locally if there is no progress on this today. Thank you!

Issue #9849, opened 2019-12-02
[Feature Request] Pagination Accessibility


JohnMica:
### Problem to solve
Accessibility issues with the pagination buttons  / currently selected page

### Proposed solution
introduce options for the buttons to have aria-* attributes
introduce options for the page-numbers to have aria-* attributes

jacekkarczmarczyk:
Please specify which exactly attributes should be added

JohnMica:
seems that the least attributes needed are
- wrapping element `aria-label="Pagination Navigation"`
- buttons (prev / next) - `aria-label="Prev / Next page"`
- page numbers - `aria-label="page X"`
- current page - `aria-current="true"`

If there are supposed to be more attributes I will investigate and let you know about them. The ones mentioned above should be the minimum needed, I guess.

JohnMica:
perhaps the values of aria-attributes could be editable - I'm thinking of language translations that could allow them to be modified ?

jacekkarczmarczyk:
They should just go to language files https://github.com/vuetifyjs/vuetify/tree/master/packages/vuetify/src/locale

abnersajr:
I would like to take this one. Is it possible to assign me to this one?
I had some previous experience with this.
I also agree that we can use `i18n` to the values.

abnersajr:
PR is opened. Please feel free to review and request changes if necessary.

wusongliang:
Thank you. Has the problem been solved?

johnleider:
Closed via #9926

If you have any additional questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

PR #9926, opened 2019-12-10
feat(VPagination): add a11y support

abnersajr:
## Description

This PR adds `aria-*` attributes improving accessibility of the pagination component.
Another feature is the component localization with translations to all available locations.

## Motivation and Context

It fixes the issue #9849 following all the instructions requested.

## How Has This Been Tested?

I manually tested and also the unit test was fixed to work with the new changes.

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-row justify="center">
        <v-col cols="8">
          <v-container class="max-width">
            <v-pagination
              v-model="page"
              class="my-4"
              :length="15"
            ></v-pagination>
          </v-container>
        </v-col>
      </v-row>
  </v-container>
</template>

<script>
export default {
  data: () => ({
    page: 1,
  })
}
</script>

```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/locale/en.ts
```diff
@@ -44,4 +44,13 @@ export default {
     am: 'AM',
     pm: 'PM',
   },
+  pagination: {
+    ariaLabel: {
+      wrapper: 'Pagination Navigation',
+      next: 'Next page',
+      previous: 'Previous page',
+      page: 'Goto Page {0}',
```
jacekkarczmarczyk:
```suggestion
      page: 'Goto page {0}',
```

packages/vuetify/src/locale/en.ts
```diff
@@ -44,4 +44,13 @@ export default {
     am: 'AM',
     pm: 'PM',
   },
+  pagination: {
+    ariaLabel: {
+      wrapper: 'Pagination Navigation',
```
jacekkarczmarczyk:
```suggestion
      wrapper: 'Pagination navigation',
```

packages/vuetify/src/locale/en.ts
```diff
@@ -44,4 +44,13 @@ export default {
...
+      next: 'Next page',
+      previous: 'Previous page',
+      page: 'Goto Page {0}',
+      currentPage: 'Current Page, Page {0}',
```
jacekkarczmarczyk:
```suggestion
      currentPage: 'Current page: {0}',
```
This sounds better for me, applies for other languages too

packages/vuetify/src/locale/pl.ts
```diff
@@ -44,4 +44,13 @@ export default {
     am: 'AM',
     pm: 'PM',
   },
+  pagination: {
+    ariaLabel: {
+      wrapper: 'Nawigacja paginacyjna',
```
jacekkarczmarczyk:
```suggestion
      wrapper: 'Paginacja',
```

packages/vuetify/src/components/VPagination/VPagination.ts
```diff
@@ -179,22 +185,30 @@ export default mixins(Colorable, Themeable).extend({
...
+        }],
+        class: this.classes,
+      }, children)
+    },
   },
 
   render (h): VNode {
     const children = [
-      this.genIcon(h, this.$vuetify.rtl ? this.nextIcon : this.prevIcon, this.value <= 1, this.previous),
+      this.genIcon(h, this.$vuetify.rtl ? this.nextIcon : this.prevIcon, this.value <= 1, this.previous, this.$vuetify.lang.t('$vuetify.pagination.ariaLabel.previous')),
```
johnleider: This should be a prop so the user can change it.

packages/vuetify/src/components/VPagination/VPagination.ts
```diff
@@ -179,22 +185,30 @@ export default mixins(Colorable, Themeable).extend({
...
       this.genItems(h),
-      this.genIcon(h, this.$vuetify.rtl ? this.prevIcon : this.nextIcon, this.value >= this.length, this.next),
+      this.genIcon(h, this.$vuetify.rtl ? this.prevIcon : this.nextIcon, this.value >= this.length, this.next, this.$vuetify.lang.t('$vuetify.pagination.ariaLabel.next')),
```
johnleider: This should be a prop so the user can change it.

packages/vuetify/src/components/VPagination/VPagination.ts
```diff
@@ -179,22 +185,30 @@ export default mixins(Colorable, Themeable).extend({
...
-        modifiers: { quiet: true },
-        name: 'resize',
-        value: this.onResize,
-      }],
-      class: this.classes,
-    }, children)
+    return h('nav', {
+      attrs: {
+        role: 'navigation',
+        'aria-label': this.$vuetify.lang.t('$vuetify.pagination.ariaLabel.wrapper'),
```
johnleider: This should be a prop so the user can change it.

johnleider:
I've moved this to the `dev` branch since it's a feature.

abnersajr:
@johnleider I think in case we decide to go as `props` I think this will lose the idea of be localized.
What I can consider is make the change through props as an alternative. And let the localized as default.
I would prefer make this change in another PR. Even it's nothing big, but to be separated the ideas.
What you think?

jacekkarczmarczyk:
@abnersajr prop can have the default value pointing to the locales, this will give possibility to change the value globally (in locales) or per component (in prop). See for example https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VAlert/VAlert.ts#L44

johnleider:
@abnersajr pulse check on this.

abnersajr:
@johnleider sorry. Things are going crazy this year begin. I will solve it by tomorrow.👊

abnersajr:
@johnleider @jacekkarczmarczyk all comments done.

abnersajr:
I made a mess with the commits, I'm trying to solve this. In case someone have a suggestion please let me know.

johnleider:
Is this non-breaking? If so, I believe we can get it in as a bug fix seeing as it pertains to a11y.

PR #9961, opened 2019-12-13
fix(VSelect): fix duplicated ID

### LINKED ISSUE #9819: [Bug Report] Selects dropdown has duplicate id.
### Environment
**Vuetify Version:** 2.1.10
**Vue Version:** 2.6.10
**Browsers:** Firefox 70.0
**OS:** Windows 10

### Steps to reproduce
Open devtools you will see the dropdown has duplicate id. https://www.screencast.com/t/klzE3y52

### Expected Behavior
Id should be unique

### Actual Behavior
Have tow elements with same id

### Reproduction Link
<a href="https://codepen.io/hkeyjun/pen/mddNXqG" target="_blank">https://codepen.io/hkeyjun/pen/mddNXqG</a>

abnersajr:
Added an Index based ID avoiding duplication

fix #9819

## Description

Select options were being generated with duplicated id. It was fixed appending to the old ID already generated an Index.

## Motivation and Context

Duplicated IDs aren't allowed in HTML. This PR fix #9819 

## How Has This Been Tested?

This was tested in the playground with developer tools.
## Markup:

<details>

```vue
<template>
  <div id="app">
  <v-app id="inspire">
    <v-container fluid>
      <v-row align="center">
        <v-col class="d-flex" cols="12" sm="6">
          <v-select
            :items="items"
            label="Standard"
          ></v-select>
        </v-col>
  
        <v-col class="d-flex" cols="12" sm="6">
          <v-select
            :items="items"
            filled
            label="Filled style"
          ></v-select>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</div>
</template>

<script>
export default {
  data: () => ({
    items: ['Foo', 'Bar', 'Fizz', 'Buzz'],
  }),
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelectList.ts
+++ b/packages/vuetify/src/components/VSelect/VSelectList.ts
@@ -127,9 +127,7 @@ export default mixins(Colorable, Themeable).extend({
       return `<span class="v-list-item__mask">${escapeHTML(text)}</span>`
     },
     genLabelledBy (item: object) {
-      const text = escapeHTML(this.getText(item).split(' ').join('-').toLowerCase())
-
-      return `${text}-list-item-${this._uid}`
+      return `list-item-${this._uid}`
     },
     getMaskedCharacters (text: string): {
       start: string
--- a/packages/vuetify/src/components/VSelect/VSelectList.ts
+++ b/packages/vuetify/src/components/VSelect/VSelectList.ts
@@ -31,6 +31,8 @@ import mixins from '../../util/mixins'
 import { VNode } from 'vue'
 import { PropValidator } from 'vue/types/options'
 
+type ListTile = { item: any, disabled?: null | boolean, value?: boolean, index: number };
+
 /* @vue/component */
 export default mixins(Colorable, Themeable).extend({
   name: 'v-select-list',
@@ -149,7 +151,7 @@ export default mixins(Colorable, Themeable).extend({
       index,
       disabled = null,
       value = false,
-    }: { item: any, disabled?: null | boolean, value?: boolean, index: number }): VNode | VNode[] | undefined {
+    }: ListTile): VNode | VNode[] | undefined {
       if (!value) value = this.hasItem(item)
 
       if (item === Object(item)) {
--- a/packages/vuetify/src/components/VSelect/VSelectList.ts
+++ b/packages/vuetify/src/components/VSelect/VSelectList.ts
@@ -207,7 +207,7 @@ export default mixins(Colorable, Themeable).extend({
         ? this.$createElement(VListItem, tile, scopedSlot)
         : scopedSlot
     },
-    genTileContent (item: any, index = 0 as string | number): VNode {
+    genTileContent (item: any, index = 0): VNode {
       const innerHTML = this.genFilteredText(this.getText(item))
 
       return this.$createElement(VListItemContent,
--- a/packages/vuetify/src/components/VSelect/VSelectList.ts
+++ b/packages/vuetify/src/components/VSelect/VSelectList.ts
@@ -146,11 +146,12 @@ export default mixins(Colorable, Themeable).extend({
       const end = text.slice(index + searchInput.length)
       return { start, middle, end }
     },
-    genTile (
-      item: object,
-      disabled = null as null | boolean,
-      value = false
-    ): VNode | VNode[] | undefined {
+    genTile ({
+      item,
+      index,
+      disabled = null,
+      value = false,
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VSelect/VSelectList.ts
```diff
@@ -206,12 +207,12 @@ export default mixins(Colorable, Themeable).extend({
         ? this.$createElement(VListItem, tile, scopedSlot)
         : scopedSlot
     },
-    genTileContent (item: any): VNode {
+    genTileContent (item: any, index = 0 as string | number): VNode {
```
jacekkarczmarczyk:
```suggestion
    genTileContent (item: any, index = 0 as string): VNode {
```

packages/vuetify/src/components/VSelect/VSelectList.ts
```diff
@@ -164,7 +165,7 @@ export default mixins(Colorable, Themeable).extend({
           // Default behavior in list does not
           // contain aria-selected by default
           'aria-selected': String(value),
-          'aria-labelledby': this.genLabelledBy(item),
+          'aria-labelledby': `${this.genLabelledBy(item)}-${index}`,
```
jacekkarczmarczyk: Since we're using now `index` I guess the `const text = escapeHTML(this.getText(item).split(' ').join('-').toLowerCase())` part in `genLabelledBy` is not necessary anymore

packages/vuetify/src/components/VSelect/VSelectList.ts
KaelWD: 0 isn't a string

KaelWD:
```suggestion
    genTileContent (item: any, index: string | number = 0): VNode {
```

jacekkarczmarczyk: removed the wrong type...

jacekkarczmarczyk: i've edited my comment (just removed the wrong type before), string is not necessary here

KaelWD: neither is `as number`

packages/vuetify/src/components/VSelect/VSelectList.ts
```diff
@@ -146,11 +146,12 @@ export default mixins(Colorable, Themeable).extend({
...
-      item: object,
-      disabled = null as null | boolean,
-      value = false
-    ): VNode | VNode[] | undefined {
+    genTile ({
+      item,
+      index,
+      disabled = null,
+      value = false,
+    }: { item: any, disabled?: null | boolean, value?: boolean, index: number }): VNode | VNode[] | undefined {
```
johnleider: This seems like it should be a specific type.

johnleider:
Requesting final comments @vuetifyjs/core-contributors @vuetifyjs/core-team

packages/vuetify/src/components/VSelect/VSelectList.ts
abnersajr: I don't think so. In one scenario with multiple selects the Id's can be duplicated. This part doesn't avoid all possible duplications, but it's not common to have multiple selects with same options.

In case I'm wrong, please let me know.

jacekkarczmarczyk: This should be handled by `_uid`
```ts
genLabelledBy (item: object) {
      const text = escapeHTML(this.getText(item).split(' ').join('-').toLowerCase())

      return `${text}-list-item-${this._uid}`
    },
```
Can you check it?
Also it's very common to have multiple selets with same options, for example you can have selects with nully value ('not applicable', 'other', etc)

abnersajr: I will recheck the use of `_uid` in this function.

abnersajr: @jacekkarczmarczyk I checked again your point and isn't necessary anymore the text. Only `list-item-${this._uid}` with the index. I didn't get your point.
Thanks for your review.

packages/vuetify/src/components/VSelect/VSelectList.ts
abnersajr: Added :)

johnleider:
@phiter can you do a once over on to make sure none of this goes against or breaks a11y support?

phiter:
@johnleider it is not beaking a11y. The `aria-labelledby` is still correct and in my tests the IDs are unique.

PR #10027, opened 2019-12-21
fix(VResponsive): refactor VResponsive to work with Firefox

### LINKED ISSUE #9659: [Bug Report] v-responsive doesn't work in Firefox
### Environment
**Vuetify Version:** 2.1.9
**Vue Version:** 2.6.10
**Browsers:** Chrome 78.0.3904.97
**OS:** Mac OS 10.14.6

### Steps to reproduce
1. Start window full screen
2. Shrink horizontally

### Expected Behavior
Card maintains aspect ratio

### Actual Behavior
Card does not maintain aspect ratio

### Reproduction Link
<a href="https://vuetifyjs.com/en/components/aspect-ratios" target="_blank">https://vuetifyjs.com/en/components/aspect-ratios</a>

### Other comments
According to [this](https://stackoverflow.com/a/46846942) StackOverflow post, this is due to how Firefox resolves the padding that Vuetify is using to maintain the aspect ratio. As such, it doesn't look like this technique will ever work in Firefox.

There is quirky behavior, like it *seems* to work if you load the page in a small window, but overall it doesn't work.

dsanders11:
## Description

Alternate method of enforcing aspect ratio which works around buggy behavior in Firefox.

## Motivation and Context

fixes #9659. @KaelWD 

## How Has This Been Tested?

Visually tested in Chrome, Firefox, and Safari. Has not been tested in IE11 or Edge.

## Markup:

<details>

```vue
<template>
<div id="app">
  <v-app id="inspire">
    <div>
      <v-text-field v-model="aspectRatio"></v-text-field>
      <v-card>
        <v-responsive :aspect-ratio="aspectRatio">
          <v-card-text>
            This card has the above aspect ratio
          </v-card-text>
        </v-responsive>
      </v-card>
    </div>
  </v-app>
</div>
</template>

<script>
new Vue({
  el: '#app',
  vuetify: new Vuetify(),
  data () { 
    return {
      aspectRatio: 16/9
    }
  }
})
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VAppBar/__tests__/__snapshots__/VAppBar.spec.ts.snap
+++ b/packages/vuetify/src/components/VAppBar/__tests__/__snapshots__/VAppBar.spec.ts.snap
@@ -14,7 +14,9 @@ exports[`AppBar.ts should render with background 1`] = `
            mode="in-out"
       >
       </div>
-      <div class="v-responsive__content">
+      <div class="v-responsive__sizer">
+        <div class="v-responsive__content">
+        </div>
       </div>
     </div>
   </div>
--- a/packages/vuetify/src/components/VImg/VImg.ts
+++ b/packages/vuetify/src/components/VImg/VImg.ts
@@ -127,6 +127,14 @@ export default VResponsive.extend({
         },
       }, [image])
     },
+    contentAspectStyle (): object | undefined {
+      const styles = VResponsive.options.computed.contentAspectStyle.call(this)
+
+      return {
+        ...styles,
+        ...(this.naturalWidth && { width: `${this.naturalWidth}px` }),
+      }
+    },
   },
 
   watch: {
@@ -224,16 +232,6 @@ export default VResponsive.extend({
 
       poll()
     },
-    genContent () {
-      const content: VNode = VResponsive.options.methods.genContent.call(this)
-      if (this.naturalWidth) {
-        this._b(content.data!, 'div', {
-          style: { width: `${this.naturalWidth}px` },
-        })
-      }
-
-      return content
-    },
     __genPlaceholder (): VNode | void {
       if (this.$slots.placeholder) {
         const placeholder = this.isLoading
@@ -274,7 +272,6 @@ export default VResponsive.extend({
     }
 
     node.children = [
-      this.__cachedSizer,
       this.__cachedImage,
       this.__genPlaceholder(),
       this.genContent(),
--- a/packages/vuetify/src/components/VImg/__tests__/__snapshots__/VImg.spec.ts.snap
+++ b/packages/vuetify/src/components/VImg/__tests__/__snapshots__/VImg.spec.ts.snap
@@ -13,7 +13,9 @@ exports[`VImg.ts should display placeholders 1`] = `
       loading...
     </div>
   </div>
-  <div class="v-responsive__content">
+  <div class="v-responsive__sizer">
+    <div class="v-responsive__content">
+    </div>
   </div>
 </div>
 `;
@@ -31,7 +33,9 @@ exports[`VImg.ts should display placeholders 2`] = `
       loading...
     </div>
   </div>
-  <div class="v-responsive__content">
+  <div class="v-responsive__sizer">
+    <div class="v-responsive__content">
+    </div>
   </div>
 </div>
 `;
@@ -47,7 +51,9 @@ exports[`VImg.ts should have aria attributes 1`] = `
        mode="in-out"
   >
   </div>
-  <div class="v-responsive__content">
+  <div class="v-responsive__sizer">
+    <div class="v-responsive__content">
+    </div>
   </div>
 </div>
 `;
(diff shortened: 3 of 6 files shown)


KaelWD:
Nice, dunno why I didn't think to try that.

dsanders11:
Took me a while to find something that worked in Firefox. It looks like they do implement the margin stuff to spec (fixed a couple years ago), but the behavior is buggy when it comes to an element affecting its parents height, so the only way to get the desired behavior in Firefox was to drop the sizer element.

KaelWD:
It's still broken.

After resizing:
image

Then a page refresh:
image

My playground:
```vue
<template>
  <v-container>
    <v-text-field v-model="aspectRatio"></v-text-field>
    <v-row>
      <v-col>
        <v-card>
          <v-responsive :aspect-ratio="aspectRatio">
            <v-card-text>
              This card has the above aspect ratio
            </v-card-text>
          </v-responsive>
        </v-card>
      </v-col>
      <v-col>
        <v-card>
          <v-responsive :aspect-ratio="aspectRatio">
            <v-card-text>
              <p v-for="i in 15">This card expands to its content</p>
            </v-card-text>
          </v-responsive>
        </v-card>
      </v-col>
    </v-row>

    <v-row>
      <v-col class="d-flex">
        <v-img src="https://picsum.photos/id/11/500/300" class="flex-grow-0"></v-img>
      </v-col>
    </v-row>

    <v-row>
      <v-col>
        <v-card>
          <v-card-text>
            <v-slider v-model="width" min="200" max="500" step="1"></v-slider>
            <v-navigation-drawer
              :width="width"
              :value="true"
              stateless
            >
              <v-img :aspect-ratio="16/9" src="https://cdn.vuetifyjs.com/images/parallax/material.jpg">
                <v-row align="end" class="lightbox white--text pa-2 fill-height">
                  <v-col>
                    <div class="subheading">Jonathan Lee</div>
                    <div class="body-1">heyfromjonathan@gmail.com</div>
                  </v-col>
                </v-row>
              </v-img>

              <v-list>
                <template v-for="(item, i) in items">
                  <v-divider v-if="item.divider" :key="i"></v-divider>
                  <v-list-item v-else :key="item.title" @click>
                    <v-list-item-action>
                      <v-icon>{{ item.icon }}</v-icon>
                    </v-list-item-action>
                  </v-list-item>
                </template>
              </v-list>
            </v-navigation-drawer>
          </v-card-text>
        </v-card>
      </v-col>
    </v-row>
  </v-container>
</template>

<script>
  export default {
    data () {
      return {
        aspectRatio: 16/9,
        width: 300,
        items: [
          { icon: 'inbox', title: 'Inbox' },
          { icon: 'star', title: 'Starred' },
          { icon: 'send', title: 'Sent mail' },
          { icon: 'drafts', title: 'Drafts' },
          { divider: true },
          { icon: 'mail', title: 'All mail' },
          { icon: 'delete', title: 'Trash' },
          { icon: 'error', title: 'Spam' },
        ],
      }
    }
  }
</script>

<style scoped>
  .v-navigation-drawer {
    transition: none !important;
  }

  .lightbox {
    box-shadow: 0 0 20px inset rgba(0, 0, 0, 0.2);
    background-image: linear-gradient(to top, rgba(0, 0, 0, 0.4) 0%, transparent 72px);
  }
</style>
```

dsanders11:
> It's still broken. You can see this on the components/aspect-ratios page if you run yarn build && yarn start

Works fine for me in both Firefox and Firefox Developer Edition. They're version 71. What version are you running?

KaelWD:
:man_facepalming: I've got a script to checkout and push PR branches and it screwed up and checked out master instead for some reason lmao

KaelWD:
It doesn't behave the same. Percent padding-top/bottom is based on the parent's width, so an image that doesn't fill its parent will have the same height as if it does. See my playground for an example, that tall image maintains its native aspect ratio on master.

KaelWD:
Keeping the sizer but making it a wrapper for the content instead of a sibling seems to work.

dsanders11:
> Percent padding-top/bottom is based on the parent's width

Ah, you're right, that's an oversight. So it the padding needs to be applied to a child of `.v-responsive`, not it itself.

> Keeping the sizer but making it a wrapper for the content instead of a sibling seems to work.

Yep, looks like that might do it.

BTW, I'm seeing that in `VImg`, the content is being rendered as a sibling of `.v-responsive__content` which seems a bit strange, but it's that way on master. Is that intended?

KaelWD:
> the content is being rendered as a sibling

Meaning the __image element? Yeah that's fine, it's absolutely positioned so it doesn't affect layout anyway.

dsanders11:
> Meaning the __image element? Yeah that's fine, it's absolutely positioned so it doesn't affect layout anyway.

Makes sense, just looked a bit odd on first glance.

> I already pushed the sizer change to your branch if you want to take a look.

Tested it, looks good to me. I tried it on Chrome, Firefox, and Safari. So the only ones I can't vouch for testing on are IE11 and Edge.

KaelWD:
Ok cool. We have a browserstack account so I'll check them in the morning.

PR #10060, opened 2019-12-27
Update VSelect.ts

arsen-s:
fixed clearableCallback, set value to null instead of undefined


johnleider:
Thank you for your interest in improving Vuetify. After careful consideration the team has decided that this is not functionality that we are looking to implement at this time.

Issue #10094, opened 2020-01-01
[Bug Report] VResponsive/VImg [2.1.16] changes have broken "align-end"


dancarasco:
### Environment
**Vuetify Version:** 2.1.16
**Last working version:** 2.1.15
**Vue Version:** 2.6.10
**Browsers:** Chrome 79.0.3945.88
**OS:** Mac OS 10.14.6

### Steps to reproduce
Add "align-end" to a v-img embedded in a v-card and the alignment won't work. Margin is way out. This can even be seen in the (now broken) examples in the Card component docs under "Grid": https://vuetifyjs.com/en/components/cards#grids

### Expected Behavior
Text should be at the bottom of the card to due align-end being used.

### Actual Behavior
Text appears half way up the card or completely out of the bounds.

### Reproduction Link
<a href="https://codepen.io/trickster223/pen/RwNLYaL" target="_blank">https://codepen.io/trickster223/pen/RwNLYaL</a>

### Other comments
This has broken the v-img and grid usage in the v-card documentation under Grid: https://vuetifyjs.com/en/components/cards#grids

The *breaking* changes I am referring to can be seen in these two commits:

--removed--
--removed--

Tested & working prior to these commits.

johnleider:
This change caused the regression: https://github.com/vuetifyjs/vuetify/commit/e92d6b79115c28b6cae71b7bc70bbb295c12ef89

The above referenced commits have nothing to do with the issue.

dancarasco:
@johnleider Thanks for confirming

PR #10095, opened 2020-01-01
fix(VResponsive): tweak styles to work with Firefox

### LINKED ISSUE #9659: [Bug Report] v-responsive doesn't work in Firefox

dsanders11:
## Description

Alternate method of enforcing aspect ratio which works around buggy behavior in Firefox.

This is yet another attempt at fixing the issue. Unfortunately the previous fix, #10027 had unexpected regressions due to how it changed the DOM structure.

This fix is purely CSS, so it should be less prone to unexpected behavior. Can confirm it does not have the issues from #10094.

It seems the bug in Firefox is isolated to giving `flex: 0 0 0px` to the sizer element. Allowing it to have a size avoids the bug in Firefox, and the negative margin pulls the content back into the correct place. According to the flex spec, you should be able to use `visibility: collapse` on the sizer element to get this same result without the negative margin, [but the MDN documentation points out that Chrome and Safari don't implement this correctly](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Mastering_Wrapping_of_Flex_Items#Collapsed_items). As such, the negative margin is the better alternative.

## Motivation and Context

fixes #9659. @KaelWD @johnleider 

## How Has This Been Tested?

Visually tested in Chrome, Firefox, and Safari. **Has not been tested in IE11 or Edge.**

Confirmed everything looks sane on the Card, Image, and Aspect Ratios documentation pages.

## Markup:

<details>

```vue
<template>
<div id="app">
  <v-app id="inspire">
    <div>
      <v-text-field v-model="aspectRatio"></v-text-field>
      <v-card>
        <v-responsive :aspect-ratio="aspectRatio">
          <v-card-text>
            This card has the above aspect ratio
          </v-card-text>
        </v-responsive>
      </v-card>
    </div>
  </v-app>
</div>
</template>

<script>
new Vue({
  el: '#app',
  vuetify: new Vuetify(),
  data () { 
    return {
      aspectRatio: 16/9
    }
  }
})
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VResponsive/VResponsive.sass
+++ b/packages/vuetify/src/components/VResponsive/VResponsive.sass
@@ -11,6 +11,9 @@
     flex: 1 0 0px
     max-width: 100%
 
+  &__sizer ~ &__content
+    margin-left: -100%
+
   &__sizer
     transition: padding-bottom 0.2s map-get($transition, 'swing')
-    flex: 0 0 0px
+    flex: 1 0 0px


johnleider:
Will review shortly, thank you.

johnleider:
Verified that this works in Edge/IE as well.

PR #10129, opened 2020-01-06
docs(VDataTable): add options prop example

### LINKED ISSUE #9048: [Documentation] update v-data-table options prop
### Problem to solve
Vuetify 2 introduces new props for data table like page, items-per-page, sort-by and sort-desc. Devs will assume that those options will have to synced individually which adds to the complexity of implementation. Options prop also allows sync (missing from documentation) and can handle all of these props. The options -prop is not documented (docs just say refer to examples). I had to convert 20+ tables from Vuetify 1.5 and this made the conversion somewhat painful. If you are using server side pagination you actually have to use the options -prop because you cannot easily watch all the individual sort and page props for changes (as they update in some random order when user click on header sort options).

### Proposed solution
Document v-data-table options property and mention .sync modifier can be used.

nekosaur:
## Description
add missing options prop example

## Motivation and Context
closes #9048

## How Has This Been Tested?
manually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-generator/src/maps/v-data.js
+++ b/packages/api-generator/src/maps/v-data.js
@@ -21,7 +21,7 @@ const DataPagination = {
 const DataProps = [
   { name: 'items', source: 'v-data' },
   { name: 'itemKey', source: 'v-data' },
-  { name: 'options', source: 'v-data' },
+  { name: 'options', source: 'v-data', type: 'DataOptions', example: DataOptions },
   { name: 'sortBy', source: 'v-data' },
   { name: 'sortDesc', source: 'v-data' },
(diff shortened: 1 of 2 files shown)


WhatFreshHellIsThis:
I'm sorry but I must be missing something here, this commit doesn't appear to add an example?!

nekosaur:
example is perhaps badly worded but that's what it's called in api-generator lingo. It adds a definition of the prop object to the API section, similarly to what exists for headers prop right now.

WhatFreshHellIsThis:
Ah..ok, so it will thoroughly show how to use an external server with the component?

PR #10163, opened 2020-01-09
fix(VDialog): override width/max-width when fullscreen

dsanders11:
## Description

Use the `.v-dialog--fullscreen` class to enforce full width, rather than relying on removing the styles applied by the `max-width` and `width` props.

## Motivation and Context

Due to a gotcha in Vue's SSR, when hydrating from server-rendered DOM, existing styles on an element won't be removed unless they get added by the client, then removed. If the style was already on an element, but the hydrated version doesn't have that style in its style binding, then the style is left as is. This means the hydrated version is left in a weird state where it has ghost styles rendered by the server which shouldn't be there.

This is problematic when using a screen size breakpoint with the `fullscreen` prop on `VDialog`. When rendered on the server, it's not considered fullscreen, and it will be rendered with the width styles inline on the element. When hydrated on a mobile client which does meet the breakpoint for `fullscreen`, it would skip the width styles on the client, leading to the situation described in the above paragraph where the client shouldn't have those styles inline, but Vue doesn't remove them due to the gotcha. So this would lead to a fullscreen `VDialog` which was still enforcing a max width.

## How Has This Been Tested?

Manually.

## Markup:

<details>

```vue
<template>
  <v-row justify="center">
    <v-dialog v-model="dialog" fullscreen hide-overlay transition="dialog-bottom-transition" max-width="400px">
        <v-btn color="primary" dark v-on="on">Open Dialog</v-btn>
      </template>
      <v-card>
        <v-toolbar dark color="primary">
          <v-btn icon dark @click="dialog = false">
            <v-icon>mdi-close</v-icon>
          </v-btn>
          <v-toolbar-title>Settings</v-toolbar-title>
          <v-spacer></v-spacer>
          <v-toolbar-items>
            <v-btn dark text @click="dialog = false">Save</v-btn>
          </v-toolbar-items>
        </v-toolbar>
        <v-list three-line subheader>
          <v-subheader>User Controls</v-subheader>
          <v-list-item>
            <v-list-item-content>
              <v-list-item-title>Content filtering</v-list-item-title>
              <v-list-item-subtitle>Set the content filtering level to restrict apps that can be downloaded</v-list-item-subtitle>
            </v-list-item-content>
          </v-list-item>
          <v-list-item>
            <v-list-item-content>
              <v-list-item-title>Password</v-list-item-title>
              <v-list-item-subtitle>Require password for purchase or use password to restrict purchase</v-list-item-subtitle>
            </v-list-item-content>
          </v-list-item>
        </v-list>
        <v-divider></v-divider>
        <v-list three-line subheader>
          <v-subheader>General</v-subheader>
          <v-list-item>
            <v-list-item-action>
              <v-checkbox v-model="notifications"></v-checkbox>
            </v-list-item-action>
            <v-list-item-content>
              <v-list-item-title>Notifications</v-list-item-title>
              <v-list-item-subtitle>Notify me about updates to apps or games that I downloaded</v-list-item-subtitle>
            </v-list-item-content>
          </v-list-item>
          <v-list-item>
            <v-list-item-action>
              <v-checkbox v-model="sound"></v-checkbox>
            </v-list-item-action>
            <v-list-item-content>
              <v-list-item-title>Sound</v-list-item-title>
              <v-list-item-subtitle>Auto-update apps at any time. Data charges may apply</v-list-item-subtitle>
            </v-list-item-content>
          </v-list-item>
          <v-list-item>
            <v-list-item-action>
              <v-checkbox v-model="widgets"></v-checkbox>
            </v-list-item-action>
            <v-list-item-content>
              <v-list-item-title>Auto-add widgets</v-list-item-title>
              <v-list-item-subtitle>Automatically add home screen widgets</v-list-item-subtitle>
            </v-list-item-content>
          </v-list-item>
        </v-list>
      </v-card>
    </v-dialog>
  </v-row>
</template>

<script>
  export default {
    data () {
      return {
        dialog: false,
        notifications: false,
        sound: true,
        widgets: false,
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDialog/VDialog.sass
+++ b/packages/vuetify/src/components/VDialog/VDialog.sass
@@ -65,6 +65,8 @@
   overflow-y: auto
   top: 0
   left: 0
+  width: 100% !important
+  max-width: 100% !important
 
   > .v-card
     min-height: 100%
--- a/packages/vuetify/src/components/VDialog/VDialog.ts
+++ b/packages/vuetify/src/components/VDialog/VDialog.ts
@@ -255,14 +255,10 @@ export default baseMixins.extend({
       on: {
         click: (e: Event) => { e.stopPropagation() },
       },
-      style: {},
-    }
-
-    if (!this.fullscreen) {
-      data.style = {
+      style: {
         maxWidth: this.maxWidth === 'none' ? undefined : convertToUnit(this.maxWidth),
         width: this.width === 'auto' ? undefined : convertToUnit(this.width),
-      }
+      },
     }
 
     children.push(this.genActivator())


packages/vuetify/src/components/VDialog/VDialog.sass
```diff
@@ -65,6 +65,8 @@
   overflow-y: auto
   top: 0
   left: 0
+  width: 100% !important
```
johnleider: We'll have to come up with a different solution that doesn't required **!important**

dsanders11: What issue is caused by `!important` in this spot?

I'm not sure there is a different solution in just CSS, since the width is being applied as a direct style on the element, AFAIK only `!important` can override it.

johnleider: What I mean is, why do you need !important here? What does it solve?

dsanders11: It's stated in the description for the PR:

> Use the .v-dialog--fullscreen class to enforce full width, rather than relying on removing the styles applied by the max-width and width props.

The element will have a `style` prop which has `width` and `max-width`. Due to the gotcha in Vue SSR described in the issue, the hydrated client is unable to remove those styles, leading to an odd state where the element has the `v-dialog--fullscreen` class but also the `style` prop. So the change with `!important` is to ensure that if the `v-dialog--fullscreen` class is applied, the dialog is indeed fullscreen.

johnleider: Then why not add the SSR check here: https://github.com/vuetifyjs/vuetify/blob/2b975f575f47ba75aa5d0dd9fba859595c7b9ba4/packages/vuetify/src/components/VDialog/VDialog.ts#L271

dsanders11: I'm not sure what check you're proposing. If it would be a check to not apply the styles if on the server, then I believe that would fix the issue of getting into a broken state. However, it would cause the dialog to be rendered incorrectly until hydration completed. If the developer is aiming to have two possibilities for the dialog, either fullscreen or with width N, the server-rendered markup would be neither, as it would not apply the width style until hydration is done.

I wrote the PR as it is because it's a CSS solution, which keeps the rendered DOM from the server and client consistent, which is less likely to cause surprises for developers.

KaelWD: Yeah that's fine, I don't see anyone wanting to override the width on a fullscreen dialog.

johnleider:
Any update on this?

johnleider:

PR #10315, opened 2020-01-23
feat(elevation): create new effect

johnleider:
## Description
A compositional implementation of the elevatable mixin.

## How Has This Been Tested?
Cannot unit test v3 alpha atm

## Types of changes

## Checklist:


johnleider:
Requesting initial comments @vuetifyjs/core-contributors @vuetifyjs/core-team

packages/vuetify/src/effects/elevation/__tests__/elevation.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  ElevationProps,
+  useElevation,
+} from '../'
+
+// Utilities
+import { mount } from '@vue/test-utils'
+
+describe('elevation.ts', () => {
+  /* tslint:disable-next-line */
+  const MockComponent = defineComponent({
```
jacekkarczmarczyk: Unit test should test the effect itself in the first place, not its usage in component/setup

packages/vuetify/src/effects/elevation/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  defaults: Partial<ElevationProps> = {}
+) => ({
+  elevation: {
+    type: Number,
+    default: defaults.elevation,
+  },
+})
+
+// Effect
+export function useElevation ({ elevation }: ElevationProps) {
```
jacekkarczmarczyk: Destructing props disables reactivity

packages/vuetify/src/effects/elevation/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  const classes: Dictionary<boolean> = {}
+
+  if (
+    elevation != null &&
+    !isNaN(parseInt(elevation))
+  ) {
+    classes[`elevation-${elevation}`] = true
+  }
+
+  const elevationClasses = computed(() => classes)
```
jacekkarczmarczyk: why not just 
```ts
const elevationClasses = computed(() => {
  if (props.elevation == null || isNan(parseInt(props.elevation))) return {}

  return { [`elevation-${props.elevation}`]: true }
})
```

jacekkarczmarczyk: Additionaly there could be a warning (emitted in watcher) if elevation is `NaN` or `< 0` or `> MAX_ELEVATION`, maybe only in dev mode

packages/vuetify/src/effects/elevation/__tests__/elevation.spec.ts
johnleider: Unit testing doesn't even work right now.

packages/vuetify/src/effects/elevation/index.ts
johnleider: Refactored the function. I like the assertion, we should push that to a post 3 release though.

packages/vuetify/src/effects/elevation/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    default: defaults.elevation,
+  },
+})
+
+// Effect
+export function useElevation ({ elevation }: ElevationProps) {
+  const classes: Dictionary<boolean> = {}
+
+  if (
+    elevation != null &&
```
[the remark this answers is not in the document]
johnleider: This has been refactored.

packages/vuetify/src/effects/elevation/__tests__/elevation.spec.ts
johnleider: Unit tests don't work in v3, no effort really put into them.

packages/vuetify/src/effects/elevation/index.ts
```diff
@@ -0,0 +1,22 @@
+// Setup
+import { computed } from 'vue'
+
+// Types
+interface ElevationProps {
```
nekosaur: I know this is on me, but do we maybe want to change name of interface to not be the same as the function? I don't know how people feel about prefix `I`?

jacekkarczmarczyk: what function?

nekosaur: https://github.com/vuetifyjs/vuetify/pull/10315/files#diff-5669027a74cdf9363f9f184608be74e9R10

jacekkarczmarczyk: Ah yeah, missed that
Alternative is to change the function name, i.e. `ElevationPropsFactory`

jacekkarczmarczyk: Also we're using camelCase for functions, here it's PascalCase. So `elevationProps` function would solve the issue as well, but since it's actually a factory then the name should reflect that (maybe `createElevationProps` would be better)

nekosaur: True. I think I had some thought to make it Pascal but now I don't rememeber

johnleider: I believe from our conversation majority sided with `makeElevationProps` due to it being 1 letter less than `createElevationProps`. Any final feedback before this is implemented?

nekosaur: 2 less. I am fine with make, but as noted would also be fine with just `elevationProps()`

johnleider: I'm good with either as well. I like elevationProps as it is more simple and I like makeElevationProps because it is descriptive. Overall I say I'd lean towards elevationProps.

packages/vuetify/src/effects/elevation/index.ts
```diff
@@ -0,0 +1,22 @@
...
+) => ({
+  elevation: {
+    type: Number,
+    default: defaults.elevation,
+  },
+})
+
+// Effect
+export function useElevationClasses (props: ElevationProps) {
+  return computed(() => (0 ^ props.elevation!) ? { [`elevation-${props.elevation}`]: true } : {})
```
nekosaur: I just realized we're just returning the computed directly here. I think it would be better to enforce that we always return an object from composition functions (which is also what I wrote in code guidelines). So this would be:

```
const elevationClasses = computed(...)
return { elevationClasses }

// usage
const { elevationClasses } = useElevationClasses(props)

const classes = computed(() => ({
	...elevationClasses.value,
})
```

Otherwise when we need to return multiple values from a composition function, someone might do this:

```
return computed(() => ({
  foo: ...,
  bar: ...
})

// But when using it you won't be able to destructure it without calling .value
const { foo, bar } = useFoo(props) // won't work
const { foo, bar } = useFoo(props).value // this means we might loose reactivity of foo and bar unless we're careful of what we're returning. also it doesn't look as nice :)

// and if not destructuring
const baz = useFoo(props)

const classes = computed(() => ({
  ...baz.value.foo, // again, doesn't look as nice
})
```

johnleider: This should be resolved with https://github.com/vuetifyjs/vuetify/pull/10315/commits/71069230b7e3dfaa762d78dd3efab10161e5b0f6

PR #10325, opened 2020-01-24
feat(VForm): add form level disabled prop

### LINKED ISSUE #1707: [Feature Request] disable prop for v-form that disables all child input elements
What will it allow you to do that you can't do today?
Disable all fields in a form with one prop

How will it make current work-arounds straightforward?
Remove need to add a disabled prop to every item on a form.

What potential bugs and edge cases does it help to avoid?
Forgetting to add a disabled prop to a given form element

Logikgate:
## Description
A disabled VForm now disables all child inputs. This fixes #1707.
This change was implemented so that the child inputs could still have their own disabled props but the top level VForm disabled prop takes precedence over the child input disabled state. 

## Motivation and Context
It is common to want all of the inputs on the form to be disabled all at once.

## How Has This Been Tested?
Visually tested with all components that inherit from VInput
Added unit test to check VTextfield

## Markup:

```
<v-form disabled>
      <v-text-field></v-text-field>
</v-form>
```

## Types of changes

## Checklist:

###


KaelWD:
Related: #2033

As discussed there I'd prefer this done with provide/inject rather than updating data imperatively.

Logikgate:
@KaelWD Ah yes I missed that conversation in that PR. I'm willing to refactor this if you provide me a little bit of guidance. 

The way this is working right now is it is piggybacking on the provide/inject of the validatable mixin.  I can make make a separate mixin to do the same thing but I'm unsure how that will change the implementation of form level disabled prop being able to override the input's disabled prop.  Or are you thinking that it would work in the opposite direction? i.e. the VForm would inject into the VInputs

KaelWD:
See VExpansionPanel for example. VForm provides itself and the controls have a `isDisabled` computed property that checks `this.form.disabled`.

packages/vuetify/src/mixins/registrable/index.ts
```diff
@@ -15,10 +16,10 @@ export type Registrable<T extends string, C extends VueConstructor | null = null
 
 export function inject<
   T extends string, C extends VueConstructor | null = null
-> (namespace: T, child?: string, parent?: string): Registrable<T, C> {
+> (namespace: T, child?: string, parent?: string, nestingOptional?: boolean): Registrable<T, C> {
```
KaelWD: Child/parent are only used for the warning, if it's optional you just don't use those arguments.

Logikgate: Got it. I did this because inputs are often used without a v-form so most users will see a warning that states a component "must" be used inside a v-form so even though it's a warning it reads more like an error which is misleading. Also without this 322 tests fail.

I'm good with removing this and updating all of the tests to mock the provide but I just want to make sure showing this warning to everyone will be desired behavior. Maybe the warning could be reworded to read less like an error?

Logikgate: Thanks I missed that. Fixed now 👍

Logikgate:
@KaelWD I am updating the test to cover all of the components that should be set to disabled if the form is disabled.  There are two inputs that do not inherit from v-input and therefore do not get disabled by the form level input:

VRadio: This only matters if it was used without a group but using it without a group doesn't sense so maybe it doesn't matter. It looks like if you use a VRadio without a VRadioGroup no warning is shown though, should I adjust this so the warning shows?

https://github.com/vuetifyjs/vuetify/blob/d8b3adc45d6146e7351b3607612d9b889610e309/packages/vuetify/src/components/VRadioGroup/VRadio.ts#L29

VSimpleCheckbox: Should this inherit from v-input? Or should I just do the provide/inject to make the form level disable work?

https://github.com/vuetifyjs/vuetify/blob/d8b3adc45d6146e7351b3607612d9b889610e309/packages/vuetify/src/components/VCheckbox/VSimpleCheckbox.ts#L12

KaelWD:
> should I adjust this so the warning shows?

I think that was intentional, but yeah it probably doesn't make sense to use it in a form without radio-group. 

> VSimpleCheckbox

Not supposed to be a full replacement for VCheckbox, it's for performance critical stuff like data table selection. Oh also it's functional so that won't work anyway, just ignore it.

packages/vuetify/src/components/VForm/VForm.ts
```diff
@@ -25,11 +25,18 @@ export default mixins(
...
 
+  provide (): object {
+    return {
+      form: this,
+    }
+  },
+
   inheritAttrs: false,
 
   props: {
```
johnleider: sort children alphabetically

packages/vuetify/src/components/VForm/VForm.ts
```diff
@@ -25,11 +25,18 @@ export default mixins(
 ).extend({
   name: 'v-form',
 
+  provide (): object {
+    return {
+      form: this,
```
johnleider: make this one line with the return

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
```diff
@@ -1,27 +1,58 @@
+// Libraries
 import Vue from 'vue'
+import Vuetify from '../../../framework'
+
+// Components
+import VForm from '../VForm'
```
johnleider: Is there anything of value added by testing all of the input components? All validation should be handled in the validatable mixin.

Logikgate:
@johnleider thanks for the review! I incorporated your comments; moved the isDisabled computed prop to the Registratable mixin, VForm test now only tests one component and fixed the couple of formatting issues

johnleider:
Please resolve conflicts.

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
```diff
@@ -1,28 +1,49 @@
...
 import { wait } from '../../../../test'
 
-const inputOne = Vue.component('input-one', {
-  render (h) {
-    return h(VTextField, {
-      props: [v => !!v || 'Required'],
-    })
-  },
-})
+Vue.use(Vuetify)
```
johnleider, line 18: What is the purpose of this?

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
```diff
@@ -245,4 +266,28 @@ describe('VForm.ts', () => {
 
     expect(validate).toHaveBeenCalledTimes(2)
   })
+
+  it('should disable all inputs', async () => {
+    const inputs = [
```
johnleider, line 271: move to one line

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
```diff
@@ -245,4 +266,28 @@ describe('VForm.ts', () => {
...
+      VTextField,
+    ]
+
+    const wrapper = mountFunction({
+      propsData: {
```
johnleider, line 276: move to one line

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
```diff
@@ -245,4 +266,28 @@ describe('VForm.ts', () => {
...
+        disabled: true,
+      },
+      slots: {
```
johnleider, line 279: move to one line

packages/vuetify/src/components/VForm/__tests__/VForm.spec.ts
Logikgate: It came in via the dev branch merge. The wait function is used on line 178.

packages/docs/src/lang/en/components/Forms.json
```diff
@@ -23,6 +23,7 @@
     }
   },
   "props": {
+    "disabled": "If enabled this will set all child inputs to disabled",
```
jacekkarczmarczyk, line 26: Current message ("if enabled ... to disabled") sounds weird to me, maybe "If set to `true`" or just "Disables all child inputs"

PR #10333, opened 2020-01-26
fix(vdata): prevent array value from sort-desc

### LINKED ISSUE #9267: [Bug Report] Datatable - Sorting the same column multiple times turn the sort-desc parameter from boolean to array after the 4th click
### Environment
**Vuetify Version:** 2.1.1
**Vue Version:** 2.6.10
**Browsers:** Chrome 76.0.3809.132
**OS:** Linux x86_64

### Steps to reproduce
1. Pick a column header.
2. Click it's header to sort.  sort-desc = false (Works as expected)
3. Click the same header to reverse your sort. sort-desc = true  (Works as expected)
4. Click the same header again.  sort-desc = undefined  (Works as expected)
5. Click the same header again. sort-desc = [ false ] the type will change to Array!

### Expected Behavior
In the 4th click, the value should be false not [ false ]

### Actual Behavior
The sort-desc value changed to array in the 4th click

### Reproduction Link
<a href="https://codepen.io/alaarihan/pen/yLLyMda?&editable=true&editors=101" target="_blank">https://codepen.io/alaarihan/pen/yLLyMda?&editable=true&editors=101</a>

whoistobias:
VData would emit correct values for the first 3 sorts of a data-table column, followed by emitting
array values for every sort following. VData now correctly removes the array before emitting the
value.

fixes #9267

### DIFF
--- a/packages/vuetify/src/components/VData/VData.ts
+++ b/packages/vuetify/src/components/VData/VData.ts
@@ -219,7 +219,7 @@ export default Vue.extend({
       this.updateOptions({ sortDesc: wrapInArray(sortDesc) })
     },
     'internalOptions.sortDesc' (sortDesc: boolean[], old: boolean[]) {
-      !deepEqual(sortDesc, old) && this.$emit('update:sort-desc', Array.isArray(this.sortDesc) ? sortDesc : sortDesc[0])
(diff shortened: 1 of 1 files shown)


nekosaur:
Please use the provided PR template.

PR #10337, opened 2020-01-26
feat(color): create composition functions

nekosaur:
## Description

color mixin rewritten as composition functions. takes a computed containing the color value. this is because we have a number of use-cases where the color is derived from some logic, so being able to do both `useTextColor(toRef(props, 'color'))` and `useTextColor(computed(() => ...))` through a single interface simplifies things.

there are no makeProps functions since we have a lot of varied color props (color, fill-color, bg-color, icon-color, progress-color, etc) and the prop declaration for a color is very simple.

## How Has This Been Tested?

added unit tests, playground

## Markup:

<details>

```vue
<template>
  <v-app>
    <box color="primary" bg="secondary"></box>
    <box color="#ff0000" bg="#00ff00"></box>
  </v-app>
</template>

<script>
  import { h, toRef } from 'vue'
  import { useTextColor, useBackgroundColor } from '../src/composables'

  const Box = {
    props: {
      color: String,
      bg: String,
    },
    setup (props) {
      const { textColorClasses, textColorStyles } = useTextColor(toRef(props, 'color'))
      const { backgroundColorClasses, backgroundColorStyles } = useBackgroundColor(toRef(props, 'bg'))
      return () => h('div', {
        class: ['box', textColorClasses.value, backgroundColorClasses.value],
        style: [textColorStyles.value, backgroundColorStyles.value],
      }, ['box'])
    }
  }

  export default {
    components: { Box }
  }
</script>

<style lang="sass">
  .box
    width: 100px
    height: 100px
</style>

```
</details>

## Types of changes

## Checklist:


KaelWD:
Something like this would be more like we have currently instead of having to explicitly bind class and style:
```js
const color = useColor(props)
return () => cloneVnode(h('div'), color.value)
```

nekosaur:
I thought about something like `const color = useBackgroundColor(props, 'color')`, but we have a number of more complex scenarios, e.g.

```
// VAlert
computedColor (): string {
  return this.color || this.type
},

/// selectable
computedColor (): string | undefined {
  if (!this.isActive) return undefined
  if (this.color) return this.color
  if (this.isDark && !this.appIsDark) return 'white'
  return 'accent'
}

// VRating
getColor (props: ItemSlotProps): string {
  if (this.isHovering) {
    if (props.isHovered || props.isHalfHovered) return this.color
  } else {
    if (props.isFilled || props.isHalfFilled) return this.color
  }

  return this.backgroundColor
}
```

KaelWD:
Well you have this line already:
```js
const color = isRef(props) ? props : computed(() => props.color)
```

If it's a ref just use it as-is?

packages/vuetify/src/effects/__tests__/color.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { reactive, computed } from 'vue'
+
+import {
+  ColorProps,
+  useTextColor,
+  useBackgroundColor,
+} from '../color'
+
+describe('color.ts', () => {
```
johnleider: I really like this approach. Does anyone @vuetifyjs/core-contributors @vuetifyjs/core-team object to this being documented as a standard for writing tests for effects (and any similar functionality).

To clarify, I'm referring to the nested describes.

sh7dm: I think nested describes are very good thing and they help keeping tests organized.

phiter: Yeah I also use nested describes a lot, in my views for example where I have form validations.

    describe MyView.vue
        describe form validations
            test field is required

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
...
+  const backgroundColor = computed(() => {
+    const color: string = isProps(props) ? props[prop] : props.value
+    const isCss = isCssColor(color)
+
+    if (!color) return {}
+
+    if (isCss) {
+      return {
+        style: {
+          'background-color': color,
```
johnleider: Should be on one line.

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+      }
+    }
+
+    return {
+      class: {
+        [color]: true,
```
johnleider: Should be on one line.

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
+import { computed, ComputedRef, isRef } from 'vue'
+
+export interface ColorProps {
+  color?: string
```
johnleider: I think we should make the props text-color and by-color. Right now color can mean either and it’s confusing.

jacekkarczmarczyk: That would be separate interfaces for both, any since there are actually more color props maybe it should be generic?

https://www.typescriptlang.org/play/#code/C4TwDgpgBAwg9gGzgJwArLmAzgHnZgOQEMBbCLKCAD2AgDsATCrYZASzoHMA+KAXigBvAFBQxUANpgMYYmSgco+WaXIBdAPwAuKC3ZcA3MIC+w4QGM4dFlEtI0MigLwy55SjXpNdrDjwAUdijKWDrKbhQAPkquqlgSagA0UAwQAGZEAK4IwKFKRMjAbEQIOPD2IS6Ecdy8AoLGAJQ6AEoQlsgMVSpkWMlEdCB1vCLitojBjvxQAILIyEQgAHRsWHMLIIETDphYjVAa4xVTOhJBO9hqRqLiyBDAmch0R5O7S3cMmeYQ-v7Su8lzvs+CMbmMxHcHk8oAB5ABGACt2sAlkQsFg2Jw6H9HMlRuCCWdtmodPiCeSoKBIDoAMq+LiJMEU8GpDLZYA6VlZHLxc5JJkU0yCxpM4x44xQNFQNodLrhOL9QbcEWmCxWGxc9lOISU6jAcooHQAcjScDgRqgpks1mAUCsEGU03OIX8RtoNANyCNyU1PMaRmtNmAAHc4CEndsXRI3XrPd6oEa4Zw40kUulubl-WZA4gIEskJx-PblCKcwg8wX-CGw44RcJizIlu79dsjA3MEtTXAjNWQksk56e6G+12DEA

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const colorProps = (defaults: Partial<ColorProps> = {}) => ({
+  color: {
+    type: String,
+    default: defaults.color,
+  },
+})
+
+const isCssColor = (color?: string | false): boolean => !!color && !!color.match(/^(#|(rgb|hsl)a?\()/)
```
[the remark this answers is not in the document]
jacekkarczmarczyk: I don't think we need it need anywhere in the lib, but could be exported for tests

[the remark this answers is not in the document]
jacekkarczmarczyk: See my comment about generics

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
...
+const isProps = <T>(props: T | ComputedRef<string>): props is T => !isRef(props)
+
+type Color = { class: Record<string, boolean> } | { style: Record<string, string> } | {}
+
+type TextColor = {
+  textColor: ComputedRef<Color>
+}
+
+export function useTextColor (props: ColorProps | ComputedRef<string>): TextColor
```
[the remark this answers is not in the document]
jacekkarczmarczyk: These are just declarations, not functions, personally I like groupping like this, New line between actual functions is ok though

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,83 @@
...
+  color: {
+    type: String,
+    default: defaults.color,
+  },
+})
+
+const isCssColor = (color?: string | false): boolean => !!color && !!color.match(/^(#|(rgb|hsl)a?\()/)
+const isProps = <T>(props: T | ComputedRef<string>): props is T => !isRef(props)
+
+type Color = { class: Record<string, boolean> } | { style: Record<string, string> } | {}
```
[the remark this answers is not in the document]
jacekkarczmarczyk: See my comment about generics

packages/vuetify/src/effects/color.ts
jacekkarczmarczyk: `Readonly<Ref>` for usage with normal (not computed) refs? And `ColorPropValue` for a bit more DRY

```ts
type ColorPropValue = string | undefined
export function useTextColor (props: ColorProps | Readonly<Ref<ColorPropValue>>): TextColor

useTextColor(computed(() => 'black'))
useTextColor(ref(undefined))
```

packages/vuetify/src/effects/__tests__/color.spec.ts
jacekkarczmarczyk: Might be ok for smaller files, but for bigger maybe just split the file into smaller ones? We're not restricted to having 1 file per `__test__`

nekosaur: Agreed with jacek, it's a case-by-case decision I think.

packages/vuetify/src/effects/color.ts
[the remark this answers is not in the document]
nekosaur: I don't think there's a need. The function is tested indirectly

packages/vuetify/src/effects/color.ts
nekosaur: I'm not sure that's needed (generic solution). Looking at our current instances of setTextColor and setBackgroundColor, most usage seems to be either directly with `this.color`, or a conditional/computed.

splitting color into text-color and bg-color seems big enough that it needs its own RFC for discussion

packages/vuetify/src/effects/__tests__/color.spec.ts
johnleider: To clarify, this was more for effects and similar files such as helper and utility ones.

packages/vuetify/src/effects/color.ts
johnleider: Valid point @nekosaur , let's punt on this feedback.

packages/vuetify/src/effects/color.ts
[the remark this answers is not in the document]
nekosaur: Possibly. Don't think it's worth generalizing at this point though

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,79 @@
...
+  color: {
+    type: String,
+    default: defaults.color,
+  },
+})
+
+const isCssColor = (color: string): boolean => !!color && !!color.match(/^(#|(rgb|hsl)a?\()/)
+const isProps = <T>(props: T | RefValue): props is T => !isRef(props)
+
+type PropValue = string | null | undefined
```
jacekkarczmarczyk: move type definitions before consts

packages/vuetify/src/effects/color.ts
```diff
@@ -0,0 +1,79 @@
...
+type RefValue = Readonly<Ref<PropValue>>
+type Color = { class: Record<string, boolean> } | { style: Record<string, string> } | {}
```
jacekkarczmarczyk: First i though that `ComputedRef<Color>` is a ref for color ('#000' string)
`ColorData` is imho more accurate name

packages/vuetify/src/effects/color.ts
jacekkarczmarczyk: what's the usecase for `null`?

nekosaur: `useTextColor(computed(() => null))`

packages/vuetify/src/util/colorUtils.ts
```diff
@@ -20,7 +20,7 @@ export type Hex = string
 export type Hexa = string
 export type Color = string | number | {}
 
-export function isCssColor (color?: string | false): boolean {
+export function isCssColor (color?: string | null | false): boolean {
   return !!color && !!color.match(/^(#|var\(--|(rgb|hsl)a?\()/)
```
KaelWD, line 24:
```suggestion
  return !!color && /^(#|var\(--|(rgb|hsl)a?\()/.test(color)
```

packages/vuetify/src/composables/color.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return {
+      'background-color': color.value,
+    }
+  })
+
+  const backgroundColorClasses = computed(() => {
+    if (!color.value || cssColor.value) return null
+
+    return `bg-${color.value}`
```
johnleider, line 53:
```suggestion
    if  return null

    return (!color.value || cssColor.value) ? null : `bg-${color.value}`
```

packages/vuetify/src/composables/color.ts
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  return { backgroundColorStyles, backgroundColorClasses }
```
johnleider, line 56:
```suggestion
  return { backgroundColorClasses, backgroundColorStyles }
```

packages/vuetify/src/composables/__tests__/color.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { reactive, toRef } from 'vue'
+
+import {
+  useTextColor,
+  useBackgroundColor,
+} from '../color'
```
johnleider, line 6:
```suggestion
// Composables
import { useBackgroundColor, useTextColor } from '../color'

// Utilities
import { reactive, toRef } from 'vue'
```

johnleider (changes requested):

packages/vuetify/src/composables/color.ts
```diff
@@ -0,0 +1,57 @@
...
+  const color: Ref<ColorValue> = isRef(props) ? props : computed(() => name && props[name])
+  const cssColor = computed(() => isCssColor(color.value))
+
+  const textColorStyles = computed(() => {
+    if (!color.value || !cssColor.value) return {}
+
+    return {
+      color: color.value,
+      'caret-color': color.value,
+    }
```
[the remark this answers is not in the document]
KaelWD: This is a lot less readable imo

KaelWD (approved):

johnleider (approved):

PR #10417, opened 2020-02-01
fix(VTreeview): add proper a11y support

johnleider:
## Description
Apply visual styles for focus states. 

## Motivation and Context
Improve a11y

## How Has This Been Tested?
TODO

## Markup:

<details>

```vue
<template>
  <div class="ma-12 pa-12">
    <v-treeview
      :items="items"
      selectable
    ></v-treeview>
  </div>
</template>

<script>
  export default {
    data: () => ({
      items: [
        {
          id: 1,
          name: 'Applications :',
          children: [
            { id: 2, name: 'Calendar : app' },
            { id: 3, name: 'Chrome : app' },
            { id: 4, name: 'Webstorm : app' },
          ],
        },
        {
          id: 5,
          name: 'Documents :',
          children: [
            {
              id: 6,
              name: 'vuetify :',
              children: [
                {
                  id: 7,
                  name: 'src :',
                  children: [
                    { id: 8, name: 'index : ts' },
                    { id: 9, name: 'bootstrap : ts' },
                  ],
                },
              ],
            },
            {
              id: 10,
              name: 'material2 :',
              children: [
                {
                  id: 11,
                  name: 'src :',
                  children: [
                    { id: 12, name: 'v-btn : ts' },
                    { id: 13, name: 'v-card : ts' },
                    { id: 14, name: 'v-window : ts' },
                  ],
                },
              ],
            },
          ],
        },
        {
          id: 15,
          name: 'Downloads :',
          children: [
            { id: 16, name: 'October : pdf' },
            { id: 17, name: 'November : pdf' },
            { id: 18, name: 'Tutorial : html' },
          ],
        },
        {
          id: 19,
          name: 'Videos :',
          children: [
            {
              id: 20,
              name: 'Tutorials :',
              children: [
                { id: 21, name: 'Basic layouts : mp4' },
                { id: 22, name: 'Advanced techniques : mp4' },
                { id: 23, name: 'All about app : dir' },
              ],
            },
            { id: 24, name: 'Intro : mov' },
            { id: 25, name: 'Conference introduction : avi' },
          ],
        },
      ],
    }),
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTreeview/VTreeviewNode.ts
+++ b/packages/vuetify/src/components/VTreeview/VTreeviewNode.ts
@@ -1,3 +1,4 @@
+import { keyCodes } from './../../util/helpers'
 // Components
 import { VExpandTransition } from '../transitions'
 import { VIcon } from '../VIcon'
@@ -214,8 +215,27 @@ const VTreeviewNode = baseMixins.extend<options>().extend({
           'v-treeview-node__toggle--open': this.isOpen,
           'v-treeview-node__toggle--loading': this.isLoading,
         },
+        props: { value: this.isOpen },
+        attrs: {
+          'aria-expanded': String(this.isOpen),
+          'aria-label': this.text,
+          role: 'button',
+        },
         slot: 'prepend',
         on: {
+          keydown: (e: KeyboardEvent) => {
+            if (
+              this.disabled ||
+              ![
+                keyCodes.enter,
+                keyCodes.space,
+              ].includes(e.keyCode)
+            ) return
+
+            if (this.isLoading) return
+
+            this.checkChildren().then(() => this.open())
+          },
           click: (e: MouseEvent) => {
             if (this.disabled) return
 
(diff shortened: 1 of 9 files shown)


packages/vuetify/src/components/VTreeview/VTreeviewNode.ts
```diff
@@ -214,8 +215,27 @@ const VTreeviewNode = baseMixins.extend<options>().extend({
           'v-treeview-node__toggle--open': this.isOpen,
           'v-treeview-node__toggle--loading': this.isLoading,
         },
+        props: { value: this.isOpen },
+        attrs: {
+          'aria-expanded': String(this.isOpen),
+          'aria-label': this.text,
```
phiter: `aria-label` should be used when there is no visible description for the element anywhere on the page. Since there is a visible text (the node text), `aria-labbeledby` [should be used instead](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute).

For that every text node in the treeview will need a unique ID.

phiter: The checkbox (if selectable is enabled) also needs `aria-labelledby`. Currently NVDA just reads "button" when you focus on it.

phiter:
I know this is not within the scope of the changes that were made but VTreeview could use some extra work.

Treeview has a defined [design pattern by w3](https://www.w3.org/TR/wai-aria-practices/#TreeView) that should be followed.

Currently the entire thing is built with divs when it should be an `ul` with `li`, also we're missing aria roles. Also the keyboard navigation in a tree component should not be via tab, but via keyboard only.

`v-treeview` should have `role="tree"` and it has to be an `ul`.
`v-treeview-node` should have `role="treeitem"` and be an `li`. Also should have `aria-expanded` if it has children.
`v-treeview-node__children` should have `role="group"` and be an `ul`. From here on just repeat the same for the children.

This allows the screenreader to detect which level a user is at when navigating a treeview.

[Here is an implementation ](https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-1/treeview-1a.html) that we can use to understand the aria roles we need.

I didn't find any example with select in w3 site but syncfusion has a react component for this which has the select feature, we could see how they did it.  

Here is the documentation: https://ej2.syncfusion.com/react/documentation/treeview/accessibility/  
Here is an example: https://www.syncfusion.com/react-ui-components/react-treeview

PR #10476, opened 2020-02-05
refactor(proxyable): created proxyable composition function

nekosaur:
## Description

refactored proxyable mixin to composition function. 

## TODO
- since it depends on `getCurrentInstance`, there's currently no way of writing unit tests. will have to wait for VTU to support vue 3.0 so that it can be tested in an actual component
- types are a bit iffy

## Motivation and Context

vuetify 3.0

## How Has This Been Tested?

playground

## Markup:

<details>

```vue
<script lang="ts">
  import { h, defineComponent, ref } from 'vue'
  import { useValueProxy } from '../src/effects/proxy'

  const Foo = defineComponent({
    props: {
      text: String,
      value: {
        type: String,
        default: 'x',
      },
    },
    setup (props, context) {
      const { proxy } = useValueProxy(props, context, 'value')

      return () => h('div', {
        onClick: () => {
          proxy.value = proxy.value === 'x' ? ' ' : 'x'
        },
      }, [props.text, ' - ', '[', proxy.value, ']'])
    },
  })

    components: {
      Foo,
    },
    setup () {
      const foo = ref('x')
      const bar = ref('x')

      return () => [
        h(Foo, {
          text: 'value prop defined, no change handler',
          value: foo.value,
        }),
        h(Foo, {
          text: 'value prop defined, with change handler',
          value: bar.value,
          'onUpdate:value': (v: any) => bar.value = v,
        }),
        h(Foo, {
          text: 'no value prop defined, no change handler',
        }),
        h(Foo, {
          text: 'no value prop defined, change handler',
          'onUpdate:value': (v: any) => console.log('no value prop defined, change handler triggered with', v),
        }),
      ]
    },
  })
</script>

```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/effects/proxy.ts
```diff
@@ -0,0 +1,33 @@
...
+    get () {
+      if (propIsDefined.value) return props[prop]
+      else return internal.value as unknown as T[P]
+    },
+    set (v) {
+      const value = (propIsDefined.value ? props[prop] : internal.value) as T[P]
+      if (deepEqual(v, value)) return
+
+      internal.value = v as any
+      context.emit(`update:${event || prop}`, v)
```
jacekkarczmarczyk: `event || prop` should be kebab-cased

PR #10682, opened 2020-02-26
fix(vselect): height changing on select in dense mode

### LINKED ISSUE #10635: [Bug Report] v-select dense input bug
### Environment
**Vuetify Version:** 2.2.14
**Vue Version:** 2.6.11
**Browsers:** Chrome 80.0.3987.116
**OS:** Linux x86_64

### Steps to reproduce
The codepen is ready to be seen, there are a select with selected value, an autocomplete with selected value and a select and an autocomplete without sleected value, just to show that the height is right until value is selected

### Expected Behavior
dense v-select keeps its height once value is selected

### Actual Behavior
dense v-select gets higher when a value is selected

### Reproduction Link
<a href="https://codepen.io/rsaleri/pen/rNVMvXG" target="_blank">https://codepen.io/rsaleri/pen/rNVMvXG</a>

### Other comments
I put autocomplete in the codepen just as a reference

amoliski:
The VSelect element creates a v-select__selection div that has a margin. When in dense mode, that
margin causes the overall size of the element to change. This fix adds a v-select__selection--dense
class that removes the margin. See https://codepen.io/rsaleri/pen/rNVMvXG for bug reproduction

fix #10635

## Motivation and Context
Fixes [v-select dense input bug](https://github.com/vuetifyjs/vuetify/issues/10635)

## How Has This Been Tested?
Visually tested using sandbox

```vue
<template>
  <v-container>

    <v-row>
      <v-col>
        <v-select
          label="dialog"
          :items="list"
          v-model="select"
        ></v-select>
      </v-col>
      <v-col>
        <v-autocomplete
          label="autocomplete"
          :items="list"
          v-model="autocomplete"
        ></v-autocomplete>
      </v-col>
      <v-col>
        <v-select
          label="dialog 2"
          :items="list"
          v-model="select_2"
          dense
        ></v-select>
      </v-col>
      <v-col>
        <v-autocomplete
          label="autocomplete 2"
          :items="list"
          v-model="autocomplete_2"
          dense
        ></v-autocomplete>
      </v-col>
    </v-row>

  </v-container>
</template>

<script>
export default {
    data() {
        return {
            list: [
                'element 1',
                'element 2',
                'element 3',
                'element 4',
                'element 5',
                'element 6',
            ],
            autocomplete: 'element 1',
            select: 'element 1',
            autocomplete_2: null,
            select_2: null
        };
    }
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.sass
+++ b/packages/vuetify/src/components/VSelect/VSelect.sass
@@ -7,6 +7,9 @@
   .v-select__selection--disabled
     color: map-deep-get($material, 'text', 'disabled')
 
+  .v-select__selection--dense
+    margin: 0px 4px 0px
+
 .v-select
   position: relative // For **attach** prop
 
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -379,6 +379,7 @@ export default baseMixins.extend<options>().extend({
         staticClass: 'v-select__selection v-select__selection--comma',
         class: {
           'v-select__selection--disabled': isDisabled,
+          'v-select__selection--dense': this.dense,
         },
         key: JSON.stringify(this.getValue(item)),
       }), `${this.getText(item)}${last ? '' : ', '}`)


packages/vuetify/src/components/VSelect/VSelect.sass
```diff
@@ -7,6 +7,9 @@
   .v-select__selection--disabled
     color: map-deep-get($material, 'text', 'disabled')
 
+  .v-select__selection--dense
+    margin: 0px 4px 0px
```
jacekkarczmarczyk: is this something that we could get from sass vars?

amoliski: Do you mean adding a `$select-selections-margin-dense: 0 4px 0 !default;` to 
`vuetify/packages/vuetify/src/components/VSelect/_variables.scss`, or do you mean using a variable that already exists like `$spacer` from` /styles/settings/_variables.scss`?

jacekkarczmarczyk: I meant using some existing variable from `(VSelect/VTextField/VInput)/_variables.scss` that makes displaying autocomplete properly, but can't find anything that matches. So adding `$select-selections-margin-dense: 0 4px 0 !default` to select variables should be fine

Also please add `chips` and `small-chips` variants to the playground - does your change work well with them?

amoliski: https://codepen.io/amoliski/pen/NWqjdyR
Before the change, the empty and the filled select have different heights, but the empty/filled/autocomplete small-chips all have the same (slightly taller) height.

With my fix:
With fix

Without my fix:
Without fix

I _think_ this is the desired behavior- with chips is slightly taller than without in both dense and standard mode to account for the chips padding. At the very least, it matches the autocomplete behavior.

amoliski: Actually looking at it again, the real solution might be to ADD padding to the empty dense select instead of removing it from the non-empty dense select. With the added padding, it will line up with the dense autcomplete, and without the padding, it might be _too_ dense.

With Fix: Both dense selects are shorter than the dense autocompletes
With fix 2

Without fix, only the empty select is shorter than the autocomplete
Without Fix 2

amoliski: It looks like all it actually takes to fix this is to add `:not(.v-select)` to line 89 of `VTextField.sass`

```
&.v-input--dense
    padding-top: 0

    &:not(.v-text-field--outlined):not(.v-text-field--single-line):not(.v-select) //Here
      input
        padding: $text-field-dense-padding

    &[type=text]::-ms-clear
      display: none
```

With that dense text field padding removed, everything lines up as it should:
actual fix

amoliski:
I closed this PR and opened a new one with a better fix [here](https://github.com/vuetifyjs/vuetify/pull/10694)

PR #10751, opened 2020-03-05
docs: statically generate routes

KaelWD:
## Description
SSG version of the docs

## Motivation and Context
Support now 2.

## Markup:
-NA-

## Types of changes

## Checklist:


packages/docs/build/prerender.js
```diff
@@ -0,0 +1,135 @@
...
+async function readBundle () {
+  return JSON.parse(await readFile('../dist/vue-ssr-server-bundle.json'))
+}
+
+async function readManifest () {
+  return JSON.parse(await readFile('../dist/vue-ssr-server-bundle.json'))
+}
+
+async function run () {
+  const routes = await promise(generateRoutes).catch(log)
```
KaelWD: These were async to not block webpack, you can just replace it with `readFileSync` now that it's standalone.

packages/docs/build/prerender.js
```diff
@@ -0,0 +1,135 @@
...
+        const util = require('util')
+        const mkdirp = require('mkdirp')
+        const { performance } = require('perf_hooks')
+        const { createBundleRenderer } = require('vue-server-renderer')
+
+        const writeFile = util.promisify(fs.writeFile)
+
+        const renderer = createBundleRenderer(bundle, {
+          runInNewContext: false,
+          manifest,
```
KaelWD: `{ manifest }` !== `{ clientManifest }`

packages/docs/build/prerender.js
KaelWD: Also promise() doesn't reject so why the .catch()?

KaelWD: And generateRoutes is already a promise

packages/docs/build/prerender.js
```diff
@@ -0,0 +1,135 @@
...
+  return chunks
+}
+
+function promise (cb) {
+  return new Promise(resolve => {
+    let value
+
+    try {
+      value = cb()
+    } catch (err) {
```
KaelWD: `process.on('unhandledRejection', ...`?

KaelWD: and `try { returnsAPromise()` doesn't catch shit unless you await.

packages/docs/src/components/base/Markdown.vue
```diff
@@ -116,7 +116,6 @@
       if (wantsList) code = code.map(c => `- ${c}\n`).join('')
 
       if (typeof code !== 'string') {
-        console.log(code, typeof code)
```
KaelWD: I think this is reporting errors: d74a68cf

now.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "vuetifyjs.com",
```
KaelWD: This isn't deprecated?

MajesticPotatoe: When running `now` it says it is
now

KaelWD: Yeah I got that after updating now-cli.

packages/docs/src/components/base/Markdown.vue
MajesticPotatoe: Just reporting when code is anything other than an string. we can put it back if you want, only really removed it because it was outputting on render when building.

KaelWD: Already put it back, it's reporting actual problems in the language files.

KaelWD:
Test deployment: vuetify-bk31asemn.now.sh

PR #10880, opened 2020-03-18
feat(VTabs): adhere to w3 aria keyboard navigation for tabs

### LINKED ISSUE #10416: [Bug Report][2.2.8] VTab keyboard navigation is not following w3 aria design pattern
### Environment
**Vuetify Version:** 2.2.8
**Vue Version:** 2.6.11
**Browsers:** Chrome 79.0.3945.130
**OS:** Windows 10

### Steps to reproduce
1. Focus on the tab component via keyboard
2. Try to navigate with arrows

### Expected Behavior
It should follow the [w3 specs](https://www.w3.org/TR/wai-aria-practices/#tabpanel) for tab navigation:

> ## **Tab:**

> When focus moves into the tab list, places focus on the active tab element. When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is typically either the first focusable element inside the tab panel or the tab panel itself.
>
> #### When focus is on a tab element in a horizontal tab list:
> <kbd>Left Arrow</kbd>: moves focus to the previous tab. If focus is on the first tab, moves focus to the last tab. Optionally, activates the newly focused tab (See note below).
> <kbd>Right Arrow</kbd>: Moves focus to the next tab. If focus is on the last tab element, moves focus to the first tab. Optionally, activates the newly focused tab (See note below).
> When focus is on a tab in a tablist with either horizontal or vertical orientation:
> <kbd>Space</kbd> or <kbd>Enter</kbd>: Activates the tab if it was not activated automatically on focus.
> <kbd>Home</kbd> (Optional): Moves focus to the first tab. Optionally, activates the newly focused tab (See note below).
> <kbd>End</kbd> (Optional): Moves focus to the last tab. Optionally, activates the newly focused tab (See note below).
> <kbd>Shift + F10</kbd>: If the tab has an associated pop-up menu, opens the menu.
> <kbd>Delete</kbd> (Optional): If deletion is allowed, deletes (closes) the current tab element and its associated tab panel, sets focus on the tab following the tab that was closed, and optionally activates the newly focused tab. If there is not a tab that followed the tab that was deleted, e.g., the deleted tab was the right-most tab in a left-to-right horizontal tab list, sets focus on and optionally activates the tab that preceded the deleted tab. If the application allows all tabs to be deleted, and the user deletes the last remaining tab in the tab list, the application moves focus to another element that provides a logical work flow. As an alternative to Delete, or in addition to supporting Delete, the delete function is available in a context menu.

### Actual Behavior
The navigation is done through tab only. Arrows won't work.

### Reproduction Link
<a href="https://codepen.io/phiter/pen/KKwOVrM" target="_blank">https://codepen.io/phiter/pen/KKwOVrM</a>

### Other comments
So basically we have two possible forms of activation: [manual](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html) and [automatic](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html). I personally think we could make this a prop and let the developer decide.

We already have the aria roles for the tab list and tabs themselves, but we're missing the `role="tabpanel` (which is one for each item instead of the tab container).

The active tab should have `tabindex="0"` and the other ones `tabindex="-1"`, which makes it possible to tab into the content instead of the other tabs.

I don't think this is something hard to implement and I also think this could break the current flow of some users, if they're used to use tab to navigate through tabs, but I'm not sure if this should be considered a "breaking change".

Leespiker:
## Description

adhere to [w3 aria specs](https://www.w3.org/TR/wai-aria-practices/#tabpanel) for keyboard navigation

## Motivation and Context

resolves #10416 

## How Has This Been Tested?

unit tests & visually

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-card
      v-for="activationMode in activationModes"
      :key="activationMode"
      class="mb-8"
    >
      <v-tabs
        v-model="currentItem"
        :activation-mode="activationMode"
        fixed-tabs
      >
        <v-tab
          :key="item"
          :disabled="index === 1 || index === 3"
          :href="'#tab-' + item"
        >
          {{ item }}
        </v-tab>
      </v-tabs>

      <v-tabs-items v-model="currentItem">
        <v-tab-item
          v-for="item in items"
          :key="item"
          :value="'tab-' + item"
        >
          <v-card flat>
            <v-card-title>
              {{ item }}
            </v-card-title>
            <v-card-text>
              {{ text }}
            </v-card-text>
            <v-card-text>
              <v-row justify="center">
                <v-col cols="auto">
                  <v-sheet>
                    <v-chip-group
                      active-class="primary--text"
                      column
                    >
                      <v-chip
                        v-for="tag in tags"
                        :key="tag"
                      >
                        {{ tag }}
                      </v-chip>
                    </v-chip-group>
                  </v-sheet>
                </v-col>
              </v-row>
            </v-card-text>
          </v-card>
        </v-tab-item>
      </v-tabs-items>
    </v-card>
  </v-container>
</template>

<script>
export default {
  data: () => ({
    activationModes: ['manual', 'automatic'],
    currentItem: 'tab-Web',
    items: ['Web', 'Shopping', 'Videos', 'Images'],
    tags: ['Work', 'Home Improvement', 'Vacation'],
    text: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.',
  }),
}
</script>
```
</details>

## Types of changes

## Checklist:


phiter:
I tested it and it works great! It follows the spec just as I mentioned in my issue.

packages/vuetify/src/mixins/routable/index.ts
```diff
@@ -135,6 +136,9 @@ export default Vue.extend({
 
       return { tag, data }
     },
+    keydown (e: KeyboardEvent): void {
```
johnleider, line 139: You don't need to declare void.

packages/vuetify/src/components/VTabs/VTab.ts
```diff
@@ -72,38 +90,98 @@ export default baseMixins.extend<options>().extend(
...
+      const targetIndex = (currentIndex - 1 + this.items.length) % this.items.length
+
+      return this.items[targetIndex]
+    },
+    keydown (e: KeyboardEvent): void {
+      this.$emit('keydown', e)
+
+      const { down, end, enter, home, left, right, space, up } = keyCodes
+      const isVertical = this.vertical
+      let targetTab
```
johnleider, line 135: 👏 space 👏 space 👏 space 👏

johnleider: image

packages/vuetify/src/components/VTabs/VTabs.ts
```diff
@@ -187,7 +191,10 @@ export default baseMixins.extend<options>().extend({
           height: convertToUnit(this.height),
         },
         props: {
+          activationMode: this.activationMode,
           activeClass: this.activeClass,
+          ariaLabel: this.$attrs['aria-label'],
```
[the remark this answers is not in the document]
Leespiker: The idea is to pass along the `aria-label` and `aria-labelledby` from `v-tabs` to `v-tabs-bar` since the user does not have direct access to `v-tabs-bar`. I can make them attributes in `v-tabs-bar` too if that's what you're thinking.

packages/vuetify/src/components/VTabs/VTabsBar.ts
```diff
@@ -27,10 +27,26 @@ export default mixins(
...
     return {
+      activationMode: this.activationMode,
+      items: this.items,
       tabsBar: this,
+      vertical: this.vertical,
     }
   },
 
+  props: {
+    activationMode: String,
```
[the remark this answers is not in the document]
Leespiker: `activationMode`, `ariaLabel` and `ariaLabelledBy` in an `accessible` mixin and `vertical` in an `orientable` mixin?

packages/vuetify/src/components/VTabs/VTabsBar.ts
```diff
@@ -27,10 +27,26 @@ export default mixins(
 
   provide () {
     return {
+      activationMode: this.activationMode,
```
KaelWD, line 30: Children can use `this.tabsBar.activationMode`

packages/vuetify/src/components/VTabs/__tests__/VTab.spec.ts
```diff
@@ -89,28 +89,139 @@ describe('VTab.ts', () => {
     expect(toggle).toHaveBeenCalledTimes(1)
   })
 
-  it('should respond to clicks and mousedown.enter', () => {
-    const event = { preventDefault: jest.fn() }
-    const toggle = jest.fn()
+  it('should toggle on click, keydown.enter, and keydown.space', () => {
```
johnleider, line 92: This seems like a good candidate for https://jestjs.io/docs/en/api#testeachtablename-fn-timeout

packages/vuetify/src/components/VTabs/__tests__/VTab.spec.ts
```diff
@@ -89,28 +89,139 @@ describe('VTab.ts', () => {
...
+    expect(wrapper.vm.getFirstTab()).toBe(firstTab)
+    expect(wrapper.vm.getLastTab()).toBe(lastTab)
+    expect(wrapper.vm.getNextTab()).toBe(nextTab)
+    expect(wrapper.vm.getPrevTab()).toBe(prevTab)
+  })
 
-    expect(event.preventDefault).toHaveBeenCalledTimes(2)
-    expect(toggle).toHaveBeenCalledTimes(3)
+  // eslint-disable-next-line max-statements
+  it('should focus and possibly click tab on keydown depending on activation mode', () => {
```
johnleider, line 152: This too: https://jestjs.io/docs/en/api#testeachtablename-fn-timeout

packages/vuetify/src/components/VTabs/VTabsBar.ts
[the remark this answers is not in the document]
johnleider: Disregard for now.

phiter:
@Leespiker I tested the component and it's working well!

It follows the specs correctly and I didn't find any issues with it. I tested it using NVDA on Windows 10.

I think the only thing missing now would be to have an example of automatic/manual activation in the docs, also the docs doesn't have any examples with disabled tabs. Even though it's out of the scope of your PR, it would be cool if you can add that.

Also I noticed in the docs the [tabs with icons only](https://vuetifyjs.com/en/components/tabs/#desktop-tabs) are missing a descriptive text. Again this is out of your PR's scope, but since we're working on a11y it would also be nice to add that. For each icon, just add a text wrapper by a `<span class="d-sr-only"`

johnleider:
We talked about this and are going to push the logic to a v3 implementation.

PR #10906, opened 2020-03-21
docs: add a message when adblock is detected

KaelWD:
idk about the tooltip you might want to change that


johnleider:
If anyone selectively hides the card ad, there is nothing that we can do. Otherwise, they will never be blocked.

PR #10907, opened 2020-03-21
docs: compile page layouts from pug

johnleider:
## Description
Created a parser that converts Pug files into our current structure for pages located [here](https://github.com/vuetifyjs/vuetify/tree/master/packages/docs/src/data). This is setup to allow incremental updates to the docs and follows this pattern currently:

```pug
// headers must use the h2 tag
h2 vue-cli-header

// plain text
| vue-cli-text-1

// Alerts, value is the type
alert(value="warning") vue-cli-alert-1

// components receive values from attributes
component(value="...")

code(lang="bash" src="vue-add-capacitor")

// Image src goes in value, alt text in block
img(value="quick-start/vue_ui.png") ui-img-1
```

## How Has This Been Tested?
Compiled the output data file.

## Types of changes

## Checklist:


sh7dm:
Oh, great! Maybe snippets also can be extracted? It'll be awesome to have smth like this:
```markdown
<!-- snippets have 3 parts -->
<!-- `{file type} {snippet name} {code}` -->
`'` bash vue-add-capacitor
ls /
`'`
```

MajesticPotatoe:
> Oh, great! Maybe snippets also can be extracted?

I believe this is just to replace the json structure of data files. Snippet directory would still exist, and seems like it already parses / links the snippet file.

johnleider:
> Oh, great! Maybe snippets also can be extracted? It'll be awesome to have smth like this:
> 
> ```gfm
> 
> 
> `'` bash vue-add-capacitor
> ls /
> `'`
> ```
> 
> then the compiler could extract and move code to snippets directory

We wouldn't want to do the snippets inline because they are reusable.

KaelWD:
I don't like how it's still JSON, you've just created a new DSL on top of the existing one and it's even more limiting.

johnleider:
@KaelWD I did it as such so that it's easy to start converting as opposed to having to change the underlying structure of how the docs are built. We can do that, but it's out of the scope of this.

johnleider:
Requesting **Initial** comments from @vuetifyjs/core-contributors @vuetifyjs/core-team

packages/docs/build/pug-to-json.js
```diff
@@ -0,0 +1,80 @@
...
+    compileDebug: false,
+    filename: this.resourcePath,
+    filters: {
+      code: text => text
+    },
+    plugins: [{
+      preFilters (ast) {
+        const parents = []
+        walk(ast, node => {
+          if (!pageSection || ['h2', 'up-next', 'usage', 'usage-new', 'api', 'examples'].includes(node.name)) {
```
KaelWD: Not sure about this bit, it seems like most components except markup and alerts are in a new section. Might be a better idea to be explicit about this?

packages/docs/build/pug-to-json.js
```diff
@@ -0,0 +1,80 @@
...
+            case 'Tag':
+              if (node.name === 'h2') {
+                pageSection.children.push({
+                  type: 'heading',
+                  lang: node.block.nodes[0].val
+                })
+              } else if (node.name === 'code') {
+                const lang = eval(node.attrs.find(attr => attr.name === 'lang').val)
+                const src = eval(node.attrs.find(attr => attr.name === 'src').val)
+                // TODO: inline code
```
KaelWD: We could keep the edit button and even have it link to the correct line by using `node.line`

PR #10909, opened 2020-03-22
feat: allow all console messages to be disabled

### LINKED ISSUE #10888: [Feature Request] Optionally hide error messages, warnings and logs (like "Image load failed")
### Problem to solve
Some messages like "Image load failed" message (and others) pollute the console. It will be hard to debug when there are tens or hundreds of errors and warning and logs in the console.

### Proposed solution
I know it is a bad practice, but sometimes, it is needed. I would suggest to have a way to disable some type(s) of errors/warnings. 

#### Messages
1. Image load failed (#6755)
1. Translation key "..." not found, falling back to default

Feel free to comment your suggestions.

KaelWD:
## Description

resolves #10888

## Types of changes

### DIFF
--- a/packages/docs/src/data/drawerItems.json
+++ b/packages/docs/src/data/drawerItems.json
@@ -35,6 +35,7 @@
       { "text": "accessibility", "to": "accessibility" },
       { "text": "aLaCarte", "to": "a-la-carte" },
       { "text": "breakpoints", "to": "breakpoints" },
+      { "new": true, "text": "globalConfig", "to": "global-config" },
       { "text": "icons", "to": "icons" },
       { "text": "i18n", "to": "internationalization" },
       { "text": "rtl", "to": "rtl" },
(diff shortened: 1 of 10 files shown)


jacekkarczmarczyk:
maybe regex|bool so that only some messages could be silenced?

KaelWD:
I think that's overkill. Most of the warnings are about misused or deprecated props so you'd enable this in production if you haven't updated everything yet. Trackers like bugsnag can have their own filters, and so does devtools: https://superuser.com/a/1028554/448703

ibrahimBeladi:
Would be nice to have the option to select which ones. It will be helpful to omit some messages and not others. Like I would hide deprecated messages and show the rest if I am amid the process of migrating and some error came up.
But, if it is time consuming, we can delay it for another time/version.

KaelWD:
Yeah we could add another boolean that does the same just for the `deprecate` function.

jacekkarczmarczyk:
Then why not just 
```
if (Vuetify.config.silent === true || (Vuetify.config.silent && Vuetify.config.silent.test(message))) return;
```

johnleider:

phiter:
I think that's great. I use some an alternative image in VImg when the original image fails to load, so there's always console errors for that, specially in my unit tests.

Having the option to disable those is a good thing I think.

Issue #9588, opened 2019-11-05
[Bug Report] v-dialog backdrop is not removed when the "shown" property is quickly changed


chrishamm:
### Environment
**Vuetify Version:** 2.1.9
**Last working version:** 1.5.20
**Vue Version:** 2.6.10
**Browsers:** Firefox 70.0
**OS:** Linux x86_64

### Steps to reproduce
Set the "shown" property to false really quickly when the dialog has been mounted

### Expected Behavior
Backdrop should be hidden when the value of the "shown" property is set to false

### Actual Behavior
Backdrop is sometimes persistent, which prevents users from interacting with the UI

### Reproduction Link
<a href="https://codepen.io/chrishamm/pen/vYYReNo" target="_blank">https://codepen.io/chrishamm/pen/vYYReNo</a>

### Other comments
I use a similar piece of code to indicate when the page is being loaded and while an initial AJAX call is being made. If the AJAX call returns really quickly, the backdrop is not deleted. This seems to be easier to reproduce in Safari than in Firefox or Chrome although the demo has been tested in Firefox.

KaelWD:
Related to #8625, there's no transitionend event if  it hasn't ended yet.

PR #10975, opened 2020-03-30
fix(overlayable): overlay must be removed if quickly deactivated

aleksandr-rakov:
fix for #9588
playground
https://codepen.io/aleksandrrakov/pen/qBdgYdq


packages/vuetify/src/mixins/overlayable/__tests__/overlayable.spec.ts
```diff
@@ -59,6 +59,20 @@ describe('Overlayable.ts', () => {
     expect(wrapper.vm.overlay).toBeFalsy()
   })
 
+  it('should be removed', async () => {
```
johnleider, line 62: Your fix works but this unit test works without it.

johnleider: disregard

PR #11035, opened 2020-04-05
feat(framework): create new decoration helper classes

johnleider:
## Description
Create new helper classes for **text-decoration**:

```css
.text-decoration-none- *no decoration*
.text-decoration-underline
.text-decoration-overline
.text-decoration-line-through
```

## Motivation and Context
Requested feature. It's quick and easy. Possible that I can avoid the duplicated values, @KaelWD will have to chime in. As far as using *.text-normal* as opposed to *.text-none* is because that value is used currently by the **text-transform** helper classes.

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <div class="ma-12 pa-12">
    <a
      v-for="val in ['normal', 'underline', 'overline', 'line-through']"
      :key="val"
      href="#"
      class="d-block my-4"
      :class="`text-${val}`"
      v-text="val"
    ></a>
  </div>
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/styles/settings/_utilities.scss
+++ b/packages/vuetify/src/styles/settings/_utilities.scss
@@ -312,6 +312,16 @@ $utilities: map-deep-merge(
       class: text,
       values: left right center justify start end
     ),
+    "text-decoration": (
+      property: text-decoration,
+      class: text,
+      values: (
+        normal: none,
+        underline: underline,
+        overline: overline,
+        line-through: line-through,
+      )
+    ),
     "white-space": (
       property: white-space,
       class: text,
--- a/packages/vuetify/src/styles/settings/_utilities.scss
+++ b/packages/vuetify/src/styles/settings/_utilities.scss
@@ -314,7 +314,7 @@ $utilities: map-deep-merge(
     ),
     "text-decoration": (
       property: text-decoration,
-      class: text,
+      class: text-decoration,
       values: (
         normal: none,
         underline: underline,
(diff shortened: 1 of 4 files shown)


johnleider:
Requesting **Initial** comments from @vuetifyjs/contributors @vuetifyjs/core-team

jacekkarczmarczyk:
Generally looks good, naming is in line with other classes, however i'd consider adding some infix instead of putting different kinds of styling to `text-*` group, so for this particular set it would be `text-decoration-*`. My main concerns are:

- autocompletion - if you type `text-` you'll see a number of different classes you can use (align, color, decoration, white space, overflow etc), if you had `text-decoration-*` you can type `text-d` and have just 4 suggestions 
- readability - `text-decoration-none` says much more than `text-normal`, also `text-normal` and `text-none` can be confusing

packages/vuetify/src/styles/settings/_utilities.scss
```diff
@@ -312,6 +312,16 @@ $utilities: map-deep-merge(
       class: text,
       values: left right center justify start end
     ),
+    "text-decoration": (
+      property: text-decoration,
+      class: text-decoration,
+      values: (
+        normal: none,
```
jacekkarczmarczyk, line 319: so now i'm not sure if it should be `text-decoration-none` here, it will match the prop value but still may be confused with `text-none` (the one without infix), thoughts?

johnleider: I'll just add both.

johnleider:

Issue #9993, opened 2019-12-17
[Bug Report] hide-on-scroll on Safari => works the reverse way


guillim:
### Environment
**Vuetify Version:** 2.2.0-beta.0
**Vue Version:** 2.6.10
**Browsers:** Safari
**OS:** Mac OS 10.15.0

### Steps to reproduce
1 - Open Safari
2 - Go to the [official documentation Playground](https://vuetifyjs.com/en/components/app-bars#playground)
3 - Select Hide-on-scroll
4 - Scroll to the bottom and observe

### Expected Behavior
We expect the documentation behavior: _Hides the app-bar when scrolling. Will still show the extension slot._

### Actual Behavior
works the other way around:
- scrolling at the bottom, app-bar appears
- scrolling at the top, app-bar disappears

### Reproduction Link
<a href="https://codepen.io/guillim/full/qBEqwKB" target="_blank">https://codepen.io/guillim/full/qBEqwKB</a>

### Other comments
it is a problem since the behavior is opposite to other browsers

KaelWD:
Unable to reproduce in safari 13 on Catalina

guillim:
I just updated the CodePen to make it even more straightforward. 

Here is what I have on codePen when I use Safari
![Screenshot 2019-12-18 at 14 02 59](https://user-images.githubusercontent.com/13087916/71088662-40dd8f00-219f-11ea-92c4-1ef48ebacf14.png)

Here is what I have on codePen when I use Chrome
![Screenshot 2019-12-18 at 14 07 16](https://user-images.githubusercontent.com/13087916/71088861-cbbe8980-219f-11ea-90a3-2aca32b637d0.png)

Don't you have this on your side?

KaelWD:
Nope, the toolbar is shown at first then hides when I scroll down. This is in browserstack though so maybe it's different on an actual mac. Can you make sure you don't have any extensions enabled?

guillim:
> Nope, the toolbar is shown at first then hides when I scroll down. This is in browserstack though so maybe it's different on an actual mac. Can you make sure you don't have any extensions enabled?

There are no extension installed on my Safari browser on my mac (Catalina 10.15)

I can't access BrowserStack since it's a paid option for my configuration, there may be a bug with it for the reproduction

CodeSkills:
[...]

PR #11105, opened 2020-04-13
fix(VAppBar): unhide when scroll is zero

matthieusieben:
## Description

Solves an issue that can make the app-bar hidden (with `hideOnScroll` enabled) when the page is scrolled all the way up.

## Motivation and Context

The `VAppBar` has:
https://github.com/vuetifyjs/vuetify/blob/6b8b837bbb65469132da2a553b2ed7d9a462b584/packages/vuetify/src/components/VAppBar/VAppBar.ts#L214

The `onScroll` listener will also be called through the `v-scroll` directive (since `VAppBar` uses the `Scrollable` mixin).

When `onScroll` is called twice (basically when the `VAppBar` is mounted), the value of `this.isScrollingUp` is `false` (because the `currentScroll` is `0` the two times `onScroll` is called).

When that happens (`currentScroll` is `0` **and** `this.isScrollingUp` is `false`), and the `hideOnScroll` option is enabled, the following line makes the `VAppBar` disappear:

https://github.com/vuetifyjs/vuetify/blob/6b8b837bbb65469132da2a553b2ed7d9a462b584/packages/vuetify/src/components/VAppBar/VAppBar.ts#L261-L263

## How Has This Been Tested?

Visually

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VAppBar/VAppBar.ts
+++ b/packages/vuetify/src/components/VAppBar/VAppBar.ts
@@ -259,7 +259,7 @@ export default baseMixins.extend({
       if (this.currentThreshold < this.computedScrollThreshold) return
 
       if (this.hideOnScroll) {
-        this.isActive = this.isScrollingUp
+        this.isActive = this.isScrollingUp || this.currentScroll === 0
       }
 
       this.savedScroll = this.currentScroll


johnleider:
Can you please provide a Playground Template (from `vuetify/dev/Playground.vue`. Thank you!

matthieusieben:
This seems to be kind of a race condition. I was not able to reproduce this in a Playground. And I have no idea (or time to debug further) why this occurs in my setup. I understand that you would not want to merge this. At least check if it makes sense to add this little double check, knowing that it can fix a potential race condition.

matthieusieben:
Actually my fix might not be the right one (since it should probably be comparing with the threshold, and not just `0`). Not being able to provide a link to reproduce, I guess you can just close this PR...
My workaround is currently the following:

```
 // using the v-scroll="onScroll" directive

    onScroll() {
      // @hack: https://github.com/vuetifyjs/vuetify/pull/11105
      const { appBar } = this.$refs
      if (appBar && appBar.currentScroll === 0) {
        this.$nextTick(() => {
          // Trigger after VAppBar's scroll handler
          appBar.isActive = true
        })
      }
    },
```

matthieusieben:
Might actually be linked to https://github.com/vuetifyjs/vuetify/issues/9993 but I get this on Chrome, not Safari

matthieusieben:
Adding the following watchers to `VAppBar` yield some strange results:

```js

    computedScrollThreshold: {
      handler(val) {
        console.error('computedScrollThreshold', val) // -9 (!!!!!)
      },
      immediate: true,
    },
    scrollThreshold: {
      handler(val) {
        console.error('scrollThreshold', val) // => undefined
      },
      immediate: true,
    },
```

Note that I am using this with NuxtJS with SSR (i see the -9 value both in server & client logs)

matthieusieben:
Wrong way to fix

PR #11160, opened 2020-04-19
docs: fix incorrect official MD icon page

DavidMulder0:
The previously linked page is not the official material design icons page. It's a community maintained page which lists both community icons and third party icons. It's not a bad page to occasionally use, but the third party icons make the 'For a list of all available icons' very confusing (when the link goes to a page with with unofficial not-included icons).


DavidMulder0:
Okay, correction, my bad, the fact that you guys are actually including a third party community curated list of unofficial icons (which misses official icons) didn't occur to me.

Issue #8474, opened 2019-08-12
[Feature Request] ability to modify VDataTable cell's properties without using `item` slot


cluidold:
### Problem to solve
When i create a datatable, i pass a headers and a item array to it.
If i want to add a style to a specific column (eg. color, font-size, alignment, disable sorting...) i have to use the template "<template v-slot:item.COLUMNNAME>...</template>". If i add a style in e extra <div> the style works, but the elements in the resulting table were nested like following.

< td>
   < div style:"color=red">Text< /div>
< /td>

instead of 

< td  style:"color=red">
   Text
< /td>

### Proposed solution
Add a property for passing the styles directly via the slot to prevent nesting of resulting elements.

makana:
IMO the headers config object needs a 'itemClass: string | string[]' config or just pass the current "class" config to the td elements and not just the headers. Also adding an additional class using the "value" option makes sense, i.e. th.header-<value> and td.item-<value>. So it would be easy to style the cells for a specific column. Currently it is uncomfortable because you either have to override the complete item slot to customize or use something like td:nth-child(2) to get cell elements.

MoritzMix:
+1 for this feature.
In my case i simply want to print a table and hide cells that contain control elements.
For now i have to do css workarounds as mentioned,

varadekd:
Any update on this?? Also can someone tell me to specify table row width using percentage

ktiedt:
> Any update on this?? Also can someone tell me to specify table row width using percentage

A row is as wide as the table....  do you mean columns? Those are controlled by percentage or width using the header array and the `width` property.

stevelacey:
BootstrapVue does a great job with this, it'd be great to have it work similar here

nekosaur:
@stevelacey In BootstrapVue, is this `_cellVariants` you're talking about?

stevelacey:
@nekosaur yes, anything that works like that, i.e. allowing the developer to pass the data along with their item objects would be :ok_hand: :ok_hand: it was very easy to grasp and implement solutions with

peteruithoven:
@ktiedt while usually using the headers to control the width is a good idea, you don't always want headers. (There even is a `hide-default-header` option.)

I love the cellClass idea.

ktiedt:
> @ktiedt while usually using the headers to control the width is a good idea, you don't always want headers. (There even is a `hide-default-header` option.)
> 
> I love the cellClass idea. 

[...]

PR #11254, opened 2020-04-28
feat(v-data-table): add item-class property

### LINKED ISSUE #8643: [Feature Request] ability to modify VDataTable row's properties without using `item` slot
### Problem to solve
No means of directly modifying the cell/row

### Proposed solution
Add a prop that can be applied to the template slotting the item in the data table which can apply changes to the cell conditionally. Such as background colors for the cell.

florealcab:
item-class property is  a function which takes 1 argument: the item data. The functions can return one class name, or a list of class names, or an object with, for each class name, true or false).
It can also be a string. In this case, for each item, the row classes are the ones defined inside the key of the same name in the item.

Fix #8643

## Description

Add a property to v-data-table component (not required) : item-class as described above.

Example :
```
<v-data-table
    :headers="my_headers"
    :item-class= "row_classes"
    ...
>
</v-data-table>

// JS :
...
methods: {
    row_classes(item) {
        {disabled: item.disabled, marked: item.marked}
    },
}
...
```

Another example:
```
<v-data-table
    :headers="my_headers"
    :item-class= "row_classes"
    ...
>
</v-data-table>

// JS :
...
data() {return {
    global_disable: true
}}
methods: {
    row_classes(item) {
        classes = []
        if (item.disabled or this.global_disable) {
            classes.push("disable")
        } else if (item.marked) {
            classes.push("marked")
        }
        return classes
    },
}
...
```

Example as a string:
```
<v-data-table
    :headers="my_headers"
    :items="table_data"
    item-class= "my_classes"
    ...
>
</v-data-table>

// JS :
...
data() {return {
    global_disable: true
    table_data: [{
        name: "Taste",
        date: "2020-04-29",
        my_classes: "first",
    },{
        name: "Pasta",
        date: "2020-04-30",
        my_classes: ["first", "second"],
    },{
        name: "Rice",
        date: "2020-03-29",
    }]
}}
...
```

Fix issue #8643 

## Motivation and Context

Sometimes we want to apply class to some rows on the table.

## How Has This Been Tested?

Unit tests added. Tested also on a real case.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


nekosaur:
I think I'd rather see something a bit more generic/open-ended.

It could take a simple string, an array of string, or an object with signature `Record<string, boolean>` just like the regular `class` attr. And/or it can take a function that returns any of the above.

For the sake of API (and implementation) simplicity, I'd probably vote for

`(item: any) => string | string[] | Record<string, boolean>`

When it comes to the name of the prop it should probably be called `item-class` to align it with other props, and slot names.

You will also need to add at least one unit test, and update the documentation

nekosaur:
I think you misunderstood me. The `item-class` prop should take only a function with the signature `(item: any) => string | string[] | Record<string, boolean>`

jacekkarczmarczyk:
> The item-class prop should take only a function

Maybe it could also take a string - item's prop that keeps the string|string[]|object:

```
<.. item-class="cls">

item: [{
  foo: 'bar',
  cls: 'red'
}, {
  foo: 'baz',
  cls: ['blue', 'large']
}]
```

nekosaur:
I'd prefer to keep the API simple. `:item-class="item => item.cls"` is still pretty easy

florealcab:
Oh, ok, I misundertood. It seems to be better like you say, I change that.

jacekkarczmarczyk:
`item => item.cls` is easy indeed, but i have 2 concerns
- it differs from other `item-*` props like `item-key` (string) or `item-text` etc from autocomplete (string | array | function), after using them one would assume that string would work also for `item-class`
- in my code i often make then function that transforms raw data from API or other source to data table's row data, imho it would be more convenient to create proper item's class in this function as well instead of creating it outside of the transforming function

packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
```diff
@@ -861,4 +861,44 @@ describe('VDataTable.ts', () => {
...
     expect(wrapper.html()).toMatchSnapshot()
   })
+
+  it('should apply class list to rows', () => {
+    const wrapper = mountFunction({
+      propsData: {
+        headers: testHeaders,
+        items: testItems,
+        itemsPerPage: 5,
+        'item-class': { 'my-class': true, 'my-other-class': true, 'my-none-class': false },
```
jacekkarczmarczyk, line 871: `itemClass` (in other places too)

florealcab:
@jacekkarczmarczyk so we all agree that item-class should be always a function. If it's a string or a list, it suppose that all rows have the class, and I don't know when it should be done. If class concern all rows, we just add it to the table. So if you're agree too, I change like @nekosaur said

nekosaur:
There can definitely be some value in keeping signatures consistent across props with similar names, but I don't think it should be a hard rule (not that you're advocating for it as such).

By the same token of similar prop names, one could argue for potential confusion between `class` and `itemClass`. If `class="foo"` applies `foo` to element, it's not too hard to imagine that `:itemClass="foo"` would apply `foo` to item. Instead of applying value of `item.foo`.

Still, I guess allowing `string | (item: any) => string | string[] | object` would only be re-using already existing convention ¯\_(ツ)_/¯

@vuetifyjs/contributors ?

jacekkarczmarczyk:
We all agree that it should at least support function, so you can implement it @florealcab

jacekkarczmarczyk:
> it's not too hard to imagine that :itemClass="foo" would apply foo to item

That would apply the same class for all items though, i don't think that's a real life use case

packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
florealcab: Done :)

florealcab:
Done. Now itemClass is a function. I updated the description of this PR to explain how it works. I hope it's okay for you like that.

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -97,6 +97,10 @@ export default VDataIterator.extend({
       type: Function as PropType<typeof defaultFilter>,
       default: defaultFilter,
     },
+    itemClass: {
+      type: Function,
+      default: (item: Object) => (Object()),
```
jacekkarczmarczyk, line 102:
```suggestion
      type: Function as PropType<XXX_PROP>, // see comment for line 205
      default: () => '',
```

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -198,6 +202,19 @@ export default VDataIterator.extend({
     calcWidths () {
       this.widths = Array.from(this.$el.querySelectorAll('th')).map(e => e.clientWidth)
     },
+    createItemClass (classes: string|string[]|Record<string, boolean>) {
+      let finalClasses = Object()
```
jacekkarczmarczyk, line 206:
```suggestion
      let finalClasses = {}
```

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -198,6 +202,19 @@ export default VDataIterator.extend({
     calcWidths () {
       this.widths = Array.from(this.$el.querySelectorAll('th')).map(e => e.clientWidth)
     },
+    createItemClass (classes: string|string[]|Record<string, boolean>) {
```
jacekkarczmarczyk, line 205: null/undefined should be allowed too 
Extract type and add to `types/index.d.ts`

It could be:

```ts
export type XXX = null|undefined|string|string[]|Record<string, boolean>
export type XXX_PROP = (item: any) => XXX
```
or

```ts
export type XXX_PROP = (item: any) => null|undefined|string|string[]|Record<string, boolean>

// usage
createItemClass (classes: ReturnType<XXX_PROP>) {...}
```

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -198,6 +202,19 @@ export default VDataIterator.extend({
...
+      let finalClasses = Object()
+      if (typeof classes === 'string') {
```
jacekkarczmarczyk, line 207: return empty object if string is empty
same for null/undefined

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -198,6 +202,19 @@ export default VDataIterator.extend({
...
+        finalClasses[classes] = true
+      } else if (classes instanceof Array) {
+        classes.forEach(klass => {
+          finalClasses[klass] = true
+        })
```
jacekkarczmarczyk, line 212:
```suggestion
        classes.forEach(class => (finalClasses[class] = true))
```

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -449,6 +466,7 @@ export default VDataIterator.extend({
         key: getObjectValueByPath(item, this.itemKey),
         class: {
           ...classes,
+          ...this.createItemClass(this.itemClass(item)),
```
jacekkarczmarczyk, line 469: maybe we could use mergeData here (and skip the `createItemClass` part, not sure how much less performant it would be

packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
```diff
@@ -861,4 +861,46 @@ describe('VDataTable.ts', () => {
...
+    expect(wrapper.html()).toMatchSnapshot()
+  })
+
+  it('should apply class unique to rows', () => {
+    const wrapper = mountFunction({
+      propsData: {
+        headers: testHeaders,
+        items: testItems,
+        itemsPerPage: 5,
+        itemClass: () => { return 'my-unique-class' },
```
jacekkarczmarczyk, line 884:
```suggestion
        itemClass: () => 'my-unique-class',
```

packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
```diff
@@ -861,4 +861,46 @@ describe('VDataTable.ts', () => {
...
     expect(wrapper.html()).toMatchSnapshot()
   })
+
+  it('should apply class list to rows', () => {
+    const wrapper = mountFunction({
+      propsData: {
+        headers: testHeaders,
+        items: testItems,
+        itemsPerPage: 5,
+        itemClass: () => { return ['my-class', 'my-other-class'] },
```
jacekkarczmarczyk, line 871:
```suggestion
        itemClass: () => ['my-class', 'my-other-class'],
```

packages/vuetify/src/components/VDataTable/__tests__/VDataTable.spec.ts
```diff
@@ -861,4 +861,46 @@ describe('VDataTable.ts', () => {
...
+  it('should apply class function to rows', () => {
+    const wrapper = mountFunction({
+      propsData: {
+        headers: testHeaders,
+        items: testItems,
+        itemsPerPage: 5,
+        itemClass: (item: Object) => {
+          return { 'first-class': item.fat < 10,
+            'second-class': item.protein > 4.0 }
+        },
```
jacekkarczmarczyk, line 900:
```suggestion
        itemClass: (item: object) => ({
          'first-class': item.fat < 10,
          'second-class': item.protein > 4.0,
        }),
```

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -198,6 +202,19 @@ export default VDataIterator.extend({
...
+      if (typeof classes === 'string') {
+        finalClasses[classes] = true
+      } else if (classes instanceof Array) {
+        classes.forEach(klass => {
+          finalClasses[klass] = true
+        })
+      } else {
+        finalClasses = classes
+      }
+      return finalClasses
```
jacekkarczmarczyk, line 216: https://www.notion.so/General-8ef7db0ab62a49f3a2f7cec7b67f3a44#6d2c6992893a4edd991c0a753682b445

packages/vuetify/src/components/VDataTable/VDataTable.ts
florealcab: If I understand how it works, 'class' item should be a key{str}: value{bool}, and mergeData returns an array, so I don't know how to do just with it?

jacekkarczmarczyk: with `mergeData` you should pass the whole `VNodeData` object, but you can wait for https://github.com/vuetifyjs/vuetify/pull/11271

florealcab: Ok I wait for that merge, so to eventually apply this change, so. All other reviews are fixed now

jacekkarczmarczyk: White `mergeClasses` you should be able to remove `createItemClass` and just do 
```js
class: mergeClasses(classes, this.itemClass(item))
```

johnleider:
@nekosaur sounds inline with what we already do in other areas. 👍

packages/vuetify/types/index.d.ts
```diff
@@ -254,3 +254,5 @@ export type DataItemsPerPageOption = (number | {
   text: string
   value: number
 });
+
+export type RowClassProp = (item: any) => null|undefined|string|string[]|Record<string, boolean>
```
jacekkarczmarczyk, line 258:
```suggestion
export type RowClassFunction = (item: any) => null|undefined|string|string[]|Record<string, boolean>
```

to be in line with https://github.com/vuetifyjs/vuetify/blob/b8677ad0d3237bc1fe935493c1dd00d1cb73bd2f/packages/vuetify/types/index.d.ts#L236 and others

packages/vuetify/src/components/VDataTable/VDataTable.ts
jacekkarczmarczyk: @florealcab merge, see example usage in https://github.com/vuetifyjs/vuetify/commit/9257de2b4c7960d1fd8846355aa54b9f5048bb12

florealcab: It works well, I remove the createItemClass method and use the new mergeClasses method instead! So fixed

florealcab:
I updated my branch from this repo and resolve conflicts, so one you approve all changes you can merge, no more conflicts.

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -450,10 +456,9 @@ export default VDataIterator.extend({
 
       return this.$createElement(this.isMobile ? MobileRow : Row, {
         key: getObjectValueByPath(item, this.itemKey),
-        class: {
-          ...classes,
-          'v-data-table__selected': data.isSelected,
-        },
+        class: mergeClasses(
+          { ...classes, 'v-data-table__selected': data.isSelected },
+          this.itemClass(item)),
```
jacekkarczmarczyk, line 461:
```suggestion
          this.itemClass(item),
        ),
```

johnleider:

packages/docs/src/lang/en/components/DataTables.json
```diff
@@ -117,6 +117,7 @@
       "headersLength": "Can be used in combination with `hide-default-header` to specify the number of columns in the table to allow expansion rows and loading bar to function properly",
       "height": "Set an explicit height of table",
       "hideDefaultHeader": "Hide the default headers",
+      "itemClass": "Function that takes item as argument and returns the list of classes to add to the row of the item",
```
sh7dm, line 120:
```suggestion
      "itemClass": "Function that takes item as argument and returns the list of classes to add to the item",
```

sh7dm: Not sure, but this text may be better. @vuetifyjs/core-team thoughts?

packages/vuetify/types/index.d.ts
```diff
@@ -254,3 +254,5 @@ export type DataItemsPerPageOption = (number | {
   text: string
   value: number
 });
+
+export type RowClassFunction = (item: any) => null|undefined|string|string[]|Record<string, boolean>
```
sh7dm, line 258:
```suggestion
export type RowClassFunction = (item: any) => null | undefined | string | string[] | Record<string, boolean>
```

packages/docs/src/lang/en/components/DataTables.json
florealcab: "Function that takes item as argument and returns the list of classes to add to table row corresponding to the item" maybe?

sh7dm: Ah, I was wrong.

nekosaur:
Add ability to pass `string` like @jacekkarczmarczyk mentioned earlier. This would map to a property on the item with that value. Use `getObjectValueByPath`.

ktiedt:
Is it possible to also extend similar functionality to the table cells? It seems odd that header classes only populate on the header cells, when they would generally be expected to also decorate the table cell classes so that they similarly be styled (and differentiated in styling by the thead or tbody selectors).

nekosaur:
@ktiedt That's a separate issue <https://github.com/vuetifyjs/vuetify/issues/8474>

florealcab:
@nekosaur now it's done. It can be a string or a function. I remove the default value of the itemClass prop, as it's not needed, all it's working if itemClass is null.

I also add a unit test to test the case of itemClass is a string.

packages/vuetify/src/components/VDataTable/VDataTable.ts
```diff
@@ -458,7 +455,7 @@ export default VDataIterator.extend({
         key: getObjectValueByPath(item, this.itemKey),
         class: mergeClasses(
           { ...classes, 'v-data-table__selected': data.isSelected },
-          this.itemClass(item)
+          typeof this.itemClass === 'function' ? this.itemClass(item) : getObjectValueByPath(item, this.itemClass)
```
jacekkarczmarczyk, line 458: use `getPropertyFromItem`, it accepts function as well so no need to checking if itemClass is function or string

florealcab: Oh, perfect, I change that! :)

florealcab: So, I restore the default the default value, as null value cause unexpected behavior now. But it works :)

stevelacey:
Any way to expand on this to support per-cell classes? I'd say this still falls short of what is possible in BootstrapVue – it's super handy to be able to supply class names along with the input data. Modifying the template to do this is a chore, and what I've seen developers end up doing is breaking the responsive behaviours by overwriting the t-body slot or hacks like `align: ' d-none'` :confused:

ktiedt:
@stevelacey read my comment and its response above ;)

https://github.com/vuetifyjs/vuetify/pull/11254#issuecomment-626929455

Issue #11289, opened 2020-05-01
[Bug Report] v-btn with input type="submit" not supported


dougalg:
### Environment
**Vuetify Version:** 2.2.26
**Vue Version:** 2.6.11
**Browsers:** Firefox 75.0
**OS:** Mac OS 10.15

### Steps to reproduce
The `type` prop is ignored here

```
<v-btn
	tag="input"
	type="submit"
	value="Log In"
	name="login"
	elevation="0"
/>
```

### Expected Behavior
The `type` prop should be rendered on the `input` button.

### Actual Behavior
The type prop is ignored.

### Reproduction Link
<a href="https://codepen.io/dougalg/pen/XWmejzv" target="_blank">https://codepen.io/dougalg/pen/XWmejzv</a>

### Other comments
In the v-btn render function I can see that the type is ignored if the tag is not a button:

```
    if (tag === 'button') {
      data.attrs!.type = this.type
      data.attrs!.disabled = this.disabled
    }
```

johnleider:
This feels like a very large change for a very niche issue. Thoughts @vuetifyjs/core-team @vuetifyjs/contributors?

jacekkarczmarczyk:
I'm not against it in general but not sure if this https://github.com/vuetifyjs/vuetify/pull/11291/files#diff-b9df24aeee7383f4d61be9e1db86b8cdR39 is the best way to do it (don't have a better idea atm so sorry for not being constructive)

dougalg:
@johnleider @jacekkarczmarczyk I've made a small change to the PR so that the fix still remains in place, but it removes the `CAN_HAVE_X` logic. Now it allows `type`/`disabled` on any tag. The caveat is that `type` needs to be manually specified if `tag` is not a button.

dougalg:
Personally, I view this is an accessibility concern. It's more semantic, and you get automatic form bindings for things like the enter key, and so on.  I think we should support `input type="submit"` as a very expected use case, it doesn't seem like a niche issue to me, but rather like supporting standard, semantic HTML.

phiter:
@dougalg actually using a button with `type="submit"` already works for submitting the form automatically on pressing enter. There is not much of a difference between the two.

[Difference between <button type="submit"> and <input type="submit">](
https://html.com/attributes/button-type/#Difference_between_ltbutton_typesubmitgt_and_ltinput_typesubmitgt)

YipingRuan:
[...]

PR #11291, opened 2020-05-01
fix(VBtn): allow `type` and `disabled` for non-button buttons

dougalg:
Sometimes we want button styles on non-button elements, such as an input, or other. Some of these elements support the `type` and/or `disabled` attributes, but currently vuetify does not allow them to get rendered unless VBtn is a "button".

https://github.com/vuetifyjs/vuetify/issues/11289

## Description

Resolve https://github.com/vuetifyjs/vuetify/issues/11289

There's no reason to explicitly remove `type` and `disabled` for non-button elements. If the concern is extraneous tags for cases where the attributes don't work, then it may make sense to list all tags that support `disabled` and all tags that support `type`, although this feels not very scalable.

## Motivation and Context

This allows the use of the semantic `<input type="submit"..` element which visually is rendered as a button.

## How Has This Been Tested?

unit tests

## Markup:

<details>

```vue
<div id="app">
  <v-app>
    <v-btn
      tag="input"
      type="submit"
      value="Log In"
      name="login"
      elevation="0"
    />
  </v-app>
</div>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/.eslintrc.js
+++ b/packages/vuetify/.eslintrc.js
@@ -30,10 +30,18 @@ module.exports = {
     'jest/prefer-to-contain': 'warn',
     'jest/prefer-to-have-length': 'warn'
   },
-  overrides: [{
-    files: 'dev/Playground.vue',
-    rules: {
-      'max-len': 'off'
+  overrides: [
+    {
+      files: 'dev/Playground.vue',
+      rules: {
+        'max-len': 'off'
+      }
+    },
+    {
+      files: '*.spec.ts',
+      rules: {
+        'max-statements': 'off'
+      }
     }
-  }],
+  ],
 }
--- a/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
+++ b/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
@@ -96,6 +96,17 @@ describe('VBtn.ts', () => {
     expect(wrapper.html()).toMatchSnapshot()
   })
 
+  it('should should render disabled on an <a> tag when using href prop', () => {
+    const wrapper = mountFunction({
+      propsData: {
+        href: 'http://www.google.com',
+        disabled: true,
+      },
+    })
+
+    expect(wrapper.html()).toMatchSnapshot()
+  })
+
   it('should render specified tag when using tag prop', () => {
     const wrapper = mountFunction({
       propsData: {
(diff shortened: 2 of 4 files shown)


nekosaur:
not sure I see the harm in just adding the attributes regardless of what type of element.

dougalg:
I agree, would prefer to just allow all types. I think the original bug was that because `type` is set by default to `button`, it was showing up as `type="button"` in places where it shouldn't, which is an accessibility issue. 

Most likely, that could be dealt with in a simpler way. We always set the type, but we only default the `type` to `button` if the component is a button, otherwise the implementer must pass it explicitly.

@johnleider what do you think? That way I think I could get rid of those constants at the top of the file.

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
```diff
@@ -96,6 +96,17 @@ describe('VBtn.ts', () => {
     expect(wrapper.html()).toMatchSnapshot()
   })
 
+  it('should should render disabled on an <a> tag when using href prop', () => {
```
nekosaur, line 99: double should

packages/vuetify/.eslintrc.js
```diff
@@ -30,10 +30,18 @@ module.exports = {
     'jest/prefer-to-contain': 'warn',
     'jest/prefer-to-have-length': 'warn'
   },
-  overrides: [{
-    files: 'dev/Playground.vue',
-    rules: {
-      'max-len': 'off'
+  overrides: [
```
jacekkarczmarczyk, line 33: that should be rather a separate PR, also i'm not sure if this is good idea, if there's too many `it(...)` in one `describe()` maybe it's a sign that they should be split into separate `discrabe()`s

dougalg: That's fair. I'll revert this. Any suggestions on how to split up the describes, or do you know of any other test files that I could look at to get an idea?

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
dougalg: Because I believe it so strongly! Hahah, thanks, will fix.

packages/vuetify/.eslintrc.js
jacekkarczmarczyk: VIcon.spec.ts, colorUtils.spec.ts etc

dougalg:
Closing this, as it seems, based on discussion in the initial bug report that it isn't really useful.

PR #11301, opened 2020-05-02
feat(VVirtualScroll): add horizontal scrolling mode

albertodeago:
Add horizontal mode to VVirtualScroll component

## Description
Add horizontal mode to VVirtualScroll component
If also fix a small bug to the component, precisely it fixes the `key` for the items (as of today it's bound to a wrong index)

## Motivation and Context
It's an improvement to the just added VVirtualScroll component. 
This will enable users to use the component when in need to display large amount of data horizontally.

## How Has This Been Tested?
Unit test added for the new functionality and update previously implemented tests

## Markup:
<details>

```vue
<template>
  <v-container>
    <v-virtual-scroll height="500px" :items="items" item-size="50">
      <template v-slot="{ item }">
        <div class="item">
          I'm item number {{ item.id }}
        </div>
      </template>
    </v-virtual-scroll>
    
    <br>
    
    <v-virtual-scroll
      :items="items"
      item-size="500"
      width="500"
      height="500"
      horizontal
    >
      <template v-slot="{ item }">
        <div class="item">
          I'm item number {{ item.id }}
        </div>
      </template>
    </v-virtual-scroll>
  </v-container>
</template>

<script>
import VVirtualScroll from '../src/components'

const fillArrayWithNumbers = n => Array.apply(null, Array(n)).map((x, i) => i);
export default {
  data: () => ({
    items: fillArrayWithNumbers(5000).map(i => ({id: i, content: "content-" + i})),
    itemH: 25
  })
}
</script>
```
</details>

## Types of changes

## Checklist:


nekosaur:
would it not be better to have v-scroll support horizontal mode as well instead of rolling your own for this component?

albertodeago:
> would it not be better to have v-scroll support horizontal mode as well instead of rolling your own for this component?

I realized I messed up with the v-scroll, my bad, going to use v-scroll as it should

> Looks fine at first glance, but fails in RTL mode

Uh, going to have a look on that too. Never taken in consideration RTL mode in any of my projects ^^

albertodeago:
Trying to fix RTL mode I'm running into some issues with browsers different behaviours, like [this](https://stackoverflow.com/questions/24276619/better-way-to-get-the-viewport-of-a-scrollable-div-in-rtl-mode/24394376#24394376)
And I see there are [packages](https://www.npmjs.com/package/normalize-scroll-left) just to solve this kind of issue

Do we have something in Vuetify to handle this (maybe you already faced with some other component) or should I try to build something for this purpose?

johnleider:
There is not good cross browser support for this as we discussed on Discord. Closing.

PR #11304, opened 2020-05-02
fix(VDatePicker): allow null value for multiple date picker

### LINKED ISSUE #9529: [Bug Report] Clear causing error when using v-date-picker with v-text-field
### Environment
**Vuetify Version:** 2.1.7
**Vue Version:** 2.6.10
**OS:** Linux x86_64

### Steps to reproduce
Create a vDatePicker with range=true in menu using a clearable vTextField as explained in the docs. Select any date/range and the click the clear button in the vTextField 

### Expected Behavior
vTextField clear button should set the target v-model to an emtpy array [ ] to prevent array errors when target v-model is an array

### Actual Behavior
vTextField clear button sets the target v-model to null which causes the error. This works fine for a vDatePicker without range property because target v-model is not an array, nor does the vDatePicker itself expect an array

### Reproduction Link
<a href="https://codepen.io/hope_feda/pen/abbVdoj" target="_blank">https://codepen.io/hope_feda/pen/abbVdoj</a>

### Other comments
I tried setting the target v-model to an empty array [ ] using the click:clear event. However, this did not work because apparently click:clear doesn't override the default behaviour of setting the target v-model to null.
I'm not sure if the solution is that the clear event should set the target v-model to an empty array if its an array originally(null when not an array), or if click:clear event should override default behaviour so we can customize how its cleared.

jacekkarczmarczyk:
## Description
Allows null/undefined value for range/multiple date pickers

fix, but deliberately dev branch

## Motivation and Context
fixes #9529

## How Has This Been Tested?
jest, playground

## Markup:

<details>

```vue
<template>
  <v-menu
    max-width="290"
  >
    <template #activator="{ on }">
      <v-text-field
        v-model="value"
        clearable
        :label="JSON.stringify(value)"
        readonly
        v-on="on"
      />
      <v-checkbox
        v-model="isRange"
        label="true = range, false = mutliple"
      />
    </template>
    <v-date-picker
      v-model="value"
      v-bind="{ [isRange ? 'range' : 'multiple']: true }"
    />
  </v-menu>
</template>

<script>
  export default {
    data: () => ({ value: null, isRange: true }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
@@ -7,19 +7,19 @@ import VDatePickerYears from './VDatePickerYears'
 
 // Mixins
 import Localable from '../../mixins/localable'
-import mixins from '../../util/mixins'
 import Picker from '../../mixins/picker'
 
 // Utils
+import mixins from '../../util/mixins'
+import isDateAllowed from './util/isDateAllowed'
+import { wrapInArray } from '../../util/helpers'
+import { daysInMonth } from '../VCalendar/util/timestamp'
+import { consoleWarn } from '../../util/console'
 import {
   createItemTypeListeners,
   createNativeLocaleFormatter,
   pad,
 } from './util'
-import isDateAllowed from './util/isDateAllowed'
-import { consoleWarn } from '../../util/console'
-import { daysInMonth } from '../VCalendar/util/timestamp'
-import { wrapInArray } from '../../util/helpers'
 
 // Types
 import {
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
@@ -279,7 +279,7 @@ export default mixins(
       this.activePicker = type.toUpperCase()
 
       if (this.value && this.value.length) {
-        const output = wrapInArray(this.value)
+        const output = this.multipleValue
           .map((val: string) => sanitizeDateString(val, type))
           .filter(this.isDateAllowed)
         this.$emit('input', this.isMultiple ? output : output[0])
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
@@ -151,7 +151,7 @@ export default mixins(
           return this.pickerDate
         }
 
-        const multipleValue = this.value ? wrapInArray(this.value) : []
+        const multipleValue = wrapInArray(this.value)
         const date = ((this.multiple || this.range) ? multipleValue[multipleValue.length - 1] : this.value) ||
           (typeof this.showCurrent === 'string' ? this.showCurrent : `${now.getFullYear()}-${now.getMonth() + 1}`)
         return sanitizeDateString(date as string, this.type === 'date' ? 'month' : 'year')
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/components/VDatePicker/VDatePicker.ts
```diff
@@ -134,25 +135,30 @@ export default mixins(
           return this.pickerDate
         }
 
-        const date = (this.multiple || this.range ? (this.value as string[])[(this.value as string[]).length - 1] : this.value) ||
+        const multipleValue = this.value ? wrapInArray(this.value) : []
```
johnleider, line 138: If we have access to `this` could we use `this.multipleValue`?

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
```diff
@@ -250,15 +256,15 @@ export default mixins(
...
+      } else if (this.isMultiple && this.multipleValue.length && (!oldValue || !(oldValue as string[]).length) && !this.pickerDate) {
         this.tableDate = sanitizeDateString(this.inputDate, this.type === 'month' ? 'year' : 'month')
       }
     },
     type (type: DatePickerType) {
       this.activePicker = type.toUpperCase()
 
       if (this.value && this.value.length) {
-        const output = (this.isMultiple ? (this.value as string[]) : [this.value as string])
+        const output = wrapInArray(this.value)
```
johnleider, line 267: Same here, could we use `multipleValue`?

johnleider:
Should this target master now that 2.3 is out?

jacekkarczmarczyk:
yes, I'll rebase

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
jacekkarczmarczyk: computed accessed in data() returns undefined, `this.$options.computed.multipleValue.call(this)` works but it doesn't make sense

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
```diff
@@ -7,18 +7,19 @@ import VDatePickerYears from './VDatePickerYears'
 
 // Mixins
 import Localable from '../../mixins/localable'
-import mixins from '../../util/mixins'
 import Picker from '../../mixins/picker'
 
 // Utils
+import mixins from '../../util/mixins'
+import isDateAllowed from './util/isDateAllowed'
+import { wrapInArray } from '../../util/helpers'
```
johnleider, line 15: order alphabetically

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
```diff
@@ -293,11 +298,11 @@ export default mixins(
...
   methods: {
     emitInput (newInput: string) {
-      if (this.range && this.value) {
-        if (this.value.length !== 1) {
+      if (this.range) {
+        if (this.multipleValue.length !== 1) {
           this.$emit('input', [newInput])
         } else {
-          const output = [...this.value, newInput]
+          const output = [this.multipleValue[0], newInput]
```
johnleider, line 305: is it possible for **multipleValue** to be 0 and if so is **undefined** an acceptable value for **output** on 305?

jacekkarczmarczyk: not possible, in this branch of `if` length of multipleValue is always equal 1

PR #11314, opened 2020-05-03
refactor(types-import): change import path for "from 'types'"

bramceulemans:
## Description

The import path from 

```typescript
from 'types'
```
has been changed to
```ts
from 'vuetity/types'
```

## Motivation and Context

The reason for this change is that PhpStorm doesn't currently recognize the 'types' path if Vuetify is used as a library.

## How Has This Been Tested?

After this change I recompiled my code and checked if the VDataTable was still function, everything seems to be working just fine.

## Types of changes

## Checklist:


jacekkarczmarczyk:
Thanks, looks better, few notes:
- can you explain what do you mean by "The reason for this change is that PhpStorm doesn't currently recognize the 'types' path if Vuetify is used as a library."? I'm using PhpStorm and didn't noticed this problem
- your first PR changes the import path to relative (`../../types`), this one uses `vuetify/types`, not sure if there is any significant difference between both, definitely `vuetify/types` is easier to automatic search/replace. If it works i'm fine with it, but if it can cause any issues it would be good to know
- any reason for making it on dev instead of master?

bramceulemans:
To answer your questions:
- I am using Laravel with Vue + Vuetify, and whenever I would try to import `DataTableHeader`, it wouldn't let me, because the absolute path `from 'types'` wouldn't get recognized.
- There should not be a difference but I figured `vuetify/types` is easier to read.
- From what I read `dev` was a better fit for "enhancements" (<https://vuetifyjs.com/en/getting-started/contributing/#pull-requests-for-vuetify>).

jacekkarczmarczyk:
you should import types from `vuetify`:
```ts
import { DataTableHeader } from 'vuetify';
```

bramceulemans:
I know, but when clicking around the '`types'` still showed up as red text, (not finding the import when using Vuetify as a library).

jacekkarczmarczyk:
The you have some other issue or i misunderstood, it works fine for me:
image

Anyway, we're slightly off topic now, we can continue it on discord or in a new issue, PR is fine for me

PR #11327, opened 2020-05-05
fix(vvirtualscroll): the key of the children is not right

albertodeago:
The key of the items is not right. It should never change to avoid unnecessary re-renders

## Description
Change key value from the index of the element based on the rendered to the index of the element based on all items

## Motivation and Context
This change will improve the component performance.

## How Has This Been Tested?
Playground example.
See example provided in the section below and test it with browser cache disabled to visually see the issue

## Markup:
<details>

```vue
<template>
  <v-container>
    <v-virtual-scroll height="750px" :items="items" item-height="250">
      <template v-slot="{ item }">
        <div class="item">
          I'm item number {{ item.id }}
          <v-img width="200" height="200" :src="'https://picsum.photos/id/'+ item.id +'/200/200'"></v-img>
        </div>
      </template>
    </v-virtual-scroll>
  </v-container>
</template>

<script>
import VVirtualScroll from '../src/components'

const fillArrayWithNumbers = n => Array.apply(null, Array(n)).map((x, i) => i);
export default {
  data: () => ({
    items: fillArrayWithNumbers(5000).map(i => ({id: i, content: "content-" + i})),
  })
}
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.ts
+++ b/packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.ts
@@ -69,13 +69,13 @@ export default Measurable.extend({
       ).map(this.genChild)
     },
     genChild (item: any, index: number) {
-      const firstToRender = this.firstToRender + index
-      const top = convertToUnit(firstToRender * this.__itemHeight)
+      const indexToRender = this.firstToRender + index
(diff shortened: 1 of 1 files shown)


johnleider:
Looks good 👍

PR #11372, opened 2020-05-11
fix(VDatePicker): throttle scroll with using scrollable prop

### LINKED ISSUE #5546: [Bug Report] Datepicker scroll is way too sensitive on mac
### Versions and Environment
**Vuetify:** 1.3.7
**Vue:** 2.5.17
**OS:** Mac OS 10.14.0

### Steps to reproduce
Open the datepicker and scroll to navigate in the calendar

### Expected Behavior
Normal scroll

### Actual Behavior
Calendar pages scrolling way too fast. 
It's impossible to scroll to the next page because the tiniest scroll movement brings you 5 or 6 pages away.

### Reproduction Link
<a href="https://vuetifyjs.com/en/components/date-pickers#example-date-dialog-and-menu" target="_blank">https://vuetifyjs.com/en/components/date-pickers#example-date-dialog-and-menu</a>

Logikgate:
Fixes #5546

## Description
Fixes #5546 by adding a 250ms throttle to the function reacting to scroll wheel events

## Motivation and Context
Fixes blurring that happens when VDatePicker is scrolled via a touchpad.  Reference #5546

## How Has This Been Tested?
Tested visually on a MacBook Pro on safari, chrome and firefox.  Passes all existing tests.

## Markup:

<details>

```vue
<v-date-picker scrollable></v-date-picker>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -365,6 +365,17 @@ export function debounce (fn: Function, delay: number) {
   }
 }
 
+export function throttle (fn: Function, limit: number) {
+  let throttling = false
+  return (...args: any[]) => {
+    if (!throttling) {
+      fn(...args)
+      throttling = true
+      setTimeout(() => throttling = false, limit)
+    }
+  }
+}
+
 export function getPrefixedScopedSlots (prefix: string, scopedSlots: any) {
   return Object.keys(scopedSlots).filter(k => k.startsWith(prefix)).reduce((obj: any, k: string) => {
     obj[k.replace(prefix, '')] = scopedSlots[k]
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -365,13 +365,13 @@ export function debounce (fn: Function, delay: number) {
   }
 }
 
-export function throttle (fn: Function, limit: number) {
+export function throttle<T extends (...args: any[]) => any> (fn: T, limit: number) {
   let throttling = false
-  return (...args: any[]) => {
+  return (...args: Parameters<T>): void | ReturnType<T> => {
     if (!throttling) {
-      fn(...args)
       throttling = true
       setTimeout(() => throttling = false, limit)
+      return fn(...args)
     }
   }
 }
--- a/packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
+++ b/packages/vuetify/src/components/VDatePicker/mixins/date-picker-table.ts
@@ -11,6 +11,7 @@ import Themeable from '../../../mixins/themeable'
 // Utils
 import isDateAllowed from '../util/isDateAllowed'
 import mixins from '../../../util/mixins'
+import { throttle } from '../../../util/helpers'
 
 // Types
 import { VNodeChildren, PropType } from 'vue'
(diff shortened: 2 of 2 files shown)


jacekkarczmarczyk:
I've tried this solution but didn't like it much. Currently you can quickly scroll through months with mouse wheel but with this PR turning the wheel even by a big angle scrolls just by 1 month.
Also if you dynamically swipe on your touchpad picker will be changing months for few more seconds after you finish swiping

Logikgate:
Is the standard use case for scrolling to be able to scroll through multiple years of months? Or is it to provide the ability to scroll between months close to the current selection?  I would guess that the latter is used more often and that use case is currently broken on sensitive trackpads.

I think we can find a throttle that satisfies both of use cases pretty well. I just started with 250ms to begin the conversation but played around with it as low as 100.  Anywhere in that range seemed to work well.  At 100ms one finger swipe of a mouse wheel with detents seems to consistently navigate four months.

One other thing to consider is maybe scrolling through multiple years of months shouldn't be taken into account since it is easy for a user to make large changes to the date by clicking on the year.

jacekkarczmarczyk:
Overall experience with this PR even if not ideal it still seems to be better than the current one, at least for touch pad users, I just hoped for ideal solution ;)

I actually didn't look how other frameworks deal with this problem, maybe I will, if you have any findings about it please share

In the mean time there's at least one issue in the code that needs to be fixed - `throttle` function doesn't preserve typings, type of arguments of returned function should be inferred from `fn`

Also if we approve it would you be interested in creating similar PR for VTimePicker?

Logikgate:
I don't know of specific frameworks but most of my interaction with date pickers around the web it seems they don't include the ability to scroll through months.

Yep I'm game to adding the same throttling to VTimePicker :)

I just updated the throttle function to maintain the argument and return types of the passed in function. Are you thinking you also want the limit on the throttle to be changed or is it good at 250ms?

packages/vuetify/src/util/helpers.ts
```diff
@@ -365,6 +365,17 @@ export function debounce (fn: Function, delay: number) {
   }
 }
 
+export function throttle<T extends (...args: any[]) => any> (fn: T, limit: number) {
```
johnleider, line 368: This type of functionality was removed last May https://github.com/vuetifyjs/vuetify/commit/5aa23f15bf20127256e64561826bcfb2c0254006

Logikgate: Does that mean you prefer to not have throttle helper function at all or that you would like to change the implementation of it to be like the one the was removed?

johnleider: I just want to understand why it was removed. Was it because after removing its use within the `v-color-picker` it had no purpose? Or was it because the throttle caused issues. I'm leaning towards the former, because this seems like it would be useful.

I'm not sure if @KaelWD can chime in but if not, I'd be interested in exploring other areas where this could be used.

johnleider:
@jacekkarczmarczyk can you do a once over please? Thank you

KaelWD:
What is `event.deltaY` and `event.wheelDelta` on the macbook?

Logikgate:
@KaelWD `event.deltaY` on my macbook pro when scrolling at a normal pace varies between 1 and 18. `event.wheelDelta` is between 3 and 45.

PR #11421, opened 2020-05-17
feat(VDataPicker): add/correct persian calendar support

### LINKED ISSUE #1892: [Bug Report] v-date-picker new month is started at the end of previous month in FA-local
### Versions
Vuetify 0.15.7 and vue2.4

### What is expected ?
In date picker month should be started with its first day. which is not working for 'FA' locals.

for current date `26 SEP 2017`, correct value in `FA` local is  `مهر ۴, ۱۳۹۶`
۴ means 4 in English and مهر is the month 7th in Persian and `۱۳۹۶ ` means 1396.
but in the date picker white space header it is showing `۱۳۹۶ شهریور`  .
 `شهریور` is the month 6 in `FA` local.
on the other hand for blue space it is showing `مهر ۴`   means 4th day of month 7 which is correct one.

Another problem is date picker table.  in codepen month 7 is started at the end of month 6.  `۱` or 1 in English is the first day of month 7 in `FA` local which should be in the first row of table but now it is on the last row after `۳۱` or 31 of month 6 in `FA`  local. 

 ### Reproduction Link
https://codepen.io/MUHU/pen/QqdVLP

Ashafaat:
## Description

some minor but effective changes are necessary for fixing the misbehaviour of date picker in fa-IR locale. this pull request applys them and has no effect if the locale is something other than fa-IR. this component needs a massive refactor for a real support for special calendars (like persian/jalali calendar) and this is just a dirty solution. but until then, this pull request works like a charm! (at least i didn't find any problem)
fixes #1892
## Motivation and Context

fixes #1892 
## How Has This Been Tested?

i have tested it manually. no unit/auto test has implemented.
## Markup:

<details>

```vue
<template>
  <v-container>
    <v-date-picker
      v-model="date"
      locale="fa-IR"
      first-day-of-week="6"
    ></v-date-picker>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      date: null,
    }),
    beforeCreate () {
      this.$vuetify.rtl = true
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
@@ -19,6 +19,7 @@ import {
 import isDateAllowed from './util/isDateAllowed'
 import { consoleWarn } from '../../util/console'
 import { daysInMonth } from '../VCalendar/util/timestamp'
+import { gregorianToJalali } from './util/jdf'
 
 // Types
 import {
@@ -134,8 +135,15 @@ export default mixins(
           return this.pickerDate
         }
 
-        const date = (this.multiple || this.range ? (this.value as string[])[(this.value as string[]).length - 1] : this.value) ||
+        let date = (this.multiple || this.range ? (this.value as string[])[(this.value as string[]).length - 1] : this.value) ||
           (typeof this.showCurrent === 'string' ? this.showCurrent : `${now.getFullYear()}-${now.getMonth() + 1}`)
+        if (this.locale === 'fa-IR') {
+          const jdate = gregorianToJalali(parseInt((date as string).split('-')[0]),
+            parseInt((date as string).split('-')[1]),
+            parseInt((date as string).split('-')[2] || 1)
+          )
+          date = `${jdate[0]}-${pad(jdate[1])}-${pad(jdate[2])}`
+        }
         return sanitizeDateString(date as string, this.type === 'date' ? 'month' : 'year')
       })(),
     }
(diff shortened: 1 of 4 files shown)


nekosaur:
1. It seems like you've accidentally included some extra commits. Please rebase with only your changes.

2. This should probably apply to more locales than just `fa-IR`? For example in the issue `ar-sa` is also mentioned.

3. Not sure what stance @johnleider has on this, but personally I'd prefer if the utility functions were at least MIT (or comparable) licensed instead of GPL/LGPL. Or if the algorithm is generally available we could roll our own.

4. I'm not a huge fan of the hard-coding being done here. But I also recognize that a more proper solution could be a much larger undertaking. Regardless if this is to go forward I think there are a couple of things that could be done cleaner. What are your thought @jacekkarczmarczyk?

5. You need to add tests for this

Ashafaat:
> 1. It seems like you've accidentally included some extra commits. Please rebase with only your changes.
> 2. This should probably apply to more locales than just `fa-IR`? For example in the issue `ar-sa` is also mentioned.
> 3. Not sure what stance @johnleider has on this, but personally I'd prefer if the utility functions were at least MIT (or comparable) licensed instead of GPL/LGPL. Or if the algorithm is generally available we could roll our own.
> 4. I'm not a huge fan of the hard-coding being done here. But I also recognize that a more proper solution could be a much larger undertaking. Regardless if this is to go forward I think there are a couple of things that could be done cleaner. What are your thought @jacekkarczmarczyk?
> 5. You need to add tests for this

1. sorry, i solved it.
2. no, `ar-sa` has a different calendar than persian, their calendar is Hijri which has a total different logic. i can implement Hijri too but i think it's better to be implemented by a arab person.
3. i can reimplement them with just changing the variable names, because the logic is static.
4.  
5. i have no idea about unit/auto testing, it will take me a lot of time get fimiliar with tools and other things. i prefer not to do it as it will not affect other locale or components and datepicker is not usable in fa-IR, so it's a improvment anyway!

johnleider:
v2.3 is going on lock soon so this won't make it into the 2.x major.

Issue #11364, opened 2020-05-11
[Feature Request] VSlideGroup: Events when reaching the start/end of the group


LoneRifle:
### Problem to solve
Items in slide groups can currently be generated programmatically using the `v-for` directive. It is desired that when a user wishes to progress beyond the end of the slide group (or before the beginning), we are notified, so that we can proceed to generate more items for the `v-for` directive, and do some housekeeping (eg, perhaps remove items that have been scrolled way past the view to avoid consuming too much memory).

Doing so will allow a Vuetify user to create dynamic scrolling slide groups, with the items being generated in response to user events.

### Proposed solution
An event could be emitted when:
  - a user scrolls to either end of the group, as well as;
  - when the user is at the end and clicks the prev/next button

PR #11440, opened 2020-05-19
feat(VSlideGroup): emit events near start/end

LoneRifle:
## Description
Implement two new VSlideGroup events - `near:start` and `near:end` -
using the following heuristic:

- When the content width is greater than twice the wrapper width,
  trigger an event if we are approaching to within one scroll of
  the start/end

- When the content width is equal to or less than the wrapper width,
  trigger an event if we are at the start/end

Fixes vuetifyjs/vuetify#11364

## Motivation and Context

Some VSlideGroup users may want to listen for events indicating the
approaching towards the start or end of a slide group. These new events
and the heuristic that triggers them are so that users may be able to act _before_
the start/end has been reached, which would be useful when eg,
dynamically generating slide group items to simulate infinite scrolling

## How Has This Been Tested?

- Implement tests that exercise this for:
  - content width > 2 * wrapper widths
  - content width < 2 * wrapper widths

## Markup:
N/A - used GitPod

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-generator/src/helpers/variables.js
+++ b/packages/api-generator/src/helpers/variables.js
@@ -287,13 +287,32 @@ const VCalendarDay = {
   future: 'boolean',
 }
 
+const VCalendarEvent = {
+  input: 'any',
+  start: VTimestamp,
+  startIdentifier: 'number',
+  startTimestampIdentifier: 'number',
+  end: VTimestamp,
+  endIdentifier: 'number',
+  endTimestampIdentifier: 'number',
+  allDay: 'boolean',
+  index: 'number',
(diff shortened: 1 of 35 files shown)


LoneRifle:
Failing test appears to be due to 137186a, due to a stale Jest snapshot for VBtn. Will rebase when needed

johnleider:
I like this feature but unfortunately it's not going to make it in for v2.3 and with v3 being a completely different structure, I'm going to close this.

With that being said, I imagine that this logic will be useful during our rebuild. Reach out to me in our community if you're interested in partaking.

PR #11454, opened 2020-05-20
fix(VTextField): allow numeric input starting with -0 

### LINKED ISSUE #9368: [Bug Report] Bug in numeric v-text-field when typing negative decimals
### Environment
**Vuetify Version:** 2.1.3
**Vue Version:** 2.6.10
**Browsers:** Chrome 78.0.3904.50
**OS:** Windows

### Steps to reproduce
Manually type "-0.1" into `<v-text-field v-model.number="value" type="number" step="any">`

### Expected Behavior
Expected value = -0.1
(or another simple way to support numeric inputs)

### Actual Behavior
While typing "-0.1", vuetify converts "-0" into "0", so user ends up with "0.1" in input, and value = 0.1.

Works well for other values:

- 10.5
- -10.5
- 0.5
- -.5
- +0.5
- -1e-5

### Reproduction Link
<a href="https://codepen.io/lukaw3d/pen/ExxyaQr" target="_blank">https://codepen.io/lukaw3d/pen/ExxyaQr</a>

### Other comments
- `step="any"` is just input's native attribute to allow decimal numbers (Firefox validation)
- `type="number"` limits valid input characters natively
- `v-model.number` converts string into numeric value for v-model

#### Potential solution

https://github.com/vuetifyjs/vuetify/blob/f0ec7dba89da409bb647f1a947159126f9838acb/packages/vuetify/src/components/VTextField/VTextField.ts#L357-L361

These DOM props get converted into strings, and `this.lazyValue = -0` stringifies into `"0"`. It would be fixed if it stringified into `"-0"` or if the old `this.$refs.input.value` was kept

florianmientjes:
## Motivation and Context
Applied the proposed solution in #9368 to return '-0' as a string.

fixes #9368

## How Has This Been Tested?
playground

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-text-field
      v-model.number="input"
      type="number"
      step="any"
      label="Try typing '-0.1'"
    >
    </v-text-field>

    {{ input }}

    <v-text-field
      v-model.number="inputText"
      label="Try typing '-0.1'"
    >
    </v-text-field>

    {{ inputText }}

  </v-container>
</template>

<script>
  export default {
    data: () => ({
      input: null,
      inputText: null,
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -377,7 +377,7 @@ export default baseMixins.extend<options>().extend({
       return this.$createElement('input', {
         style: {},
         domProps: {
-          value: Object.is(this.lazyValue, -0) ? '-0' : this.lazyValue,
+          value: (this.type === 'number' && Object.is(this.lazyValue, -0)) ? '-0' : this.lazyValue,
         },
         attrs: {
           ...this.attrs$,
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -377,7 +377,7 @@ export default baseMixins.extend<options>().extend({
       return this.$createElement('input', {
         style: {},
         domProps: {
-          value: this.lazyValue,
+          value: Object.is(this.lazyValue, -0) ? '-0' : this.lazyValue,
         },
         attrs: {
           ...this.attrs$,
--- a/packages/vuetify/src/components/VTextField/__tests__/VTextField.spec.ts
+++ b/packages/vuetify/src/components/VTextField/__tests__/VTextField.spec.ts
@@ -858,4 +858,13 @@ describe('VTextField.ts', () => { // eslint-disable-line max-statements
 
     expect(wrapper.vm.computedColor).toBe('primary')
   })
+
+  it('should keep -0 in input when type is number', async () => {
+    const wrapper = mountFunction({
+      propsData: { type: 'number', value: -0 },
+    })
+
+    const input = wrapper.find('input')
+    expect(input.element.value).toBe('-0')
+  })
 })


packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -377,7 +377,7 @@ export default baseMixins.extend<options>().extend({
       return this.$createElement('input', {
         style: {},
         domProps: {
-          value: this.lazyValue,
+          value: Object.is(this.lazyValue, -0) ? '-0' : this.lazyValue,
```
johnleider, line 380: I can't find a better way to implement this, but we shouldn't perform this comparison if the `type` isn't `number`.

florianmientjes: I added a type check before the comparison but I am not sure this follows the code style of the project.

This will re-introduce the bug if the VTextField type is "text" and the v-model has the .number modifier (second field in the updated playground).  This might not be an issue because I can't think of a scenario where you would use the number modifier without the number type.

johnleider: You'd be surprised lol

florianmientjes: > You'd be surprised lol

I checked our own projects, and it is indeed filled with them.

florianmientjes:
I tried to find a way to fix it for the `v-model.number` and `type="text"` combination. Since there is no way, as far as I know, to get the modifiers in a custom component I can't make this work.

Still, this fix does work when type="number" and doesn't break current functionality for type="text"

johnleider:
Please add a unit test for this and we'll get it out next week. Thank you 👍

itaross:
upvote for this PR

Issue #11184, opened 2020-04-22
[Feature Request] Allow to set opacity value of v-tooltip


Bastczuak:
### Problem to solve
Right now the opacity of tooltips is set hard to 90% by inline styles when active https://github.com/vuetifyjs/vuetify/blob/b120d01e04587c0e00e8762e94c43268ed2cfe63/packages/vuetify/src/components/VTooltip/VTooltip.ts#L135. But I want to change the opacity so I can have tooltips with 100% opacity for example.

### Proposed solution
Add another property called `opacity` to `v-tooltip` to be able to change the inline style opacity value.

Hyperion2400:
Please add this in the next version.

genygilles:
As a workaround, you can use this in a global (S)CSS file:

```css
.v-tooltip__content.menuable__content__active {
  opacity: 1!important;
}

```

iyush1993:
The easiest way I found was to specify the opacity in a class and assign that class to `content-class` prop of tooltip

```scss
.custom-tooltip {
    opacity: 1!important;
}
```

then in v-tooltip, assign this class:
```vue
<v-tooltip content-class='custom-tooltip'>
</v-tooltip>
```

webcoderkz:
@iyush1993 

in my case background color needed to be modified:

```
.custom-tooltip {
  opacity: 1 !important;
  background: rgba(97, 97, 97, 1);
}
```

codeflorist:
this is a current global solution without needing a custom class:

```css
.v-tooltip .v-overlay__content {
    background: rgba(var(--v-theme-surface-variant), 1) !important;
}
```

NickSomsen:
For me using `:deep()` selector worked (Vuetify `3.1.7`):

```css
:deep(.v-overlay__content) {
  background: rgba(66, 66, 66, 1) !important; /* 66, 66, 66 = --v-theme-surface-variant; 1 for opacity */
}
```

Use a custom class to target only specific tooltips:

```html
<v-tooltip class='custom-tooltip'>
</v-tooltip>
```
```css
.custom-opacity :deep(.v-overlay__content) {
  background: rgba(66, 66, 66, 1) !important;
}
```

PR #11456, opened 2020-05-21
feat(vtooltip): added opactiy prop to VTooltip

Bastczuak:
Like described in this issue https://github.com/vuetifyjs/vuetify/issues/11184, I wanted to be able
to change the opacity of tooltips instead of having a hard coded value.

re #11184


johnleider:
Unfortunately this did not make it in for v2.3 and v3 is being built from the ground up. 

Issue #11548, opened 2020-06-03
[Bug Report] v-menu improper positioning when using left and nudge-width property


hxtpoe:
### Environment
**Vuetify Version:** 2.2.30
**Vue Version:** 2.6.10
**Browsers:** Safari 13.1.1, Google Chrome, Safari, Mozilla Firefox
**OS:** Mac OS 10.15.5

### Steps to reproduce
0. Use :left="true" property to position menu.
1. Click "MENU AS POPOVER"
2. Observe displayed menu position.
3. Close menu
4. Click "MENU AS POPOVER" again
5. Observe again displayed menu position.

### Expected Behavior
Displayed menu should be aligned to the left border of a button. 

### Actual Behavior
Menu is not positioned properly during the first open. Opening menu for the second time gives proper results. Menu is positioned properly. 
Screencast: https://drive.google.com/drive/folders/1b-t9haQoWCyKeSyePJTZmUFkXAa-cD3Y?usp=sharing

### Reproduction Link
<a href="https://codepen.io/hxtpoe/pen/GRoRxma" target="_blank">https://codepen.io/hxtpoe/pen/GRoRxma</a>

hxtpoe:
PR #11549

MajesticPotatoe:
This seems more of a conflict with `nudge-width` then the left property itself

MajesticPotatoe:
related to #5870

KaelWD:
Duplicate of #5870

PR #11549, opened 2020-06-03
align the component towards the left fixed

hxtpoe:
This is a hack/fix for #11548 .

It seems that dimensions are not properly calculated during the first call. 
I wasn't able to found a root cause of that weird behavior, but calling updateDimensions twice solves the problem.

I'm not sure if this can be merged. I hope this is useful and we will find the root cause soon.

### DIFF
--- a/packages/docs/src/examples/menus/intermediate/popover.vue
+++ b/packages/docs/src/examples/menus/intermediate/popover.vue
@@ -5,6 +5,7 @@
       :close-on-content-click="false"
       :nudge-width="200"
       offset-x
+      :left="true"
     >
       <template v-slot:activator="{ on }">
         <v-btn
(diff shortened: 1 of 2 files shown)


packages/docs/src/examples/menus/intermediate/popover.vue
```diff
@@ -5,6 +5,7 @@
       :close-on-content-click="false"
       :nudge-width="200"
       offset-x
+      :left="true"
```
MajesticPotatoe, line 8:
```suggestion
      left
```

johnleider:
This is a potentially hazardous change for v2.2.x and don't think it should be introduced as we head towards our next LTS version.

Thank you for your contribution and interest in improving Vuetify.

PR #11675, opened 2020-06-18
fix(VSelect): ensure VMenu's content has scroll listener added

bart5:
## Description
Ensure that `VSelect` adds scroll listener to `VMenu`'s `content`.

## Motivation and Context
Popping `addEventListener` with `setTimeout` from the stack to the callback queue does not guarantee in any way that target element will exist the next time that we try to add the listener.
Proposed solution does guarantee that listener is eventually added.

## How Has This Been Tested?
This was tested in the particular application which uses Vuetify, where `VMenu` was repeatedly unable to generate it's `content` soon enough for `VSelect` to attach the listener, until the proposed change was introduced.

## Reproduction of the issue and demo of the fix
https://codesandbox.io/s/intelligent-saha-jw7gm?file=/src/components/Combo.vue

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -248,11 +248,15 @@ export default baseMixins.extend<options>().extend({
       this.setSelectedItems()
     },
     menuIsBooted () {
-      window.setTimeout(() => {
-        if (this.getContent() && this.getContent().addEventListener) {
-          this.getContent().addEventListener('scroll', this.onScroll, false)
+      const handler = this.onScroll
+      const menu = this.$refs.menu
+      ;(function addListener () {
+        const content = menu.$refs.content
+        if (content && content.addEventListener) {
+          return content.addEventListener('scroll', handler, false)
         }
-      })
+        menu.$once('hook:updated', addListener)
+      })()
     },
     isMenuActive (val) {
       window.setTimeout(() => this.onMenuActiveChange(val))


johnleider:
Please provide a reproduction or playground that demonstrates the bug that this resolves, thank you!

bart5:
Attached link to reproduction and demo of the proposed patch.
https://codesandbox.io/s/intelligent-saha-jw7gm?file=/src/components/Combo.vue

KaelWD:
Simplified reproduction:
```vue
<template>
  <v-container>
    <v-alert>Loaded {{ this.options.length }} options</v-alert>

    <v-combobox ref="combobox" :items="options">
      <template #append-item>
        <div v-intersect="onNextPageIntersection"/>
        <v-progress-linear :active="loading" indeterminate class="fixed-bottom"></v-progress-linear>
      </template>
    </v-combobox>
    <div class>1. Try opening the menu without loading options first.
      <br>2. Load options, and try scrolling the menu.
      <br>3. You can see that we cannot see past first 20 items.
      <br>
      <br>If you inspect `v-menu__content` you will notice that it has no `scroll` listener.
    </div>
    <v-btn @click="loadOptions">Load Options</v-btn>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      loading: false,
      options: [],
    }),
    methods: {
      onNextPageIntersection(entries, observer, isIntersecting) {
        if (isIntersecting && !this.loading) {
          this.loadOptions()
        }
      },
      loadOptions() {
        this.loading = true
        setTimeout(() => {
          this.options.push(
            ...Array(20)
              .fill(0)
              .map(() => Math.round(Math.random() * 100000))
          )
          this.loading = false
        }, 1000)
      },
    },
  }
</script>

<style>
  .fixed-bottom {
    position: sticky !important;
    bottom: 8px;
    transform: translateY(8px);
  }

  .v-select-list {
    position: relative !important;
  }
</style>
```

> `VMenu` was repeatedly unable to generate it's content soon enough for `VSelect` to attach the listener

This patch doesn't actually fix the underlying issue outside this one specific scenario, it's more that `VSelect` sets `menuIsBooted = true` on focus even though it hasn't actually opened the menu because there's nothing to display yet. You also still have to scroll back up slightly to actually display newly loaded items because `lastItem` won't go above `items.length`. 

I've fixed both of these problems in d9bb128a5, thanks for your contribution though.

bart5:
@KaelWD Yes, you are right, I'm not fixing the crux of the issue, but neither does your commit. Solution you introduced is in fact making the matter worse since you try to add the listener in the microtask.
I just pulled it and tested just in case, and it does not work.

The essence of the problem is the fact that activation of the menu is at all correlated with attaching the event listener - this is very unreliable 'proxy': "Is menu is active? Yup. Then add the listener.". (Remember that `isMenuActive` we talk about has nothing to do with the actual element - it's just a flag). Instead, scroll listener should be created whenever `content` is actually ready (this is the actual element) - that's all.
Because the problem is deeper I wanted to propose something not intrusive, yet effective.

If you think that is not the correct approach, and have more streamlined solution then by all means propose it.

Otherwise please reconsider reopening this PR. 

"You also still have to scroll back up slightly to actually display newly loaded items because lastItem won't go above items.length." - I'm aware but that has nothing to do with the main problem of listener not being added.

bart5:
@KaelWD Yes, you are right, I'm not fixing the crux of the issue, but neither does your commit. Solution you introduced is in fact making the matter worse since you try to add the listener in the microtask.
I just pulled it and tested just in case, and it does not work.

The essence of the problem is the fact that activation of the menu is at all correlated with attaching the event listener - this is very unreliable 'proxy': "Is menu is active? Yup. Then add the listener.". (Remember that `isMenuActive` we talk about has nothing to do with the actual element - it's just a flag). Instead, scroll listener should be created whenever `content` is actually ready (this is the actual element) - that's all.
Because the problem is deeper I wanted to propose something not intrusive, yet effective.

If you think that is not the correct approach, and have more streamlined solution then by all means propose it.

Otherwise please reconsider reopening this PR. 

"You also still have to scroll back up slightly to actually display newly loaded items because lastItem won't go above items.length." - I'm aware but that has nothing to do with the main problem of listener not being added.

KaelWD:
> it does not work

What did you test it with? Using that code I posted in `packages/vuetify/dev/Playground.vue` and following your directions seems to work for me now. When you click on the select for the first time without any items, `$refs.menu.$refs.content` is undefined and `$refs.menu.isBooted` is false. In the nextTick *with* items they're defined and true. 

> The essence of the problem is the fact that activation of the menu is at all correlated with attaching the event listener

Yeah ideally VMenu would just pass events through and we wouldn't need any of this.

KaelWD:
```diff
diff --git a/packages/vuetify/src/components/VMenu/VMenu.ts b/packages/vuetify/src/components/VMenu/VMenu.ts
index 83765e2ca..8e3466ee8 100644
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -198,6 +198,10 @@ export default baseMixins.extend({
     this.isActive && this.callActivate()
   },
 
+  updated () {
+    console.log('menu updated')
+  },
+
   methods: {
     activate () {
       // Update coordinates and dimensions of menu
diff --git a/packages/vuetify/src/components/VSelect/VSelect.ts b/packages/vuetify/src/components/VSelect/VSelect.ts
index f120f0902..6038fcf67 100644
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -252,8 +252,12 @@ export default baseMixins.extend<options>().extend({
 
       if (!val) return
 
+      console.log('menu active', this.getContent())
+
       this.$nextTick(() => {
+        console.log('nextTick', this.getContent())
         if (!this.menuIsBooted && this.getContent() && this.$refs.menu.isBooted) {
+          console.log('listener added')
           this.getContent().addEventListener('scroll', this.onScroll, false)
           this.menuIsBooted = true
         }
```

image

bart5:
I tested your changes directly in the application I'm having problem with.
 
My reproduction was missing one important case, and the one that is actually most natural - when options are load just after we first interact with the menu. Fixed the playground:
```
<template>
  <v-container>
    <v-alert>Loaded {{ this.options.length }} options</v-alert>

    <v-combobox ref="combobox" :items="options" @click="firstLoad && loadOptions()">
      <template #append-item>
        <div v-intersect="onNextPageIntersection"/>
        <v-progress-linear :active="loading" indeterminate class="fixed-bottom"></v-progress-linear>
      </template>
    </v-combobox>
    <div class>1. Try opening the menu without loading options first.
      <br>2. Load options, and try scrolling the menu.
      <br>3. You can see that we cannot see past first 20 items.
      <br>
      <br>If you inspect `v-menu__content` you will notice that it has no `scroll` listener.
    </div>
    <v-btn @click="loadOptions">Load Options</v-btn>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      loading: false,
      options: [],
      firstLoad: true,
    }),
    methods: {
      onNextPageIntersection(entries, observer, isIntersecting) {
        if (isIntersecting && !this.loading) {
          this.loadOptions()
        }
      },
      loadOptions() {
        if (this.firstLoad) this.firstLoad = false
        this.loading = true
        setTimeout(() => {
          this.options.push(
            ...Array(20)
              .fill(0)
              .map(() => Math.round(Math.random() * 100000))
          )
          this.loading = false
        }, 1000)
      },
    },
  }
</script>

<style>
  .fixed-bottom {
    position: sticky !important;
    bottom: 8px;
    transform: translateY(8px);
  }

  .v-select-list {
    position: relative !important;
  }
</style>
```
Sorry for the confusion.

bart5:
In case you'd like to reconsider I rebased, cleaned up this a bit and opened a new PR.
https://github.com/vuetifyjs/vuetify/pull/11725

KaelWD:
Ah right I missed that use case. I'm still not entirely sold on repeatedly listening to the updated event like that though, it feels kinda messy.

Issue #4867, opened 2018-08-15
[Feature Request] Server side pagination for v-autocomplete/v-select (loading more items on scroll)


ricardovanlaarhoven:
V-select and V-autocomplete load more items on scroll, this only loads more items into the HTML, but the `items` prop has to be filled already. It would be nice to be able to load this from the server.

### What problem does this feature solve?
Some API's endpoints are paginated, for now when i have an autocomplete i add a parameter like perPage=10000 to API request parameters so everything works, except this is really slow. 

### What is your proposed solution?
Add a prop which indicates if you want to use server side pagination, when the `onScroll` method from `v-select` at this point it raises the `this.lastItem`, it should fire an event which then can load x more items from the server.

ricardovanlaarhoven:
When i had this issue i created a PR which fixed my issue.
Some time past since then.
My PR was closed "Unfortunately this is not functionality that we are looking to implement at this time."
And i have had some contact with @johnleider

He told me that there are ways to do this with the current functionality. And i think this is the right place to show the people who are struggling with this can find the solution.

Could anyone who knows the solution post it here?
At this point i request 10.000's of items from the api whithout pagination as a dirty bugfix

johnleider:
This is what I was referring to: https://vuetifyjs.com/en/components/selects#example-slots

massimolavermicocca:
Hi, any news on this subject yet? Is there a workaround to implement it using current vuetify release?

kanteankit:
@massimolavermicocca 
You can use the `append` slot to add a button/menu-item to the end of the dropdown, that says `Load More`. When user clicks on it, you can make an API call to fetch more data and push it to the list

massimolavermicocca:
@kanteankit Could you please provide a sandbox with that? the `append` slot seems to draw object next to the autocomplete itself.
Example: [https://codepen.io/anon/pen/omZNgo?editors=1010](url)

KaelWD:
He means `append-item`

tyteen4a03:
Bumping this - are there any workarounds right now?

NothingWeAre:
This solution somewat buggy.
When added items count reach multipliers of 20 new items does not show, until user tries to scroll up.
This is related to the internal value of `lastItem`

ricardovanlaarhoven:
@tyteen4a03 It is possible with the `append-item` mentioned by @KaelWD but i prefer a more build-in solution by vuetify like the PR i've made.
[...]

Issue #2660, opened 2017-11-28
[Bug Report] V-Select scrolling behaviour


asiFarran:
### Versions and Environment
**Vuetify:** 0.17.3
**Vue:** 2.5.9
**Browsers:** Chrome 62.0.3202.94
**OS:** Windows 7

### Steps to reproduce
Give the select component a largish number of items (anything above 50 will do).

Open the select and start scrolling

### Expected Behavior
The scrollbar thumb should be of constant height (representing the relationship between the overall height of the item list and the height of the popup element), and scrolling should be fluid and linear

### Actual Behavior
The scrollbar's thumb starts off too high and as you begin to scroll, changes it's height and jumps to a different location on its track - basically disappearing from under the mouse cursor - thus leading to a jittery and uneven scrolling action making it almost impossible to reach the bottom of the list

### Reproduction Link
[https://codepen.io/anon/pen/RjYmJb?editors=1010](https://codepen.io/anon/pen/RjYmJb?editors=1010)

KaelWD:
v-select lazy-loads items 20 at a time for performance reasons. This could be done with a set number of items at some point in the future though.

asiFarran:
I understand the perf concern but shouldn't the UX concern take precedence here? 

The current strategy makes the component practically unusable when in effect (i.e over 20 items) so what's the point of having nice perf in a context that no-one would ever use unless they like to hear their users complain.

cars10:
Yeah, that behaviour is really annoying and not at all user friendly. Any updates on that?

KaelWD:
It was removed in 1.1 bit had to be added back because of the massive performance impact of you have a lot of records. We're planning to add proper list virtualisation at some point, but probably not in 1.1

cars10:
How about changing it into a prop? In my current project for example I have a fixed amount of about 150 records, would be cool if I could set the value myself.

cars10:
Great, thank you! Using `auto` will indeed render the whole list, guess i missed that in the documentation. Sadly this seems to break `max-height` (now having no effect), but i will use it anyway. Thanks!

Webifi:
I think this could be solvable if we could easily change the default value of 20 for VSelect.lastItem to something else via a prop. 

Could it be as simple as just adding something like this to VSelect.ts?:

```
props: {
  ...
  defaultLastItem: Number,
  ...
},
mount () {
  ...
[...]

PR #11725, opened 2020-06-25
fix(VSelect): ensure VMenu's content has scroll listener added

bart5:
## Description
Ensure that `VSelect` adds scroll listener to `VMenu`'s `content`.

## Motivation and Context
Arbitrary one-time delay of `this.getContent().addEventListener('scroll', this.onScroll, false)` does not guarantee that target element will exist the next time that we try to add the listener.
Proposed solution does guarantee that listener is eventually added.

## How Has This Been Tested?
In the playground.

## Reproduction

```vue
<template>
  <v-container>
    <v-alert>Loaded {{ this.options.length }} options</v-alert>

    <v-combobox ref="combobox" :items="options" @click="firstLoad && loadOptions()">
      <template #append-item>
        <div v-intersect="onNextPageIntersection"/>
        <v-progress-linear :active="loading" indeterminate class="fixed-bottom"></v-progress-linear>
      </template>
    </v-combobox>
    <div class>1. See that no options are loaded. Open the menu.
      <br>2. You notice that some options loaded. Scroll down.
      <br>3. You can see that new batch of options loads yet we cannot see past the first 20.
      <br>
      <br>If you inspect `v-menu__content` you will notice that it has no `scroll` listener.
    </div>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      loading: false,
      options: [],
      firstLoad: true,
    }),
    methods: {
      onNextPageIntersection(entries, observer, isIntersecting) {
        if (isIntersecting && !this.loading) {
          this.loadOptions()
        }
      },
      loadOptions() {
        if (this.firstLoad) this.firstLoad = false
        this.loading = true
        setTimeout(() => {
          this.options.push(
            ...Array(20)
              .fill(0)
              .map(() => Math.round(Math.random() * 100000))
          )
          this.loading = false
        }, 1000)
      },
    },
  }
</script>

<style>
  .fixed-bottom {
    position: sticky !important;
    bottom: 8px;
    transform: translateY(8px);
  }

  .v-select-list {
    position: relative !important;
  }
</style>
```

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -100,6 +100,7 @@ export default baseMixins.extend({
       resizeTimeout: 0,
       selectedIndex: null as null | number,
       tiles: [] as HTMLElement[],
+      onContentReadyOnce: () => {},
     }
   },
 
@@ -199,6 +200,12 @@ export default baseMixins.extend({
   },
 
   methods: {
+    setOnContentReadyOnce<T extends () => void>(onReady: T) {
+      this.onContentReadyOnce = () => {
+        this.$nextTick(onReady)
+        this.onContentReadyOnce = () => {}
+      }
+    },
     activate () {
       // Update coordinates and dimensions of menu
       // and its activator
@@ -295,6 +302,7 @@ export default baseMixins.extend({
     },
     genTransition (): VNode {
       const content = this.genContent()
+      this.onContentReadyOnce()
 
       if (!this.transition) return content
 
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -100,7 +100,6 @@ export default baseMixins.extend({
       resizeTimeout: 0,
       selectedIndex: null as null | number,
       tiles: [] as HTMLElement[],
-      onContentReadyOnce: () => {},
     }
   },
 
@@ -200,12 +199,6 @@ export default baseMixins.extend({
   },
 
   methods: {
-    setOnContentReadyOnce<T extends () => void>(onReady: T) {
-      this.onContentReadyOnce = () => {
-        this.$nextTick(onReady)
-        this.onContentReadyOnce = () => {}
-      }
-    },
     activate () {
       // Update coordinates and dimensions of menu
       // and its activator
@@ -302,7 +295,6 @@ export default baseMixins.extend({
     },
     genTransition (): VNode {
       const content = this.genContent()
-      this.onContentReadyOnce()
 
       if (!this.transition) return content
 
@@ -360,6 +352,11 @@ export default baseMixins.extend({
           keydown: this.onKeyDown,
         },
       } as VNodeData
+
+      if (this.$listeners.scroll) {
+        options.on = options.on || {}
+        options.on.scroll = this.$listeners.scroll
+      }
 
       if (!this.disabled && this.openOnHover) {
         options.on = options.on || {}
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -254,13 +254,20 @@ export default baseMixins.extend<options>().extend({
         this.menuIsBooted = true
         const handler = this.onScroll
         const menu = this.$refs.menu
-        ;(function addListener () {
+        const addListener = () => {
           const content = menu.$refs.content
           if (content && content.addEventListener) {
-            return content.addEventListener('scroll', handler, false)
+            content.addEventListener('scroll', handler, false)
           }
-          menu.$once('hook:updated', addListener)
-        })()
+        }
+        this.$refs.menu.setOnContentReadyOnce(addListener)
+        // ;(function addListener () {
+        //   const content = menu.$refs.content
+        //   if (content && content.addEventListener) {
+        //     return content.addEventListener('scroll', handler, false)
+        //   }
+        //   menu.$once('hook:updated', addListener)
+        // })()
       }
     },
     items: {
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -249,26 +249,6 @@ export default baseMixins.extend<options>().extend({
     },
     isMenuActive (val) {
       window.setTimeout(() => this.onMenuActiveChange(val))
-
-      if (!this.menuIsBooted) {
-        this.menuIsBooted = true
-        const handler = this.onScroll
-        const menu = this.$refs.menu
-        const addListener = () => {
-          const content = menu.$refs.content
-          if (content && content.addEventListener) {
-            content.addEventListener('scroll', handler, false)
-          }
-        }
-        this.$refs.menu.setOnContentReadyOnce(addListener)
-        // ;(function addListener () {
-        //   const content = menu.$refs.content
-        //   if (content && content.addEventListener) {
-        //     return content.addEventListener('scroll', handler, false)
-        //   }
-        //   menu.$once('hook:updated', addListener)
-        // })()
-      }
     },
     items: {
       immediate: true,
@@ -529,6 +509,7 @@ export default baseMixins.extend<options>().extend({
             this.isMenuActive = val
             this.isFocused = val
           },
+          scroll: this.onScroll,
         },
         ref: 'menu',
       }, [this.genList()])
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -250,14 +250,19 @@ export default baseMixins.extend<options>().extend({
     isMenuActive (val) {
       window.setTimeout(() => this.onMenuActiveChange(val))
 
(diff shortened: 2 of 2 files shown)


KaelWD:
Duplicate of #11675

KaelWD:
Aaand github won't let me reopen the original because it was force-pushed.

bart5:
"I'm still not entirely sold on repeatedly listening to the updated event like that though, it feels kinda messy."
I agree. It is a trade-off. We pay several invocations of addListener for the certainty of it being effective eventually. In my tests I saw the update running 3 or 4 times before it could actually attach the listener. 

However, it seems to be the quickest and easiest solution, and frankly, pretty reliable.

I pushed another proposition with more direct approach.

KaelWD:
What about just having VMenu emit the scroll event itself?

```diff
diff --git a/packages/vuetify/src/components/VMenu/VMenu.ts b/packages/vuetify/src/components/VMenu/VMenu.ts
index 83765e2ca..958d29aea 100644
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -350,6 +350,7 @@ export default baseMixins.extend({
             if (this.closeOnContentClick) this.isActive = false
           },
           keydown: this.onKeyDown,
+          scroll: this.$listeners.scroll,
         },
       } as VNodeData
 
diff --git a/packages/vuetify/src/components/VSelect/VSelect.ts b/packages/vuetify/src/components/VSelect/VSelect.ts
index f120f0902..b55a90ae0 100644
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -112,7 +112,6 @@ export default baseMixins.extend<options>().extend({
   data () {
     return {
       cachedItems: this.cacheItems ? this.items : [],
-      menuIsBooted: false,
       isMenuActive: false,
       lastItem: 20,
       // As long as a value is defined, show it
@@ -249,15 +248,6 @@ export default baseMixins.extend<options>().extend({
     },
     isMenuActive (val) {
       window.setTimeout(() => this.onMenuActiveChange(val))
-
-      if (!val) return
-
-      this.$nextTick(() => {
-        if (!this.menuIsBooted && this.getContent() && this.$refs.menu.isBooted) {
-          this.getContent().addEventListener('scroll', this.onScroll, false)
-          this.menuIsBooted = true
-        }
-      })
     },
     items: {
       immediate: true,
@@ -518,6 +508,7 @@ export default baseMixins.extend<options>().extend({
             this.isMenuActive = val
             this.isFocused = val
           },
+          scroll: this.onScroll,
         },
         ref: 'menu',
       }, [this.genList()])
```

bart5:
I think it is by far the most sane solution.

packages/vuetify/src/components/VMenu/VMenu.ts
```diff
@@ -350,6 +350,9 @@ export default baseMixins.extend({
             if (this.closeOnContentClick) this.isActive = false
           },
           keydown: this.onKeyDown,
+          ...this.$listeners.scroll
+            ? { scroll: this.$listeners.scroll }
```
KaelWD, line 354: Does `undefined` do something bad?

bart5: Vue throws because of it and there are some `TypeErrors` later along the way.

bart5: We could `scroll: this.$listeners.scroll || false` (yes, false works :man_shrugging:) or `scroll: this.$listeners.scroll || (() => {})` but it leaves us with listener that does nothing.

bart5: Alternatively:
```js
      if (this.$listeners.scroll) {
        options.on = options.on || {}
        options.on.scroll = this.$listeners.scroll
      }
```

bart5:
We can also get rid of `scroll` listener completely:
```diff
diff --git a/packages/vuetify/src/components/VMenu/VMenu.ts b/packages/vuetify/src/components/VMenu/VMenu.ts
index a85202633..97b166f56 100644
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -174,6 +174,10 @@ export default baseMixins.extend({
     },
     isContentActive (val) {
       this.hasJustFocused = val
+
+      if (!val) {
+        requestAnimationFrame(() => (this.$refs.content.scrollTop = 0))
+      }
     },
     listIndex (next, prev) {
       if (next in this.tiles) {
@@ -350,9 +354,6 @@ export default baseMixins.extend({
             if (this.closeOnContentClick) this.isActive = false
           },
           keydown: this.onKeyDown,
-          ...this.$listeners.scroll
-            ? { scroll: this.$listeners.scroll }
-            : {},
         },
       } as VNodeData
 
diff --git a/packages/vuetify/src/components/VSelect/VSelect.ts b/packages/vuetify/src/components/VSelect/VSelect.ts
index 5f84be822..818979ddd 100644
--- a/packages/vuetify/src/components/VSelect/VSelect.ts
+++ b/packages/vuetify/src/components/VSelect/VSelect.ts
@@ -217,7 +217,7 @@ export default baseMixins.extend<options>().extend({
     virtualizedItems (): object[] {
       return (this.$_menuProps as any).auto
         ? this.computedItems
         ? this.computedItems
-        : this.computedItems.slice(0, this.lastItem)
+        : this.computedItems.slice(0, this.getLastItem())
     },
     menuCanShow: () => true,
     $_menuProps (): object {
@@ -509,7 +509,6 @@ export default baseMixins.extend<options>().extend({
             this.isMenuActive = val
             this.isFocused = val
           },
-          scroll: this.onScroll,
         },
         ref: 'menu',
       }, [this.genList()])
@@ -625,7 +624,7 @@ export default baseMixins.extend<options>().extend({
       })
       const item = this.allItems[index]
       if (index !== -1) {
-        this.lastItem = Math.max(this.lastItem, index + 5)
+        this.lastItem = Math.max(this.getLastItem(), index + 5)
         this.setValue(this.returnObject ? item : this.getValue(item))
         this.$nextTick(() => this.$refs.menu.getTiles())
         setTimeout(() => this.setMenuIndex(index))
@@ -715,21 +714,19 @@ export default baseMixins.extend<options>().extend({
 
       VTextField.options.methods.onMouseUp.call(this, e)
     },
-    onScroll () {
-      if (!this.isMenuActive) {
-        requestAnimationFrame(() => (this.getContent().scrollTop = 0))
+    getLastItem (): number {
+      if (!this.isMenuActive || !this.getContent() || this.lastItem > this.computedItems.length) {
+        return this.lastItem
       } else {
-        if (this.lastItem > this.computedItems.length) return
-
         const showMoreItems = (
           this.getContent().scrollHeight -
           (this.getContent().scrollTop +
           this.getContent().clientHeight)
         ) < 200
-
         if (showMoreItems) {
           this.lastItem += 20
         }
+        return this.lastItem
       }
     },
     onSpaceDown (e: KeyboardEvent) {
```

KaelWD:
How would it load more items when you scroll then?

bart5:
As you can see above we can put `updateLastItem` into `items` watcher.
```diff
@@ -253,15 +253,15 @@ export default baseMixins.extend<options>().extend({
     items: {
       immediate: true,
       handler (val) {
...
```

bart5:
Because Vuetify itself is not loading items it's enough to put `updateLastItem` into `items` watcher handler.
If items change `virtualizedItems`  will be changed accordingly.

Also, now thinking about it, we should add to that `updateLastItem` cap on `lastItem` in case the number of item decreased (e.g. new, narrower search). Otherwise if I gradually load e.g. 500 items scrolling down, then close the menu, run search, get some 400 results, we will try to render all those at once.

bart5:
Actually, I think we should completely reset `lastItem` to default once menu is closed. Why keep all those nodes in the DOM?

bart5:
I now realised that it doesn't work for lists loaded as a single batch.
I think it should be solved by delegating decision to the user whether their items list is paginated or not.

KaelWD:
There's a feature request for proper pagination (#4867), anything we do here is still just a hacky workaround. 

> Why keep all those nodes in the DOM?

I think we've considered making v-menu remove itself from the DOM when closed.

> on reopening the menu, `lastItem` should be set back to `computedItems` length

You'd have to store the last value of `lastItem`, otherwise I could open the menu, see 20 items, close and reopen it and now I've got all 1000. 

We should just focus on the relevant bug here, `lastItem` needs to be replaced with the new virtual scroller anyway (#2660)

packages/vuetify/src/components/VMenu/VMenu.ts
KaelWD: Yeah I like that better, and we're already doing it for the mouse handlers.

packages/vuetify/src/components/VMenu/VMenu.ts
```diff
@@ -349,10 +349,14 @@ export default baseMixins.extend({
             if (target.getAttribute('disabled')) return
             if (this.closeOnContentClick) this.isActive = false
           },
-          keydown: this.onKeyDown,
```
KaelWD, line 352: ?

PR #11741, opened 2020-06-27
docs(v-select): add items prop object the disabled option

EdenK:
Add items prop object the disabled option in the description in english and in the example also,
mention the item-disabled prop in the description as its relevant

### DIFF
--- a/packages/docs/src/lang/en/components/Selects.json
+++ b/packages/docs/src/lang/en/components/Selects.json
@@ -67,7 +67,7 @@
     "hideSelected": "Do not display in the select menu items that are already selected",
     "itemColor": "Sets color of selected items",
     "itemDisabled": "Set property of **items**'s disabled value",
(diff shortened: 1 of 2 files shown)


packages/docs/src/lang/en/components/Selects.json
```diff
@@ -67,7 +67,7 @@
     "hideSelected": "Do not display in the select menu items that are already selected",
     "itemColor": "Sets color of selected items",
     "itemDisabled": "Set property of **items**'s disabled value",
-    "items": "Can be an array of objects or array of strings. When using objects, will look for a text and value field. This can be changed using the **item-text** and **item-value** props.  Objects that have a **header** or **divider** property are considered special cases and generate a list header or divider; these items are not selectable.",
+    "items": "Can be an array of objects or array of strings. When using objects, will look for a text, value and disabled keys. This can be changed using the **item-text**, **item-value** and **item-disabled** props.  Objects that have a **header** or **divider** property are considered special cases and generate a list header or divider; these items are not selectable.",
```
KaelWD, line 70: Items can have `header` or `divider` too (both `boolean`)

EdenK: I added to the example the `divider = boolean` and the `header = string` options

Issue #7392, opened 2019-06-04
[Feature Request] Hide Text Input Counter when field does not have focus


blakenorman37:
### Problem to solve
Removing the clutter of the screen when the page has many text fields. The counter does not add value to the user when they are not using the text field.

### Proposed solution
Add the ability to auto-hide the counter when the text field does not have focus, similar to how the hint text currently works.

johnleider:
in v3 the counter is now hidden by default and can be enabled by using **persistent-counter**.

ashunimh:
how can i hide counter from text field but show hint?

PR #11838, opened 2020-07-11
feat(VTextField): hide counter when input is not focused

digital88:
## Description
As per #7392 feature request, added a 'counter-show-on-focus' prop to v-text-field that, when set, modifies behaviour of counter display - counter becomes visible only when input has focus. Also added test for new feature.

## Motivation and Context
resolves https://github.com/vuetifyjs/vuetify/issues/7392

## How Has This Been Tested?
Updated existing unit tests for v-text-field and added new test for this feature.
Test imitates input focus and blur events and looks for el with class '.v-counter' to confirm it exists or not exists.

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-text-field
      value="text"
      :counter="true"
      counter-show-on-focus
      :error="true"
      :error-messages="'error!'"
    />
  </v-container>
</template>

<script>
  export default {
    data: () => ({
    //
    }),
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -66,6 +66,7 @@ export default baseMixins.extend<options>().extend({
     },
     counter: [Boolean, Number, String],
     counterValue: Function as PropType<(value: any) => number>,
+    counterShowOnFocus: Boolean,
     filled: Boolean,
     flat: Boolean,
     fullWidth: Boolean,
@@ -308,6 +309,10 @@ export default baseMixins.extend<options>().extend({
     },
     genCounter () {
       if (!this.hasCounter) return null
+
+      if (this.counterShowOnFocus && !this.isFocused) {
+        return null
+      }
 
       const max = this.counter === true ? this.attrs$.maxlength : this.counter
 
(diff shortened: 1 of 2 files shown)


johnleider:
My concern with this is the different implementation for hint. By default, hint only shows on focus but has a **persistent-hint** property that shows it all the time. By default, counter shows but this proposal would add a **counter-show-on-focus** property.

Ideally counter would work exactly like hint and have a **persistent-counter** property. However, this would be a breaking visual change and would have to go into v3. If that's the case, is there value in implementing this for v2 when the approach for the next version may differ?

Any insight from the @vuetifyjs/core-team ?

KaelWD:
> However, this would be a breaking visual change and would have to go into v3

Eh, we removed placeholders in v2.4 because it made more sense.

johnleider:
In light of that, should this be pushed to v3?

johnleider:
Personally I like the idea of having hint and counter operate in the same manner. I believe that the Default behavior should be hidden for both, with **persistent-hint**, **persistent-counter**, and maybe even **persistent-details**, which would counter nicely to the **hide-details** prop.

johnleider:
This has been implemented in v3 under the prop **persistent-counter**. By default, the counter is now hidden.

PR #11900, opened 2020-07-21
fix(helpers): searchItems() type sig

morphatic:
## Description

TypeScript complains and fails (on my system) with `error TS2769: No overload matches this call` when running `yarn build lib`. The search algorithm in [the `searchItems()` function](https://github.com/vuetifyjs/vuetify/blob/dev/packages/vuetify/src/util/helpers.ts#L338) uses `Object.keys(item)` to filter the list of items, which will _only_ succeed with objects. Since the function is designed to filter objects, it seems reasonable that the type signature reflects this.

Current signature: 

```ts
export function searchItems<T extends any = any> (items: T[], search: string): T[] {}
```

Suggested signature: 

```ts
export function searchItems<T extends object = object> (items: T[], search: string): T[] {}
```

## Motivation and Context

This type signature was introduced in b03b507 (2019-12-02). It affects `es5/lib` builds for packages that extend Vuetify.

## How Has This Been Tested?

Ran `yarn build lib`. There was no error and libs were built as expected.

## Types of changes

## Checklist:
- [x] I've added relevant changes to the documentation (applies to new features and breaking changes in core library) N/A


jacekkarczmarczyk (commented):
- `ban-types` rule suggests using `Record<string, unknown>` instead of `object` - https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options
- `Object.keys()` works with arrays as well, can the table item be an array (assuming `value` prop in `headers` are numeric)?
- `DataSearchFunction` definition could be updated as well

morphatic:
@jacekkarczmarczyk like this?

```ts
export function searchItems<T extends Record<string, unknown> = Record<string, unknown>> (items: T[], search: string): T[] {}
```

My typescript fu is not all that strong. I don't fully understand the syntax around generics. I think I grok `<T extends Record<string, unknown>>` but I'm not clear about how/why the assignment gets done inside of it, i.e. `<T extends X = X>`. Does that just assign a default type to the generic?

FWIW, when I use this signature and run `yarn build:lib` the typescript compiler complains about a lot of missing/nonexistent types, but the lib appears to get built. Not exactly sure how to test more thoroughly than that.

Let me know if I should update the PR with your suggested syntax.

jacekkarczmarczyk:
> FWIW, when I use this signature and run yarn build:lib the typescript compiler complains about a lot of missing/nonexistent types

can't reproduce it, builds fine for me both with your and mine signature, can you show the error?

> but I'm not clear about how/why the assignment gets done inside of it

if lets you skip setting the generic (not sure what's the proper expression for this), so if you want to apply generic it needs to extend `Record<string, unknown>` but if you skip it `searchExpression` will set the `T` to `Record<string, unknown>` by default

KaelWD:
Your env must be cooked, `any` is working fine for everyone else.

KaelWD:
Any updates on this or should I close it?

PR #11921, opened 2020-07-23
feat(VDataTable): Add shift key logic for multiple selection

### LINKED ISSUE #9958: [Feature Request] Selecting multiple items on VDataTable with Shift-key
### Problem to solve
When using "show-select"s in a VDataTable there is currently no native way of quickly selecting multiple, neighboring items. 

e.g. If you want to select the first 15 items in a table, you have to individually click on each checkbox.

### Proposed solution
Listen for "Shift"-clicks. 

Once you have marked an item, as soon as you mark another item (while holding the shift key) all the items in between will get selected as well.

For mobile devices instead of listening to shift click, you could listen to double clicks on the second item.

### Code example
Here is some code, it's not perfect, but should illustrate what would be nice to have by default in the data table component.

```
// select everything between last 2 selected items

shiftClick(item) {
  if (this.selectedItems.length >= 2) {
    const secondToLastItemId = this.selectedItems[this.selectedItems.length - 2].id;
    const secondToLastItemIndex = this.items
      .map(rowData => rowData.id)
      .indexOf(secondToLastItemId);
    const lastItemIndex = this.items
      .map(rowData => rowData.id)
      .indexOf(item.id);
    let startIndex = null;
    let endIndex = null;

    if (lastItemIndex > secondToLastItemIndex) {
      startIndex = secondToLastItemIndex + 1;
      endIndex = lastItemIndex - 1;
    } else {
      startIndex = lastItemIndex + 1;
      endIndex = secondToLastItemIndex - 1;
    }
    for (let i = startIndex; i <= endIndex; i += 1) {
      if (
        !this.selectedItems.some(
          element => element.id === this.items[i].id
        )
      ) {
        this.selectedItems.push(this.items[i]);
      }
    }
  }
}
```

``` 
doubleClick(item) {
  if (this.selectedItems.length !== 0) {
    this.selectedItems.push(item);
    this.shiftClick(item);
  }
}
```

Aramis13:
## Description
Add a listener for shift key press.
Add logic for selecting multiple check-boxes.

## Motivation and Context
resolves #9958.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] The PR title is no longer than 64 characters.
- [ x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [ x] My code follows the code style of this project.


packages/vuetify/src/components/VDataIterator/VDataIterator.ts
```diff
@@ -68,6 +68,7 @@ export default mixins(
     selection: {} as Record<string, any>,
     expansion: {} as Record<string, boolean>,
     internalCurrentItems: [] as any[],
+    shiftKeyDown: false as boolean,
```
KaelWD, line 71:
```suggestion
    shiftKeyDown: false,
```

packages/vuetify/src/components/VDataIterator/VDataIterator.ts
```diff
@@ -148,7 +149,24 @@ export default mixins(
...
+    window.addEventListener('keyup', this.onKeyUp)
+  },
+  beforeDestroy () {
+    window.removeEventListener('keydown', this.onKeyDown)
+    window.removeEventListener('keyup', this.onKeyUp)
+  },
+
   methods: {
+    onKeyDown (e: KeyboardEvent): void {
+      if (e.keyCode !== 16) return
```
KaelWD, line 163:
```suggestion
      if (e.keyCode !== keyCodes.shift) return
```

KaelWD (changes requested):

nekosaur (changes requested):
There are a few behaviors here that I feel are not standard when it comes to shift selecting.

1. Select last entry without holding shift. Then select first entry while holding shift. My expectation here is that it would select from last entry to first entry, but that is not the case.

2. Select first entry without holding shift. Then select last entry while holding shift. The entire page is select. Then while still holding shift, select a middle entry. Here I would expect that everything from middle entry to the last entry would be deselected, but the opposite happens where everything from first entry to middle entry is deselected.

packages/vuetify/src/components/VDataIterator/VDataIterator.ts
```diff
@@ -218,6 +211,25 @@ export default mixins(
...
+          emit && this.$emit('item-selected', { currentItem, value })
+        }
+      } else if (index > this.lastEntry) {
+        for (let i = this.lastEntry; i <= index; i++) {
+          const currentItem = this.selectableItems[i]
+          const key = getObjectValueByPath(currentItem, this.itemKey)
+          if (value) selection[key] = currentItem
+          else delete selection[key]
+          emit && this.$emit('item-selected', { currentItem, value })
+        }
```
nekosaur, line 230: Try to refactor this into something more DRY. The only difference between the ifs is the for loop definition

nekosaur (approved):

PR #11970, opened 2020-07-31
fix(helpers): coerce name value to string in groupItems

### LINKED ISSUE #12069: [Bug Report] The group function in DataTable doesn't work when some item's data is null
### Environment
**Vuetify Version:** 2.3.7
**Vue Version:** 2.6.11
**Browsers:** Chrome 79.0.3945.130
**OS:** Windows

### Steps to reproduce
Just modify the default example in the official document.
Change some item's field to undefined or null

Turns out 'undefined' works normal, 
but 'null' field will cause the components crash!

### Expected Behavior
I hope that null will work's too, because some back-end restful api will return null for you.

### Actual Behavior
Crashed

### Reproduction Link
<a href="https://codepen.io/oraant/pen/mdPELgJ" target="_blank">https://codepen.io/oraant/pen/mdPELgJ</a>

### LINKED ISSUE #12501: [Bug Report] Group by in data table does not allow null values for the group-by item property
### Environment
**Vuetify Version:** 2.3.15
**Vue Version:** 3.0.1
**Browsers:** Chrome 86.0.4240.75
**OS:** Windows 10

### Steps to reproduce
Open pen, data table does not display. 

### Expected Behavior
Databable should display, null values of the the item property should not result in error. Also no group should be created for null value

### Actual Behavior
No table is displayed at all.

### Reproduction Link
<a href="https://codepen.io/djr021320/pen/QWEdZpm" target="_blank">https://codepen.io/djr021320/pen/QWEdZpm</a>

### Other comments
Error is raised in groupItems helper function for group for items with null values as no items are added because they have value null.

wol-soft:
## Description
If the property which is used as groupBy contains null a TypeError is thrown (TypeError: Cannot read property 'items' of undefined) in https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/util/helpers.ts#L278 as the variable `current` is initialized with null.

Reproduction: https://codepen.io/WOL-Soft/pen/VweojBr

With this fix all items which are provided with null values or which completely miss the property used for grouping are grouped under a 'empty' group. The null group can be identified via the property `name` of the `ItemGroup` (name will be set to an empty string).

Fixes #12069
Fixes #12501 

## How Has This Been Tested?

* unit tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -268,10 +268,10 @@ export function groupItems<T extends any = any> (
 ): ItemGroup<T>[] {
   const key = groupBy[0]
   const groups: ItemGroup<T>[] = []
-  let current = null
+  let current = undefined
   for (var i = 0; i < items.length; i++) {
     const item = items[i]
-    const val = getObjectValueByPath(item, key)
+    const val = getObjectValueByPath(item, key, null)
     if (current !== val) {
       current = val
       groups.push({
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -268,7 +268,7 @@ export function groupItems<T extends any = any> (
 ): ItemGroup<T>[] {
   const key = groupBy[0]
   const groups: ItemGroup<T>[] = []
-  let current = undefined
+  let current
   for (var i = 0; i < items.length; i++) {
     const item = items[i]
     const val = getObjectValueByPath(item, key, null)
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -275,7 +275,7 @@ export function groupItems<T extends any = any> (
     if (current !== val) {
       current = val
       groups.push({
-        name: val || '',
+        name: val ?? '',
         items: [],
       })
     }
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/util/helpers.ts
```diff
@@ -264,14 +264,14 @@ export function groupItems<T extends any = any> (
...
+  let current
   for (var i = 0; i < items.length; i++) {
     const item = items[i]
-    const val = getObjectValueByPath(item, key)
+    const val = getObjectValueByPath(item, key, null)
     if (current !== val) {
       current = val
       groups.push({
-        name: val,
+        name: val || '',
```
KaelWD, line 274:
```suggestion
        name: val ?? '',
```

wol-soft:
Hi @KaelWD,

is there anything I can do to move this PR forward?

Cheers

wol-soft:
And again the question: can I do anything to move this PR forward? Is the solution of creating an 'empty' group containing all null items not the way to go?

The PR is nearly 4 months old.

Another issue was opened which addresses the TypeError currently thrown.

Fixes #12501 

@johnleider @nekosaur

johnleider (dismissed):

johnleider (approved):

Issue #12018, opened 2020-08-07
[Bug Report] transitions events do not fire on v-expand-transition


matthieusieben:
### Environment
**Vuetify Version:** 2.3.8
**Vue Version:** 2.6.11
**Browsers:** Chrome 84.0.4147.105
**OS:** Mac OS 10.15.6

### Steps to reproduce
Just add transitions listener (`afterEnter`, ...) to a `v-expand-transition` elements

### Expected Behavior
The listener should be triggered

### Actual Behavior
The listeners is not triggered

### Reproduction Link
<a href="https://codepen.io/matthieusieben/pen/xxVbXLY" target="_blank">https://codepen.io/matthieusieben/pen/xxVbXLY</a>

KaelWD:
Two problems here:
 1. `mergeData` is mutating the `functions` object so on each render another handler is added
 2. vue's transition component doesn't account for array event listeners: https://github.com/vuejs/vue/blob/4de4649d/src/platforms/web/runtime/modules/transition.js#L126

matthieusieben:
Actually `mergeListeners` is the culprit. It mutates arrays instead of creating new once.

matthieusieben:
Array listeners seem to be working just fine. They must be compiled into a function somewhere along the way.

KaelWD:
> Array listeners seem to be working just fine.

I get this in your codepen:
![image](https://user-images.githubusercontent.com/16421948/89629983-08cbbd80-d8e2-11ea-99ef-09feec8e33ce.png)

matthieusieben:
Hey @KaelWD any estimate as to when we could see this bugfix live ?

flyingL123:
@matthieusieben @KaelWD I just ran into this same problem. It still doesn't seem to be working properly with 2.3.9. I don't get an error anymore. Instead the callback is just never called:

https://codepen.io/flyingl123/pen/eYZvEjp?editors=1011

Is this a known issue or is this expected to be fixed already?

flyingL123:
@matthieusieben @KaelWD I'm having trouble understanding if this is fixed or not. Can you offer any help? As you can see in the linked codepen in my previous comment it does not appear to work in 2.3.9. I am finding however that the `enter` hook does work. But not `beforeEnter` or `afterEnter`. If you edit the codepen from `@before-enter` to just `@enter`, the callback is fired.

matthieusieben:
[...]

PR #12019, opened 2020-08-07
fix(VExpandTransition): Properly trigger transition listeners

matthieusieben:
## Description

fixes https://github.com/vuetifyjs/vuetify/issues/12018

## Motivation and Context

https://github.com/vuetifyjs/vuetify/issues/12018

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-btn @click="visible = !visible">toggle</v-btn>
    <v-expand-transition @afterEnter="afterEnter">
      <div v-if="visible" style="height: 100px; background: red">
        {{text}}
      </div>
    </v-expand-transition>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      visible: false,
      text: 'afterEnter was NOT triggered'
    }),
    methods: {
      afterEnter(el) {
        this.text = 'afterEnter WAS triggered'
      }
    }
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/util/mergeData.ts
+++ b/packages/vuetify/src/util/mergeData.ts
@@ -148,7 +148,7 @@ export function mergeListeners (
 
       if (dest[event]) {
         // Merge previous and new listeners. Note that dest[event] must not be
-        // altered, event if it was already an array. This is due to the fact
+        // altered, even if it was already an array. This is due to the fact
         // that neither "target" or "source" must be altered.
         dest[event] = [...wrapInArray(dest[event]), ...wrapInArray(listeners[event])]
       } else {
--- a/packages/vuetify/src/util/mergeData.ts
+++ b/packages/vuetify/src/util/mergeData.ts
@@ -150,7 +150,7 @@ export function mergeListeners (
         // Merge previous and new listeners. Note that dest[event] must not be
         // altered, even if it was already an array. This is due to the fact
         // that neither "target" or "source" must be altered.
-        dest[event] = [...wrapInArray(dest[event]), ...wrapInArray(listeners[event])]
+        dest[event] = ([] as Function[]).concat(dest[event], listeners[event])
       } else {
         // Straight assign.
         dest[event] = listeners[event]
--- a/packages/vuetify/src/util/mergeData.ts
+++ b/packages/vuetify/src/util/mergeData.ts
@@ -139,19 +139,24 @@ export function mergeListeners (
   if (!target) return source
   if (!source) return target
 
-  let event: string
+  const dest: { [key: string]: Function | Function[] } = {}
 
-  for (event of Object.keys(source)) {
-    // Concat function to array of functions if callback present.
-    if (target[event]) {
-      // Insert current iteration data in beginning of merged array.
-      target[event] = wrapInArray(target[event])
-      ;(target[event] as Function[]).push(...wrapInArray(source[event]))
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/util/mergeData.ts
```diff
@@ -139,19 +139,24 @@ export function mergeListeners (
...
+  for (const listeners of [target, source]) {
+    for (const event in listeners) {
+      if (!listeners[event]) continue
+      if (Array.isArray(listeners[event]) && listeners[event].length === 0) continue
+
+      if (dest[event]) {
+        // Merge previousa and new listeners. Note that dest[event] must not be
+        // altered, event if it was already an array. This is due to the fact
+        // that neither "target" or "source" must be altered.
+        dest[event] = [...wrapInArray(dest[event]), ...wrapInArray(listeners[event])]
```
KaelWD, line 153: `[].concat(dest[event], listeners[event])` should do the same thing without creating arrays just to spread them again.

packages/vuetify/src/util/mergeData.ts
```diff
@@ -139,19 +139,24 @@ export function mergeListeners (
...
-      target[event] = source[event]
+  const dest: { [key: string]: Function | Function[] } = {}
+
+  for (const listeners of [target, source]) {
+    for (const event in listeners) {
+      if (!listeners[event]) continue
+      if (Array.isArray(listeners[event]) && listeners[event].length === 0) continue
+
+      if (dest[event]) {
+        // Merge previousa and new listeners. Note that dest[event] must not be
```
KaelWD, line 150:
```suggestion
        // Merge previous and new listeners. Note that dest[event] must not be
```

packages/vuetify/src/util/mergeData.ts
```diff
@@ -139,19 +139,24 @@ export function mergeListeners (
...
+        // altered, event if it was already an array. This is due to the fact
```
KaelWD, line 151:
```suggestion
        // altered, even if it was already an array. This is due to the fact
```

KaelWD:
If you feel like adding a test for this too that'd be even better.

packages/vuetify/src/util/mergeData.ts
matthieusieben: right

matthieusieben: I was doing like this because `listeners[event]` was potentially `null`, which is no longer the case with the cheks above.

matthieusieben: I do get a typing error though ...

matthieusieben:
Will this do ?

packages/vuetify/src/util/mergeData.ts
```diff
@@ -141,19 +141,19 @@ export function mergeListeners (
...
   const dest: { [key: string]: Function | Function[] } = {}
 
-  for (const listeners of [target, source]) {
-    for (const event in listeners) {
-      if (!listeners[event]) continue
-      if (Array.isArray(listeners[event]) && listeners[event].length === 0) continue
+  let i = 2
+
+  while (i--) {
+    for (const event in arguments[i]) {
```
KaelWD, line 147: Not a big deal but `arguments` isn't type-safe, it's basically `any[]`

PR #12119, opened 2020-08-25
Update customize-rows.vue 'v-slot' directive doesn't support any modifier

vanarok:
[vue/valid-v-slot]
'v-slot' directive doesn't support any modifier.eslint-plugin-vue

### DIFF
--- a/packages/docs/src/examples/data-tables/intermediate/customize-rows.vue
+++ b/packages/docs/src/examples/data-tables/intermediate/customize-rows.vue
@@ -4,7 +4,7 @@
     :items="desserts"
     class="elevation-1"
   >
-    <template v-slot:item.calories="{ item }">
+    <template v-slot:item[calories]="{ item }">
       <v-chip :color="getColor(item.calories)" dark>{{ item.calories }}</v-chip>
     </template>
   </v-data-table>


vanarok:
> This doesnt work. you would need to do something like
> `` v-slot:[`item.calories`]="{ item } ``
I have no more errors with my change, yours doesn't work for me.
[vue/valid-v-slot]
Dynamic argument of 'v-slot' directive cannot use that slot parameter.eslint-plugin-vue

In general, these errors do not affect the Vue template, either before or after my change, but eslint-plugin-vue insists on this error in Visual Studio Code.

MajesticPotatoe:
I was just providing an example of a solution to the posted linting error. Applying this change to the actual docs example actually doesn't render the slot, even if it removes the lint error.
https://codepen.io/majesticpotatoe-the-bashful/pen/YzqZEvw

The original code should probably stay as is (as it is functional, even if the eslint is throwing an error). The issue has been reported upstream. https://github.com/vuejs/eslint-plugin-vue/issues/1165. Will likely wait till that is resolved before taking action, as there are plenty other examples aside of this one that use the same syntax and would need to be updated.

PR #12155, opened 2020-08-31
docs: import docs-next

johnleider:
## Description
Merge docs-next


KaelWD:
- Missing latest commit from docs-next
- Should replace the docs package, not add a new one
- missing crowdin rewrite in vercel.json
- `src/pages/*` used to be ignored, why was that changed?
- The shared package doesn't do anything? Shared dependencies should be in the root package.json and shared configs should also be in the root directory

tsconfig.json
```diff
@@ -63,8 +63,19 @@
     "emitDecoratorMetadata": true             /* Enables experimental support for emitting type metadata for decorators. */
   },
   "exclude": [
-    "**/*.spec.ts",
+    ".git",
```
KaelWD: > Specifies an array of filenames or patterns that should be skipped when resolving `include`

None of these are in an `include`d path

MajesticPotatoe (approved):
🥇

PR #12239, opened 2020-09-16
fix(VTimePicker): add flex-basis property to container for safari 52 support

gtsiolis:
## Description

This will add the `flex-basis` property on the time picker (VTimePicker) component to render ok on browsers like Firefox 52.

| Before | After |
|-|-|
| image | image |

See [relevant MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis):

> The `flex-basis` CSS property sets the initial main size of a flex item.

## Motivation and Context

The time picker component was not rendering ok on browsers like Firefox 52.

## How Has This Been Tested?

Tried this on Firefox 52 and latest Firefox, Chrome.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
+++ b/packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
@@ -32,6 +32,7 @@
   &__container
     display: flex
     flex-direction: column
+    flex-basis: 290px
     justify-content: center
     padding: $time-picker-clock-padding
 


packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
```diff
@@ -32,6 +32,7 @@
   &__container
     display: flex
     flex-direction: column
+    flex-basis: 290px
```
gtsiolis, line 35: Alternatively, we could remove all flexbox-related properties since this will probably won't have any effect. Deleted code is the best code. 🏀 

```suggestion
```

gtsiolis (commented):
@jacekkarczmarczyk WDYT? /cc @nekosaur

packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
jacekkarczmarczyk: 290px is just the default value, does it work with different widths? and with full-width?
I'll check it later anyway (but only on current browsers)

gtsiolis: Thanks @jacekkarczmarczyk! It seems to work fine on all browsers and widths unless I'm missing something component-specific. Let me know if any further changes are needed. 🔍

johnleider: This fix is concise and resolves the issue. More global changes should be iterated on in a separate PR.

Thank you!

johnleider (approved):

packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
gtsiolis: Thanks @johnleider @jacekkarczmarczyk! 🙏

PR #12244, opened 2020-09-18
feat(VIcon): specify a global default icon component

### LINKED ISSUE #7821: [Feature Request] Iconify Support
### Problem to solve
Currently, the icons supported require loading the full CSS for icons, although this certainly works it adds quite a bit of bloat to the loading times and on very slow mobile networks can cause serious performance impacts. Generally speaking this the way everyone has always loaded icons and I say that continue to do so. However, we also need an alternative that allows loading the SVG icons.

### Proposed solution
I am proposing support for https://iconify.design/ this would allow for potentially much faster load times (only load the icons on the page) and it would allow for the use of many different font packs without loading in any extra CSS or any bloat. The JS that runs it is also very small and in my own testing results in much better icon load times.

I propose that we continue to use the `icon=""` prop. Maybe in the same way we currently handle Material Design icons and font awesome we could use something like `icon="iconify-mdi:google"` where `iconify-` informs the vuetify icon render to use the iconify format. and the `mdi:google` gets passed to the `data-icon=""` part of the iconify format.

Another option would be to completely replace the current icon system and completely replace it with iconify as iconify already includes the Google Material Design Icons (whats currently default) as well as font-awesome, MDI and many more. Plus it could potentially actually decrease the amount of code logic required to create the icon HTML.

dnldsht:
## Description
Allows to use (globally) a different component inside v-icon
fixes #7821

## Motivation and Context
It would be really great to use Iconify with vuetify. 
We can also take advantage of [PurgeIcons](https://github.com/antfu/purge-icons) by @antfu to make our bundle sizes more compact.

## How Has This Been Tested?
I've tested it on the playground, If you like the idea I can write some test and add some docs.

## Markup:

`Playground.vue`

```vue
<template>
  <v-container>
    <v-icon small>mdi:ab-testing</v-icon>
    <v-icon x-large color="green">mdi:about-outline</v-icon>
    <v-icon large color="red">mdi:about-outline</v-icon>
    
    <v-icon x-small>$test</v-icon>
    <v-icon x-small>$test1</v-icon>
    <!--  -->
  </v-container>
</template>
```

custom `Icon.vue` file
```vue
<script>
import Iconify from '@iconify/iconify'

export default {
props: {
    icon: {
      type: String,
      required: true,
    },
  },
  render(createElement) {
    const icon = Iconify.getSVGObject(this.icon, {});
    if(!icon) return createElement('span', this.icon)
    console.log(this.icon, icon)
    return createElement('svg', {
        attrs: icon.attributes,
        domProps: {
            innerHTML: icon.body
        }
    });
  }
}
</script>
```

`vuetify.js`
```js
import Vue from 'vue'
import Vuetify from 'vuetify'
import * as locales from '../src/locale'

import Icon from './Icon'

Vue.use(Vuetify)

export default new Vuetify({
  lang: {
    locales,
  },
  icons: {
    component: Icon,
    iconfont: 'mdi',
    values: {
      test: {
        component: Icon,
        props: {
          icon: 'mdi:1-2-3',
        },
      },
      test1: 'mdi-1-2-3',
    },
  },
})
```

## Types of changes

## Checklist:


gilnd:
Well done! 
I like this addition.

KaelWD (approved):

PR #12252, opened 2020-09-20
chore: remove fibers recommendation and upgrade sass

DRoet:
## Description

Discussed previously on discord, this removes the `fibers` recommendation and also brings support for node `^14` for building the docs/Vuetify

I've tested it against the Vuetify build and removing fibers causes a max. 3 second decrease in performance, but the sass upgrade brings it back to current baseline, I don't think we lose much here.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


nekosaur:
I assume it's still possible to use fibers with the `sass` upgrade, so that we don't invalidate existing configs?

DRoet:
Yeah people can still optionally add it in their own config

johnleider (approved):

PR #12287, opened 2020-09-25
Allow removal of all Sass side effects from components

### LINKED ISSUE #12282: [Feature Request] Allow module importing without (Sass) side-effects
### Problem to solve
We're trying to extend Vuetify to export shared components within our company.

However, this is proving quite difficult for several reasons.

One of the main issues is that to use custom variables, we have to modify webpack's (or vue-cli's) sass-loader. It's very non-trivial then to import that source into another repo (as that repo would _also_ have to modify webpack settings, including sass-loader, to inject this variables file).

### Proposed solution
This WOULD be circumventable if the modules themselves could be imported as components, _and_ their related Sass could be opted to NOT be imported as a side-effect with the module.

As in:
```vue
<!-- Button.vue -->
<template>
  <v-btn>{{ label }}</v-btn>
</template>

<script>
import { VBtn } from 'vuetify/lib'
export default {
  components: { VBtn }
  props: ['label']
}
</script>

<style lang="scss">
@import "styles/variables.scss";  // our custom vars
@import "~vuetify/src/components/VBtn/VBtn.sass";
</style>
```

So, one trivially easy way to enable this is to remove `import './VBtn.sass'` from `VBtn.ts` and add it to `index.ts`.

That would mean that you could import without side effects like:
```js
import VBtn from 'vuetify/lib/components/VBtn/VBtn'
```
and you could _still_ import with side effects like:
```js
import { VBtn } from 'vuetify/lib'
// or import { VBtn } from 'vuetify/lib/components/VBtn'
```

So IMO it's a trivial one-line change that would allow more flexibility in how / when styles are imported, and would allow us to extend Vuetify to multiple repos. Downstream repos, using this pattern would need neither vuetify-loader or a modification of sass-loader to import this source (using vue-cli).

matthew-dean:
## Description

Fixes #12282 - I moved all Sass imports from components (and presets) to the `index.ts` of components. In a normal use case, this should result in no changes. However, it allows you to separately import the Sass file of a component from the component itself, which does the following:
1. This makes modifying sass-loader unnecessary if using a custom `variables.scss` file. (This can be placed in your `.vue` file before the component's styles.)
2. This makes it possible to export a Vuetify-based component library without downstream repos needing to add vuetify-loader and make changes to their `sass-loader` setup.

## Motivation and Context

See: #12282

## How Has This Been Tested?

Unit tests for the `vuetify` package all pass. I was unable, however, at any time, to build the entire project, despite the contributing guide. It generated endless Unhandled promise rejection errors. So I couldn't visually confirm there were no breaking changes; however, when importing this as a local package (instead of the NPM `vuetify`, it works exactly as described above).

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VAlert/VAlert.ts
+++ b/packages/vuetify/src/components/VAlert/VAlert.ts
@@ -1,6 +1,3 @@
-// Styles
-import './VAlert.sass'
-
 // Extensions
 import VSheet from '../VSheet'
 
--- a/packages/vuetify/src/components/VAlert/index.ts
+++ b/packages/vuetify/src/components/VAlert/index.ts
@@ -1,3 +1,6 @@
+// Styles
+import './VAlert.sass'
+
 import VAlert from './VAlert'
 
 export { VAlert }
--- a/packages/vuetify/src/components/VApp/VApp.sass
+++ b/packages/vuetify/src/components/VApp/VApp.sass
@@ -1,4 +1,4 @@
-@import '../../styles/styles.sass'
+@import '../../styles/main.sass'
 
 // Theme
 +theme(v-application) using ($material)
--- a/packages/vuetify/src/components/VApp/VApp.ts
+++ b/packages/vuetify/src/components/VApp/VApp.ts
@@ -1,6 +1,3 @@
-// Styles
-import './VApp.sass'
-
 // Mixins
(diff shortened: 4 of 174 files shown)


matthew-dean:
Hmm, I need to fix a number of things, will close for now.

PR #12148, opened 2020-08-30
fix(VIcon): render a button element around clickable component icons

KaelWD:
fixes #10623

Font icons and SVG strings had a `<span>` or `<button>` put around them, but icon components were just rendered directly. This PR normalises the rendering behaviour between different types of icons. This probably fixes #9999 properly too as the click listener is now on the wrapping button instead of the provided component. 

## Markup:

Switch to `iconfont: 'faSvg'` in `dev/vuetify.js` first

```vue
<template>
  <v-container>
    <v-text-field clearable value="value" label="clearable"></v-text-field>
  </v-container>
</template>
```

IDK if anyone has a complete icon playground, mostly just relying on snapshot tests for this. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

PR #12319, opened 2020-09-30
fix(VIcon): set svg dimensions in style instead of attributes

### LINKED ISSUE #12372: [Bug Report] @mdi/js icons size prop overwritten by .v-icon__svg class
### Environment
**Vuetify Version:** 2.3.13
**Vue Version:** 2.6.12
**Browsers:** Chrome 85.0.4183.121
**OS:** Mac OS 10.15.6

### Steps to reproduce
I cant reproduce, could not install @mdi/js in codepen/jsfiddle etc.
reproduce:
Follow official guide to install @mdi/js https://vuetifyjs.com/en/features/icons/#material-design-icons---js-svg
Add icont to project:
  <v-icon size="200px" left>{{ mdiPaw }}</v-icon>
Icon won't be resized and you can see .v-icon__svg is overwritting size behavior.

### Expected Behavior
Expected size to be increased

### Actual Behavior
Nothing happens, only span around is increasing its size

### Reproduction Link
<a href="https://codepen.io/albertb/pen/PozoQJj" target="_blank">https://codepen.io/albertb/pen/PozoQJj</a>

KaelWD:
Fixes regression from #12148 preventing size props from working

fixes #12372

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-btn color="primary" icon>
      <v-icon large>{{ mdiAccountCircle }}</v-icon>
    </v-btn>

    <v-btn color="primary" icon>
      <v-icon>{{ mdiAccountCircle }}</v-icon>
    </v-btn>
  </v-container>
</template>

<script>
  import { mdiAccountCircle } from '@mdi/js'

  export default {
    data: () => ({
      mdiAccountCircle,
    }),
  }
</script>
```
</details>

## Types of changes

### DIFF
--- a/packages/vuetify/src/components/VIcon/VIcon.ts
+++ b/packages/vuetify/src/components/VIcon/VIcon.ts
@@ -153,17 +153,23 @@ const VIcon = mixins(
       return h(this.hasClickListener ? 'button' : this.tag, data, newChildren)
     },
     renderSvgIcon (icon: string, h: CreateElement): VNode {
-      const fontSize = this.getSize()
       const svgData: VNodeData = {
         class: 'v-icon__svg',
         attrs: {
           xmlns: 'http://www.w3.org/2000/svg',
           viewBox: '0 0 24 24',
-          height: fontSize || '24',
-          width: fontSize || '24',
           role: 'img',
           'aria-hidden': true,
         },
+      }
+
+      const size = this.getSize()
+      if (size) {
+        svgData.style = {
+          fontSize: size,
+          height: size,
+          width: size,
+        }
       }
 
       return h(this.hasClickListener ? 'button' : 'span', this.getSvgWrapperData(), [
--- a/packages/vuetify/src/components/VIcon/__tests__/__snapshots__/VIcon.spec.ts.snap
+++ b/packages/vuetify/src/components/VIcon/__tests__/__snapshots__/VIcon.spec.ts.snap
@@ -6,8 +6,6 @@ exports[`VIcon for component icon should render an svg icon 1`] = `
 >
   <svg xmlns="http://www.w3.org/2000/svg"
        viewbox="0 0 24 24"
-       height="24"
-       width="24"
        role="img"
        aria-hidden="true"
        class="v-icon__svg"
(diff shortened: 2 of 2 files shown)


icleolion:
Yeah I had a shock after I updated and all the icons were huge!  Glad I saw this PR. I was about to start adjusting the size of them all again as I thought it was just a necessary consequence of the fix.

.gitattributes
```diff
@@ -1 +1 @@
-* text=auto
+* text=auto eol=lf
```
johnleider, line 1: This apparently was fixed:

- <https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#L248>
- <https://stackoverflow.com/questions/29435156/what-will-text-auto-eol-lf-in-gitattributes-do>

packages/vuetify/src/components/VIcon/VIcon.ts
```diff
@@ -153,19 +153,25 @@ const VIcon = mixins(
...
           xmlns: 'http://www.w3.org/2000/svg',
           viewBox: '0 0 24 24',
-          height: fontSize || '24',
-          width: fontSize || '24',
           role: 'img',
           'aria-hidden': true,
         },
       }
 
+      const size = this.getSize()
```
johnleider, line 166:
```suggestion
      const size = this.getSize()

```

johnleider (changes requested):

.gitattributes
KaelWD: > The
> combination has been fixed to be equivalent to doing
> $ git config core.autocrlf true

Yeah that's what we want (except with `input` instead). The release notes are saying that `text=auto eol=lf` previously behaved like `text eol=lf` which marks every file as text instead of detecting text or binary.

KaelWD: This commit was also supposed to be to master because I am the dumb

packages/vuetify/src/components/VIcon/VIcon.ts
KaelWD: Why? That's how it is everywhere else and in the original code.

PR #12344, opened 2020-10-03
feat: add global prop defaults, update bootstrapping

nekosaur:


packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
+  }
+
+  return {
+    install,
   }
+}
+
+// Would be nice to have PropOptions here
+const isFactory = (type: any) => {
+  const arr = Array.isArray(type) ? type : [type]
```
johnleider, line 75: could we use the `wrapInArray` helper function?

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
     }
 
-    // rtl is not installed and
-    // will never be called by
-    // the init process
-    this.framework.rtl = Boolean(this.preset.rtl) as any
+    app.provide(VuetifySymbol, vuetify)
+  }
+
+  return {
```
johnleider, line 68: Could be a single line.

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
+const isFactory = (type: any) => {
+  const arr = Array.isArray(type) ? type : [type]
+  return arr.find((t: any) => t === Array || t === Object)
+}
 
-  mergePreset ({ preset, ...userPreset }: VuetifyUserPreset) {
-    return mergeDeep(mergeDeep(Preset, preset), userPreset) as VuetifyPreset
+export const makeProps = <P extends Record<string, Prop<any>>>(props: P) => {
+  for (const key in props) {
+    const propOptions = (props[key] as any)
```
johnleider, line 81:
```suggestion
    const propOptions = (props[key] as any)

```

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
-  // Instantiate a VuetifyService
-  use (Service: VuetifyService) {
-    const property = Service.property
+  return props
+}
 
-    if (property in this.framework) return
+const generateDefault = (propName: string, localDefault: any, type: any) => {
+  return () => {
+    const vm = getCurrentInstance()
```
johnleider, line 90:
```suggestion
    const vm = getCurrentInstance()

```

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
+    if (!vm || !vm.type.name) {
+      console.warn('Unable to get current component instance when generating default prop value')
```
johnleider, line 92:
```suggestion
      console.warn('Unable to get current component instance when generating default prop value')

```

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
+      return localDefault
+    }
```
johnleider, line 94:
```suggestion
    }

```

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
 }
+
+// TODO: Fix typings and shit
+const getGlobalDefault = <
+  C extends string,
+  P extends string
+>(component: C, prop: P) => {
+  const vuetify = useVuetify()
+  const key = component as keyof VuetifyComponentDefaults
+  const defaults = vuetify.defaults[key] as any
```
johnleider, line 109:
```suggestion
  const defaults = vuetify.defaults[key] as any

```

packages/vuetify/src/framework.ts
```diff
@@ -1,83 +1,112 @@
...
+  if (defaults == null) return null
```
johnleider, line 110:
```suggestion
  if (defaults == null) return null

```

johnleider (changes requested):

KaelWD (approved):

PR #12348, opened 2020-10-04
feat(VMenu): add jump to first/last item keyboard navigation 

janKollars:
pressing HOME in a v-menu selects the first selectable list item, END the last

## Description

I am not aware of any related open issues.

## Motivation and Context
VSelect uses the VMenu component. On long lists it is convenient to use HOME/END to navigate to the top/bottom. This is an optional [behaviour](https://www.w3.org/TR/wai-aria-practices/#listbox) according to the W3C and is implemented in their [examples](https://www.w3.org/TR/wai-aria-practices/#examples-6). The `<select>` HTML element as implemented in Chrome/Firefox (Windows/Linux) provides this functionality too.\
Moreover, a menu as defined by the [W3C](https://www.w3.org/TR/wai-aria-practices/#menu) is required to be  navigable using these two buttons.

## How Has This Been Tested?

A new unit test was created for checking the new behaviour.

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-select
      :items="[
        {
          text: 'foo',
          disabled: true,
        },
        'bar',
        'hello',
        'world'
      ]"
      label="pick one"
    />
  </v-container>
</template>
```
</details>

## Types of changes

## Checklist:

The last checklist item is not necessary in my opinion. Please correct me if I am wrong.


packages/vuetify/src/components/VMenu/VMenu.ts
```diff
@@ -432,7 +452,7 @@ export default baseMixins.extend({
         this.$nextTick(() => activator && activator.focus())
       } else if (
         !this.isActive &&
-        [keyCodes.up, keyCodes.down].includes(e.keyCode)
+        [keyCodes.up, keyCodes.down, keyCodes.home, keyCodes.end].includes(e.keyCode)
```
KaelWD, line 455: https://www.w3.org/TR/wai-aria-practices/#menubutton only mentions up/down, and none of the sections you linked explicitly state that home/end should also open the menu.

janKollars: You are absolutely correct! While focusing a select, home and end should only select the first/last selectable item and should not open the menu.
Sorry for the inconvenience. I will fix this as soon as I can.

packages/vuetify/src/components/VMenu/VMenu.ts
```diff
@@ -424,6 +428,22 @@ export default baseMixins.extend({
       this.listIndex--
       if (tile.tabIndex === -1) this.prevTile()
     },
+    lastTile () {
+      const tile = this.tiles[this.tiles.length - 1]
```
johnleider, line 432:
```suggestion
      const tile = this.tiles[this.tiles.length - 1]

```

packages/vuetify/src/components/VMenu/VMenu.ts
```diff
@@ -424,6 +428,22 @@ export default baseMixins.extend({
...
+      if (!tile) return
+
+      this.listIndex = this.tiles.length - 1
+
+      if (tile.tabIndex === -1) this.prevTile()
+    },
+    firstTile () {
+      const tile = this.tiles[0]
```
johnleider, line 440:
```suggestion
      const tile = this.tiles[0]

```

johnleider (changes requested):

janKollars:
I incorporated the feedback. When you test the changes, you might notice a bug: The select loses the focus after up/down/home/end was pressed. This problem was not introduced by my changes.
I created an issue for this #12572

KaelWD (approved):

PR #7441, opened 2019-06-09
feat(VTextField, VSelect): add clearable mixin (#4144, #5722, #6751)

bludnic:
## Description
Add clearable mixin to customize reset value by using `reset-value` prop.

Tested with these components: VTextField, VTextarea, VSelect, VAutocomplete, VCombobox, VOverflowBtn.

## Motivation and Context

Solves these issues: #4144, #5722, #6751. And maybe #2752.

## How Has This Been Tested?

unit, visually

## Markup:

<details>

```vue
<template>
  <v-container>
    <!-- VTextField -->
    <v-layout
      wrap
      class="w300"
    >
      <h3 class="headline">TextField</h3>
      <v-text-field
        v-model="text"
        clearable
        reset-value="reset"
      ></v-text-field>
    </v-layout>

    <v-divider></v-divider>

    <!-- VTextarea -->
    <v-layout
      wrap
      class="w300"
      mt-5
    >
      <h3 class="headline">VTextarea</h3>
      <v-textarea
        v-model="text"
        clearable
        reset-value="reset"
      ></v-textarea>
    </v-layout>

    <v-divider></v-divider>

    <!-- VSelect -->
    <v-layout
      wrap
      class="w300"
      mt-5
    >
      <h3 class="headline">VSelect</h3>
      <v-flex xs12>
        <v-select
          v-model="select"
          :items="items"
          clearable
          reset-value="Vue"
        ></v-select>
      </v-flex>
    </v-layout>

    <v-divider></v-divider>

    <!-- VSelect (multiple) -->
    <v-layout
      wrap
      class="w300"
      mt-5
    >
      <h3 class="headline">VSelect (multiple)</h3>
      <v-flex xs12>
        <v-select
          v-model="selectMultiple"
          :items="items"
          :reset-value="['Vue']"
          clearable
          multiple
        ></v-select>
      </v-flex>
    </v-layout>

    <v-divider></v-divider>

    <!-- VAutocomplete -->
    <v-layout
      wrap
      class="w300"
      mt-5
    >
      <h3 class="headline">VAutocomplete</h3>
      <v-flex xs12>
        <v-autocomplete
          v-model="select"
          :items="items"
          :search-input.sync="search"
          placeholder="Search"
          hide-details
          clearable
          :reset-value="'Vue'"
        ></v-autocomplete>
      </v-flex>

      <v-flex xs12>
        <v-alert
          type="info"
          value="true"
        >
          {{ select }}
        </v-alert>
      </v-flex>
[...]

Issue #4431, opened 2018-06-30
[Bug Report] (v-autocomplete) not able to select an item that is a null value or an empty string


DRoet:
### Versions and Environment
**Vuetify:** 1.1.0
**Vue:** 2.5.16
**Browsers:** Chrome 67.0.3396.99
**OS:** Windows

### Steps to reproduce
Click on the first item of the autocomplete element in the codepen

### Expected Behavior
The first element gets (visually) selected

### Actual Behavior
The element doesn't (visually) get selected and looks like an empty state

### Reproduction Link
[https://codepen.io/anon/pen/MXLKwO](https://codepen.io/anon/pen/MXLKwO)

### Other comments
note: the v-model value actually does change correctly

DRoet:
This also happens in V-Select when v-model is already set to null (empty string '' does work however)

codepen: 
https://codepen.io/anon/pen/QxPYyv

Issue #4144, opened 2018-05-25
[Feature Request] Pressing the clearable icon on <v-text-field> with clearable prop sets the model to null not to empty string


zawnk:
### Versions and Environment
**Vuetify:** 1.0.18
**Vue:** 2.5.16
**Browsers:** Vivaldi 1.96.1147.42
**OS:** Windows 10

### Steps to reproduce
Create a `<v-text-field>` with `clearable` and the model for it. Fill in some text. Press X icon (default) to clear the text, check the model.

### Expected Behavior
Model is an empty string `''`.

### Actual Behavior
Model is `null`.

### Reproduction Link
[https://codepen.io/anon/pen/YLoaOo?editors=1111](https://codepen.io/anon/pen/YLoaOo?editors=1111)

### Other comments
It gets annoying when you have String functions on that model and they return errors because they can't be applied to `null` if a user in the front-end uses the clearable icon. The bug also applies on the newest alpha.

ChaStPierre:
do we have a workaround? 
Thanks

johnleider:
You would need to extend the component and rewrite its reset functionality.

duoweb:
Hi! I found a solution to your problem. Use mask props!

duoweb:
But I have the opposite problem. I have a object from backend with null values. and if the user enters a value and clicks clearable, the value will change to an empty string instead of null. Provided that the mask is used in the field.

e0:
It looks like there is a new prop that solves this problem now. In version > 1.1, there is a prop called `clear-icon-cb`, which can be used like `:clear-icon-cb="handleClear"`.

robodude:
I ran into this myself today.

Check out the differences in using `clear-icon-cb` vs `@click:clear`

https://codepen.io/anon/pen/RerOXe?editors=1011

With `@click:clear` crashes because even though I set the value to `""`, the new handler must be setting the value to `null` afterwards.

NoahCardoza:
I "third" an empty-string as well for the same reasons. It creates problems because `null` isn't the same as `''`, which is what I'd expect to find in a text/string input field.

cb109:
Issue just popped up on Discord again. I think it's a good idea to make this configurable, as John suggested.

matteotrapani:
> I can see this being added as a prop in the future, **reset-value**, but currently this is not considered a bug.

Any news about **reset-value**?
It would be great for me because I must manage clearable on VSelect.
[...]

PR #12373, opened 2020-10-07
fix(inputs): reset value to null instead of undefined

### LINKED ISSUE #7429: [Bug Report] Form reset should set value to null
### Versions and Environment
**Vuetify:** 1.5.14
**Vue:** 2.6.10
**Browsers:** Chrome 75.0.3770.80
**OS:** Linux x86_64

### Steps to reproduce
Press "Reset" button

### Expected Behavior
Resetting form should set input value to null (according to docs) 

### Actual Behavior
Resetting form set input value to undefined

### Reproduction Link
<a href="https://codepen.io/anon/pen/zQVmob?editors=1010#0" target="_blank">https://codepen.io/anon/pen/zQVmob?editors=1010#0</a>

KaelWD:
## Description

This fixes some inconsistencies when resetting input fields. 

- **v-text-field** would reset to `null` when clear was clicked, but `undefined` in a form.  
- **v-select** and variants, and **v-file-input** would reset to `undefined` or `[]` if multiple.  
- **v-autocomplete** and **v-combobox** would reset `search-input` to `null`, unless clear was clicked then it would be `undefined` instead

With this change these will all be `null` or `[]`, allowing values to be sent as JSON without having to check for `undefined` first. 

This is technically a breaking change if someone was relying on `form.reset()` to reset inputs to `undefined` but this is unlikely and can easily be worked around by setting the values manually afterwards.  
**v-select** also now re-selects an item with `value: null` if the clear button is clicked (#4431, unit test). Again, unlikely to actually be a problem (either intended behaviour or not using both clearable and a null item) and easily worked around by implementing the clear button yourself with `append-icon` and `@click:clear`. 

The consistency is worth these low-impact changes in my opinion. 

Some sort of clear-value prop (#4144, #7441) would be even better, but the PR was abandoned and apparently had problems with non-falsy values. 

fixes #7429

## Types of changes

---

Untested outside of existing unit tests btw, might be scuffed


johnleider (dismissed):

sindilevich (changes requested):
Hello, this is a great PR that tremendously streamlines handling inputs' data!
I thought to highlight, however, that some traces of `undefined` are still left unaltered. I can't tell whether they'd matter at all, but thought it'd better to point them out:
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L54
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L99
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L180
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L288
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VFileInput/VFileInput.ts#L69-L70
- https://github.com/vuetifyjs/vuetify/blob/4a1bf59bfe83a395fac8bc07480b2beee5eb37be/packages/vuetify/src/components/VSelect/VSelect.ts#L122-L124

Hope it helps, and thank you.

KaelWD:
Good catch, thanks.

johnleider (approved):
It's outside of this issue's scope but I think we should explore exposing the reset process so the user could set it to whatever they want.

sindilevich:
> It's outside of this issue's scope but I think we should explore exposing the reset process so the user could set it to whatever they want.

Letting user a way to control the reset value - is a great idea! Hope someone could come up with an appropriate PR.

Issue #11340, opened 2020-05-07
[Documentation] Examples should be in plain markup


apfrod:
### Problem to solve
the code examples for things like what classes to use are very hard to understand because they've all been made with script.

eg
```
v-for="a in align"
      :key="a"
      :class="`d-flex align-${a} mb-6`"
```
i cannot read that at a glance to see the valid syntax in the simplest case.

### Proposed solution
Just list them plainly as examples, in longhand, so the reader can quickly refer to the actual expected syntax.

AngeloBonini:
Hi, mind I take this one?

MajesticPotatoe:
> Hi, mind I take this one?

go ahead

santosh:
How are things going with this one?

ClintOxx:
I’m curious about the progress on this too

caweidmann:
Hello, I'd be keen to tackle this one if no one has started on this already.

caweidmann:
Can I tackle this ticket?

KareemDa:
how about seperating this issue into multiple ones as there is over 200 files to change in order to remove all `v-for` in documentation example

MajesticPotatoe:
> how about separating this issue into multiple ones as there is over 200 files to change in order to remove all `v-for` in documentation example

Little impractical. Can simple reference this ticket in each pr (or one pr for all of it)

nekosaur:
Also keep in mind that not all v-for usage in documentation examples is "bad"

caweidmann:
I haven't yet gotten around to it, been caught up with work. This has shifted in priorities for me, however, I'm still keen to tackle this. Might only get to it in a few weeks time.

KareemDa:
> I haven't yet gotten around to it, been caught up with work. This has shifted in priorities for me, however, I'm still keen to tackle this. Might only get to it in a few weeks time.

okay, I can help with some pages, will start fixing them nad make PRs for each section.

KareemDa:
I've started working on this task, but I have a question on [flex page](http://localhost:8082/en/styles/flex):
Currently, we're generating all flex items using, v-for. 

Isn't it understandable in this case? should use plain markup in this example too?

johnleider:
I don't want to remove this styling from everything. There is a benefit to it and I think having a healthy mix helps the user implement different use-cases.

ClintOxx:
The flex/grid pages is what needs this change the most

KareemDa:
[...]

PR #12393, opened 2020-10-10
Docs: Plain markup docs

ricardovanlaarhoven:
## Description
I've tried to do something for #11340 but i'm not sure this is the way vuetify wan't to go, so i only did it for 2 examples.

## Motivation and Context
#11340

## How Has This Been Tested?
just opened the docs locally and checked if they have the same result as online

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/examples/flex/flex-align.vue
+++ b/packages/docs/src/examples/flex/flex-align.vue
@@ -1,9 +1,7 @@
 <template>
   <div>
     <v-card
-      v-for="a in align"
-      :key="a"
-      :class="`d-flex align-${a} mb-6`"
+      class="d-flex align-start mb-6"
       color="grey lighten-2"
       flat
       height="100"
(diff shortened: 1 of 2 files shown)


KaelWD:
We have a usage component now that outputs markup based on selected props:  
https://vuetifyjs.com/en/components/alerts/#usage  
https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/components/examples/Usage.vue

I wonder if it could be used here too instead of duplicating everything.

ricardovanlaarhoven:
I'm not getting it to work with the current Usage component.
1. it's dedicated into turning the came of the usage into an element (in my case `<flex></flex>`)
2. when simply creating a new option `class` this is rendered as an attribute to the first element. And this should be on a lower element
3. i think the best way would be one usage example with multiple selects (one for justify, one for align etc) and this would require a big change to the usage component

I've tried to add an extra option with classes with something like this
`attributeArray.class = Object.values(this.usageClasses).join(' ');`
but i'm not sure how to get an extra object next to options passed from the flex usage to the base usage component.
It seems a bit like magic how this component get's its data

Perhaps with a bit more information i could get it to work

MajesticPotatoe:
>i think the best way would be one usage example with multiple selects (one for justify, one for align etc) and this would require a big change to the usage component

The usage component was originally designed for components and prop usage, not class helpers. It is however still in its infancy (not all components use it, and it lacks the ability for some things like slot/events/functions). May be something simple as giving it `type="class"` and it just outputting a div and passing with the selected classes. Let me see about playing around with that.

ricardovanlaarhoven:
So you'll create this new component or make this adjustment?

Thenit's up to you if you merge this PR and create a new issue (For me) or if i have to change this PR once this component is done

MajesticPotatoe:
>So you'll create this new component or make this adjustment?
Thenit's up to you if you merge this PR and create a new issue (For me) or if i have to change this PR once this component is done

For now well take this as is. Changing/adding the usage component is a PR in itself and more than just this would be updated by it.

MajesticPotatoe (approved):

PR #12471, opened 2020-10-21
feat(Layout): add new layout system

johnleider:
## Description
POC implementation for RFC https://github.com/vuetifyjs/rfcs/tree/rfc/rename-application-service

## Markup:

<details>

```vue
<template>
  <div>
    <VLayout
      group="0"
      priority="0"
      width="200"
      position="left"
      style="background: blue;"
    >
    </VLayout>

    <VLayout
      group="0"
      priority="1"
      height="200"
      position="top"
      style="background: red;"
    >
    </VLayout>

    <VLayout
      group="1"
      priority="0"
      width="200"
      position="right"
      style="background: orange;"
    >
    </VLayout>

    <VLayout
      group="2"
      priority="0"
      height="200"
      position="bottom"
      style="background: yellow;"
    >
    </VLayout>
  </div>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      // const vuetify = useVuetify()
    },
  }
</script>

<style lang="sass">
  .v-layout
    position: fixed
    opacity: .6
</style>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VApp/VApp.ts
+++ b/packages/vuetify/src/components/VApp/VApp.ts
@@ -1,23 +1,143 @@
+/* eslint-disable vue/one-component-per-file */
 // Styles
-import './VApp.sass'
 
-import { defineComponent, computed, h, mergeProps } from 'vue'
+import mergeData, { mergeStyles } from '../../util/mergeData'
+import type { InjectionKey } from 'vue'
+import {
+  computed,
+  defineComponent,
+  h,
+  inject,
+  getCurrentInstance,
+  onMounted,
+  onBeforeUnmount,
+  Prop,
+  provide,
+  ref,
+  Ref,
+  watch,
+} from 'vue'
+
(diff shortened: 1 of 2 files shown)


nekosaur:
I feel that there are a number of things missing here, but I'm not sure what your planned scope is for this poc, so apologies in advance if I'm jumping the gun before everything is implemented. Or maybe we just have different ideas of what a poc should be.

- there is no `v-main` equivalent shown
- dynamically removing `v-layout` does not work correctly
- dynamically changing width/height of `v-layout` does not work correctly
- having a child app layout inside another is not shown
- `priority` is shown in the markup but does not seem to have an effect when changed

When it comes to the implementation, my main comment is that I am not a fan of the callback-style updating of values. I would prefer to see something that feels more idiomatically Vue, using the reactivity API.

johnleider:
The team has consolidated the best ideas from this PR into #13067.

PR #12516, opened 2020-10-26
feat(VAppBar): don't truncate title with shrink-on-scroll

### LINKED ISSUE #12514: [Bug Report] Items in <v-app-bar fixed shrink-on-scroll> block <v-toolbar-title> on low resolution even when the app-bar is expanded
### Environment
**Vuetify Version:** 2.3.15
**Vue Version:** 2.6.12
**Browsers:** Firefox 81.0
**OS:** Windows 10

### Steps to reproduce
1. Open page without scrolling down (which would cause the "Vuetify.js" title to move up from the second line below icons due to `shrink-on-scroll`)
2. Reduce the page width by resizing window to move the icons on right side to be above the title

### Expected Behavior
There's no intersection, and so the title should be properly displayed

### Actual Behavior
The title becomes shorter, like "Vueti...", for no apparent reason to the visitor:

Vueti...

Paradoxically, more of the text gets shown due to font size change after a scroll once there's less space.

### Reproduction Link
<a href="https://codesandbox.io/s/distracted-meninsky-osl9g?file=/layouts/default.vue" target="_blank">https://codesandbox.io/s/distracted-meninsky-osl9g?file=/layouts/default.vue</a>

### Other comments
* This behavior isn't customizable or documented (just like the inability to specify title scaling details, or any icon hiding priority).
* The documentation also doesn't properly explain that we have to put all button icons inside [`<v-toolbar-items>`](https://vuetifyjs.com/en/api/v-toolbar-items/) (which has no API description) before putting them inside `<v-toolbar-title>`, and even then the positions aren't perfect.
* `shrink-on-scroll` seems to automatically apply `prominent`, which the docs don't explicitly mention, other than "Shrinks a prominent toolbar".
* I also misunderstood this feature by assuming that it provides an additional slot below for items while not shrunk, but then realized that the only available `extension` slot is always visible (as a "third row"), so I'd like to ask, is it a design decision not to allow a similar behavior, or is it reasonable to be considered in a future?

KaelWD:
## Description

Adds a new v-app-bar-title component which uses absolute positioning to avoid
being truncated when there isn't enough space in the main part of the toolbar

closes #12514

## Markup:

<details>

```vue
<template>
  <v-container style="height: 1200px">
    <v-app-bar
      fixed
      shrink-on-scroll
    >
      <v-app-bar-nav-icon />
      <v-app-bar-title>Vuetify.js</v-app-bar-title>
      <v-spacer />
      <v-btn icon>
        <v-icon>mdi-chart-box-outline</v-icon>
      </v-btn>
      <v-btn icon>
        <v-icon>mdi-cog-outline</v-icon>
      </v-btn>
      <v-responsive max-width="260">
        <v-text-field
          class="ml-1"
          dense
          flat
          placeholder="Search Anywhere"
          hide-details
          rounded
          solo-inverted
          prepend-inner-icon="mdi-magnify"
        />
      </v-responsive>
      <v-btn icon>
        <v-icon>mdi-menu</v-icon>
      </v-btn>
    </v-app-bar>
  </v-container>
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/pages/en/components/app-bars.md
+++ b/packages/docs/src/pages/en/components/app-bars.md
@@ -35,7 +35,7 @@ A styled icon button component created specifically for use with [v-toolbar](/co
 
 ### v-app-bar-title
 
-A modified [v-toolbar-title](/components/toolbars) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
+A modified [v-toolbar-title](/components/toolbars/) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
 
 ## Caveats
 
--- a/packages/docs/src/pages/en/components/app-bars.md
+++ b/packages/docs/src/pages/en/components/app-bars.md
@@ -25,12 +25,17 @@ The `v-app-bar` component is used for application-wide actions and information.
 
 - [v-app-bar](/api/v-app-bar)
 - [v-app-bar-nav-icon](/api/v-app-bar-nav-icon)
+- [v-app-bar-title](/api/v-app-bar-title)
 
 ## Sub-components
 
 ### v-app-bar-nav-icon
 
 A styled icon button component created specifically for use with [v-toolbar](/components/toolbars) and `v-app-bar`. Typically seen on the left side of a toolbar as a hamburger menu, it is often used to control the state of a navigation drawer. The `default` slot can be used to customize the icon and function of this component. This is a **functional** component.
+
+### v-app-bar-title
+
+A modified [v-toolbar-title](/components/toolbars) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
 
 ## Caveats
 
--- a/packages/docs/src/pages/en/components/app-bars.md
+++ b/packages/docs/src/pages/en/components/app-bars.md
@@ -35,7 +35,7 @@ A styled icon button component created specifically for use with [v-toolbar](/co
 
 ### v-app-bar-title
 
-A modified [v-toolbar-title](/components/toolbars/) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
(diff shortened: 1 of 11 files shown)


packages/docs/src/pages/en/components/app-bars.md
```diff
@@ -25,13 +25,18 @@ The `v-app-bar` component is used for application-wide actions and information.
...
 
 ## Sub-components
 
 ### v-app-bar-nav-icon
 
 A styled icon button component created specifically for use with [v-toolbar](/components/toolbars) and `v-app-bar`. Typically seen on the left side of a toolbar as a hamburger menu, it is often used to control the state of a navigation drawer. The `default` slot can be used to customize the icon and function of this component. This is a **functional** component.
 
+### v-app-bar-title
+
+A modified [v-toolbar-title](/components/toolbars) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
```
johnleider, line 38:
```suggestion
A modified [v-toolbar-title](/components/toolbars/) specifically for use with the `shrink-on-scroll` prop. `v-toolbar-title` will be truncated on small screens (see [issue #12514](https://github.com/vuetifyjs/vuetify/issues/12514)) but this component uses absolute positioning to be fully visible when expanded. Don't use `v-app-bar-title` without `shrink-on-scroll` as it does add a resize watcher and some extra calculations.
```

johnleider: In addition, I really like the idea of having a consistent component name and would love it if possible to use it as a semantic alternative to `v-toolbar-title`, even if the user didn't enable **shrink-on-scroll**.

packages/vuetify/src/components/VAppBar/VAppBar.ts
```diff
@@ -98,28 +102,25 @@ export default baseMixins.extend({
         'v-app-bar--shrink-on-scroll': this.shrinkOnScroll,
       }
     },
+    scrollRatio (): number {
+      const threshold = this.computedScrollThreshold
```
johnleider, line 106:
```suggestion
      const threshold = this.computedScrollThreshold

```

packages/vuetify/src/components/VAppBar/VAppBarTitle.ts
```diff
@@ -0,0 +1,80 @@
+// Mixins
+import { inject } from '../../mixins/registrable'
+
+// Types
+import { VNode } from 'vue'
+import { ExtractVue } from '../../util/mixins'
+import VAppBar from './VAppBar'
+
+// Helpers
```
johnleider, line 9:
```suggestion
// Utilities
```

packages/vuetify/src/components/VAppBar/VAppBarTitle.ts
```diff
@@ -0,0 +1,80 @@
...
+    content: Element
+    placeholder: Element
+  }
+}
+
+export default base.extend<options>().extend({
+  name: 'v-app-bar-title',
+
+  data: () => ({
+    width: 0,
```
johnleider, line 26: move under https://github.com/vuetifyjs/vuetify/pull/12516/files#diff-fa84974a767a521a8eb6cab9dce6857158f2b5f15b1d8bb7b97c42f43c379f58R28

packages/vuetify/src/components/VAppBar/VAppBarTitle.ts
```diff
@@ -0,0 +1,80 @@
...
+    },
+  },
+
+  mounted () {
+    this.updateDimensions()
+  },
+
+  methods: {
+    updateDimensions (): void {
+      const dimensions = this.$refs.placeholder.getBoundingClientRect()
```
johnleider, line 55:
```suggestion
      const dimensions = this.$refs.placeholder.getBoundingClientRect()

```

johnleider (changes requested):

packages/docs/src/pages/en/components/app-bars.md
KaelWD: > would love it if possible to use it as a semantic alternative to v-toolbar-title

You technically can, but I explained why you shouldn't here: https://github.com/vuetifyjs/vuetify/issues/12514#issuecomment-717143516

PR #12542, opened 2020-11-01
feat: initial theme implementation

nekosaur:
## Markup:

<details>

```vue
<template>
  <v-app>
    <foo/>
    <div>
      <div v-for="color in colors" :class="['box', `bg-${color}`, `on-${color}`]" :key="color">{{ color }}</div>
    </div>
    <div>
      <div v-for="color in textColors" :class="['box', `text-${color}`, 'border', `border-${color}`]" :key="color">{{ color }}</div>
    </div>
    <v-theme-provider class="padding">
      <foo/>
      <div>
        <div v-for="color in colors" :class="['box', `bg-${color}`, `on-${color}`]" :key="color">{{ color }}</div>
      </div>
      <div>
        <div v-for="color in textColors" :class="['box', `text-${color}`, 'border', `border-${color}`]" :key="color">{{ color }}</div>
      </div>
      <v-theme-provider v-model:theme="asd" class="padding">
        <foo/>
        <div>
          <div v-for="color in colors" :class="['box', `bg-${color}`, `on-${color}`]" :key="color">{{ color }}</div>
        </div>
        <div>
          <div v-for="color in textColors" :class="['box', `text-${color}`, 'border', `border-${color}`]" :key="color">{{ color }}</div>
        </div>
        <v-theme-provider class="padding">
          <div>
            <div v-for="color in colors" :class="['box', `bg-${color}`, `on-${color}`]" :key="color">{{ color }}</div>
          </div>
          <div>
            <div v-for="color in textColors" :class="['box', `text-${color}`, 'border', `border-${color}`]" :key="color">{{ color }}</div>
          </div>
        </v-theme-provider>
      </v-theme-provider>
    </v-theme-provider>
  </v-app>
</template>

<script>
  import { defineComponent, h, onBeforeMount, ref } from 'vue'

  const Foo = defineComponent({
    setup (props) {
      const theme = useTheme()

      const next = () => {
        const keys = Object.keys(theme.themes.value)
        theme.current.value = keys[(keys.indexOf(theme.current.value) + keys.length - 1) % keys.length]
      }

      return () => h('div', [h('div', { onClick: next }, 'next'), h('input', { value: theme.current.value, onInput: e => theme.current.value = e.target.value })])
    },
  })

  export default {
    name: 'Playground',
    components: { Foo },
    setup () {
      const theme = useTheme()
      onBeforeMount(() => {
        theme.setTheme('contrast', {
          dark: true,
          colors: {
            background: '#000000',
            surface: '#222222',
            primary: '#eeeeee',
            secondary: '#ffff00',
            success: '#00ff00',
            warning: '#ffff00',
            error: '#ff0000',
            info: '#0000ff',
            'primary-lighten-1': '#ff00ff',
          },
        })
      })

      const asd = ref('contrast')

      return {
        asd,
        colors: ['surface', 'primary-darken-1', 'primary', 'primary-lighten-1', 'primary-lighten-2', 'secondary-lighten-1', 'secondary', 'secondary-darken-1', 'success', 'warning', 'error', 'info'],
        textColors: ['text', 'primary-darken-1', 'primary', 'primary-lighten-1', 'primary-lighten-2', 'secondary-lighten-1', 'secondary', 'secondary-darken-1', 'success', 'warning', 'error', 'info'],
      }
    },
  }
</script>

<style>
  .box {
    width: 150px;
    height: 150px;
    display: inline-block;
  }

  .border {
    border: 4px solid;
  }

  .padding {
    padding-left: 1rem
  }
</style>
```
</details>

### Multiple themes with dark/light variants

<details>

```vue
<template>
  <div>
    <v-app>
      <label>THEME</label>
      <select v-model="current">
        <option value="first">first</option>
        <option value="second">second</option>
      </select>
      <label>DARK</label><input type="checkbox" v-model="dark" />
      <div class="box bg-primary"/>
      <div class="box bg-secondary"/>
    </v-app>
  </div>
</template>

<script>

  const THEMES = {
    first: {
      dark: {
        dark: true,
        colors: {
          primary: '#FF00FF',
          secondary: '#FFFF00',
          background: '#000000'
        }
      },
      light: {
        dark: false,
        colors: {
          primary: '#00FFFF',
          secondary: '00FF00',
          background: 'FFFFFF',
        }
      }
    },
    second: {
      dark: {
        dark: true,
        colors: {
          primary: '#444444',
          secondary: '#888888',
          background: '#222222'
        }
      },
      light: {
        dark: false,
        colors: {
          primary: '#222222',
          secondary: '#444444',
          background: '#EEEEEE',
        }
      }
    }
  }

  export default {
    setup (props) {
      const theme = useTheme()
      const current = ref('first')
      const dark = ref(false)

      watch([current, dark], () => {
        theme.setTheme('dark', THEMES[current.value].dark)
        theme.setTheme('light', THEMES[current.value].light)
        theme.current.value = dark.value ? 'dark' : 'light'
      }, {
        immediate: true,
      })

      return {
        theme,
        current,
        dark,
      }
    }
  }
</script>

<style>
  .box {
    width: 100px;
    height: 100px;
  }
</style>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,231 @@
+// Utilities
+import { capitalize, computed, inject, provide, ref, watch } from 'vue'
+import { useVuetify } from '@/framework'
+import { colorToInt, colorToRGB, contrastRatio } from '../util'
+
+// Types
+import type { InjectionKey, Ref } from 'vue'
+
+interface BaseColors {
+  background: string
```
renoirb, line 10: How about we have a naming convention similar to ${name}Bg ${name}Text. Because, after all, regardless of the naming, the text is likely going to be against another color. So, ´Bg', ´Text' would be required for any color name. alphaBg, alphaText, primaryBg, primaryText, etc.

packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,231 @@
...
+  surface: string
+  primary: string
+  primaryVariant: string
+  secondary: string
+  secondaryVariant: string
```
renoirb, line 15: Following up previous comment, what would be the purpose and difference between secondary and secondaryVariant?  Is it for grading (e.g. like TailwindCSS would)? Or an abstract term with no relationship. What if there would be more than one variant (dark, sephia, loght,...)

packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,231 @@
...
+  onPrimaryVariant: string
+  onSecondary: string
+  onSecondaryVariant: string
+  onSuccess: string
+  onWarning: string
+  onError: string
+  onInfo: string
+}
+
+interface InternalTheme extends BaseColors, OnColors {
```
renoirb, line 35: export type IInternalTheme = BaseColors & OnColors

Or could be 

const COLORS_SLOTS = [´foo','bar'] as const
export IColorType = typeof COLORS_SLOTS[number]
export IBadeColor = Record<IColorType, string>

Could work too. They're both 1level deep Record.

(Yeah, I should do PR review from a computer)

renoirb (commented):
A few comments, i'll come back later

packages/vuetify/src/composables/theme.ts
MajesticPotatoe: Its material design color system 
https://material.io/design/color/the-color-system.html#color-theme-creation

renoirb: Oh, right.
So in that naming scheme, it might already have a naming convention w.r.t. telling which color name contrasts enough for text against another.

KaelWD: They seem to allow for multiple variants: 
image

packages/vuetify/src/composables/theme.ts
KaelWD: Text would be `onBackground`

packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,260 @@
...
+  const themes = ref<Record<string, ThemeOption>>(combinedOptions.themes)
+  const variations = ref(combinedOptions.variations)
+
+  const toHex = (v: number) => `#${v.toString(16).repeat(3)}`
+
+  const genOnColor = (color: string) => {
+    // naive solution
+    const goal = 8
+    let curr = 0
+    while (contrastRatio(color, toHex(curr)) < goal && curr < 255) {
```
KaelWD, line 156: Last time is tried to implement this: 414cfe1d52f9cb8d45009cbc844efb71e72a299b

I don't recall where the 0.18 value came from.

renoirb:
I've been trying to run it and it says that it doesn't find `v-container`. Is this deliberate?

I'm going to try again with another ticket, in case it's a matter of me not doing it right.

Steps to reproduce:
1. git clone ... vuetify
1. `cd vuetify`
1. `git checkout -b poc/theme -t origin/poc/theme`
1. `yarn dev`

Screen Shot 2020-11-26 at 6 51 18 PM

MajesticPotatoe:
@renoirb 

```
yarn
yarn build vuetify
yarn dev
```

renoirb:
Screen Shot 2020-11-26 at 7 32 33 PM

nekosaur:
Sorry, I forgot to add the Playground markup. You will need to copy it into your `dev/Playground.vue` file.

renoirb:
Is this experiment should also make sure all use-cases shown has good contrast ratios too?

Is this PR also about making sure all contrasts are OK?

KaelWD:
No, that example is explicitly using `text-*` colours on a `--v-theme-background` element. For automatic contrast you use matching `bg-` and `on-` classes, eg. `class="bg-primary on-primary"`

KaelWD:
- `on-` colours are generating variants that are the same as the base, variants should probably be removed for these
- add reserved light/dark key to themes for dark mode elevation

nekosaur:
> * `on-` colours are generating variants that are the same as the base, variants should probably be removed for these

Not sure about this. variants might be sufficiently different in contrast to trigger a switch from black to white or the reverse

KaelWD:
Yeah I just realised they're `on-(x-lighten)` not `on-x-(lighten)`

packages/vuetify/src/util/colorUtils.ts
```diff
@@ -235,15 +236,48 @@ export function RGBtoInt (rgba: RGBA): ColorInt {
...
+  // TODO: why this false positive?
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  lab[0] = lab[0] + amount * 10
+  return sRGB.fromXYZ(CIELAB.toXYZ(lab))
+}
+
+export function darken (value: ColorInt, amount: number): ColorInt {
+  const lab = CIELAB.fromXYZ(sRGB.toXYZ(value))
+  lab[0] = lab[0] - amount * 10
+  return sRGB.fromXYZ(CIELAB.toXYZ(lab))
```
johnleider, line 259:
```suggestion

  return sRGB.fromXYZ(CIELAB.toXYZ(lab))
```

packages/vuetify/src/util/colorUtils.ts
```diff
@@ -235,15 +236,48 @@ export function RGBtoInt (rgba: RGBA): ColorInt {
   return (rgba.r << 16) + (rgba.g << 8) + rgba.b
 }
 
+export function colorToRGB (color: string) {
+  const int = colorToInt(color)
+  return {
```
johnleider, line 241:
```suggestion

  return {
```

packages/vuetify/src/util/colorUtils.ts
```diff
@@ -235,15 +236,48 @@ export function RGBtoInt (rgba: RGBA): ColorInt {
...
+    b: (int & 0xFF),
+  }
+}
+
+export function lighten (value: ColorInt, amount: number): ColorInt {
+  const lab = CIELAB.fromXYZ(sRGB.toXYZ(value))
+  // TODO: why this false positive?
+  // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+  lab[0] = lab[0] + amount * 10
+  return sRGB.fromXYZ(CIELAB.toXYZ(lab))
```
johnleider, line 253:
```suggestion

  return sRGB.fromXYZ(CIELAB.toXYZ(lab))
```

packages/vuetify/src/util/colorUtils.ts
```diff
@@ -235,15 +236,48 @@ export function RGBtoInt (rgba: RGBA): ColorInt {
...
-  const [, y1] = toXYZ(RGBtoInt(c1))
-  const [, y2] = toXYZ(RGBtoInt(c2))
+export function getContrast (first: Color, second: Color) {
+  const l1 = getLuma(first)
+  const l2 = getLuma(second)
 
-  return (Math.max(y1, y2) + 0.05) / (Math.min(y1, y2) + 0.05)
+  const light = Math.max(l1, l2)
+  const dark = Math.min(l1, l2)
+  return (light + 0.05) / (dark + 0.05)
```
johnleider, line 282:
```suggestion

  return (light + 0.05) / (dark + 0.05)
```

johnleider (changes requested):
I think it would be valuable to add a small doc-block for theme.ts for all of the helper functions. It would look similar to other areas in which we use this type of comment:

```js
/** Name
 *  Lorem ipsum...
 *  More info...
 */
`export const useTheme = () => {}`
```

packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,288 @@
...
+        warning: '#FB8C00',
+      },
+    },
+  },
+}
+
+const parseThemeOptions = (options?: ThemeOptions): InternalThemeOptions => {
+  if (options == null) return defaultThemeOptions
+  if (options === false) return { ...defaultThemeOptions, isDisabled: true } as InternalThemeOptions
+  return {
```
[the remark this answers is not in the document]
nekosaur: That should be done outside of this PR

packages/vuetify/src/composables/theme.ts
```diff
@@ -0,0 +1,288 @@
+// Utilities
+import { computed, inject, provide, ref, watch } from 'vue'
+import { colorToInt, colorToRGB, createRange, intToHex, lighten, darken, getLuma } from '@/util'
+
+// Types
+import type { InjectionKey, Ref, SetupContext } from 'vue'
+
+interface BaseColors {
```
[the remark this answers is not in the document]
nekosaur: I don't agree with this.

Hardcoding `primaryVariant` and `secondaryVariant` as keys does not play well with the variants generation, including the ability to turn them off completely. The `BaseColors` represent the minimum amount of colors that a user needs to define for nothing to break out-of-the-box, i.e. colors that Vuetify depends on internally. Any colors defined outside of those (including the generated variations) are entirely in user-space.

If users want to add `primaryVariant`, `secondaryVariant`, or rely on the generated variations, or disable them completely, is (and should be) outside of our control.

I have changed the default behaviour of the generated variations to only generate `primary-darken-1`, `secondary-darken-1` to match the spec example color setup.

johnleider (changes requested):

johnleider (commented):
I made a few changes to code styling and resolved an issue with VApp's testing. I think that this is ready to get in and start testing. If I were to have any more feedback, it would be related to how you declare functions inside of the createTheme method. Unless there is an explicit need for it to be assigned to a variable, this would help keep the codebase more consistent.

PR #12549, opened 2020-11-03
fix(overlayable): overlay in VDialog should be inserted to attach prop

### LINKED ISSUE #6425: [Bug Report] v-dialog overlay still placed in root when using attach
### Versions and Environment
**Vuetify:** 1.5.0
**Vue:** 2.6.2
**Browsers:** Chrome 72.0.3626.81
**OS:** Windows 10

### Steps to reproduce
Create a DOM element with a class/id. Create a VDialog and set the attach attribute with the selector. As described here at the 'attach' property: https://vuetifyjs.com/en/components/dialogs#attach

### Expected Behavior
VDialog and the overlay have the target DOM element as parent. (this.$el.parentNode)

### Actual Behavior
VDialog and overlay have the v-app DOM element as parent. (document.querySelector('[data-app]'))

### Reproduction Link
<a href="https://codepen.io/Paulsky/pen/GzQZvX" target="_blank">https://codepen.io/Paulsky/pen/GzQZvX</a>

sudhakar-s:
fixes #6425 

now overlay will attach to respective location if provided in attach prop

### DIFF
--- a/packages/vuetify/src/mixins/overlayable/index.ts
+++ b/packages/vuetify/src/mixins/overlayable/index.ts
@@ -8,6 +8,7 @@ import {
   addPassiveEventListener,
   getZIndex,
 } from '../../util/helpers'
+import { consoleWarn } from '../../util/console'
 
 // Types
 import Vue from 'vue'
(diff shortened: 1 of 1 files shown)


sudhakarselva:
if everything is ok, Have to fix  type issue, new to typescript, so not sure where how to add attach in overlayable .

sudhakar-s:
if everything is ok, Have to fix type issue, new to typescript, so not sure where how to add attach in overlayable . BTW i am using it in production using patch package

sudhakar-s:
**Note:-** 

1. Attaching overlay element inside attach prop element fixes css stacking order issue mainly.
2. We are attaching overlay inside the respective attach prop, but its css position value is still fixed.
3. If we are going with this fix, users have to validate their app bcuz they might be using some workaround already for this issue or it was fine for them with the old behaviour.

Love to hear some input & feedback and I will work on updating the fix to align with the same.
 Thanks & Happy coding.

johnleider:
Overlayable itself doesn't have access to an **attach** property. This change only works for `v-dialog` because it implements the **detachable** mixin, which provides the property as well as all of the logic you wrote for this PR: https://github.com/vuetifyjs/vuetify/blob/b1c3f81bb62bc218b1c18236f7252412722b32fc/packages/vuetify/src/mixins/detachable/index.ts#L118

Any changes made for a fix should take **detachable** into account.

sudhakarselva:
@johnleider  Thanks for the input, I will look into it  today.

sudhakarselva:
@johnleider  

As per my understanding detachable is responsible for rendering contents for v-dialog, v-menu and v-tooltip and overlayable is used by v-dialog and v-navigation-drawer. one solution maybe we can add prop attach to overlayable.

Note:- 
Overlay or content might not look as expected, since we are using position fixed for v-dialog__content and v-overlay. In that case, we can add workaround by adding [transform, perpective or filter](https://developer.mozilla.org/en-US/docs/Web/CSS/position) to ancestor(attach) style.

KaelWD:
I think this is all too much of a mess to try fixing in v2. I've already resolved it in the rewrite, if we have time after v3 release I might consider backporting some of those changes.

PR #12571, opened 2020-11-08
fix(VSlideGroup): a 'window is not defined' error

mcdmaster:
## Description

Unwanted error messages produced in the format below have even become a showstopper for the development lifecycle with application referring to `VSlideGroup.setWidth()`
VSlideGroup-Error

## Motivation and Context

Eliminate the unwanted error messages so that the development activities get on track

## How Has This Been Tested?

`yarn jest --findRelatedTestings VSlideGroup.ts`

## Markup:

<details>
<p class="codepen" data-height="265" data-theme-id="light" data-default-tab="js,result" data-user="mcdmaster" data-slug-hash="mdEGgzL" style="height: 265px; box-sizing: border-box; display: flex; align-items: center; justify-content: center; border: 2px solid; margin: 1em 0; padding: 1em;" data-pen-title="VSlideGroup.setWidths()">
  <span>See the Pen <a href="https://codepen.io/mcdmaster/pen/mdEGgzL">
  VSlideGroup.setWidths()</a> by mcdmaster (<a href="https://codepen.io/mcdmaster">@mcdmaster</a>)
  on <a href="https://codepen.io">CodePen</a>.</span>
</p>
<script async src="https://static.codepen.io/assets/embed/ei.js"></script>
</details>

## Types of changes

## Checklist:


.env.local
```diff
@@ -0,0 +1,6 @@
+COSMIC_BUCKET_SLUG=vuetify-vslidegroup
+COSMIC_READ_KEY=5FIzwbW5bEHpNLRcL4U0yELZAvHv8HVCd0DCUeiNUtA7OJKTgg
+COSMIC_PREVIEW_SECRET=fK2g0BBjfP6gkfX4FyTcID2RsJzvLdBpDJCYRV8Rgg3cUtvjlq
```
renoirb: I'm guessing you don't want to commit those, or want this merged 😨

packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
```diff
@@ -382,18 +382,20 @@ export const BaseSlideGroup = mixins<options &
...
-        this.widths = {
-          content: content ? content.clientWidth : 0,
-          wrapper: wrapper ? wrapper.clientWidth : 0,
-        }
-
-        this.isOverflowing = this.widths.wrapper < this.widths.content
-
-        this.scrollIntoView()
-      })
+      if (typeof window === 'undefined') {
```
renoirb, line 385: Avoid calling global window, instead use a pattern calling the `$el.ownerDocument.defaultView` which is the current `WindowProxy` (or `Window` global in its frame context).

Also, avoid re-defining object properties, in Vue 2 we have `Vue.set` and `this.$set` methods in case.

```patch
diff --git a/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts b/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
index 1a6693be7..c81c0270e 100644
--- a/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
+++ b/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
@@ -385,5 +385,7 @@ export const BaseSlideGroup = mixins<options &
-      if (typeof window === 'undefined') {
-        this.widths = { content: 0, wrapper: 0 }
-        setTimeout(() => this.setWidths(), 1000)
-      } else if (this.widths === { content: 0, wrapper: 0 }) {
-        window.requestAnimationFrame(() => {
+      // defaultView represent window. But this code is sensitive to
+      // only running when it is client-side, but using Vue.js' so
+      // we do not rely on global object, and will make this code
+      // ALSO work when SSRing
+      const defaultView: Window | null = this.$el && this.$el.ownerDocument ? this.$el.ownerDocument.defaultView : null
+      if (defaultView) {
+        defaultView.requestAnimationFrame(() => {
@@ -391,4 +393,6 @@ export const BaseSlideGroup = mixins<options &
-          this.widths = {
-            content: typeof content !== 'undefined' ? content.clientWidth : 0,
-            wrapper: typeof wrapper !== 'undefined' ? wrapper.clientWidth : 0,
-          }
+          // Avoid re-defining existing objects, this'll override reactive attributes
+          this.widths.content = content && content.clientWidth ? content.clientWidth : 0
+          this.widths.wrapper = wrapper && wrapper.clientWidth ? wrapper.clientWidth : 0
+          // Otherwise, use Vue.set or this.$set
+          // this.$set(this.widths, 'content', typeof content !== 'undefined' ? content.clientWidth : 0)
+          // this.$set(this.widths, 'wrapper', typeof wrapper !== 'undefined' ? wrapper.clientWidth : 0)
@@ -397,0 +402,4 @@ export const BaseSlideGroup = mixins<options &
+      } else {
+        this.widths.content = 0
+        this.widths.wrapper = 0
+        setTimeout(() => this.setWidths(), 1000)
```

.env.local
mcdmaster: I DID NOT want to. Thank you.

packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
mcdmaster: That is why I modified the source, in which the `window` (not "`Window`") object was called without prior assignment, allocation, reference etc.

mcdmaster: Just reflected upon.
Much appreciated your insight

renoirb: When I said capitalized Window, i was making reference to the type.

mcdmaster (commented):
Seems looks like fine. All workflows are to be migrated afterwards.

johnleider:
Couldn't we simply check if we're in the browser and if not, return? We added these recently: https://github.com/vuetifyjs/vuetify/blob/b1c3f81bb62bc218b1c18236f7252412722b32fc/packages/docs/src/util/globals.js#L1

mcdmaster:
Thanks @johnleider 
Supposed that it can be a resolution.
I intend to close this

Issue #8199, opened 2019-07-30
[Feature Request] Data table: missing props.index in item slot


vschimpf:
### Problem to solve
I need that index to do several customizations that depend on the current item index (which depends on the current sort).
I use it for programatically select the first item on the table, customize the row hover behavior, customize the row selection, specify tabindex, and more. All these inside my own generic table component that I use for serveral tables (like a wrapper).
It is a requirement for me to be able to migrate to version 2.0.x

### Proposed solution
Inside the slot "item" the props should also contain the "index" property, like the previous version has.

redijedi:
This is a blocker for my current project. How does one integrate this change? I do not see the `src/maps/v-data-table.js` referenced in the PR.

vschimpf:
> 
> 
> This is a blocker for my current project. How does one integrate this change? I do not see the `src/maps/v-data-table.js` referenced in the PR.

I think the changes are not merged to master yet, let's see if they are included in the next release..

PR #12605, opened 2020-11-12
feat(VDataTable): index prop in item* slots

### LINKED ISSUE #10646: [Feature Request] VDataTable: add missing index variable in "item.<name>" scoped slot
### Problem to solve
It would be very useful for adding a column with row index, and using index for custom stuff in a cell.

### Proposed solution
Inside the scoped slot "item.\<name>", the props should also contain the "index" property, like in the "item" scoped slot (cf #8199 and #8283).

jacekkarczmarczyk:
## Description
Adds `index` to `item`, `item.<name>`, `expanded-item` slots

## Motivation and Context
fixes #10646

## How Has This Been Tested?
playground

## Markup:

<details>

```vue
<template>
  <v-container>
    <h2>#item.&lt;name&gt;</h2>

    <v-data-table
      :headers="tableHeaders"
      :items="tableData"
    >
      <template #item.index="{ index }">
        {{ index }}
      </template>
      <template #item.name="{ item }">
        {{ item.name }}
      </template>
    </v-data-table>

    <h2>#item</h2>

    <v-data-table
      :headers="tableHeaders"
      :items="tableData"
    >
      <template #item="{ index, item }">
        <tr><td>{{ index }}</td><td>{{ item.name }}</td></tr>
      </template>
    </v-data-table>

    <v-data-iterator :items="tableData">
      <template #item="{ index, item }">
        <div>{{ index }}: {{ item.name }}</div>
      </template>
    </v-data-iterator>
  </v-container>
</template>

<script>
  export default {
    data () {
      return {
        tableHeaders: [
          {
            text: 'No',
            sortable: false,
            value: 'index',
          },
          {
            text: 'Name',
            sortable: true,
            value: 'name',
          },
        ],
        tableData: Array(50).fill(null).map((_, index) => ({ name: `Data ${index}` })),
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


KaelWD:
Is this the index in the table, or the index in `items`? Index in items would be more useful but conflicts with #8283.

jacekkarczmarczyk:
in `items` (look at playground, first page has indices 0-9, next one 10-19)
does it make sense to leave `index` as index in table and add other prop for index in `items`?

KaelWD:
That's a breaking change then, although the current behaviour seems pretty useless so I don't know if it would even affect anyone.

jacekkarczmarczyk:
> Does it account for sorting?

No, I'm not sure if it's even possible (especially with server side data). I think if someone wants to take sorting/filtering into account he should just put `index` as a property of `item`

> That's a breaking change then

So is "Reset/clear on inputs" already in 2.4, which imho may have bigger impact than this. As I mentioned before it would be safe to add another property (`globalIndex`?), but I feel like it's not worth it
@vuetifyjs/core-team thoughts?

KaelWD:
> especially with server side data

More so that you can do something like `items.splice(index, 1)`, you do get `item` in the slot though so I guess an `indexOf` isn't that hard. 

Looking at #10646 and #8199 it seems like they would want what you've done here, but maybe without the pagination?

KaelWD (approved):

PR #12634, opened 2020-11-17
feat(VTextField): add `counter` scoped slot

### LINKED ISSUE #9548: [Feature Request] Add slot for (.v-counter) in inputs like (v-text-field)
### Proposed solution
```vue
<template v-slot:counter="{ current_length, max }">
  <custom-element>
    {{ current_length }}
    <custom-separator></custom-separator>
    {{ max }}
  </custom-element>
</template>
```
> Developers might also use this slot to display some messages in the other corner beside the (**.v-messages**) slot.

Thank you :)

KaelWD:
resolves #9548

## Types of changes

### DIFF
--- a/packages/docs/src/examples/v-text-field/prop-counter.vue
+++ b/packages/docs/src/examples/v-text-field/prop-counter.vue
@@ -35,10 +35,11 @@
         >
           <v-text-field
             v-model="title"
-            :rules="rules"
-            counter="25"
-            filled
-            label="Filled"
+            :rules="wordsRules"
+            counter="5"
+            hint="This field counts words instead of characters"
+            label="Custom counter from prop"
+            :counter-value="v => v.trim().split(' ').length"
           ></v-text-field>
         </v-col>
 
@@ -48,11 +49,15 @@
         >
           <v-text-field
             v-model="title"
-            :rules="rules"
-            counter="25"
-            label="Outlined"
-            outlined
-          ></v-text-field>
+            :rules="wordsRules"
+            counter="5"
+            hint="This field counts words instead of characters"
+            label="Custom counter from slot"
+          >
+            <template v-slot:counter="{ props }">
+              <v-counter v-bind="props" :value="title.trim().split(' ').length"></v-counter>
+            </template>
+          </v-text-field>
         </v-col>
       </v-row>
     </v-container>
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -312,14 +312,14 @@ export default baseMixins.extend<options>().extend({
...
-        },
-      })
+      const props = {
+        dark: this.dark,
+        light: this.light,
+        max,
+        value: this.computedCounterValue,
+      }
+
+      return this.$scopedSlots.counter?.({ props }) ?? this.$createElement(VCounter, { props })
```
johnleider, line 322: Should we use the getSlot helper fn?

johnleider (commented):
Looks good, only one comment regarding process.

packages/docs/src/examples/v-text-field/prop-counter.vue
```diff
@@ -35,10 +35,11 @@
...
-            :rules="rules"
-            counter="25"
-            filled
-            label="Filled"
+            :rules="wordsRules"
+            counter="5"
+            hint="This field counts words instead of characters"
+            label="Custom counter from prop"
+            :counter-value="v => v.trim().split(' ').length"
           ></v-text-field>
```
shamilfrontend, line 43: `/>` ?

packages/docs/src/examples/v-text-field/prop-counter.vue
```diff
@@ -48,11 +49,15 @@
...
-            label="Outlined"
-            outlined
-          ></v-text-field>
+            :rules="wordsRules"
+            counter="5"
+            hint="This field counts words instead of characters"
+            label="Custom counter from slot"
+          >
+            <template v-slot:counter="{ props }">
+              <v-counter v-bind="props" :value="title.trim().split(' ').length"></v-counter>
```
shamilfrontend, line 58: `/>`?

packages/docs/src/examples/v-text-field/prop-counter.vue
Shinigami92:
```suggestion
          />
```

packages/docs/src/examples/v-text-field/prop-counter.vue
Shinigami92:
```suggestion
              <v-counter v-bind="props" :value="title.trim().split(' ').length"/>
```

packages/docs/src/examples/v-text-field/prop-counter.vue
KaelWD: Self-closing tags don't work in DOM templates so we don't use them in examples.

PR #12638, opened 2020-11-17
feat(VCalendar): update parser to return full category

nquinn721:
Update parser to return full category object

## Description
The parser is currently returning {categoryName}. It needed to return full category

## Motivation and Context
During interval styling we want to read the whole category instead of just the name

## How Has This Been Tested?
Wrote a unit test for interval styling to make sure full category is being passed.

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-calendar
      color="primary"
      type="category"
      :categories="['Jeff', 'Nate']"
      category-show-all
      :events="events"
      :interval-style="intervalStyle"
    />
  </v-container>
</template>

<script>
export default {
  data: () => ({
    events: [
      {
        name: "Nate",
        category: "Nate",
        start: new Date(),
        end: new Date()
      }
    ]
  }),
  methods: {
    intervalStyle(obj){
      console.log(obj.category)
    }
  }
};
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VCalendar/__tests__/VCalendarCategory.spec.ts
+++ b/packages/vuetify/src/components/VCalendar/__tests__/VCalendarCategory.spec.ts
@@ -48,7 +48,7 @@ describe('VCalendarCategory', () => {
     expect(wrapper.find('.v-calendar-category__column-header').text()).toEqual('20')
   })
 
-  it('should pass entire cateogry to interal', () => {
+  it('should pass entire cateogry to interval style method', () => {
     function intervalStyle (obj) {
       expect(obj.category).toEqual({ name: 'Nate', age: 20, categoryName: 'Nate' })
     }
--- a/packages/vuetify/src/components/VCalendar/__tests__/VCalendarCategory.spec.ts
+++ b/packages/vuetify/src/components/VCalendar/__tests__/VCalendarCategory.spec.ts
@@ -24,7 +24,7 @@ describe('VCalendarCategory', () => {
     }
   })
 
-  it('should test categoryText prop as a string', async () => {
+  it('should test categoryText prop as a string', () => {
     const wrapper = mountFunction({
       propsData: {
         categories: [{ name: 'Nate' }],
@@ -35,7 +35,7 @@ describe('VCalendarCategory', () => {
     expect(wrapper.find('.v-calendar-category__column-header').text()).toEqual('Nate')
   })
 
-  it('should test categoryText prop as a function', async () => {
+  it('should test categoryText prop as a function', () => {
     const wrapper = mountFunction({
       propsData: {
         categories: [{ name: 'Nate', age: 20 }],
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VCalendar/__tests__/VCalendarCategory.spec.ts
```diff
@@ -47,4 +47,18 @@ describe('VCalendarCategory', () => {
 
     expect(wrapper.find('.v-calendar-category__column-header').text()).toEqual('20')
   })
+
+  it('should pass entire cateogry to interal', () => {
```
nekosaur, line 51:
```suggestion
  it('should pass entire category to internal', () => {
```

packages/vuetify/src/components/VCalendar/util/parser.ts
```diff
@@ -21,7 +21,7 @@ export function getParsedCategories (
       const categoryName = typeof v === 'string' ? v
         : typeof v === 'object' && v && typeof v.categoryName === 'string' ? v.categoryName
           : parsedCategoryText(v, categoryText)
-      return { categoryName }
+      return Object.assign({ categoryName }, v)
```
nekosaur, line 24: Unless I'm mistaken, `v` here is sometimes a string and sometimes an object?

nekosaur (changes requested):

packages/vuetify/src/components/VCalendar/util/parser.ts
nquinn721: Yes, object.assign will just add each letter as indexed values, but instead of mutating the object depending on the type of v and adding several lines of code, it's just simpler this way. If v is a string, we only need categoryName, otherwise we want all the props of v on the object.

nekosaur: I would prefer to be verbose and correct here. You could for example return early if v (bad variable name btw, please use `category` or something similar) is a string, and be left with an object

packages/vuetify/src/components/VCalendar/util/parser.ts
```diff
@@ -17,11 +17,12 @@ export function getParsedCategories (
...
-      const categoryName = typeof v === 'string' ? v
-        : typeof v === 'object' && v && typeof v.categoryName === 'string' ? v.categoryName
-          : parsedCategoryText(v, categoryText)
-      return { categoryName }
+    return categories.map((category: CalendarCategory) => {
+      const categoryName = typeof category === 'string' ? category
+        : typeof category === 'object' && category && typeof category.categoryName === 'string' ? category.categoryName
+          : parsedCategoryText(category, categoryText)
+      if (typeof category === 'string') return { categoryName }
+      return Object.assign({ categoryName }, category)
```
nekosaur, line 25:
```suggestion
      if (typeof category === 'string') return { categoryName: category }
      
      const categoryName = typeof category.categoryName === 'string' ? category.categoryName : parsedCategoryText(category, categoryText)

      return { ...category, categoryName }
```

nekosaur (approved):

Issue #11253, opened 2020-04-28
[Bug Report] Title attribute is not working in v-select


chans-me:
### Environment
**Vuetify Version:** 2.2.25
**Vue Version:** 2.6.11
**Browsers:** Chrome 78.0.3904.97
**OS:** Linux x86_64

### Steps to reproduce
The normal behaviour of title attribute is to display the title text on hover , similar to v-text-field. But in v-select it is not working, please check the codepen

### Expected Behavior
On hover select box, title message should display

### Actual Behavior
it is not showing the title message

### Reproduction Link
<a href="https://codepen.io/chansv/pen/ExVXEQJ?editors=1000" target="_blank">https://codepen.io/chansv/pen/ExVXEQJ?editors=1000</a>

KaelWD:
v-select passes all attributes down to the `<input>` element like v-text-field does (#12362), but it has `pointer-events: none` to prevent you from typing in it.

PR #12776, opened 2020-12-12
fix(inputs): display title attribute on root element

### LINKED ISSUE #12775: [Bug Report] Title attribute not showing on label hover, and in some components at all
### Environment
**Vuetify Version:** 2.3.21
**Vue Version:** 2.6.11
**Browsers:** Firefox 83.0
**OS:** Windows 10

### Steps to reproduce
Try hovering on components like [`v-text-field`, `v-textarea`, `v-switch`, `v-checkbox`, `v-radio`, `v-select`, `v-combobox`, `v-autocomplete`] with [`title` and `label`] props, and watch if the title tooltip appears. It doesn't work in these places:

image

### Expected Behavior
* When we set up a `title` on `v-text-field` with a `label`, I assume the most logical use-case is for the user to be able to intuitively hover on the label (gaining accessibility), considering that Vuetify makes the ending disappear when there isn't enough space (white I'd personally also vote to allow overriding, as we may prefer to use vertical space to display all of it); but also in the scenarios when we want a custom title.
* Of course, the title is expected to also work properly in all other components.

### Actual Behavior
* Not only doesn't `v-text-field`'s `label` support hover (once it moves away from the input), it actually moves position outside the element. When outlined prop is used, it stays within the element, however the hover doesn't trigger on those "boundaries" either.
* Another issues appear in `v-switch` and `v-radio`, in total a combination of about three issues: not displaying at all, not displaying for all elements under parent, and being blocked by a ripple element.
* It doesn't work for `v-select` at all, and has issues in several other related components.

### Reproduction Link
<a href="https://codepen.io/cTAlENtATI/pen/KKgagzB" target="_blank">https://codepen.io/cTAlENtATI/pen/KKgagzB</a>

### Other comments
Don't worry, **I've already fixed this** and sent a PR #12776 to dev branch, so please help me test the stability.

There's a related issue for v-select: https://github.com/vuetifyjs/vuetify/issues/11253 where the `pointer-events: none` implementation was described as a workaround to achieve expected behavior. In my implementation I enabled `pointer-events: auto` on active label, but there may be a better solution; I was unsuccessful in trying to use something like `& :not(&--active)` or `& :not(.v-label--active)` with `pointer-events: none` instead, intending to avoid setting `auto` as an override in case the developer explicitly changes it in a parent component, so you may also reconsider this.

scscgit:
Display hover effect of title attr even on label of v-text-field, which may have moved outside (above) the input when not outlined, by enabling pointer-events only when it's active (not otherwise).

Fix another v-switch bug of ripple effect's event blocking hover.

Fix displaying title at all in v-radio. (Also fixed ripple here.)

Move title attr to a parent object, making it trigger even on the threshold of input boundaries, increasing its accessibility scope to avoid blind zones. Remove the duplicate title in v-text-field, v-radio, v-switch, and v-checkbox after moving to parent element; specifically VInput, or in the v-radio's case via its own parent. Modified VInput's genInputSlot(): it's being called by components VAutocomplete, VRadioGroup, VSelect, and VTextField (maybe more).

This also fixed v-select's issue of title not appearing.

## Description

Fixes #12775, #11253

## Motivation and Context

Title issues, details are in the issue #12775.

## Implementation details

There's a related issue for v-select: #11253 where the `pointer-events: none` implementation was described as a workaround to achieve expected behavior. In my implementation I enabled `pointer-events: auto` on active label, but there may be a better solution; I was unsuccessful in trying to use something like `& :not(&--active)` or `& :not(.v-label--active)` with `pointer-events: none` instead, intending to avoid setting `auto` as an override in case the developer explicitly changes it in a parent component, so you may also reconsider this.

The title attribute was moved to the parent, and then removed from child elements by copying the `{ ...this.attrs$ }` and `delete`-ing `title` field. This shouldn't be necessary, as the duplicate doesn't cause any harm, but will produce cleaner output.

## How Has This Been Tested?

visually, but without exhaustive testing, so I'd appreciate any help.

## Markup:

<details>

```vue
template>
  <v-container>
    <v-text-field title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-text-field>
    <v-text-field outlined title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-text-field>
    <v-textarea title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-textarea>
    <v-switch title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-switch>
    <v-checkbox title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-checkbox>
    <v-radio title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-radio>
    <v-select title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-select>
    <v-combobox title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-combobox>
    <v-autocomplete title="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details" label="A very long label that doesn't fit on the screen, so the user may have to hover on the label to read more details"></v-autocomplete>
  </v-container>
</template>
```
</details>

## Types of changes

## Checklist:

* This may introduce potential issues and isn't a fix for a critical functionality, so I'd conservatively classify it as a new feature (of label tooltips) and submit to dev first

### DIFF
--- a/packages/vuetify/src/components/VInput/VInput.ts
+++ b/packages/vuetify/src/components/VInput/VInput.ts
@@ -154,6 +154,9 @@ export default baseMixins.extend<options>().extend({
     genControl () {
       return this.$createElement('div', {
         staticClass: 'v-input__control',
+        attrs: {
+          title: this.attrs$.title,
+        },
       }, [
         this.genInputSlot(),
         this.genMessages(),
--- a/packages/vuetify/src/components/VInput/VInput.ts
+++ b/packages/vuetify/src/components/VInput/VInput.ts
@@ -154,9 +154,7 @@ export default baseMixins.extend<options>().extend({
     genControl () {
       return this.$createElement('div', {
         staticClass: 'v-input__control',
-        attrs: {
-          title: this.attrs$.title,
-        },
+        attrs: { title: this.attrs$.title },
       }, [
         this.genInputSlot(),
         this.genMessages(),
--- a/packages/vuetify/src/components/VRadioGroup/VRadio.ts
+++ b/packages/vuetify/src/components/VRadioGroup/VRadio.ts
@@ -140,6 +140,9 @@ export default baseMixins.extend<options>().extend({
       }, getSlot(this, 'label') || this.label)
     },
     genRadio () {
+      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
+      const radioAttrs = { ...this.attrs$ }
+      delete radioAttrs.title
       return this.$createElement('div', {
         staticClass: 'v-input--selection-controls__input',
       }, [
@@ -151,7 +154,7 @@ export default baseMixins.extend<options>().extend({
         this.genInput({
           name: this.computedName,
           value: this.value,
-          ...this.attrs$,
+          ...radioAttrs,
         }),
         this.genRipple(this.setTextColor(this.rippleState)),
       ])
@@ -179,6 +182,9 @@ export default baseMixins.extend<options>().extend({
       on: mergeListeners({
         click: this.onChange,
       }, this.listeners$),
+      attrs: {
+        title: this.attrs$.title,
+      },
     }
 
     return h('div', data, [
--- a/packages/vuetify/src/components/VRadioGroup/VRadio.ts
+++ b/packages/vuetify/src/components/VRadioGroup/VRadio.ts
@@ -140,9 +140,8 @@ export default baseMixins.extend<options>().extend({
       }, getSlot(this, 'label') || this.label)
     },
     genRadio () {
-      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
-      const radioAttrs = { ...this.attrs$ }
-      delete radioAttrs.title
+      const { title, ...radioAttrs } = this.attrs$
+
       return this.$createElement('div', {
         staticClass: 'v-input--selection-controls__input',
       }, [
@@ -182,9 +181,7 @@ export default baseMixins.extend<options>().extend({
       on: mergeListeners({
         click: this.onChange,
       }, this.listeners$),
-      attrs: {
-        title: this.attrs$.title,
-      },
+      attrs: { title: this.attrs$.title },
     }
 
     return h('div', data, [
--- a/packages/vuetify/src/components/VSwitch/VSwitch.ts
+++ b/packages/vuetify/src/components/VSwitch/VSwitch.ts
@@ -77,9 +77,8 @@ export default Selectable.extend({
       ]
     },
     genSwitch (): VNode {
-      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
-      const switchAttrs = { ...this.attrs$ }
-      delete switchAttrs.title
+      const { title, ...switchAttrs } = this.attrs$
+
       return this.$createElement('div', {
         staticClass: 'v-input--selection-controls__input',
       }, [
--- a/packages/vuetify/src/components/VSwitch/VSwitch.ts
+++ b/packages/vuetify/src/components/VSwitch/VSwitch.ts
@@ -77,12 +77,15 @@ export default Selectable.extend({
       ]
     },
     genSwitch (): VNode {
+      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
+      const switchAttrs = { ...this.attrs$ }
+      delete switchAttrs.title
       return this.$createElement('div', {
         staticClass: 'v-input--selection-controls__input',
       }, [
         this.genInput('checkbox', {
           ...this.attrs,
-          ...this.attrs$,
+          ...switchAttrs,
         }),
         this.genRipple(this.setTextColor(this.validationState, {
           directives: [{
--- a/packages/vuetify/src/components/VTextField/VTextField.ts
+++ b/packages/vuetify/src/components/VTextField/VTextField.ts
@@ -376,9 +376,7 @@ export default baseMixins.extend<options>().extend({
     genInput () {
       const listeners = Object.assign({}, this.listeners$)
       delete listeners.change // Change should not be bound externally
-      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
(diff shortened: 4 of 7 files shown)


johnleider (changes requested):
Thank you for your contribution. I think this is a good start but there is a lot of duplication between the components that I'm sure can be DRYed up.

scscgit:
@johnleider As I'm not too knowledged about the Vuetify's component/mixin hierarchy, do you have any recommendation about the placements of those title attr exclusions?
* Also I think someone should make the decision whether moving the label to parent is the best solution (or alternatively using it in all children, which could even be parametrized, making label-only or input-only attributes possible; btw. I think in one case I've already had to place the title on both child components due to a lack of parent).
* On the other hand, the `:title.prop="'text'"` workaround that I've been recommended in the issue seems to make most of these changes redundant; but only as long as having title on parent is the expected behavior, which removes the option of only using it on a child. It'd be probably better to just take the input-specific CSS (pointer-events) fixes and then either add a new option of putting title on label, or just recommend the .prop within docs for new users.

packages/vuetify/src/components/VRadioGroup/VRadio.ts
```diff
@@ -140,6 +140,9 @@ export default baseMixins.extend<options>().extend({
       }, getSlot(this, 'label') || this.label)
     },
     genRadio () {
+      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
+      const radioAttrs = { ...this.attrs$ }
```
johnleider, line 144:
```suggestion
      const { title, ...radioAttrs } = this.attrs$
```

packages/vuetify/src/components/VRadioGroup/VRadio.ts
```diff
@@ -140,6 +140,9 @@ export default baseMixins.extend<options>().extend({
...
+      delete radioAttrs.title
```
johnleider, line 145:
```suggestion

```

packages/vuetify/src/components/VInput/VInput.ts
```diff
@@ -154,6 +154,9 @@ export default baseMixins.extend<options>().extend({
     genControl () {
       return this.$createElement('div', {
         staticClass: 'v-input__control',
+        attrs: {
+          title: this.attrs$.title,
+        },
```
johnleider, line 159:
```suggestion
        attrs: {  title: this.attrs$.title  },
```

packages/vuetify/src/components/VRadioGroup/VRadio.ts
```diff
@@ -179,6 +182,9 @@ export default baseMixins.extend<options>().extend({
       on: mergeListeners({
         click: this.onChange,
       }, this.listeners$),
+      attrs: {
+        title: this.attrs$.title,
+      },
```
johnleider, line 187:
```suggestion
      attrs: {  title: this.attrs$.title  },
```

packages/vuetify/src/components/VSwitch/VSwitch.ts
```diff
@@ -77,12 +77,15 @@ export default Selectable.extend({
       ]
     },
     genSwitch (): VNode {
+      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
+      const switchAttrs = { ...this.attrs$ }
+      delete switchAttrs.title
```
johnleider, line 82:
```suggestion
      const { title, ...radioAttrs } = this.attrs$

```

packages/vuetify/src/components/VTextField/VTextField.ts
```diff
@@ -376,14 +376,17 @@ export default baseMixins.extend<options>().extend({
     genInput () {
       const listeners = Object.assign({}, this.listeners$)
       delete listeners.change // Change should not be bound externally
+      // title attr is used on parent directly, so that its hover effect includes VLabel; this avoids a duplicate
+      const inputAttrs = { ...this.attrs$ }
+      delete inputAttrs.title
```
johnleider, line 381:
```suggestion
      const { title, ...inputAttrs } = this.attrs$

```

packages/vuetify/src/styles/components/_selection-controls.sass
```diff
@@ -76,6 +76,7 @@
     left: -12px
     top: calc(50% - 24px)
     margin: 7px
+    pointer-events: none // Enable title attr hover effect, as ripple blocks the input
```
johnleider, line 79: Have you tested this in all of our supported browsers?

johnleider (changes requested):

packages/vuetify/src/styles/components/_selection-controls.sass
scscgit: Actually now that you mentioned it, I've noticed that it causes a new bug of ripple not disappearing (probably on `mouseout`); but because we've already moved the title to the parent `v-input`, the ripple already inherits the title hover effect, so this workaround isn't necessary anymore, and I've removed it in a new commit.

johnleider (changes requested):
Is there a specific reason for not having a separate computed property that reduces the repeating of process.

scscgit:
@johnleider if you mean the process of removing `title` from `this.attrs$`, its used only once per component. Would you prefer moving each of them to a separate computed property, or is there some viable place for a mixin inheritance where I should move them?

KaelWD (approved):

Issue #12692, opened 2020-11-28
[Task] v-sheet


johnleider:
# Overview

- Related Tasks: ~~#12789~~
- Related Issues: N/A
- Related Pull Requests: ~~#12542~~

A conversion should not start until *all* assigned issues with the milestone v2.2.x are resolved and in the `next` branch. Reference our [Coding Guidelines](https://github.com/vuetifyjs/vuetify/wiki/coding-guidelines) for information regarding the team's practices or reach out to us on [Discord](https://community.vuetifyjs.com).

renoirb:
✋🏾 ,

Branch pushed?

renoirb:
As I understand it, a `v-sheet` is basically a 2D shape we might want to add border radius in its corners, and an elevation and drop shadow.

Since there's something in the works for border-radius (roundable) there might be something missing for elevation?

johnleider:
Yes, this is on a separate branch, but I went on vacation. Will follow-up soon.

PR #12801, opened 2020-12-17
feat(border-radius) add test refactor with type guards

renoirb:
Related to vuetifyjs/vuetify#12692 vuetifyjs/vuetify#12789

## Description

Moved tests, extracted logic into stateless functions, an attempt at making things explicit. (Maybe be too verbose now though).

* Added border-radius test case copy-pasta from rounded mixin

## Motivation and Context

* Introducing a way to be more specific with what's acceptable as input w.r.t. typing
* Separate computation from Vue, test without Vue, then mount into appropriate effect
 
## How Has This Been Tested?

* Separated computation tested without runtime (Vue), then in integration

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/__tests__/border-radius.spec.ts
+++ b/packages/vuetify/src/composables/__tests__/border-radius.spec.ts
@@ -1,0 +1,49 @@
+import {
+  useBorderRadius,
+  borderRadiusClassNames,
+} from '../border-radius'
+import type {
+  BorderRadiusProps,
+} from '../border-radius'
+
+describe('border-radius', () => {
+  const assertions = [
+    // Rounded with 0
+    [{ rounded: false }, ['rounded-0']],
+    [{ rounded: '0' }, ['rounded-0']],
+    [{ rounded: 0 }, ['rounded-0']],
+    [{ rounded: 'tile' }, ['rounded-0']],
+    // Rounded with a word
+    [{ rounded: 'circle' }, ['rounded-circle']],
+    [{ rounded: 'shaped' }, ['rounded-shaped']],
+    [{ rounded: 'pill' }, ['rounded-pill']],
+    // Rounded only
+    [{ rounded: true }, ['rounded']],
+    [{ rounded: '' }, ['rounded']],
+    // When should return nothing
+    [{ rounded: 1 }, []],
+    [{ rounded: { toString: () => 'pill' } }, []],
+    [undefined, []],
+    [{ rounded: null }, []],
+    // Rounded with more than one words, ensure sorting
+    [
+      { rounded: 'tr-xl br-lg' },
+      ['rounded-br-lg', 'rounded-tr-xl'],
+    ],
+    [{ rounded: 'sm-3 4' }, ['rounded-4', 'rounded-sm-3']],
+    // All strings are accepted?
+    [{ rounded: 'foo-bar-bazz-buzz' }, ['rounded-foo-bar-bazz-buzz']],
+    [{ rounded: '!' }, ['rounded-!']],
+  ] as [BorderRadiusProps, string[]][]
+
+  it.each(assertions)(
+    'should return correct rounded classes',
+    (propsData, expected) => {
+      const vm = useBorderRadius(propsData)
+      const subject = borderRadiusClassNames(propsData)
+
+      expect(vm.borderRadiusClasses.value).toMatchObject(expected)
+      expect(subject).toMatchObject(expected)
+    },
+  )
+})
--- a/packages/vuetify/src/composables/border-radius.ts
+++ b/packages/vuetify/src/composables/border-radius.ts
@@ -1,63 +1,23 @@
-// Utilities
 import { computed } from 'vue'
-import type { Prop } from 'vue'
-
-import propsFactory from '@/util/propsFactory'
 
 // Types
-export type IRounded = typeof radius[number]
+import type { Prop } from 'vue'
+import type { IRounded } from './rounded'
+
+// Utils
+import { roundedClassNames, isRounded } from './rounded'
+import propsFactory from '@/util/propsFactory'
+
 export interface BorderRadiusProps {
   rounded?: IRounded
 }
 
-const radius = [
-  true,
-  false,
-  '',
-  0,
-  '0',
-  'xs',
-  'sm',
-  'lg',
-  'xl',
-  'circle',
-  'pill',
-  'shaped',
-  'tile',
-] as const
-
 export const makeBorderRadiusProps = propsFactory({
   rounded: {
     type: [Boolean, Number, String],
-    validator: (v: any) => radius.includes(v),
+    validator: isRounded,
   } as Prop<IRounded>,
 })
-
-export const roundedClassNames = (
-  props: BorderRadiusProps,
-): string[] => {
-  const classes = []
-
-  if (props && 'rounded' in props) {
-    const rounded = props.rounded
-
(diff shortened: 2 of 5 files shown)


renoirb:
Hi @johnleider here are a few things I've done today

renoirb:
Adding @KaelWD, because I wanna chat strict typing :)

nekosaur:
This seems extremely verbose. What problem is this trying to solve exactly?

renoirb:
> This seems extremely verbose. What problem is this trying to solve exactly?

This is an illustration (not necessarily part of that specific task).

In the case of one usage (`borderRadiusClasses`) is making it look ridiculous. Indeed.

It's meant as an illustration of how we could structure code when it gets a bit more than here. 

But the moment we have 2+ things that has more than what's shown as `if(...) {} else if(...) {} else if(...) {}` for more than 2 times, that could be a pattern.

Also, notice the `ClassNameTransformer`, we have a lot of places where we have to pass and flip on/off classNames, and how to dump them into render functions.  In my past projects I'd have such helpers.  I've added one here as part of my proposal.

nekosaur:
Could you show some specific scenario where this _would_ be useful then? I still don't understand what problem this solves.

As to `ClassNameTransformer`, unless I'm misunderstanding what it's for, it's not needed. VNode `class` data can be any combination of `{},'str',[]`, it will be merged

packages/vuetify/src/composables/border-radius.ts
```diff
@@ -26,14 +29,15 @@ export const makeBorderRadiusProps = propsFactory({
...
 })
 
-// Effect
-export function useBorderRadiusClasses (props: BorderRadiusProps) {
-  const borderRadiusClasses = computed(() => {
+export const borderRadiusClassNames = (
+  props: BorderRadiusProps,
+): string[] => {
+  const classes = []
+  if (props && 'rounded' in props) {
```
johnleider, line 39:
```suggestion

  if (props && 'rounded' in props) {
```

packages/vuetify/src/composables/border-radius.ts
```diff
@@ -26,14 +29,15 @@ export const makeBorderRadiusProps = propsFactory({
...
     type: [Boolean, Number, String],
     validator: (v: any) => radius.includes(v),
-  },
+  } as Prop<IRounded>,
 })
 
-// Effect
-export function useBorderRadiusClasses (props: BorderRadiusProps) {
-  const borderRadiusClasses = computed(() => {
+export const borderRadiusClassNames = (
```
johnleider, line 35: I don't think this needs to be a separate function. Can just move into the code block on line 62

packages/vuetify/src/composables/__tests__/border-radius.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    [{ rounded: 'foo-bar-bazz-buzz' }, ['rounded-foo-bar-bazz-buzz']],
+    [{ rounded: '!' }, ['rounded-!']],
+  ] as [BorderRadiusProps, string[]][]
+
+  it.each(assertions)(
+    'should return correct rounded classes',
+    (propsData, expected) => {
+      const vm = useBorderRadius(propsData)
+      expect(vm.borderRadiusClasses.value).toMatchObject(expected)
+      const subject = borderRadiusClassNames(propsData)
```
johnleider, line 44:
```suggestion

      const subject = borderRadiusClassNames(propsData)
```

johnleider (changes requested):

packages/vuetify/src/composables/border-radius.ts
renoirb: Sure.

But I wanted to point out here that it's always cheaper to run test without the framework.  In Vue, and Vue 3 it's cheap. But if we introduce this here, later on, if things gets heavier, we won't have to change the whole code base.

This is from a person who was in an heavy code base where every test were also "testing the framework".

packages/vuetify/src/composables/border-radius.ts
```diff
@@ -1,15 +1,18 @@
 // Utilities
 import { computed } from 'vue'
+import type { Prop } from 'vue'
```
johnleider, line 3: move to line 7

packages/vuetify/src/composables/border-radius.ts
```diff
@@ -1,15 +1,18 @@
...
 import propsFactory from '@/util/propsFactory'
 
 // Types
+export type IRounded = typeof radius[number]
 export interface BorderRadiusProps {
-  rounded?: boolean | number | string
+  rounded?: IRounded
```
renoirb, line 9: I'm using `IFoo` because Microsoft's ESLint work recommends using this naming, and I gotten used doing it.

packages/vuetify/src/composables/border-radius.ts
renoirb: K

renoirb: Wait, are you sure you want to mix import from exports?

Also, if we make a separation from what we're importing from public (vue), from internal package, from what we're exporting.  This won't scale well.

packages/vuetify/src/composables/border-radius.ts
```diff
@@ -1,15 +1,18 @@
 // Utilities
 import { computed } from 'vue'
+import type { Prop } from 'vue'
 import propsFactory from '@/util/propsFactory'
 
 // Types
+export type IRounded = typeof radius[number]
```
renoirb, line 7: This line is based on the const a few lines below, that export type would be best AFTER not before, like it is at the moment.

packages/vuetify/src/composables/border-radius.ts
jacekkarczmarczyk: there's an eslint rule for new lines after const/let

renoirb: OK

packages/vuetify/src/composables/rounded.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+  }
+
+  return classes.sort()
+}
+
+export const isRounded = (input: any): input is IRounded => {
+  return ROUNDED.includes(input)
+}
```
renoirb, line 50: Notice the complete need for using Vue for testing.

renoirb:
Related to #12886 12886

renoirb:
Maybe this should be an RFC?

Quoting what I’ve said on Discord

---


1. Have logic written without dependency, only strictly and [typed pure functions][PUREFN],  Then, mount  Vue.js [later for easier testing][LATEMOUNTING]
2. Leverage TypeScript and typing (e.g. based on [array of possible values][ARRAYBASEDTYPE]),  [Leveraging Node.js’s assert API][LEVERAGINGASSERT] and [User Defined Guards][USERDEFNGUARDS]

[PUREFN]: https://github.com/vuetifyjs/vuetify/pull/12801/files#r553057616
[LATEMOUNTING]: https://github.com/vuetifyjs/vuetify/pull/12886#discussion_r553057336
[LEVERAGINGASSERT]: https://github.com/vuetifyjs/vuetify/pull/12801#issuecomment-748156253
[ARRAYBASEDTYPE]: https://github.com/vuetifyjs/vuetify/pull/12801#discussion_r546802674
[USERDEFNGUARDS]: https://github.com/vuetifyjs/vuetify/pull/12801

PR #12889, opened 2021-01-04
refactor(dimensions): update code to unify process

johnleider:
## Description
Updated naming convention and process for composable and its unit tests to bring it more in line with current standards.

## Motivation and Context
Working on #12692 and ran into consistency issues with how we handle classes and styles.

## How Has This Been Tested?
jest

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/dimensions.ts
+++ b/packages/vuetify/src/composables/dimensions.ts
@@ -1,63 +1,42 @@
+// Utilities
 import { computed } from 'vue'
-
-// Types
-import type { Prop } from 'vue'
-
-// Utils
-import { convertToUnit, keys } from '@/util/helpers'
+import { convertToUnit } from '@/util/helpers'
 import propsFactory from '@/util/propsFactory'
 
-// Props
-const allDimensionsProps = {
-  height: {
-    type: [Number, String],
-  },
-  maxHeight: {
-    type: [Number, String],
-  },
-  maxWidth: {
-    type: [Number, String],
-  },
-  minHeight: {
-    type: [Number, String],
-  },
-  minWidth: {
-    type: [Number, String],
-  },
-  width: {
-    type: [Number, String],
-  },
+// Types
+export interface DimensionProps {
+  height?: number | string
+  maxHeight?: number | string
+  maxWidth?: number | string
+  minHeight?: number | string
+  minWidth?: number | string
+  width?: number | string
 }
 
-// Types
-type PropValue = string | number | null | undefined
-type PropNames = keyof typeof allDimensionsProps
+export type DimensionKey = keyof DimensionProps
+export type DimensionStyles = Partial<Record<DimensionKey, string>>
 
-// Effect
-export function dimensionsFactory<S extends PropNames> (...possibleProps: S[]) {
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/composables/dimensions.ts
```diff
@@ -1,63 +1,42 @@
...
-    return { dimensions }
-  }
-
-  return {
-    makeDimensionsProps,
-    useDimensions,
-  }
+export function useDimensionStyles (props: DimensionProps) {
+  const dimensionStyles = computed(() => {
+    const properties = Object.keys(props) as DimensionKey[]
```
KaelWD, line 30: There's no filter on `props` now, it will convert *all* of the component's props to units.

KaelWD: Oh and there's a typed `keys` function in util/helpers that doesn't need type casting.

packages/vuetify/src/composables/dimensions.ts
```diff
@@ -1,63 +1,50 @@
...
+export type DimensionStyles = Partial<Record<DimensionKey, string>>
+
+export interface DimensionProps {
+  height?: number | string
+  maxHeight?: number | string
+  maxWidth?: number | string
+  minHeight?: number | string
+  minWidth?: number | string
+  width?: number | string
+}
```
renoirb, line 17: Yeah, it's a bit manual, but the opposite would be pedantic

Screen Shot 2021-01-06 at 11 08 22 PM

packages/vuetify/src/composables/dimensions.ts
```diff
@@ -1,63 +1,50 @@
...
 
-  return {
-    makeDimensionsProps,
-    useDimensions,
-  }
+// Composables
+export const makeDimensionProps = propsFactory({
```
renoirb, line 20:
```patch
- export const makeDimensionProps = propsFactory({
+ export const makeDimensionProps = propsFactory<PropOptions<DimensionPropNumberString>>({
```

KaelWD: > Otherwise each prop seem to infer into <any>

Make sure you've run `yarn`, `propsFactory` already infers types correctly without manual annotation.

packages/vuetify/src/composables/dimensions.ts
```diff
@@ -1,63 +1,50 @@
...
-          obj.style[key] = convertToUnit(value)
-        }
-        return obj
-      }, { style: {} } as { style: Record<S, string | undefined> })
-    })
-
-    return { dimensions }
-  }
+export type DimensionKey = keyof DimensionProps
+export type DimensionStyles = Partial<Record<DimensionKey, string>>
```
KaelWD, line 8: Not everything needs to be exported.

packages/vuetify/src/composables/dimensions.ts
renoirb: I know, it was rhetorical

nekosaur (commented):
This is technically not a refactor, since it changes behaviour. The main reason for the factory function was to enable components to specify exactly which props they use. Not all components might use all of the props.

PR #12943, opened 2021-01-11
feat: update VResponsive and VImg to Vue 3

### LINKED ISSUE #4888: [Feature Requestt] Add error slot to v-img
### Problem to solve
v-progress-circular keeps on loading the loader when v-img src getting failed. 
v-progress-circular should dismiss either success or error

### Proposed solution
success slot - dismissing loader (Works fine)
error slot - will replace placeholder for error message or whatever

### LINKED ISSUE #5876: [Feature Request] v-img fallback support
### Problem to solve
As the web progresses forward, new standards of image compression arise. Since not all browsers support the latest formats, we need an effective way to provide a fallback format when the passed media cannot be rendered.

**Currently `v-img` will not render the `background-image` if the provided image format is not supported.**

Ex. This example renders in Chrome 70, but not in Firefox 63. 

```
<v-img src="img/awesomeWebPImage.webp" />
```

Before using Vuetify I solved this problem with the HTML5 `<picture>` element. I would love to do the same with `v-img`!
```
<picture>
  <source srcset="img/awesomeWebPImage.webp" type="image/webp">
  <source srcset="img/creakyOldJPEG.jpg" type="image/jpeg"> 
  <img src="img/creakyOldJPEG.jpg" alt="Alt Text!">
</picture>
```

### Proposed solution
As a developer using `v-img`, the easiest approach for myself is allowing `src` prop to contain an array of image sources. When an array of sources is passed the browser could fallback in the order of sources provided.

Therefore the snippet from above would become:
```
<v-img :src="['img/awesomeWebPImage.webp', 'img/creakyOldJPEG.jpg']" />
```

This would be the resulting css:
```
.img {
  background-image: image("img/awesomeWebPImage.webp", "img/creakyOldJPEG.jpg")
}
```
Source: https://codepen.io/sandip_rb/pen/ZWNEwr

### LINKED ISSUE #6633: [Bug Report] V-Img doesn't re-show lazy-src when src changes.
### Versions and Environment
**Vuetify:** 1.5.3
**Vue:** 2.6.7
**Browsers:** Chrome 72.0.3626.109
**OS:** Mac OS 10.14.3

### Steps to reproduce
When dynamically changing the src of a v-img it seems like the lazy-src is not rendered while the new img src is loading. 

### Expected Behavior
When the v-img :src value is changed, the lazy-src shows until the new img is loaded.

### Actual Behavior
When the v-img :src value is changed, the new img is loaded and (if it is slow) you see the img drawn as the network loads it (top to bottom). 

### Reproduction Link
<a href="https://codepen.io/DanDC/pen/gEpZBG" target="_blank">https://codepen.io/DanDC/pen/gEpZBG</a>

### Other comments

This is more clear on a local site where the images are remote and you slow down network speeds. On the code pen above I don't think I see a flash of the lazy-src but it is hard to be certain with a fast network. When I test on my local machine and degrade the network it is much more obvious.

If there is a better way to verify if the lazy-src is shown or not let me know and I can edit the codepen.

### LINKED ISSUE #7017: [Bug Report] Image component blink after loading on Safari
### Versions and Environment
**Vuetify:** 1.5.12
**Vue:** 2.6.10
**Browsers:** Safari, Other
**OS:** Mac OS 10.14.3

### Steps to reproduce
Load a `v-img` component from Safari. 

I have used an example with multiple images from the doc, but it happens as well with a single image.

### Expected Behavior
To show the image with no blinking.

### Actual Behavior
Image is loading, then blink (the image disappear for a few ms then re-appear).

### Reproduction Link
<a href="https://codepen.io/thomaskientz/pen/eoMNeX" target="_blank">https://codepen.io/thomaskientz/pen/eoMNeX</a>

### Other comments
Video of the blinking :

https://www.dropbox.com/s/jpm63bjedjdhgvp/vuetify%20image%20blink%20issue%20safari.mov?dl=0

### LINKED ISSUE #12425: [Bug Report] [Safari] v-img with svg files results in a warning. [Vuetify] Failed to decode image, trying to render anyway. Original error: Invalid image type.
### Environment
**Vuetify Version:** 2.3.10
**Vue Version:** 2.6.12
**Browsers:** Safari
**OS:** iOS

### Steps to reproduce
It is very easy to reproduce this issue, just set the svg file as the source for v-img and open it in Safary on your iPhone.

### Expected Behavior
To display image

### Actual Behavior
The image is displayed but a warning is displayed in the console.

### Reproduction Link
<a href="https://codepen.io/alekstar79/pen/BazKKqa?editors=1010" target="_blank">https://codepen.io/alekstar79/pen/BazKKqa?editors=1010</a>

### Other comments
This issue already existed [Issues #4969](https://github.com/vuetifyjs/vuetify/issues/4969) but for some reason it was closed as resolved.

KaelWD:
- No more keyed elements, should fix #7017
- Show `lazy-src` again when `src` changes, resolves #6633
- Removed decode failure warning, resolves #12425
- Added error slot, resolves #4888
- Added `loadstart` event
- Use `` and `object-fit` instead of `background-image`
- Use `<picture>` when the `sources` slot is provided, resolves #5876

BREAKING CHANGE: Removed `contain`, it's now the default. Added `cover`.
BREAKING CHANGE: Reordered `v-intersect` arguments, `isIntersecting` is now the first arg
BREAKING CHANGE: Changed `v-img` class names
BREAKING CHANGE: No longer using `background-image`

## Markup:

<details>

```vue
<template>
  <v-app class="v-application--is-ltr">
    <v-container>
      <v-img
        ref="img"
        src="https://picsum.photos/id/11/500/300"
        lazy-src="https://picsum.photos/id/11/10/6"
        width="500"
        height="300"
      >
        <template #sources>
          <source srcset="https://picsum.photos/id/11/500/300.webp">
        </template>
      </v-img>

      <v-img
        src="https://picsum.photos/id/11/500/300"
        lazy-src="https://picsum.photos/id/11/10/6"
        width="500"
        height="300"
      >
        Content
      </v-img>

      <v-img
        src="https://picsum.photos/id/11/500/300"
        lazy-src="https://picsum.photos/id/11/10/6"
        width="500"
        height="300"
      />

      <v-img
        src="https://bad.src/not/valid"
        lazy-src="https://picsum.photos/id/11/100/60"
        max-width="500"
        max-height="300"
      >
        <template #placeholder>
          <v-row
            class="fill-height ma-0"
            align="center"
            justify="center"
          >
            Loading...
          </v-row>
        </template>
      </v-img>

      <v-img
        max-width="800"
        src="https://picsum.photos/id/11/500/300"
        lazy-src="https://picsum.photos/id/11/10/6"
      />
    </v-container>
  </v-app>
</template>
```
</details>


packages/vuetify/src/directives/intersect/index.ts
```diff
@@ -5,12 +5,12 @@ import type {
 } from 'vue'
 
 type ObserveHandler = (
+  isIntersecting: boolean,
```
KaelWD, line 8: Not *really* related but the original order of this makes no sense, most of the time you only care about `isIntersecting` so it should be the first argument.

packages/vuetify/src/components/VImg/__tests__/__snapshots__/VImg.spec.ts.snap
```diff
@@ -1,198 +1,238 @@
...
-       style="background-image: url(LOAD_SUCCESS_SRC); background-position: center center;"
-       name="fade-transition"
-       mode="in-out"
-  >
-  </div>
-  <div class="v-responsive__content"
-       style="width: 1600px;"
-  >
+  <transition-stub>
+    <img class="v-img__img v-img__img--contain"
```
KaelWD, line 228: `` isn't showing up in jest for some reason.

packages/vuetify/src/components/VImg/VImg.ts
```diff
@@ -1,49 +1,54 @@
...
 export interface srcObject {
-  src: string
+  src?: string
   srcset?: string
-  lazySrc: string
+  lazySrc?: string
   aspect: number
 }
 
 const hasIntersect = typeof window !== 'undefined' && 'IntersectionObserver' in window
```
KaelWD, line 42: We still need this check because safari sucks.

packages/vuetify/src/util/helpers.ts
```diff
@@ -415,3 +414,27 @@ export function getUid () {
   return getUid._uid++
 }
 getUid._uid = 0
+
+export function flattenFragments (nodes: VNode[]): VNode[] {
```
KaelWD, line 418: Didn't end up using this because I decided to pass all the `<source>` elements straight through instead of parsing them, might be useful for something else though.

glen-84:
Is it possible to use the [loading attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading)? Will it be applied to the `img` element?

KaelWD:
I experimented with that, but only Firefox emits a loadstart event for it. I'm still using the IntersectionObserver from V2 which behaves like `loading="lazy"`

PR #12976, opened 2021-01-15
feat(position.ts): add new composable

johnleider:
## Description
Added new composable position

## Motivation and Context
To port positionable mixin to v3

## How Has This Been Tested?

jest

## Types of changes

## Checklist:


packages/vuetify/src/composables/position.ts
```diff
@@ -0,0 +1,58 @@
...
+import { convertToUnit } from '@/util/helpers'
+import propsFactory from '@/util/propsFactory'
+
+// Types
+export interface PositionProps {
+  bottom?: boolean | string
+  left?: boolean | string
+  position?: 'static' | 'relative' | 'fixed' | 'absolute' | 'sticky'
+  right?: boolean | string
+  top?: boolean | string
```
jacekkarczmarczyk, line 13:
```suggestion
  bottom?: boolean | string | undefined
  left?: boolean | string | undefined
  position?: 'static' | 'relative' | 'fixed' | 'absolute' | 'sticky' | undefined
  right?: boolean | string | undefined
  top?: boolean | string | undefined
```

vue defines optional props as `foo?: type | undefined`

image


Also looks like there is `ExtractPropTypes` tool which can be used instead

packages/vuetify/src/composables/__tests__/position.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+describe('position.ts', () => {
+  it.each([
+    [{ position: undefined }, []],
+    [{ position: false }, []],
+    [{ position: null }, []],
+    [{ position: 'absolute' }, ['position-absolute']],
+    [{ position: 'fixed' }, ['position-fixed']],
+  ])('should have proper classes', (props, expected) => {
+    const { positionClasses } = usePosition(props as PositionProps)
```
jacekkarczmarczyk, line 15: i guess casting is here so that `position: null` doesn't throw ts error, but since `null` is not a valid value of prop i don't think testing it (and casting) is necessary

packages/vuetify/src/composables/__tests__/position.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it.each([
+    [{ top: true }, { top: '0px' }],
+    [{ right: 20 }, { right: '20px' }],
+    [{ bottom: '20' }, { bottom: '20px' }],
+    [{ left: undefined }, {}],
+    [{ top: null, right: 50 }, { right: '50px' }],
+    [{ right: false, left: '50' }, { left: '50px' }],
+  ])('should have proper styles', (props, expected) => {
+    const { positionStyles } = usePosition(props as PositionProps)
```
jacekkarczmarczyk, line 28: same as for `should have proper classes` test

packages/vuetify/src/composables/position.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const positionClasses = computed(() => {
+    const classes: string[] = []
+
+    if (!props.position) return classes
+
+    classes.push(`position-${props.position}`)
+
+    return classes
+  })
```
nekosaur, line 41:
```suggestion
  const positionClasses = computed<string[]>(() => {
    if (!props.position) return []

    return [`position-${props.position}`]
  })
```

packages/vuetify/src/composables/position.ts
KaelWD: `something?: string` already includes `undefined`, that's probably just the IDE.

packages/vuetify/src/composables/position.ts
KaelWD:
```suggestion
  const positionClasses = computed(() => {
    return props.position ? `position-${props.position}` : undefined
  })
```

KaelWD:
Delete the positionable mixin.

packages/vuetify/src/composables/position.ts
jacekkarczmarczyk: ah ineed looks like it's IDE

johnleider: > `something?: string` already includes `undefined`, that's probably just the IDE.

This was my understanding as well.

johnleider:

packages/vuetify/src/composables/position.ts
```diff
@@ -0,0 +1,52 @@
...
+// Types
+export interface PositionProps {
+  bottom?: boolean | string
+  left?: boolean | string
+  position?: 'static' | 'relative' | 'fixed' | 'absolute' | 'sticky'
+  right?: boolean | string
+  top?: boolean | string
+}
+
+const positionValues = ['static', 'relative', 'fixed', 'absolute', 'sticky'] as const
```
nekosaur, line 16: This could be used as the basis for the `position` property in PositionProps as well

packages/vuetify/src/composables/position.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// Composables
+export const makePositionProps = propsFactory({
+  bottom: [Boolean, String],
+  left: [Boolean, String],
+  position: {
+    type: String as PropType<Partial<typeof positionValues[number]>>,
+    validator: /* istanbul ignore next */ (v: any) => positionValues.includes(v),
```
nekosaur, line 24: Is this worth it just to artificially increase coverage?

johnleider: Okay Mr. OCD about classes as a string.

Issue #4641, opened 2018-07-17
[Bug Report] Vertical Dividers Regression - Width


slathrop:
### Versions and Environment
Vuetify: 1.1.5
Browsers: Chrome, Firefox, Edge
OS: Windows

### Steps to reproduce
See example in Vuetify docs website [here](https://vuetifyjs.com/en/components/dividers#example-vertical) and note that the alignment/spacing of "My Home" text is off due to 100% width of vertical divider that is directly inside v-toolbar (in .v-toolbar__content)

### Expected Behavior
`.v-divider` to have `width: auto`

### Actual Behavior
`.v-divider` has `width: 100%` when directly inside a v-toolbar

### Reproduction Link
https://vuetifyjs.com/en/components/dividers#example-vertical
---

### Additional Comments:
Chrome is telling me that `v-divider` class is incorrectly getting `width: 100%` due to [this](https://github.com/vuetifyjs/vuetify/blob/c1404ab099748f83755cc954a994c77564529d69/src/stylus/components/_dialogs.styl#L50) line of code

AlexSHoffman:
Does `.v-divider` even need a `width` rule? 

It looks like `flex` rule is handling the width by itself.

![output](https://user-images.githubusercontent.com/13764106/42848455-4d4cae20-8a28-11e8-9359-910fe6f49356.gif)

PR #12993, opened 2021-01-20
feat(VDivider): port to v3

### LINKED ISSUE #12054: [Bug Report] v-divider (inset) use margin instead of max size and remove space.
### Environment
**Vuetify Version:** 2.3.8
**Vue Version:** 2.6.11
**Browsers:** Chrome 84.0.4147.125
**OS:** Windows 10

### Steps to reproduce
1. Create a layout with a grid. 
2. Add divider.
3. See that divider takes space and forces wrapping layout. Remove space with negative margin.
4. See that the max-width / max-height calculation does not work correctly. Use margin instead.

### Expected Behavior
It's possible to remove the space of a border in this case of a line divider. Could be an option.

### Actual Behavior
The border takes space and breaks by layout. Removing that space results in a wrong compution of max-width / max-height.

### Reproduction Link
<a href="https://codepen.io/infacto/pen/rNexgJN" target="_blank">https://codepen.io/infacto/pen/rNexgJN</a>

### Other comments
Regardless of my case, consider to use margin only without max-width / max-height.
What do you think about? I think this is a more solid solution.
And removing space by negative margin also could be a great solution. Maybe as option (attribute like `inset`). 

#### My Steps

**Without any hacks**
1-without-hacks

**Remove space.**
2-hack-remove-space

**Use margin instead of max-width / max-height**
3-hack-use-margin

#### My final SCSS

```scss
.v-divider--inset {
  max-width: initial !important;
  max-height: initial !important;
  &:not(.v-divider--vertical) {
    margin-top: -1px !important;
    margin-left: 8px !important;
    margin-right: 8px !important;
  }
  &.v-divider--vertical {
    margin-left: -1px !important;
    margin-bottom: 8px !important;
  }
}
```

This only adds some styles to the build-in styles of Vuetify.
For example: Margin left and top is already set that framework.
So, set all styles in the final solution to have complete styles.

I like Vuetify and value you expert opinion. Margin only (without max-size) is a more solid way to do that, or not? And divider without spaces are more flexible to use in layout. What do you think about?

johnleider:
## Description
More to follow..

fixes #12054

## Motivation and Context
Port v-divider to v3

## How Has This Been Tested?
jest

## Markup:

<details>

```vue
https://gist.github.com/johnleider/9b501dc94fa13e75ed339f425613f115

```
</details>

## Types of changes

## Checklist:


johnleider:
Requesting **Initial** comments from @vuetifyjs/contributors @vuetifyjs/core-team

packages/vuetify/src/components/VDivider/VDivider.ts
```diff
@@ -1,42 +1,43 @@
...
-      },
-      attrs: {
-        role: 'separator',
-        'aria-orientation': orientation,
-        ...this.$attrs,
-      },
-      on: this.$listeners,
+  setup (props, { attrs }) {
+    const { themeClasses } = useTheme()
+    const classes = computed(() => {
```
nekosaur, line 20: Not sure if there's any gain here using a `computed` instead of just putting this class object directly in render function?

packages/vuetify/src/components/VDivider/VDivider.ts
```diff
@@ -1,42 +1,43 @@
...
+      return {
+        'v-divider--inset': props.inset,
+        'v-divider--vertical': props.vertical,
+      }
     })
+    const role = attrs.role ?? 'separator'
```
nekosaur, line 26: This will not be reactive

johnleider: Tested this locally and it works. What am I missing?

```vue
<template>
  <v-app>
    <div class="ma-12 pa-12">
      <button @click="onClick">click me</button>

      <div class="my-5"></div>

      <v-divider :role="role"></v-divider>
    </div>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
      role: undefined,
    }),

    methods: {
      onClick () {
        this.role = this.role === undefined ? 'button' : undefined
      },
    },
  }
</script>
```

packages/vuetify/src/components/VDivider/VDivider.ts
```diff
@@ -1,42 +1,40 @@
...
-        'aria-orientation': orientation,
-        ...this.$attrs,
-      },
-      on: this.$listeners,
-    })
+  setup (props, { attrs }) {
+    const { themeClasses } = useTheme()
+    const role = attrs.role ?? 'separator'
+    const ariaOrientation = role === 'separator'
+      ? props.vertical ? 'vertical' : 'horizontal'
```
jacekkarczmarczyk, line 22: No idea why `attrs.role` seems to be reactive here, but `props.vertical` is not (`aria-orientation` doesn't change to `vertical`):

```vue
<template>
  <v-app>
    <div style="width: 100px; height: 100px">
      <v-divider :vertical="vertical" />
    </div>
    <button @click="vertical = !vertical">toggle vertical</button>
  </v-app>
</template>

<script>
  export default {
    data: () => ({ vertical: false }),
  }
</script>
```

packages/vuetify/src/components/VDivider/__tests__/__snapshots__/VDivider.spec.ts.snap
```diff
@@ -1,8 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`VDivider should render component and match snapshot 1`] = `
-<hr role="separator"
-    aria-orientation="horizontal"
-    class="v-divider theme--light"
+exports[`VDivider should match a snapshot 1`] = `
+<hr class="v-divider  v-theme--light"
+    ariaorientation="horizontal"
```
DRoet, line 5: shouldn't the result still be
```suggestion
    aria-orientation="horizontal"
```
or is this intentional?

packages/vuetify/src/components/VDivider/VDivider.sass
```diff
@@ -1,39 +1,32 @@
 @import './_variables.scss'
 
-+theme(v-divider) using ($material)
-  border-color: map-get($material, 'dividers')
-
 .v-divider
+  border-style: $divider-border-style
+  border-width: $divider-border-width
   display: block
-  flex: 1 1 0px
```
KaelWD, line 8: This was here to fix #4641

packages/vuetify/src/components/VDivider/__tests__/__snapshots__/VDivider.spec.ts.snap
johnleider: It only does this in jest because I'm using camelCase. I can change it to dash case if it matters to anyone.

packages/vuetify/src/components/VDivider/VDivider.sass
johnleider: I checked it and it's no longer needed in v2.4, so no point in bringing it to v3.

packages/vuetify/src/components/VDivider/__tests__/__snapshots__/VDivider.spec.ts.snap
jacekkarczmarczyk: It seems to be a vue issue, already resolved but not released yet
https://github.com/vuejs/vue-next/issues/583

jacekkarczmarczyk: it does get kebab-cased in codepen https://codepen.io/jkarczm/pen/vYXMbWp?editors=1010, test-related issue?

johnleider:

packages/vuetify/src/components/VDivider/VDivider.ts
```diff
@@ -1,42 +1,58 @@
...
   props: {
-    inset: Boolean,
-    vertical: Boolean,
+    ...makeBorderProps(),
+    ...makeDimensionProps({
+      height: undefined,
+      maxHeight: undefined,
+      maxWidth: undefined,
+      width: undefined,
+    }),
```
nekosaur, line 27: What is the expected behavior when these props in this component? Is it the width/height of the actual divider line? Or is it the width/height of the component/element?

If it is the element dimension then I would expect the actual divider line to be centered correctly inside it, which it's not since it uses border. I'm not sure I understand this scenario though. If I needed to pad a divider my first instinct would be to use padding/margin classes.

If it's the divider dimension then it currently doesn't work since again it uses border to draw it. In this scenario I would probably suggest that we use props `width` and `length` instead. `width` would be the thickness of the divider line, and `length` the length of it, which works in both vertical and horizontal directions.

(also this could just be `makeDimensionProps()`)

nekosaur: @jacekkarczmarczyk proposed `thickness` instead of `width` to reduce confusion

nekosaur (commented):
I'm not entirely on board with the current API of v-divider. Assuming everything works correctly, the following markup

```vue
<v-divider width="100px" height="100px" border></v-didiver>
```

Would result in a 100x100px box with border on all sides. That seems very far from what `v-divider` should be and do.

Additionally, markup like this

```vue
<v-divider width="100px"></v-divider>
```

Would result in something like this in horizontal mode

image


And this in vertical mode

image

since width/height are orientation dependent.

In my mind, the props

```
border
width
height
maxWidth
maxHeight
```

could be replaced by

```
thickness
length
color (if we want to keep that functionality from border)
```

and better represent the function of the component

packages/vuetify/src/components/VDivider/VDivider.sass
johnleider: Changing this to 100% so these work in flex containers. The only other way I could get it to work was using `min-width: 1px` which felt wrong.

packages/vuetify/src/components/VDivider/__tests__/VDivider.spec.ts
```diff
@@ -1,105 +1,50 @@
...
 
-  it('should have no aria-orientation if presentation role is set in attrs', () => {
-    const wrapper = mountFunction({
-      attrs: { role: 'presentation' },
-    })
+  it.each([
+    [{}, ['horizontal', 'separator']],
+    [{ vertical: true }, ['vertical', 'separator']],
+    [{ role: 'foo' }, [undefined, 'foo']],
+  ])('should have correct styles for %s', (propsData, expected) => {
```
nekosaur, line 43:
```suggestion
  ])('should have correct attributes for %s', (propsData, expected) => {
```

nekosaur (approved):

PR #13067, opened 2021-02-06
feat(layout): add new layout system and components

nekosaur:
## Description
Adds new layout system and composable, components

The system works in an outside to in fashion, where each added layout-item takes up some amount of space in the layout. Each subsequent layout-item is placed below all the previous ones, and only has whatever remaining space in the layout left to occupy.

The following markup adds items to the layout system in the order they are defined in the markup.

```vue
<v-layout>
  <v-layout-item position="top" size="100" />
  <v-layout-item position="left" size="300" />
  <v-content>
    content area here
  </v-content>
</v-layout>
```

This works fine until you need to dynamically remove/add items in a layout (e.g. with a v-if). Then you need to explicitly tell the system the order of the items.

```vue
<v-layout :layout="['drawer', 'bar']">
  <v-layout-item name="bar" position="top" size="100" />
  <v-layout-item name="drawer" position="left" size="300" />
  <v-content>
    content area here
  </v-content>
</v-layout>
```

Here the items appear as `['bar', 'drawer']` in the markup, but they will always be rendered in the order `['drawer', 'bar']`.

The composables that power these components can be re-used for v-app and the other layout components.

Nested layouts are possible, see the playground markup

Here's what a couple of final markups would look like. v-app uses the layout composable so is serving the same function as v-layout. Each 'app' level component like toolbar and navigation-drawer uses the layout item composable.

```vue
<v-app>
	<v-toolbar />
    <v-navigation-drawer />
	<v-content>
		...
	</v-content>
</v-app>
```

Here's the discord layout example we have on docs.

```vue
<v-app :layout="['system-bar', 'server-list', 'channel-list', 'channel-bar', 'sidebar', 'input']">
	<v-toolbar name="system-bar" size="25" />
    <v-navigation-drawer name="server-list" size="75" />
    <v-navigation-drawer name="channel-list" size="150" />
	<v-toolbar name="channel-bar" size="50" />
    <v-navigation-drawer name="sidebar" right size="150" />
	<v-footer name="input" size="75" />
	<v-content>
		...
	</v-content>
</v-app>
```

## How Has This Been Tested?
- [ ] tests are needed

## Markup:

<details>

```vue
<template>
  <div>
    <v-layout :layout="layout" full-height>
      <v-layout-item name="drawer1" position="left" :size="width" v-if="bar">drawer1</v-layout-item>
      <v-layout-item name="drawer2" :position="pos">drawer2</v-layout-item>
      <v-layout-item name="toolbar" position="top" size="100">toolbar</v-layout-item>
      <v-layout-item name="foo" :size="200" position="bottom">foo</v-layout-item>
      <v-layout-item name="foo2" :size="100" position="top">foo2</v-layout-item>
      <!-- <v-footer id="innerfoot">inner footer</v-footer> -->
      <v-content>
        <button @click="bar = !bar">v-if drawer1</button>
        <button @click="toggle">reverse layout</button>
        <button @click="dir">switch drawer2 position</button>

        <v-layout :style="{ margin: '100px' }">
          <v-layout-item position="bottom" size="150">innerfoot</v-layout-item>
          <v-layout-item position="top" size="100">foo</v-layout-item>
          <v-layout-item position="left" size="50">bar</v-layout-item>
          <v-content>
            asdasdzxczxc
          </v-content>
        </v-layout>
      </v-content>
    </v-layout>
  </div>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const bar = ref(true)
      const width = ref(300)
      const layout = ref(['toolbar', 'drawer1', 'drawer2', 'foo', 'foo2'])
      const pos = ref('right')

      const toggle = () => {
        layout.value = layout.value.slice().reverse()
      }

      const dir = () => {
        pos.value = pos.value === 'left' ? 'right' : 'left'
      }

      setTimeout(() => {
        width.value = 200
      }, 2000)

      return {
        bar,
        width,
        layout,
        toggle,
        dir,
        pos,
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


glen-84:
How does this relate to #12471?

johnleider:
> How does this relate to #12471?

It's simply another implementation of the same feature. We are taking the best concepts from both and combining them.

nekosaur:
new playground

```vue
<template>
  <v-layout full-height>
    <v-layout-item :priority="1" position="top" size="100" />
    <v-layout-item :priority="Number(prio)" position="left" size="300" v-if="show" />
    <v-layout-item :priority="5" position="right" size="50" />
    <v-layout-item :priority="7" position="bottom" size="75" />
    <v-content>
      content area here
      <button @click="show = !show">show</button>
      <input v-model="prio">
    </v-content>
  </v-layout>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const show = ref(true)
      const prio = ref('3')

      return {
        show,
        prio,
      }
    },
  }
</script>
```

glen-84:
Is this sort of similar to [source order](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items#the_order_property) in flexbox/grid? If so, perhaps `order` would work as the property name?

I don't mean that it's _literally_ using the CSS `order` property, just that it's maybe being used for a similar purpose.

nekosaur:
`order` would technically work just as well, it would just be the reversed `priority` value

johnleider (approved):

PR #13096, opened 2021-02-12
feat(transition): update to vue 3, use material easings

KaelWD:
https://material.io/design/motion/speed.html#easing

Replaced the `$transition` map with three variables based on the material easings.  
Updated the transition classes to use accel/decel easings.

**BREAKING CHANGE:** Removed the `$transition` SASS map
**BREAKING CHANGE:** Removed the `.transition-<name>` classes

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <button @click="toggle">Toggle</button>

      <v-expand-transition>
        <div
          v-if="show"
          style="width: 200px; height: 200px"
          class="elevation-6"
        ></div>
      </v-expand-transition>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    setup () {
      const show = ref(false)
      function toggle () {
        show.value = !show.value
      }

      return {
        show,
        toggle,
      }
    },
  }
</script>
```
</details>


packages/vuetify/src/components/transitions/createTransition.ts
```diff
@@ -1,40 +1,18 @@
...
-  dest: Function | Function[] = [],
-  ...transitions: (Function | Function[])[]
-) {
-  /* eslint-disable-next-line no-array-constructor */
-  return Array<Function>().concat(dest, ...transitions)
-}
+import type { FunctionalComponent } from 'vue'
+import { defineComponent, TransitionGroup, Transition, h } from 'vue'
 
 export function createSimpleTransition (
```
johnleider, line 4: Should this be named `createCssTransition` to be more declarative?

packages/vuetify/src/components/transitions/createTransition.ts
```diff
@@ -45,59 +23,49 @@ export function createSimpleTransition (
       },
     },
 
-    render (h, context): VNode {
-      const tag = `transition${context.props.group ? '-group' : ''}`
-      const data: VNodeData = {
-        props: {
+    setup (props, { slots }) {
+      return () => {
+        const tag = props.group ? TransitionGroup : Transition
```
johnleider, line 28: can the user still supply a custom tag when using **group**?

johnleider (changes requested):

packages/vuetify/src/components/transitions/createTransition.ts
KaelWD: yes

packages/vuetify/src/components/transitions/createTransition.ts
```diff
@@ -45,59 +26,49 @@ export function createSimpleTransition (
...
-        }),
-        context.children
-      )
+    setup (props, { slots }) {
+      return () => {
+        return h(Transition as any, {
+          name,
+          // mode: props.mode, TODO: vuejs/vue-next#3104
+          ...functions,
+        }, () => slots.default?.())
```
johnleider, line 70: I assume this is intended. Can you explain what we have to return a function that returns the default slot instead of just returning the default slot.

johnleider (approved):

PR #13115, opened 2021-02-15
feat(VNavigationDrawer): port to v3

johnleider:
## Description
port to v3

## Motivation and Context
port to v3

## How Has This Been Tested?
pending

## Markup:

<details>

https://gist.github.com/johnleider/2961f4bcd3831fc37f8ea2ff4061137e
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -0,0 +1,137 @@
...
+          {
+            hasImg && (
+              <div class="v-navigation-drawer__img">
+                { slots.img
+                  ? slots.img?.({ src: props.src })
+                  : (<img src={ props.src } />)
+                }
+              </div>
+            )
+          }
```
KaelWD, line 116:
```suggestion
          { hasImg && (
            <div class="v-navigation-drawer__img">
              { slots.img
                ? slots.img?.({ src: props.src })
                : (<img src={ props.src } />)
              }
            </div>
          )}
```

KaelWD:
```suggestion
          { hasImg && (
            <div class="v-navigation-drawer__img">
              { slots.img?.({ src: props.src }) ?? (<img src={ props.src } />) }
            </div>
          )}
```

packages/vuetify/src/components/VNavigationDrawer/_variables.scss
```diff
@@ -1,4 +1,27 @@
...
+$navigation-drawer-elevation: 0 !default;
+$navigation-drawer-fixed-z-index: 6 !default;
+$navigation-drawer-floating-border: none !default;
+$navigation-drawer-img-height: inherit !default;
+$navigation-drawer-img-object-fit: cover !default;
+$navigation-drawer-img-width: inherit !default;
+$navigation-drawer-overflow-scrolling: touch !default;
+$navigation-drawer-positions: absolute fixed !default;
+$navigation-drawer-temporary-elevation: 16 !default;
+$navigation-drawer-transition-duration: 0.2s !default;
```
KaelWD, line 18: This and the timing function should be synced with v-main.

johnleider: As far as I can see, it is? 
image

johnleider:
Requesting **Initial** comments from @vuetifyjs/core-team

nekosaur:
I think it would be very helpful if you provided descriptions of some of the props available here. Some of my immmediate questions are:

1. What's the difference between `mobile` and `temporary`? They seem to have the same effect, even om small screens.
2. `temporary` and `floating` are too close in meaning imo. It would not be hard to imagine that someone sees `floating` and thinks that it does what `temporary` actually does.
3. What does `permanent` do?
4. What does `stateless` do? It sounds a little bit like `temporary` as well.
5. `bottom` makes nav drawer take up entire screen
6. On the topic of `bottom` (and `right`), I am not a fan of using the position composable but changing the meaning of the props. In nav drawer, `bottom` and `right` are used only as boolean props to indicate where drawer should show (and the logic and calculation of that is done by layout composable), but they are defined as `[Boolean, Number, String]`. I would much rather see a `position` prop that takes `'left' | 'right' | 'bottom'`. 
7. Because it uses position composable it also has `top` prop which doesn't make sense.
8. `makeLayoutItemProps` should not set default name. That would immediately trip up the layout system when two nav drawers are used.
9. What's the use case for `fixed`? Right now it gets overwritten by layout which sets absolute position in styles

johnleider:
1. `mobile` is a temporary prop to test the mobile functionality of a component and will go away once the display composable is done.
2. Agreed, only done for v2 feature parity. `floating` can be removed for `border="0"`
3. We can remove stateless, but permanent should stay.
  a. That will work as well
4. It can go
5. Should only happen on landscape, I can remove for now and implement it once display is in
6. I'm not opposed to this
7. The same as 6
8. Okay
9. So that being fixed isn't tied to the usage of `app`, at least in v2

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { makeTagProps } from '@/composables/tag'
+import makeProps from '@/util/makeProps'
+
+export default defineComponent({
+  name: 'VNavigationDrawer',
+
+  props: makeProps({
+    ...makeBorderProps(),
+    ...makeBorderRadiusProps(),
+    ...makeDimensionProps({ width: 256 }),
```
nekosaur, line 26: Using this composable means it will get all of the width/height props but using them (except for width) won't have any effect because they will be (rightly) overwritten by layout styles

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -0,0 +1,142 @@
...
+    ...makeElevationProps(),
+    ...makeLayoutItemProps(),
+    ...makePositionProps(),
+    ...makeTagProps({ tag: 'nav' }),
+    expandOnHover: Boolean,
+    mobile: Boolean,
+    modelValue: Boolean,
+    permanent: Boolean,
```
nekosaur, line 34: I'm still not convinced of this prop, but 🤷. I feel like the proxy composable was made specifically for this exact scenario, where a user wants to absolutely control some state of a component without the component itself doing something that makes the internal value change.

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    const isActive = useProxiedModel(props, 'modelValue')
+    const isHovering = ref(false)
+    const size = computed(() => Number(props.rail ? props.railWidth : props.width))
+    const width = computed(() => {
+      return (props.rail && props.expandOnHover && isHovering.value)
+        ? props.width
+        : size.value
+    })
+    const styles = useLayoutItem(
```
nekosaur, line 60: nitpick but could be `layoutStyles`

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { makeElevationProps, useElevation } from '@/composables/elevation'
+import { makeLayoutItemProps, useLayoutItem } from '@/composables/layout'
+import { makePositionProps, usePosition } from '@/composables/position'
+import { useProxiedModel } from '@/composables/proxiedModel'
+import { useTheme } from '@/composables/theme'
+
+// Utilities
+import { computed, defineComponent, onBeforeMount, ref, toRef } from 'vue'
+import { convertToUnit } from '@/util/helpers'
+import { makeTagProps } from '@/composables/tag'
```
nekosaur, line 17: should be in composables section

nekosaur (changes requested):

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
johnleider: There is a certain level of simplicity that comes from a boolean prop. This would be a good example usage of prop aliases. For now I'd like to keep it and revisit if/when necessary.

nekosaur: The alternative would also just be one prop `:model-value="true"`

nekosaur (approved):

Issue #2324, opened 2017-10-26
High CPU usage for v-progress-bar component. Ridiculous HIGH


ultraN:
### Steps to reproduce
Just go to https://vuetifyjs.com/components/progress, open your chrome task manager sort by cpu usage.
### Versions

Vue 2.5.2
Vuetify 0.16.9
Webpack 3
Npm 5.xx
Ndoe 8.xx
OS: macOS 10.13
Browser: Chrome 62.0.3202.62
CPU: Core i7 (I7-3740QM)@2.7GHz, 4 cores 8 threads

### What is expected ?

Low or reasonable cpu usage. Not like ~30% all the time.

### What is actually happening ?

![qq20171026-150503](https://user-images.githubusercontent.com/3374628/32034591-1cc024e2-ba5f-11e7-9d09-ce652407f0b9.png)
As shown in the picture, the progress component is taking significant system resources to render/calculate the spinning circle.
### Reproduction Link

https://vuetifyjs.com/components/progress

nekosaur:
This has been brought up before (https://github.com/vuetifyjs/vuetify/issues?utf8=%E2%9C%93&q=cpu%20progress). It seems to vary pretty wildly between setups. For me (CPU i5 3570, Win 10, Chrome 62.0.3202.62) it hovers around ~10%.

The docs page is also not representative of most real use cases for progress bars. That said, as a comparison, how's the CPU usage of this page (https://material.angularjs.org/latest/demo/progressCircular) for you? It's pretty similar for me at around 10%.

KaelWD:
There's what, 15 loaders on that page? I don't think 2% each is too high, you aren't likely to have more than one or two on an actual page. There isn't really any way to create this animation without repaints anyway, as it is constantly changing length.

ultraN:
@nekosaur @KaelWD I used docs page just because it's fastest way to represent this. In our own project, a single circle will take ~25% cpu resources. This metric is measured under 15'' retina screen.

KaelWD:
What's the resolution of this "retina" screen?

ultraN:
@KaelWD 2880 x 1800

nekosaur:
Please create a codepen reproduction or github repo with your setup in which a single circle consumes 25% CPU

peluprvi:
[...]

Issue #785, opened 2017-06-19
v-btn loader CPU usage


Christilut:
### Steps to reproduce
This is probably the same as #553 but I noticed it again

CPU usage is steady around 50% and it is very noticeable. Both my laptops are experiencing massive performance issues when a browser tab with a spinning loader is visible. My XPS 13 (only has onboard graphics card, also Linux mint) won't even re-render Visual Studio Code and on my XPS 15 the entire system just slows down and it is very noticeable.

On the v-progress demo page it wasn't a big deal because it's not really a practical example but for 1 spinner to cause this is crazy.

### Versions
Vue 2.3.4, latest vuetify
Linux mint 18

### What is expected ?
No performance loss

### What is actually happening ?
![image](https://user-images.githubusercontent.com/3471363/27275635-8f53a48a-54d8-11e7-893c-1fe6864b01b6.png)

### Reproduction Link
https://codepen.io/anon/pen/pweyzp

cawa-93:
![screenshot_20170619_130004](https://user-images.githubusercontent.com/1662812/27280053-3b22fb56-54ef-11e7-907d-8ba75d6618a8.png)

johnleider:
Can you check out https://dev.vuetifyjs.com and see if this issue still persist for you. We have made a few adjustments which should improve this already from a recent ticket.

Christilut:
This is on my XPS13, I'll try on the XPS15 tomorrow

![image](https://user-images.githubusercontent.com/3471363/27305692-f4386920-5542-11e7-92ea-9960bcf61789.png)

Can you fork the codepen with the vuetify dev version so I can test that too?

johnleider:
You can just go to https://dev.vuetifyjs.com

cawa-93:
@johnleider The first message says about a large load with one button and one loader. At https://dev.vuetifyjs.com we have a lot of loaders, which is why browser indicators may not be objective. It will be fair to create an identical template in the codepen with one button and one loader using the dev vuetifyjs

dohomi:
@cawa-93 @Christilut you can use the dev version of vuetify (also in Codepen):
``` "vuetify": "https://github.com/vuetifyjs/vuetify-builds.git#dev ```
Here is the folder on github: https://github.com/vuetifyjs/vuetify-builds/tree/dev/dist
Hope this helps

Christilut:
Same codepen with dev version: https://codepen.io/anon/pen/EXXpqV

[...]

Issue #720, opened 2017-06-08
High CPU usage on documentation site


areading314:
### Steps to reproduce
* Go to API documentation page: https://vuetifyjs.com/vuetify/quick-start
* Watch CPU usage -- its still reasonable
* Click once on  the Components menu item on the left hand side to expand long list of components in sidebar
* CPU usage will spike while Components menu is open

### Versions

* Arch linux using Version 59.0.3071.86 (Official Build) (64-bit)

* NOTE: This did not happen in Firefox 53 

### What is expected ?

* Reasonable CPU usage

### Reproduction Link

https://vuetifyjs.com/vuetify/quick-start

EDIT: Clarification: Behavior observed in Google Chrome browser.

johnleider:
There are a few things missing from this equation. Firstly, what are the specs of your computer? You say reasonable CPU usage, but don't define what your usage is. On my work computer I see the % go from ~4% to ~7% when opening that section, but that is irrelevant since this is not a slow computer.

Please understand that even though this ticket is following the guidelines for reporting, it gives absolutely no information that I could use to look into this situation.

fenivana:
At "buttons" page ( https://vuetifyjs.com/components/buttons ), stay and do nothing, my CPU usage is around 7% (Surface Book, i5). This is unreasonable because the page isn't doing anything visually. I don't know whether there's some bug in the example codes, or in the components.

![image](https://user-images.githubusercontent.com/11750532/26943319-b452f84a-4cb7-11e7-8926-aff856de9350.png)

johnleider:
Thank you, that is much more helpful.

johnleider:
Looks like it is being caused by the data-table.

johnleider:
This should be resolved with #727

techntrails:
Performance issues are not resolved yet. Most of the component pages in the docs seem to have this issue. If I have a Chrome tab open with a Vuetify Datatable, my macbook battery only lasts ~25% the time in normally does

nekosaur:
This fix has not yet been released. It will be available in 0.13

techntrails:
@nekosaur I forgot to mention... I've tested this on the 0.13 and it's still not fixed, at least with the Datatable component. I've even tried disabling the linear progress bar

nekosaur:
Are you sure you're running the dev build? If so, please create a new issue using the template and as much information as possible.

techntrails:
[...]

Issue #553, opened 2017-05-17
CPU usage on progress page


Christilut:
I was just reading the API and my CPU started running at 150% permanently while on [this](https://vuetifyjs.com/components/progress) page.
I'm not very experienced with animations but maybe some optimizing can be done here?

Using Chrome 58.0.3029.110 on Linux Mint 18

cawa-93:
I agree. The load is very high
![screenshot_20170517_184308](https://cloud.githubusercontent.com/assets/1662812/26162903/afa7a298-3b30-11e7-85d2-7ef4c79a9a24.png)

johnleider:
I've made a few tweaks on the new docs to help alleviate this a bit. I went through both components to confirm no memory leaks or unnecessary re-rendering of the components due to properties changed and it all looked in order. While I did not experience the level of CPU usage that you have shown, that page on average was 1% higher CPU utilization, from 2% to 3% than other pages. This is primarily due to how many animations are rendering at once and is not indicative of real world performance.

In light of that, with the changes I've made, I'm going to close this ticket pending any additional information can be found to expedite additional dev time into this.

Brimstedt:
I experience the same problem, and if i use progress-circular or progress-linear (only one), my app uses lots of cpu as well.

Chrome performance tab shows lots of activity in 'Update layer tree'

This is on Chrome Version 59.0.3071.109 (Official Build) (64-bit) on Linux.

nekosaur:
This issue is closed. Please feel free to open a new issue with a reproduction of the problem. Thank you.

PR #13136, opened 2021-02-19
feat(VProgress): only play animations when visible

### LINKED ISSUE #12731: [Feature Request] Add lazy functionality to Progress components
### Problem to solve
There have been various reports of the Progress components spiking CPU in different scenarios.

#### Various reports:
#2324, #785, #720, #553

#### Progress components:
https://vuetifyjs.com/en/components/progress-circular
https://vuetifyjs.com/en/components/progress-linear/

### Proposed solution
A member of my team discovered that if you wrap these components with `VLazy` you can greatly reduce the performance impact - especially if multiple instances of Progress components exist on the page.

The proposed solution is to add a `lazy` prop to the Progress components that gives them lazy functionality (i.e. not rendering if not visible on the page).

This would be either by wrapping in `VLazy` or using the `v-intersect` directive - likely requiring a minimal change to the `render` function and an added `prop` of `lazy`.

https://vuetifyjs.com/en/components/lazy
https://vuetifyjs.com/en/directives/intersect/

P.S. some kind of `lazy` directive would be cool but not sure if it's feasible/wanted and would for sure have a namespace clash with `v-lazy`

Bamblehorse:
## Description

This PR resolves #12731 and deals with reducing CPU load for `VProgressCircular` and `VProgressLinear` while they aren't visible.

## Motivation and Context

The solution seeks to reduce reported CPU spikes(#2324, #785, #720, #553) by pausing the components animations when not visible on the page (through the `Intersect` directive)
## How Has This Been Tested?

Unit tests  have been updated
## Markup:

<details>

```vue
<template>
  <v-container style="height: 300vh; padding-top: 200vh;">
    <v-progress-circular
      indeterminate
    />
    <v-progress-linear
      indeterminate
      style="margin-top: 10vh;"
    />
    <v-progress-linear
      striped
      indeterminate
      reverse
      style="margin-top: 10vh;"
    />
    <v-progress-linear
      stream
      :value="25"
      buffer-value="50"
      style="margin-top: 10vh;"
    />
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      //
    }),
  }
</script>

```
</details>

## Types of changes

## Checklist:


KaelWD:
My idea was more to just apply a class that disables the CSS animation, rather than hiding elements.

Bamblehorse:
> My idea was more to just apply a class that disables the CSS animation, rather than hiding elements.

@KaelWD that would be much more simple 🙂 I'll adjust the PR and get back to you

Bamblehorse:
> My idea was more to just apply a class that disables the CSS animation, rather than hiding elements.

@KaelWD thank you for your feedback 🙂 it was helpful and made sense.

I have refactored to use the `Intersect` directive to toggle a class which pauses the animation when the element is not visible. I was going to use the `Intersectable` mixin but realized it doesn't pass the `isIntersecting` Boolean to the callback so it seemed appropriate to use the directive instead.

Let me know what you think when you have some time 🙂

nekosaur:
Don't call the directive methods directly. Use `directives` option in vnode data when using `h()`. See for example `v-lazy` component.

Bamblehorse:
> Don't call the directive methods directly. Use `directives` option in vnode data when using `h()`. See for example `v-lazy` component.

Of course 😅 thanks for the sanity check and pointer to `v-lazy` - have added it in properly

I'm wondering if this should use similar logic to `v-img` to check for intersect availability:
https://github.com/vuetifyjs/vuetify/blob/21331625b9707f19cc2004f2e0957b288a2ccad3/packages/vuetify/src/components/VImg/VImg.ts#L25
https://github.com/vuetifyjs/vuetify/blob/21331625b9707f19cc2004f2e0957b288a2ccad3/packages/vuetify/src/components/VImg/VImg.ts#L258-L269

Alternatively could set `isVisible` to be `true` by default which should catch those edge cases as well

KaelWD:
> I'm wondering if this should use similar logic to v-img to check for intersect availability

The directive itself should probably handle that?

Bamblehorse:
> > I'm wondering if this should use similar logic to v-img to check for intersect availability
> 
> The directive itself should probably handle that?

I was under the impression if this was used on an unsupported browser - say IE - the animation would never play - so I was thinking to set `isVisible` to `true` which would hopefully solve that

Bamblehorse:
@KaelWD / @nekosaur Thanks for your continued feedback 🙂 I appreciate it

I have updated in line with my comment above to set `isVisible` to `true` by default

It immediately switches to `false` on browsers that support `IntersectionObserver` so is a simple fail safe

I have added a regression test to keep it in place

KaelWD:
Looks great, thanks. Are you planning to do the same for VProgressLinear?

Bamblehorse:
> Looks great, thanks. Are you planning to do the same for VProgressLinear?

Awesome 🙂 yeah just wanted to check the changes fitted in with the codebase - I've added the equivalent changes for `VProgressLinear` in https://github.com/vuetifyjs/vuetify/pull/13136/commits/900d452beb78cd50a6ab65c3a9b2b863f0b0b87a

Let me know if you want me to separate the PRs or if the changes are similar enough that it's okay.

KaelWD (approved):

Issue #10537, opened 2020-02-10
[Documentation] Add Testing Commands to Contributing 


rachael-smith:
### Problem to solve
There is no information on how to run tests for contributors 

### Proposed solution
add commands/information about how to run tests in contributing

olegstepura:
For anyone else willing to contribute and ending up here. Looks like this is enough to run unit tests for components (assuming you are in project root folder):
```sh
cd packages/vuetify/
jest
```
of cause you need to run `yarn install` before

PR #13218, opened 2021-03-05
fix(VCarousel): VCarousel is changing context to dark mode

### LINKED ISSUE #8392: [Bug Report] carousel children has dark theme while dark: false
### Environment
**Vuetify Version:** 2.0.5
**Vue Version:** 2.6.10
**Browsers:** Chrome 75.0.3770.142
**OS:** Windows 10

### Steps to reproduce
1. set {theme: {dark: false}}
2. use v-carousel

### Expected Behavior
use setting from theme.dark

### Actual Behavior
dark and light prop are null in carousel

### Reproduction Link
<a href="https://codesandbox.io/s/bug-report-g2ukh" target="_blank">https://codesandbox.io/s/bug-report-g2ukh</a>

### Other comments
2 v-card same content just for sanity check

olegstepura:
## Description
When using `VCarousel` is an app in light mode it was changing context to dark mode

## Motivation and Context

Fixes #8392

Reason for this is this sequence:
- with current implementation of `isDark` (1)
- since `VCarousel` extends `VWindow` which is turn extends `VItemGroup`which mixes in `Themable` (2)
- and `Themable` is watching `isDark` property to set the context (3)
- and on the other hand `Themable` has defaults for `dark` and `light` with `null` (4)

<details>
  <summary>code snippets for the above</summary> 

(1):
https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/components/VCarousel/VCarousel.ts#L79-L81
(2):
https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/components/VCarousel/VCarousel.ts#L23
 
 https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/components/VWindow/VWindow.ts#L18
 
 https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/components/VItemGroup/VItemGroup.ts#L22-L25
(3):
https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/mixins/themeable/index.ts#L90-L98
(4):
https://github.com/vuetifyjs/vuetify/blob/9bec589dcd6ba4630e1ce5473753165cf238b622/packages/vuetify/src/mixins/themeable/index.ts#L28-L36
</details>

So not setting any value to `dark` or `light` on `VCarousel` leaves those 2 props as `null` and the `isDark` computed property in `VCarousel` is checking literally `null || !null`. So it becomes true in both scenarios:
- `dark=true`
- nothing specified (both `dark` and `light` are `null`s)

what indicates a bug

## How Has This Been Tested?
- added a unit-test
- run the unit test, note the https://github.com/vuetifyjs/vuetify/issues/10537 makes it harder
<details>
  <summary>Tests output</summary>
<pre>
Test Suites: 188 passed, 188 total
Tests:       64 skipped, 1581 passed, 1645 total
Snapshots:   637 passed, 637 total
Time:        15.301 s
Ran all test suites.
</pre>
</details>

## Types of changes

## Checklist:

## Workaround
If similar issue is faced it can be workaround by
```jsx
    <v-carousel
      ...
      :light="!$vuetify.theme.isDark"
    >
```

### DIFF
--- a/packages/vuetify/src/components/VCarousel/VCarousel.ts
+++ b/packages/vuetify/src/components/VCarousel/VCarousel.ts
@@ -77,7 +77,7 @@ export default VWindow.extend({
       }
     },
     isDark (): boolean {
-      return this.dark || !this.light
+      return this.dark === true || this.light === false
     },
     isVertical (): boolean {
       return this.verticalDelimiters != null
--- a/packages/vuetify/src/components/VCarousel/__tests__/VCarousel.spec.ts
+++ b/packages/vuetify/src/components/VCarousel/__tests__/VCarousel.spec.ts
@@ -125,4 +125,29 @@ describe('VCarousel.ts', () => {
 
     expect(wrapper.vm.internalHeight).toBe(300)
   })
+
+  it.only('should not change context to dark mode automatically', async () => {
+    const wrapper = mountFunction()
+
+    expect(wrapper.vm.isDark).toBe(false)
+
+    wrapper.setProps({ dark: true })
+
+    expect(wrapper.vm.isDark).toBe(true)
+
+    wrapper.setProps({ dark: false })
+
+    expect(wrapper.vm.isDark).toBe(false)
+
+    // reset previous dark
+    wrapper.setProps({ dark: null })
+
+    wrapper.setProps({ light: true })
+
+    expect(wrapper.vm.isDark).toBe(false)
+
+    wrapper.setProps({ light: false })
+
+    expect(wrapper.vm.isDark).toBe(true)
+  })
 })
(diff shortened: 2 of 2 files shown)


KaelWD:
This was done because the default controls have a dark background:
image

If we inherit the app theme then the icons are illegible:
image

We either need to remove themeable from carousel and hardcode the icon colours, or somehow only use the dark/light props for controls and inherit from the parent (not the app) for custom content.

olegstepura:
Hi, @KaelWD!

This code does not seem like an intentional darkenizing of the context.
```js
isDark (): boolean { 
   return this.dark || !this.light 
 }, 
```
 I tried to explain above what happens. Enforcing dark mode in the context of any element is not a good idea. We faced this issue using `VCarousel` for non-image content. Regular `v-card` became dark in the context of `v-carousel` and it was a hard to debug issue.

KaelWD:
I know what's happening, I wrote the thing. I just didn't expect anything other than images at the time.

PR #13319, opened 2021-03-24
feat(VItemGroup): port to v3

nekosaur:
## Description

Port v-item-group to v3. This ends up just being a simple wrapper around group composables. 

- I added `makeGroupProps` and `makeGroupItemProps` functions, but I'm pretty sure they can't really be re-used straight up because we have components that only provide some of the functionality (i.e. multiple but not mandatory). We'll see.
- Added `selectedClass` prop to both group and group item composables. Renamed it from `activeClass` to match the `isSelected` && `select` naming.

## Breaking changes

- `activeClass` renamed to `selectedClass`
- v-model prop and event updated to `model-value` / `update:model-value`
- added `'force'` option to `mandatory` prop. item group will no longer automatically select first non-disabled item in group if `:mandatory="true"`. if this behavior is wanted, use `mandatory="force"`

## Motivation and Context

v3

## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="d-flex align-center justify-center">
      <v-item-group v-model="selected2" v-slot="{ next, prev }">
        <v-item v-slot="{ toggle, selectedClass }" value="foo">
          <div :class="selectedClass" @click="toggle">foo</div>
        </v-item>
        <v-item v-slot="{ toggle, selectedClass, disabled }" value="bar" disabled>
          <div :class="[selectedClass, disabled && 'disabled']" @click="toggle">bar</div>
        </v-item>
        <v-item v-slot="{ toggle, selectedClass }" value="baz">
          <div :class="selectedClass" @click="toggle">baz</div>
        </v-item>
        <v-btn @click="prev">prev</v-btn>
        <v-btn @click="next">next</v-btn>
      </v-item-group>
      <v-item-group v-model="selected" :multiple="multiple" :mandatory="mandatory">
        <v-item v-for="i in 5" :key="i" :value="i" v-slot="{ selectedClass, toggle }">
          <div :class="selectedClass" @click="toggle">item {{ i }}</div>
        </v-item>
      </v-item-group>
    </div>
    <div>
      <pre>{{ selected }}</pre>
      <v-btn @click="mandatory = !mandatory" :color="mandatory ? 'success' : 'grey'">mandatory</v-btn>
      <v-btn @click="multiple = !multiple" :color="multiple ? 'success' : 'grey'">multiple</v-btn>
    </div>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    setup () {
      const selected = ref([])
      const selected2 = ref()
      const mandatory = ref(true)
      const multiple = ref(false)
      return { selected, selected2, mandatory, multiple }
    },
  }
</script>

<style>
.fill-height {
  height: 100%;
}

.v-item--selected {
  background: red;
}

.disabled {
  opacity: 0.1;
}
</style>
```
</details>

## Types of changes

## Checklist:


johnleider:
Requesting **Final** comments from @vuetifyjs/core-team

packages/vuetify/src/composables/__tests__/group.spec.ts
```diff
@@ -163,7 +163,7 @@ describe('group', () => {
       ])
     })
 
-    it('should set first non-disabled item as value when mandatory', async () => {
+    it.skip('should set first non-disabled item as value when mandatory', async () => {
```
jacekkarczmarczyk, line 166: since it's request for final comments `.skip` should be removed

packages/vuetify/src/composables/__tests__/group.spec.ts
```diff
@@ -323,7 +323,7 @@ describe('group', () => {
       expect(events[1][0][0]).not.toEqual(events[1][0][1])
     })
 
-    it('should set first non-disabled item as value when mandatory', async () => {
+    it.skip('should set first non-disabled item as value when mandatory', async () => {
```
jacekkarczmarczyk, line 326: `.skip`

packages/vuetify/src/composables/group.ts
```diff
@@ -56,17 +81,23 @@ export function useGroupItem (
...
+    value,
+    disabled,
   }
 }
 
 export function useGroup (
   props: GroupProps,
   injectKey: InjectionKey<GroupProvide>
 ) {
+  const isUnmounted = ref(false)
```
jacekkarczmarczyk, line 100: why is it a ref?

packages/vuetify/src/composables/group.ts
```diff
@@ -155,7 +203,12 @@ export function useGroup (
 
     const currentId = selected.value[0]
     const currentIndex = items.findIndex(i => i.id === currentId)
-    const newIndex = (currentIndex + offset) % items.length
+
+    let newIndex = (currentIndex + offset) % items.length
+
+    while (items[newIndex].disabled) {
```
jacekkarczmarczyk, line 209: won't it loop infinitely if all items are disabled?

packages/docs/src/examples/v-item-group/misc-chips.vue
```diff
@@ -23,16 +23,16 @@
...
         <v-item
           v-for="n in 8"
           :key="n"
-          v-slot="{ active, toggle }"
+          v-slot="{ selected, selectedClass, toggle }"
         >
           <v-chip
-            active-class="purple--text"
-            :input-value="active"
+            :class="[selectedClass]"
```
jacekkarczmarczyk, line 34:
```suggestion
            :class="selectedClass"
```

packages/vuetify/src/components/VItemGroup/VItemGroup.sass
```diff
@@ -4,4 +4,4 @@
   flex: 0 1 auto
   position: relative
   max-width: 100%
-  transition: $item-group-transition
+  transition: 0.28s $standard-easing
```
jacekkarczmarczyk, line 7: what's the reason for not using $item-group-transition? that way it's less customizable

packages/vuetify/src/composables/group.ts
```diff
@@ -56,17 +81,23 @@ export function useGroupItem (
...
   }
 }
 
 export function useGroup (
   props: GroupProps,
   injectKey: InjectionKey<GroupProvide>
 ) {
+  const isUnmounted = ref(false)
   const items = reactive<GroupItem[]>([])
   const selected = useProxiedModel(
```
jacekkarczmarczyk, line 102: maybe `selected` could be a `Set`? also accepting `Set` as a model would be a nice feature, it doesn't seem to be supported now

KaelWD: :+1:, set wasn't reactive in vue 2 but is now.

nekosaur:
In regards to the skipped tests, they are related to functionality that I am still undecided on.

In v2, we automatically select the first non-disabled item on mount if nothing is selected and `mandatory` prop is used. My current feeling is that this should be removed, or at the very least put behind an optional prop. We are making an assumption about user code, and it will be hard for them to counteract the behavior if it is automatic.

Slightly related to this (automatic behavior, assumptions) is functionality that I added that tries to convert model value from single to array and back again if `multiple` prop is dynamically changed. For some reason I am slightly more okay with this functionality than the first scenario, but I am still undecided.

Feedback very welcome.

packages/vuetify/src/composables/group.ts
nekosaur: good catch

packages/vuetify/src/composables/group.ts
nekosaur: no reason other than muscle memory of defining a variable inside a composable

packages/vuetify/src/composables/group.ts
nekosaur:
```
Set in => Set out
Array in => Array out
nothing in => what out?
```

jacekkarczmarczyk: hmm good question, i think i'd go for array and use `Set` only if dev explicitely uses `Set` in `v-model`

nekosaur: I am dropping this. It seems like way more hassle than it's worth. Internally things don't really get much easier or more performant (unless I'm just stupid and missing something obvious). And from a DX point of view, users can still do 
```
:model-value="[...set.values()]" @update:model-value="set = new Set($event)"
```
if they really want to store value in a set.

packages/api-generator/src/locale/en/group.json
```diff
@@ -0,0 +1,12 @@
+{
+  "props": {
+    "modelValue": "Represents the selected item(s). Will be an array if **multiple** prop is used.",
+    "mandatory": "Forces at least one item to always be selected (if available).",
+    "max": "Sets a maximum number of selections that can be made.",
```
johnleider, line 5: What about a min prop?

packages/vuetify/src/components/VItemGroup/VItem.tsx
```diff
@@ -0,0 +1,25 @@
+// Composables
+import { makeGroupItemProps, useGroupItem } from '@/composables/group'
+import { VItemGroupSymbol } from './VItemGroup'
```
johnleider, line 3: Put under a // Components comment

packages/vuetify/src/components/VItemGroup/_variables.scss
```diff
@@ -1,3 +1,3 @@
 @import '../../styles/styles.sass';
```
johnleider, line 1: Update sass structure, add index.scss and adjust imports (same as btn and other components)

packages/vuetify/src/composables/group.ts
```diff
@@ -27,11 +28,33 @@ interface GroupProvide {
...
 }
 
+export const makeGroupProps = propsFactory({
+  modelValue: {
+    type: [Number, Boolean, String, Array, Object],
+    default: undefined,
+  },
+  multiple: Boolean,
+  mandatory: [Boolean, String] as PropType<boolean | 'force'>,
+  max: Number,
```
johnleider, line 41: Thoughts on a min prop?

packages/vuetify/src/composables/group.ts
```diff
@@ -56,17 +81,23 @@ export function useGroupItem (
     return group.isSelected(id)
   })
 
+  const selectedClass = computed(() => isSelected.value && (group.selectedClass.value ?? props.selectedClass))
+
```
johnleider, line 85: Breaking your own blank line rules 🤣

packages/vuetify/src/composables/group.ts
```diff
@@ -146,29 +185,40 @@ export function useGroup (
     }
   }
 
-  function getOffsetId (offset: number) {
+  function step (offset: number) {
```
johnleider, line 188: What about step reverse? Thinking within the context of v-carousel

johnleider (changes requested):
Really looking good!

packages/api-generator/src/locale/en/group.json
nekosaur: How would `min` work? Say it's set to 2. I select one item. What happens? Is it shown as selected just as if min was not set? Are we emitting new value with it included? Probably not right, because 1 is less than 2. That means we now have to create another internal value that differs from the 'real' value to keep track of 'selected but not valid' items.

If `min` is only a validation thing like for forms, not a hard limit like `max`, then it should probably not live in this component.

packages/vuetify/src/composables/group.ts
nekosaur: ?

packages/vuetify/src/composables/group.ts
nekosaur: Not sure what you mean? There are prev and next functions

johnleider (approved):

PR #13395, opened 2021-04-08
fix(types): add declaration for vuetify/lib/components

Djaler:
## Description
Vuetify has separated `declare module` for `vuetify/lib/directives`. This add `vuetify/lib/components` too.

## Motivation and Context
Now we can do something like `import * as VuetifyComponents from 'vuetify/lib/components';` without typescript errors.

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/types/lib.d.ts
+++ b/packages/vuetify/types/lib.d.ts
@@ -1,19 +1,21 @@
 declare module 'vuetify/lib' {
-  // eslint-disable-next-line import/no-duplicates
-  import { VueConstructor, DirectiveOptions } from 'vue'
   import Vuetify from 'vuetify'
   import { Colors } from 'vuetify/lib/util/colors'
 
   export default Vuetify
 
   const colors: Colors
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
I think it looks but not sure what is the best way to test it, after adding this to node_modules/vuetify/types/lib.d.ts i still have ts error (both in IDE and during compilation)
Also the components list was updated so should be updated in this PR as well

Djaler:
Yeah, I'm able to reexport them (but still need to list them all.
We can make it vice versa - declare types in `vuetify/lib/components` and `vuetify/lib/directives` and then just do 
```
export * from 'vuetify/lib/components'
export * from 'vuetify/lib/directives'
```
in `vuetify/lib` module.
What do you think about it? @johnleider

jacekkarczmarczyk:
Sounds good for me

Djaler:
So here it is

KaelWD (approved):

Issue #13248, opened 2021-03-10
[Bug Report] Chip sizes not changing icon sizes


megaroeny:
### Environment
**Vuetify Version:** 2.4.6
**Vue Version:** 3.0.7
**Browsers:** Edge 89.0.774.45
**OS:** Windows 10

### Steps to reproduce
1. Apply a `small` size prop to the chip with an icon. 
2. You'll see the icon size doesn't change.

### Expected Behavior
Icon sizes should change and follow the SASS variables provided for $icon-sizes here: https://vuetifyjs.com/en/api/v-chip/#icon-sizes

### Actual Behavior
Icon sizes don't change between any size applied.

### Reproduction Link
<a href="https://codepen.io/megaroeny/pen/zYoyObG" target="_blank">https://codepen.io/megaroeny/pen/zYoyObG</a>

### Other comments
This is my best guess that it is a bug, but please correct me if I'm wrong.

PR #13412, opened 2021-04-12
fix(VChip): icon size based on chip size prop

daikiojm:
## Description

fixes https://github.com/vuetifyjs/vuetify/issues/13248

## How Has This Been Tested?

visually(use Playground)

before
スクリーンショット 2021-04-12 10 33 25

after
スクリーンショット 2021-04-12 10 32 02

## Markup:

<details>

```vue
<template>
 <v-app id="inspire">
    <div class="text-center">
      <v-chip
        class="ma-2"
        color="indigo"
        text-color="white"
              x-small
      >
        <v-avatar left>
          <v-icon>mdi-account-circle</v-icon>
        </v-avatar>
        Mike
      </v-chip>

      <v-chip
        class="ma-2"
        color="orange"
        text-color="white"
              large
      >
        Premium
        <v-icon right>
          mdi-star
        </v-icon>
      </v-chip>

      <v-chip
        class="ma-2"
        color="primary"
        text-color="white"
        small
      >
        1 Year
        <v-icon right>
          mdi-cake-variant
        </v-icon>
      </v-chip>


    </div>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
      hasMessage: false,
      hideDetails: false,
    }),
    computed: {
      d () {
        return this.hideDetails ? this.hideDetails : 'auto'
      },
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


KaelWD (approved):

johannes-huther:
I recently used a medium sized `v-icon` in a `small` `v-chip`, which (more or less coincidentally) fit perfectly (rounded icon matched curve of chip). After this change, the icon is `small` as well. 

I'd like to reset the icon size, but adding `medium` doesn't do the trick (`large` works, but is to large of cause). Is there a property that does that? Should `medium` work?

```vue
           <v-chip
              small
              :href="GIT_ROOT"
              id="gh-link"
              color="#24292e"
              text-color="white"
            >
              <v-icon class="pr-1" medium>mdi-github</v-icon>
              GitHub
            </v-chip>
```

Before:
image

After: 
image



*To be clear, I don't want to critizise this PR. It does totally make sense in most cases, just not in mine. I'm not sure if this is the right place to ask either, but I figured you might know since you made the change.*

gustavosbarreto:
@johannes-huther same here.

@KaelWD I think this PR should be reverted because that changes modifies the behavior of already existing apps and behaves differently than a `VIcon`  inside a `VBtn`.

Let's take this code snippet as an example:

```
    <v-chip small>
      <v-icon medium>mdi-account</v-icon>
      v-chip
    </v-chip>

    <v-btn primary small>
      <v-icon medium>mdi-account</v-icon>
      v-btn
    </v-btn>
```

before this PR:

download - 2021-06-10T100903 319

after this PR:
download - 2021-06-10T100815 856

It seems to me that the code added by this PR should be on the @daikiojm application's side not in the Vuetify library.

tlebel:
We've also run into the same issues after update; icons got really small. 

Just quick additional note that the current state is not inline anymore with Material Design Chip specs, where the icon size is closer to the previous `24px` for the default/medium size. Cf.: https://material.io/components/chips#types

jojomatik:
Interesting points @gustavosbarreto, @tlebel.

I don't know to much about Material Design, therefore I can't judge what is right or wrong. 

From my point of view this change does make sense, but adding `medium` to the `v-icon` should definitely make it possible to make it look like it looked before this change, which is currently not the case.

johannes-huther:
Interesting points @gustavosbarreto, @tlebel.

I don't know to much about Material Design, therefore I can't judge what is right or wrong.

From my point of view this change does make sense, but adding `medium` to the `v-icon` should definitely make it possible to make it look like it looked before this change, which is currently not the case.

KaelWD:
@daikiojm I'm going to revert this, it breaks more common use cases and I don't have time to look into a proper fix.

WIStudent:
If you liked what this PR did and still want it after it was reverted you can add the following to your own css
```scss
@use "~vuetify/src/components/VChip/variables";
@each $name, $size in variables.$icon-sizes {
  .v-chip.v-size--#{$name} .v-icon {
    font-size: #{map-get($size, 'font-size')}px;
    height: #{map-get($size, 'font-size')}px;
    width: #{map-get($size, 'font-size')}px;
  }
}
```

If you only want it in selected components, add it as scoped style to those:
````html
<style scoped lang="scss">
@use "~vuetify/src/components/VChip/variables";
@each $name, $size in variables.$icon-sizes {
  .v-chip.v-size--#{$name} .v-icon {
    font-size: #{map-get($size, 'font-size')}px;
    height: #{map-get($size, 'font-size')}px;
    width: #{map-get($size, 'font-size')}px;
  }
}
</style>
````

Personally I prefer the icons a bit bigger so I added 4px to each size:
```scss
@use "~vuetify/src/components/VChip/variables";
@each $name, $size in variables.$icon-sizes {
  .v-chip.v-size--#{$name} .v-icon {
    font-size: #{map-get($size, 'font-size') + 4}px;
    height: #{map-get($size, 'font-size') + 4}px;
    width: #{map-get($size, 'font-size') + 4}px;
  }
}
```

If you prefer sass, look  how the PR did it in sass and adapt it accordingly.

PR #13593, opened 2021-05-03
fix(intersect): don't unbind until the element is intersecting

KaelWD:
## Description

The handler can sometimes be called even if no entries are intersecting, if the `once` modifier is used we don't want to unbind until the element is actually visible in the viewport. 

## Types of changes


johnleider (approved):
These changes should be ported to v3 as well.

PR #13597, opened 2021-05-04
feat(VProgressCircular): port to v3

### LINKED ISSUE #11922: [Bug Report] Circular progress animation breaks under heavy load
### Environment
**Vuetify Version:** 2.3.6
**Vue Version:** 2.6.11
**Browsers:** Google Chrome
**OS:** Windows 10

### Steps to reproduce
Click the heavy load button in the codepen to simulate a heavy load, the animation on the circular progress on the left will break (the dash animation doesn't display). This problem is more prominent if you click on the button while the circular progress 'tail' is short.

### Expected Behavior
The animation continues to play even under heavy load.

### Actual Behavior
The animation doesn't display correctly.

### Reproduction Link
<a href="https://codepen.io/romankoval10/pen/eYJXBjz" target="_blank">https://codepen.io/romankoval10/pen/eYJXBjz</a>

### Other comments
In the codepen I have a circular progress on the right that disables the animation applied to the circle element inside the circular progress. This fixes the heavy load problem, but the animation now doesn't have the stroke part, only the rotate. 

Please see this [material-ui issue](https://github.com/mui-org/material-ui/issues/10327) that targets the exact same problem.

The solution could be the same as what material-ui does, which is to provide a `disableShrink` property that will clear the stroke animation applied to the circle element. The current implementation makes this hard without having to override the styles.

blalan05:
Resolves #11922

## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="ma-12 pa-12">
      <v-progress-circular model-value="50" />
      <v-progress-circular model-value="50" />
      <v-progress-circular model-value="50" indeterminate />
      <v-progress-circular model-value="50" color="blue" />
      <v-progress-circular model-value="50" color="blue" size="x-large" />
      <v-progress-circular
        model-value="50"
        color="blue"
        size="48"
        width="12"
      />
    </div>
  </v-app>
</template>

<script>
  export default {

  }
</script>

```
</details>

## Types of changes

- [X] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:


packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+export default defineComponent({
+  name: 'VProgressCircular',
+
+  props: makeProps({
+    indeterminate: Boolean,
+    rotate: {
+      type: [Number, String],
+      default: 0,
+    },
+    size: {
```
nekosaur, line 22: use size composable

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+    ...makeTagProps({ tag: 'div' }),
+  }),
+
+  setup (props, { slots }) {
+    const radius = 20
+    const isVisible = true
+
+    const { themeClasses } = useTheme()
+
+    const { colorClasses, colorStyles } = useColor(computed(() => ({
```
nekosaur, line 47: could be `useTextColor` instead

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+      default: 0,
+    },
+    ...makeTagProps({ tag: 'div' }),
+  }),
+
+  setup (props, { slots }) {
+    const radius = 20
+    const isVisible = true
+
+    const { themeClasses } = useTheme()
```
nekosaur, line 45: no need to apply theme classes

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+      return parseFloat(props.value)
+    })
+
+    const strokeDashArray = computed((): number => {
+      return Math.round(circumference.value * 1000) / 1000
+    })
+
+    const strokeDashOffset = computed((): string => {
+      return ((100 - normalizedValue.value) / 100) * circumference.value + 'px'
```
nekosaur, line 72: use `convertToUnit`

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const radius = 20
+    const isVisible = true
+
+    const { themeClasses } = useTheme()
+
+    const { colorClasses, colorStyles } = useColor(computed(() => ({
+      text: props.color,
+    })))
+
+    const calculatedSize = computed(():number => Number(props.size))
```
nekosaur, line 51: no need to explicitly type the computeds

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+              'v-progress-circular--visible': isVisible,
+              'v-progress-circular--indeterminate': props.indeterminate,
+            },
+            themeClasses.value,
+            colorClasses.value,
+          ]}
+          style={[
+            colorStyles.value,
+            { height: convertToUnit(calculatedSize.value) },
+            { width: convertToUnit(calculatedSize.value) },
```
nekosaur, line 104: combine to one object

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,142 @@
...
+    },
+    value: {
+      type: [Number, String],
+      default: 0,
+    },
+    ...makeTagProps({ tag: 'div' }),
+  }),
+
+  setup (props, { slots }) {
+    const radius = 20
```
nekosaur, line 42: should this be a static value?

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: Done, didn't know about that part of the composable. Cool

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: Understood

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: Done.

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: Ah... wasn't sure where we needed typings and when/where we don't.

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: Ah, duh! Thanks.

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: It's used with the loading animation, I think to determine how much of the circle is filled, and how much is gapped. It was static in the previous version, and when changing it, it isn't intuitive what happens when the number changes. With 10, the gap is smaller, with 100 you get 3 different section loading and 3 gaps.

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
blalan05: This is looking for a number that it uses to inform the svg viewbox height and width. Since the size composable returns strings, I'm not sure how I would get the to work...

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
nekosaur: I was wrong on this :) We've decided to apply theme classes on component level after all.

nekosaur:
- I added support for using `small, large, x-large` size values to keep things consistent with other components
- Re-added intersection observer functionality
- Added `bg-color` prop. This one I'm a bit unsure of. I think it would be very nice to be able to set background circle color (since it is already possible in progress-linear). However I'm not sure what the default value should be. The current `surface` color is just white. Something like `grey-lighten-1` looks pretty good, but then it will look different if users disable material colors

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,127 @@
...
+import './VProgressCircular.sass'
+
+// Composables
+import { makeTagProps } from '@/composables/tag'
+import { useTextColor } from '@/composables/color'
+
+// Utilities
+import { computed, defineComponent, toRef } from 'vue'
+import { convertToUnit, makeProps } from '@/util'
+import { useTheme } from '@/composables/theme'
```
ElijahKotyluk, line 11: Move composable imports to `// Composables`

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { useIntersectionObserver } from '@/composables/intersectionObserver'
+import { makeSizeProps, useSize } from '@/composables/size'
+
+export default defineComponent({
+  name: 'VProgressCircular',
+
+  props: makeProps({
```
ElijahKotyluk, line 18: Sort props alphabetically

ElijahKotyluk (changes requested):

johnleider:
Can we add different animations? Such as:

image
image
image
image

These are just a few examples / ideas. Does the @vuetifyjs/core-team think that my suggestions are within the scope of this component?

blalan05:
Adding different styles/options could be interesting, but it brings up the question of how much outside of Google MD do we want to implement/provide? 

Another question, what would the syntax be to select the style desired, which one is default?
<v-progressive-circular type="growing-dots" />?

KaelWD:
I don't think that's necessary.

KaelWD (approved):

johnleider (approved):

PR #13648, opened 2021-05-13
fix(VTimePickerClock): race condition between event and $refs

Tofandel:
In some rare cases the error `Cannot read property 'getBoundingClientRect' of undefined` is thrown, I believe it's due to the fact that the event may fire only after the element has been destroyed in some slow devices, for now it only has happened on chrome desktop

## Description

Simply gracefully exit the function without throwing in this very edge case

## Motivation and Context
Some of my users (less than 0.1%) experience this issue in production
image

image
image

## How Has This Been Tested?

This is an edge case that happens randomly with very low incidence and probably impossible to reproduce deterministically, so it hasn't

## Types of changes

## Checklist:


jacekkarczmarczyk (approved):
lgtm, no idea how it's possible that the error happens though...

PR #13654, opened 2021-05-14
test(VBtn): add cypress wrapper and better assertions.

JessicaSachs:
## Description
* This change adds a reusable wrapper for `mount` so that `createVuetify` is done under the hood on every `cy.mount` call.
* Additionally, the VBtn tests are refactored to become more meaningful.
* VBtn is also showcased using a presentational test-only grid component (we _could_ use the Vuetify grid system if we wanted to)

### API Changes:
New, more concise syntax for mounting a component with props.

```tsx
import { h } from 'vue'
import VBtn from './VBtn'

it('renders a basic button', () => {
  cy.mount(<VBtn color="success">Let's go!</VBtn>)
})
```

### Visual Changes:
The buttons are properly laid out.

Screen Shot 2021-05-13 at 8 16 02 PM

### API changes:
The previous API for writing a basic test.
```tsx
import { createVuetify } from '@/entry-bundler'
import { mount } from '@cypress/vue'
import { h } from 'vue'
import VBtn from './VBtn'

it('renders a basic button', () => {
  const vuetify = createVuetify()
  mount(VBtn, {
    global: {
      plugins: [vuetify],
    },
    slots: {
      default: 'Button',
    },
  })
})
```

### Visual Changes:
The buttons are smushed together and are not easily viewed or debugged.

Screen Shot 2021-05-13 at 8 20 45 PM

## Motivation and Context:
* By adding a `cy.mount` wrapper, you're able to inject application context like theme providers, i18n, etc.
* Adding styles to the components under test make them easier to debug and view

## How Has This Been Tested?

Using tests.

## Markup:

<details>

```vue
```
</details>

## Types of changes:

## Checklist:

## Outstanding tasks:
- [x] Fix the type errors
- [x] Refactor the "presentational templates" to be more readable
- [x] Reformat/lint


nekosaur:
Looks good. I think it would be useful to be able to pass options to `createVuetify` with a mount function like this

```ts
mount: <Props = any>(comp: Component<Props>, options?: CyMountOptions<Props>, vuetifyOptions?: any) => Cypress.Chainable<JQuery<any>>
```

but it looks like CyMountOptions is not exported. although it seems like most things in it are actually flagged as deprecated, so maybe we can just use `MountingOptions` instead?

JessicaSachs:
Alright @nekosaur I think I'm ready for a final review. Any thoughts?

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
```diff
@@ -1,3 +1,5 @@
+declare const expect: jest.Expect
```
nekosaur, line 1: Would we have to add this to every non-cypress test file?

packages/vuetify/types/cypress.d.ts
```diff
@@ -0,0 +1,9 @@
+import type { mount as cyMount } from '@cypress/vue'
+
+declare global {
+  namespace Cypress {
+    export interface Chainable {
+      mount: typeof cyMount
```
nekosaur, line 6: This does not include `vuetifyOptions` argument

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
JessicaSachs: Yes, if you want the Jest types on the expect matchers. The alternate solution is to pull out the Cypress tests into their own directory. Alternatively you can `import { expect} from 'jest'`.

packages/vuetify/types/cypress.d.ts
JessicaSachs: I don't really know how to type it. I'll try agian.

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.ts
nekosaur: I think this works for now.

packages/vuetify/types/cypress.d.ts
nekosaur: Since we're not actually using the argument yet I think we can revisit it later.

nekosaur (approved):

johnleider (approved):

Issue #13656, opened 2021-05-14
[Bug Report][2.5.0] Regression: vCarousel in v-if triggers an error


Tofandel:
### Environment
**Vuetify Version:** 2.5.0
**Last working version:** 2.4.11
**Vue Version:** 2.6.11
**Browsers:** Chrome 90.0.4430.212
**OS:** Linux x86_64

### Steps to reproduce
Put a carousel in a v-if, toggle it, notice the error in the console

### Expected Behavior
No error to occur

### Actual Behavior
`TypeError: Cannot read property 'clientHeight' of undefined`

### Reproduction Link
<a href="https://codesandbox.io/s/bug-report-forked-sq8q7?file=/src/App.vue" target="_blank">https://codesandbox.io/s/bug-report-forked-sq8q7?file=/src/App.vue</a>

### Other comments
Introduced by https://github.com/vuetifyjs/vuetify/commit/faeb8dfb8b1361f2195ea89785b747bca3e24672
Was not throwing an error in 2.4.11

Tofandel:
@KaelWD I believe the way the VCarousel theme was fixed is flawed, it should just use the `Themeable` mixin instead of using a sub theme provider, no?

nirmalraghavan:
I have this same issue with **2.4.11** and **2.2.34**
Vue Version: 2.6.11

Tofandel:
> I have this same issue with **2.4.11** and **2.2.34**
> Vue Version: 2.6.11

I cannot reproduce it in those versions with my reproduction, can you post a minimal reproduction?

mahammad-sixberries:
still the same issue with 2.5.1 using nuxt

PR #13669, opened 2021-05-17
fix(VCarousel): prevent TypeError in v-if

### LINKED ISSUE #8392: [Bug Report] carousel children has dark theme while dark: false

Tofandel:
## Description

Changes the implementation from faeb8dfb8b1361f2195ea89785b747bca3e24672 to use inject/provide instead of the theme provider

## Motivation and Context

The previous commit introduced a bug in the vCarousel #13656 

## How Has This Been Tested?

Visually and in the markup that there was no changes and with thorough test units

image

## Markup:

<details>

```vue
<template>
  <div>
    <v-btn @click="show = !show">Show</v-btn>
    <v-container v-if="show">
      <v-carousel>
        <v-carousel-item>
          <v-card> I am dark</v-card>
        </v-carousel-item>
      </v-carousel>
      <v-carousel light>
        <v-carousel-item>
          <v-card> I am light</v-card>
        </v-carousel-item>
      </v-carousel>
    </v-container>
  </div>
</template>

<script>
export default {
  data: () => ({
    show: false
  }),
}
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VCarousel/__tests__/VCarousel.spec.ts
```diff
@@ -9,18 +8,18 @@ import VProgressLinear from '../../VProgressLinear/VProgressLinear'
 // Utilities
 import {
   mount,
-  MountOptions,
+  MountOptions, shallowMount,
```
jacekkarczmarczyk, line 11:
```suggestion
  MountOptions,
  shallowMount,
```

packages/vuetify/src/components/VCarousel/__tests__/VCarousel.spec.ts
```diff
@@ -125,4 +124,57 @@ describe('VCarousel.ts', () => {
 
     expect(wrapper.vm.internalHeight).toBe(300)
   })
+
+  it('should have the correct theme', async () => {
+    let wrapper = mountFunction({}, true)
+
+    expect(wrapper.vm.isDark).toBeFalsy()
+    expect(wrapper.html()).toMatchSnapshot();
```
jacekkarczmarczyk, line 132: i think you could just check existence of `.v-carousel.theme-light/dark` element instead of generating the whole markup, not sure what are team's thoughts on this

Tofandel: Yep, I just wanted to see the html generated to make sure, whatever is preferred

Tofandel:
@KaelWD Can you have a look please?

KaelWD:
This was also intended, I wanted to decouple the carousel's theme from its content. Most of the time you want the carousel to be a dark theme so the controls have white text, but you won't want that to be inherited by anything inside it.

Tofandel:
Doesn't the content always need to be the opposite theme of the carousel and the carousel always need to be the opposite theme of it's parent then?

Eg: current theme light > carousel => dark > content => light

Eg: current theme dark > carousel => light > content => dark

Or the opposite

Eg: current theme light > carousel => light > content => dark

Eg: current theme dark > carousel => dark > content => light


We need to define what should be the default behavior because the current one is definitely not what is expected because the carousel will not use the theme of it's parent but only the app theme (And it's also throwing nice error)

KaelWD:
> current theme dark > carousel => light > content => dark

Dark theme carousel still has black backgrounds and white text, it should basically always be dark unless you're overriding this. 

> the carousel will not use the theme of it's parent but only the app theme

The carousel is always dark unless you use the `light` prop, and it passes its parent's theme to its children to avoid #8392.

Tofandel:
Okay I see what you mean now, then I know just the way to fix this

Tofandel:
@KaelWD Done, now it's the behavior you intended

Tofandel:
Mmm nevermind the buttons inherit the theme as well, so I need to set it in the carousel item only

Tofandel:
Now it's perfect :+1:

KaelWD:
That's exactly what it was already doing with VThemeProvider though?

Tofandel:
Almost yes, except you weren't inheriting the parent theme only the app theme and the v theme provider creates an error when it wraps the v carousel item

KaelWD (approved):

PR #13670, opened 2021-05-18
fix(Menuable): correctly calculate position in embedded v-app

### LINKED ISSUE #7703: [Feature Request] Make dropdown position calculation compatible for embedded Vuetify apps
### Problem to solve
If you have a Vuetify app that is embedded into a page where it is not taking up the full browser window, an app on a WordPress page for example, and you have a relative positioned element as its parent, dropdowns like VSelect will be offset.

### Proposed solution
I tried digging into the source code, but could not really track down the logic, but the only way for me to get Vuetify's dropdowns to work in an embedded app situation, I had to remove all `position: relative`s from parent nodes. Add `position: relative` to v-app didn't work, so I'm guessing the calculation is based on the window, but somehow relative position parents alter the calculation.

I propose rewriting how the positioning calculation is performed and making sure v-app is the reference used for positioning.

Tofandel:
## Motivation and Context

Fixes #7703

## How Has This Been Tested?

Visually and unit
image
image

Jest doesn't really simulate dom positioning so the unit tests are limited and sadly there doesn't seem to be any base for e2e tests

## Markup:

<details>

```vue
<template>
  <div class="wrapper">
    <v-app style="height: 4000px; margin: 20px 40px;">
      <v-app-bar
        absolute
        color="#6A76AB"
        shrink-on-scroll
        prominent
      >
        <v-app-bar-nav-icon></v-app-bar-nav-icon>

        <v-app-bar-title>Title</v-app-bar-title>

        <v-spacer></v-spacer>

        <v-btn icon>
          <v-icon>mdi-magnify</v-icon>
        </v-btn>

        <v-btn icon>
          <v-icon>mdi-heart</v-icon>
        </v-btn>

        <v-btn icon>
          <v-icon>mdi-dots-vertical</v-icon>
        </v-btn>

        <template v-slot:extension>

          <v-menu attach>
            <template v-slot:activator="{ on, attrs }">
              <v-btn
                color="primary"
                dark
                v-bind="attrs"
                v-on="on"
              >
                Dropdown
              </v-btn>
            </template>
            <v-list>
              <v-list-item
                :key="index"
              >
              </v-list-item>
            </v-list>
          </v-menu>
        </template>
      </v-app-bar>
      <v-main>
        <div style="height: 300px"></div>
        <v-container grid-list-lg fluid>
          <div style="height: 400px"></div>
          <v-autocomplete @input="val => (value = val)" :items="[1, 2, 3, 4]" label="Click Me"></v-autocomplete>
          <v-tooltip bottom>
            <template v-slot:activator="{ on, attrs }">
              <v-btn
                color="primary"
                dark
                v-bind="attrs"
                v-on="on"
              >
                Button
              </v-btn>
            </template>
            <span>Tooltip</span>
          </v-tooltip>

          <v-menu offset-y>
            <template v-slot:activator="{ on, attrs }">
              <v-btn
                color="primary"
                dark
                v-bind="attrs"
                v-on="on"
              >
                Dropdown
              </v-btn>
            </template>
            <v-list>
              <v-list-item
                :key="index"
              >
              </v-list-item>
            </v-list>
          </v-menu>
        </v-container>
      </v-main>
    </v-app>
  </div>
</template>

<script>
export default {
  name: "App",
  data: () => ({
    items: [
      { title: 'Click Me' },
      { title: 'Click Me' },
      { title: 'Click Me' },
      { title: 'Click Me 2' },
    ],
    menu: {
      value: false
    }
  })
};
</script>

<style>
.wrapper {
  position: relative;
  margin: 100px 200px;
}
</style>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/mixins/menuable/index.ts
+++ b/packages/vuetify/src/mixins/menuable/index.ts
@@ -7,9 +7,10 @@ import Detachable from '../detachable'
 // Utilities
 import mixins, { ExtractVue } from '../../util/mixins'
 import { convertToUnit } from '../../util/helpers'
-import { VNode } from 'vue'
 
 // Types
+import { VNode } from 'vue'
+
 const baseMixins = mixins(
   Stackable,
   Positionable,
@@ -30,7 +31,6 @@ interface dimensions {
 }
 
 interface options extends ExtractVue<typeof baseMixins> {
-  hasActivator: boolean
   attach: boolean | string | Element
   offsetY: boolean
   offsetX: boolean
@@ -199,19 +199,17 @@ export default baseMixins.extend<options>().extend({
   },
 
   methods: {
-    cumulativeOffset (element: HTMLElement | null, until?: string) {
+    cumulativeOffset (element: HTMLElement | null) {
       let top = 0
       let left = 0
 
       if (element) {
         element = element.offsetParent as HTMLElement | null
-        /* eslint-disable no-unmodified-loop-condition */
-        while (element && (!until || !element.matches(until))) {
+        while (element) {
           top += element.offsetTop || 0
           left += element.offsetLeft || 0
           element = element.offsetParent as HTMLElement | null
         }
-        /* eslint-enable */
       }
 
       return {
@@ -224,10 +222,10 @@ export default baseMixins.extend<options>().extend({
         offsetTop: 0,
         offsetLeft: 0,
         scrollHeight: 0,
-        top: (this.positionY || this.absoluteY),
-        bottom: (this.positionY || this.absoluteY),
-        left: (this.positionX || this.absoluteX),
-        right: (this.positionX || this.absoluteX),
+        top: this.positionY || this.absoluteY,
+        bottom: this.positionY || this.absoluteY,
+        left: this.positionX || this.absoluteX,
+        right: this.positionX || this.absoluteX,
         height: 0,
         width: 0,
       }
--- a/packages/vuetify/src/mixins/menuable/index.ts
+++ b/packages/vuetify/src/mixins/menuable/index.ts
@@ -122,6 +122,7 @@ export default baseMixins.extend<options>().extend({
         scrollHeight: 0,
       },
     },
+    relativeYOffset: 0,
     hasJustFocused: false,
     hasWindow: false,
     inputActivator: false,
@@ -168,6 +169,9 @@ export default baseMixins.extend<options>().extend({
     },
     hasActivator (): boolean {
       return !!this.$slots.activator || !!this.$scopedSlots.activator || !!this.activator || !!this.inputActivator
+    },
+    absoluteYOffset (): number {
+      return this.pageYOffset - this.relativeYOffset
     },
   },
 
@@ -254,7 +258,7 @@ export default baseMixins.extend<options>().extend({
     },
     calcYOverflow (top: number) {
       const documentHeight = this.getInnerHeight()
-      const toTop = this.pageYOffset + documentHeight
+      const toTop = this.absoluteYOffset + documentHeight
       const activator = this.dimensions.activator
       const contentHeight = this.dimensions.content.height
       const totalHeight = top + contentHeight
@@ -268,13 +272,13 @@ export default baseMixins.extend<options>().extend({
         // the overflow, don't offset
         activator.top > contentHeight
       ) {
-        top = this.pageYOffset + (activator.top - contentHeight)
+        top = this.absoluteYOffset + (activator.top - contentHeight)
       // If overflowing bottom
       } else if (isOverflowing && !this.allowOverflow) {
         top = toTop - contentHeight - 12
       // If overflowing top
-      } else if (totalHeight < this.pageYOffset && !this.allowOverflow) {
-        top = this.pageYOffset + 12
+      } else if (top < this.absoluteYOffset && !this.allowOverflow) {
+        top = this.absoluteYOffset + 12
       }
 
       return top < 12 ? 12 : top
@@ -413,6 +417,7 @@ export default baseMixins.extend<options>().extend({
         if (!activator) return
 
         const relative = this.measureAbsolute()
+        this.relativeYOffset = relative.top
         dimensions.activator = this.measure(activator)
         dimensions.activator.top -= relative.top
         dimensions.activator.left -= relative.left
--- a/packages/vuetify/src/mixins/menuable/index.ts
+++ b/packages/vuetify/src/mixins/menuable/index.ts
@@ -2,16 +2,19 @@
 import Positionable from '../positionable'
 import Stackable from '../stackable'
 import Activatable from '../activatable'
+import Detachable from '../detachable'
 
 // Utilities
 import mixins, { ExtractVue } from '../../util/mixins'
 import { convertToUnit } from '../../util/helpers'
+import { VNode } from 'vue'
 
 // Types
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -165,9 +243,26 @@ export default baseMixins.extend<options>().extend({
 
   beforeMount () {
     this.hasWindow = typeof window !== 'undefined'
+
+    if (this.hasWindow) {
+      window.addEventListener('resize', this.onResize, false)
```
mschultz4, line 248: Do we need to listen for the scroll event as well?

Tofandel: I don't think it's necessary, the calculations don't use the scroll, everything is just offset relatively to the v-app now

I'm usually reluctant to add any scroll listeners, It would be a performance drawback that wouldn't see much benefit.
Here the calculation only uses the scroll to check for overflow to avoid positioning it somewhere that's outside of the view pane. It doesn't make much sense to recalculate that after, because either the user cannot see the full menu and scrolls to see the rest (so the position should not change) or he just scrolls away and doesn't want to see the content of the menu (so why would you try to keep it in the view when he scrolls?)

Keeping in mind that the menu is absolutely positioned and not fixed

mschultz4: Ok.  I wasn't sure.  

I implemented a similar fix within my app: calculate top and side based on `getBoundingClientRect` of the v-app, then pass differences to the v-select's nudge properties.  When the view loads, the select is in the proper position, however it remains fixed in the same location after scrolling.  Perhaps I need to absolutely position it to solve the issue.

Tofandel: It seems you are right, Element.getBoundingClientRect() method returns a DOMRect object providing information about the size of an element and its position **relative to the viewport**.

So the calculation would be wrong if you recalculate it with a scrolled window, I need to do some deeper testings

Tofandel: Indeed

# Before with bounding rect
image

# With cumulative offset
image

packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -246,7 +362,7 @@ export default baseMixins.extend<options>().extend({
       }
     },
     checkActivatorFixed () {
-      if (this.attach !== false) return
+      if (this.attach) return
```
KaelWD, line 365: This is intentional, `''` is falsy but should be treated as true

packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -14,14 +14,83 @@ const baseMixins = mixins(
...
+  scrollHeight: number
+  offsetLeft: number
+}
+
 interface options extends ExtractVue<typeof baseMixins> {
+  hasActivator: boolean
   attach: boolean | string | Element
   offsetY: boolean
   offsetX: boolean
+  computedRelativeOffset: { top: number, left: number }
```
KaelWD, line 34: This is unnecessary

packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -112,10 +182,17 @@ export default baseMixins.extend<options>().extend({
   }),
 
   computed: {
+    computedRelativeOffset () {
+      if (!this.attach && this.resized !== null && this.$el) {
+        const p = this.$el.closest('.v-application') as HTMLElement
```
KaelWD, line 187: Everything else uses `[data-app]`

packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -112,10 +182,17 @@ export default baseMixins.extend<options>().extend({
   }),
 
   computed: {
+    computedRelativeOffset () {
```
KaelWD, line 185: None of this is reactive, you should calculate it at the same time as `dimensions` is then store it in data instead of a computed property.

KaelWD (changes requested):
Thanks! Can you check how this behaves with a fixed activator (in an app-bar for example) or using `attach` with a parent that has overflow scrolling.

packages/vuetify/src/mixins/menuable/index.ts
Tofandel: Interesting, in my testing the attach prop defaults to undefined instead of false as well so the results came out wrong, should I change the default value then?

packages/vuetify/src/mixins/menuable/index.ts
Tofandel: You mean the type declaration or just this method? Without them typescript is not happy at all

packages/vuetify/src/mixins/menuable/index.ts
Tofandel: You're right that's a way better approach :sweat_smile:

packages/vuetify/src/mixins/menuable/index.ts
Tofandel: Okay so it seems it's basically expecting the Detachable mixin to always be included with the Menuable one, so maybe the Menuable should extend the Detachable mixin?

Tofandel:
I added an attach in an app bar in the playground, is it what you meant? It works fine, and everything with attach should work the same as before since the calculation branches in another part of the code which doesn't use the relative calculation

KaelWD:
Am I missing something? These all work on master:
image

But are broken with this PR:
image

Tofandel:
Do you happen to have pasted the entire playground? In which case it only works if pasted within the App file otherwise you have 2 v-app nested in each other, though it should still calculate correctly, I'll have a look it might be a bug

KaelWD:
Ah gotcha. It's because you're using `this.$el.closest('[data-app]')` but detachable uses `document.querySelector('[data-app]')`. A better option for the position calculation would probably be `this.$refs.content.offsetParent`

KaelWD:
I've pushed some changes, lmk what you think.

Tofandel:
Looks perfect :+1:

KaelWD (approved):

packages/vuetify/src/mixins/menuable/index.ts
```diff
@@ -165,6 +190,16 @@ export default baseMixins.extend<options>().extend({
 
   beforeMount () {
     this.hasWindow = typeof window !== 'undefined'
+
```
Kotoriii, line 193: These next 10 lines of code make my `v-navigation-drawer` which works with `expand-on-hover` and `mini-variant` throw consistent console errors 
```
index.js:245 Uncaught TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
    at VueComponent.checkActivatorFixed (index.js:245)
    at VueComponent.updateDimensions (index.js:338)
```
for this file.

I have created a barebones [Codepen](https://codepen.io/kotoriii/pen/zYZWyXB) project to try and reproduce this outside of my local environment, but unfortunately the console error does not seem to appear there. Hence why I'm not opening a proper Issue Report on this, but raising this issue here instead. 

Perhaps the error can give you a hint of what might be wrong and if it also is / will affect other people

Kotoriii: @KaelWD @Tofandel

PR #13681, opened 2021-05-19
feat: create v-dialog and v-overlay components

### LINKED ISSUE #1068: [Feature Request] Allow setting origin of dialog's zoom animation to activator's location
### What will it allow you to do that you can't do today?
It would be nice if there is a setting to make it that when a dialog appears, it will be perceived like it is originating from the activating button. Current implementation only allows animation from screen edge or screen center.

An implementation can be seen with Angular Material:
https://material.angularjs.org/latest/demo/dialog

This would help the dialog to align with Surface Creation spec:
https://material.io/guidelines/motion/choreography.html#choreography-creation

Not sure if there is any work-around to do that.

### What potential bugs and edge cases does it help to avoid?
Just an improvement to UX IMHO.

### LINKED ISSUE #4807: [Feature Request] add `overlay-class` to v-dialog

I'd like to use a dialog as an overlay toolbox. I could use a `.v-dialog__content` style rule to change `align-items` and `justify-content` to both be `start`. However, my component evolved, and I now have two dialogs. Only one of which should be the special toolbox.

Because `content-class` is not applied to the detached overlay parent (but to its child), I cannot make a selector that only selects one of the dialogs. In my case, for `align-items`, I can use `align-self` on the child, but `justify-content` doesn't (yet) have a working corresponding `justify-self` (in my Chromium 67).

It would be nice to have something similar to `content-class` added to the overlay parent so position could be customized.

Maybe an `overlay-class` property.

### LINKED ISSUE #6425: [Bug Report] v-dialog overlay still placed in root when using attach

### LINKED ISSUE #6764: [Feature Request] clean up bootable content on close
### Problem to solve
See: https://github.com/vuetifyjs/vuetify/pull/4518, this might make more sense now that `lazy` is the new default in `2.0`.

I know a workaround is to use a `v-if` on the content, but this can cause issues with transitions

### Proposed solution
Make sure content inside a `v-dialog` etc. rerenders by default, a prop like `keep-alive` could be introduced to get the current behaviour

### LINKED ISSUE #7422: [Bug Report] v-tooltip and v-dialog activator element is removed on beforeDestroy, creating a disorienting effect
### Versions and Environment
**Vuetify:** 1.5.14
**Vue:** 2.6.10
**Browsers:** Chrome 74.0.3729.169
**OS:** Mac OS 10.14.4

### Steps to reproduce
Create a v-tooltip and change routes away from the component.

### Expected Behavior
The activator element should get destroyed at the same time as every other element

### Actual Behavior
The element is destroyed before the rest of the elements

### Reproduction Link
<a href="https://codesandbox.io/s/codesandboxnuxt-e992r" target="_blank">https://codesandbox.io/s/codesandboxnuxt-e992r</a>

### Other comments
This is especially noticeable when transitioning to pages with a lot of elements (news feeds, images, etc)

### LINKED ISSUE #8105: [Feature Request] Hide main scrollbar when dialog is open
### Environment
**Vuetify Version:** 2.0.1
**Vue Version:** 2.6.10
**OS:** Windows 10

### Steps to reproduce
Click the pink button to open dialog. 

### Expected Behavior
The main page which has a scroll bar, should hide when the dialog is open. This is because you shouldn't be able to scroll on the main page anyways, and it gives a little extra room horizontally (or vertically) for the dialog. Plus since the dialog is almost full width, it looks bad to see 2 scroll bars next to each other. 

### Actual Behavior
The main scroll bar still appears when dialog is open.

### Reproduction Link
<a href="https://codepen.io/anon/pen/VomBbd" target="_blank">https://codepen.io/anon/pen/VomBbd</a>

### Other comments
Can it be changed to hide the main page scrollbars (up/down/left/right) while dialog is open?

### LINKED ISSUE #9595: Remove internalActivator
image

### LINKED ISSUE #9632: [Bug Report] v-for order is not preserved when used inside an #activator slot.
### Environment
**Vuetify Version:** 2.1.9
**Vue Version:** 2.6.10
**Browsers:** Firefox 70.0
**OS:** Linux x86_64

### Steps to reproduce
Use a v-for in an #activator slot.

### Expected Behavior
v-for displayed in order (One, Two, Three, Four)

### Actual Behavior
The order of the items appears random (One, Four, Three, Two).

### Reproduction Link
<a href="https://codepen.io/chiln/pen/NWWzXpE?editors=1010" target="_blank">https://codepen.io/chiln/pen/NWWzXpE?editors=1010</a>

### Other comments
John provided a work around:  Wrap the v-for in a root element <div>

https://codepen.io/johnjleider/pen/vYYrpMr?editors=1011

Given the workaround, This is not a showstopper, but fixing would avoid confusion & wasted time for new folks.

### LINKED ISSUE #11263: [Bug Report] Dialog flickering when being opened
### Environment
**Vuetify Version:** 2.2.26
**Last working version:** 2.2.10
**Vue Version:** 2.6.11
**Browsers:** Chrome 81.0.4044.129
**OS:** Windows 10

### Steps to reproduce
Simply opening any dialog causes the overlay to flicker. I have noticed that sometimes if you keep opening the dialog over and over again, the problem eventually disappears. But that is quite rare.

### Expected Behavior
When opening a dialog it should not flicker.

### Actual Behavior
When opening a dialog it flickers.

### Reproduction Link
<a href="https://codepen.io/tahnik/pen/abvymaZ" target="_blank">https://codepen.io/tahnik/pen/abvymaZ</a>

KaelWD:
Closes #1068
Closes #4807 (`class` now applies to the root teleported element)
Fixes #6425
Closes #6764
Fixes #7422
Closes #8105
Closes #9595
Fixes #9632
Fixes #11263

Some unrelated changes:
- Fixed proxiedModel generic arguments
- Added MaybeTransition component to badge and img
- Sheet default colour is `surface`

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container style="height: 2500px">
      <v-dialog>
        <template #activator="{ isActive, props }">
          <v-btn v-bind="props">{{ isActive ? 'Close' : 'Open' }} dialog</v-btn>
        </template>

        <v-sheet width="800" height="800">
          <DialogContent />
          <input type="text" placeholder="input 1">
          <input type="text" placeholder="input 2">
          <input type="text" placeholder="input 3">
          <input type="text" placeholder="input 4">

          <v-dialog>
            <template #activator="{ isActive, props }">
              <v-btn v-bind="props">{{ isActive ? 'Close' : 'Open' }} dialog</v-btn>
            </template>

            <v-sheet width="200" height="200">
              Nested dialog
            </v-sheet>
          </v-dialog>
        </v-sheet>
      </v-dialog>

      <div style="height: 500px; outline: 1px solid black; overflow-y: auto; position: relative">
        <div style="height: 400px" />
        <v-dialog absolute attach>
          <template #activator="{ isActive, props }">
            <v-btn v-bind="props">{{ isActive ? 'Close' : 'Open' }} dialog</v-btn>
          </template>

          <v-sheet class="d-flex" width="200" height="200">
            <DialogContent />
          </v-sheet>
        </v-dialog>
        <div style="height: 800px" />
      </div>

      <v-dialog attach persistent>
        <template #activator="{ isActive, props }">
          <v-btn v-bind="props">{{ isActive ? 'Close' : 'Open' }} dialog</v-btn>
        </template>
        <template #default="{ isActive }">
          <v-sheet class="d-flex" width="200" height="200">
            <v-btn icon="$close" @click="isActive.value = false" />
          </v-sheet>
        </template>
      </v-dialog>

      <v-btn :ripple="false" @click="active = !active">{{ active ? 'Close' : 'Open' }} dialog</v-btn>
      <v-dialog v-slot="{ isActive }" v-model="active">
        <v-sheet class="d-flex" width="200" height="200">
          <v-btn icon="$close" @click="isActive.value = false" />
        </v-sheet>
      </v-dialog>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    components: {
      DialogContent: {
        setup (props, { attrs }) {
          return () => 'Dialog content'
        },
      },
    },

    data: () => ({
      active: false,
    }),
  }
</script>
```
</details>


packages/vuetify/src/components/VOverlay/_variables.scss
```diff
@@ -1,3 +1,3 @@
 @import '../../styles/styles.sass';
 
-$overlay-transition: $primary-transition, z-index 1ms !default;
+$overlay-opacity: 32%;
```
johnleider, line 3: in the future this could probably use disabled opacity which is .38

packages/vuetify/src/components/VSheet/VSheet.tsx
```diff
@@ -19,7 +19,10 @@ export default defineComponent({
   name: 'VSheet',
 
   props: makeProps({
-    color: String,
+    color: {
```
johnleider, line 22: remove, if sheet doesn't have a bg that's a bug

packages/vuetify/src/util/easing.ts
```diff
@@ -0,0 +1,3 @@
+export const standardEasing = 'cubic-bezier(0.4, 0, 0.2, 1)'
```
johnleider, line 1: move these into variables?

johnleider (changes requested):

packages/vuetify/src/components/VSheet/VSheet.tsx
KaelWD: It is a bug, but dialog uses sheet for its content so this was needed.

packages/vuetify/src/components/VOverlay/_variables.scss
KaelWD: material.io uses 32% for scrims

packages/vuetify/src/components/VDialog/VDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+      default: true,
+    },
+    scrollable: Boolean,
+    modelValue: Boolean,
+    ...makeDimensionProps({ width: 'auto' }),
+    ...makeTransitionProps({
+      transition: { component: VDialogTransition },
+    }),
+  }),
+
```
nekosaur, line 38: declare emits

packages/vuetify/src/components/VOverlay/VOverlay.tsx
```diff
@@ -0,0 +1,278 @@
...
+    },
+    origin: [String, Object] as Prop<string | Element>,
+    scrim: {
+      type: [String, Boolean],
+      default: true,
+    },
+    ...makeTransitionProps(),
+  }),
+
+  emits: ['click:outside'],
```
nekosaur, line 151: use object syntax

packages/vuetify/src/composables/teleport.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (_target === true) return undefined
+
+    const targetElement =
+      _target === false ? document.body
+      : typeof _target === 'string' ? document.querySelector(_target)
+      : _target
+
+    if (targetElement == null) {
+      warn(`Unable to locate target ${_target}`)
```
nekosaur, line 16: Are we using this or our own warn method?

packages/vuetify/src/components/VOverlay/VOverlay.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  setup (props, { slots, attrs, emit }) {
+    const isActive = useProxiedModel(props, 'modelValue')
+    const { teleportTarget } = useTeleport(toRef(props, 'attach'))
+    const { themeClasses } = provideTheme()
```
nekosaur, line 156: use `useTheme`

nekosaur (changes requested):
I'm using the provided playground

- Nested dialog (together with first dialog) closes when clicking inside of it.
- When nested dialog is opened there is a console error
image
- The overflow/padding of <html> is inconsistent/not working correctly? When opening first dialog, the styles are set. When opening the nested dialog, padding-right is set to zero. If dialogs are closed and first dialog is opened again, padding-right is still zero.
- Is it expected behavior that overlay scrim in the overflowed element does not cover the entire height of it when scrolled? Or should the overflow be hidden here as well?

packages/vuetify/src/composables/teleport.ts
KaelWD: Idk, this has the component stack trace but says Vue instead of vuetify

packages/vuetify/src/components/VOverlay/VOverlay.tsx
KaelWD: Overlay is moved to the root of body so it won't inherit anything without this.

KaelWD:
> Nested dialog
Stack manager is needed for this to work properly, coming later with menu probably

> overlay scrim in the overflowed element 
Idk what we can do about that, maybe have to read scrollHeight of the parent first?

packages/vuetify/src/components/VOverlay/VOverlay.tsx
nekosaur: Hm. Am I misunderstanding this?

> This also means that injections from a parent component work as expected, and that the child component will be nested below the parent component in the Vue Devtools, instead of being placed where the actual content moved to.

https://v3.vuejs.org/guide/teleport.html#using-with-vue-components

KaelWD: Misunderstood what useTheme did

packages/vuetify/src/components/VDialog/VDialog.tsx
KaelWD: Should I declare `click:outside` here too even though it isn't actually emitted by this component?

KaelWD:
> overlay scrim in the overflowed element

We can either use `position: sticky`, which doubles the scroll height when the dialog is open, or hide overflow for this element too.

packages/vuetify/src/components/VDialog/VDialog.tsx
nekosaur: No that shouldn't be necessary

nekosaur:
> We can either use `position: sticky`, which doubles the scroll height when the dialog is open, or hide overflow for this element too.

if block scroll strategy is used is should hide overflow here also, or?

slightly related, how would one set a different strategy? are you relying on passing the attrs from dialog?

KaelWD:
> how would one set a different strategy?

`scroll-strategy="close"`. We can add support for custom strategies later if we want.

packages/vuetify/src/composables/teleport.ts
johnleider: This is something we can add for beta testing in July

johnleider (approved):

PR #13719, opened 2021-05-28
fix(VListItem): allow default color to change when inactive

### LINKED ISSUE #9285: [Bug Report] Impossible to set color of List Item text
### Environment
**Vuetify Version:** 2.1.1
**Vue Version:** 2.6.10
**OS:** Mac OS 10.14.6

### Steps to reproduce
Try to set color (text/icon) of not active list item with available tools:
- `color` prop on list item;
- `active-class` prop on list item;
- `class` on list item;
- css;
- css with `!important`;
- inline css;

### Expected Behavior
image

### Actual Behavior
image

Impossible to replace light* text color of dark list item without inline css.
With inline css active tile is effected and cannot be styled to have dark* text color.

*light = the white color used for light theme backgrounds and dark theme texts.
*dark = the black color used for dark theme backgrounds and white theme texts.

image

### Reproduction Link
<a href="https://codepen.io/JesusCrow/pen/OJJPqVb?editors=0100" target="_blank">https://codepen.io/JesusCrow/pen/OJJPqVb?editors=0100</a>

### Other comments
I tried everything I could think up to create a list item that has a certain color for text and icon. I think that this is impossible without inline css. Using inline css disables me from styling active item text.

Tofandel:
## Description

Fixes #9285 by removing the !important keyword in the vuetify style sheet ~~and adding a `base-color` prop~~

## Motivation and Context

It is currently not possible to change the color of a v-list-item via an hex `color="#CCC"` prop, it is only possible to do it via a tricky css rule (needs to be very specific, and the selector with this rule is already too specific with an `!important` keyword, 

## How Has This Been Tested?

Visually

#### Before
image

#### After
image

## Markup:

<details>

```vue
<template>
      <v-list dense>
        <v-list-item-group v-model="item">
          <template v-for="(item, i) in items">
          <v-list-item
            :key="i+'0'"
            :value="0"
            v-bind="item"
          >
            <v-list-item-icon>
              <v-icon>settings</v-icon>
            </v-list-item-icon>
            <v-list-item-content>
              <v-list-item-title>normal {{ item }}</v-list-item-title>
            </v-list-item-content>
          </v-list-item>
          <v-list-item
            :key="i+'1'"
            :value="1"
            v-bind="item"
          >
            <v-list-item-icon>
              <v-icon>settings</v-icon>
            </v-list-item-icon>
            <v-list-item-content>
              <v-list-item-title>active {{ item }}</v-list-item-title>
            </v-list-item-content>
          </v-list-item>
          <v-list-item
            dark
            :key="i+'0d'"
            :value="0"
            v-bind="item"
          >
            <v-list-item-icon>
              <v-icon>settings</v-icon>
            </v-list-item-icon>
            <v-list-item-content>
              <v-list-item-title>normal dark {{ item }}</v-list-item-title>
            </v-list-item-content>
          </v-list-item>
          <v-list-item
            dark
            :key="i+'1d'"
            :value="1"
            v-bind="item"
          >
            <v-list-item-icon>
              <v-icon>settings</v-icon>
            </v-list-item-icon>
            <v-list-item-content>
              <v-list-item-title>active dark {{ item }}</v-list-item-title>
            </v-list-item-content>
          </v-list-item>
          </template>
        </v-list-item-group>
      </v-list>
</template>

<script>
export default {
data: () => ({
    item: 1,
    items: [
      { color: "#228822" },
      { color: 'amber' },
      { baseColor: "#228822" },
      { baseColor: 'amber' },
      {}
    ],
  }),
}
</script>
```
</details>

## Types of changes

## Checklist:

- [x] I added relevant changes to the documentation


Tofandel:
Should the normal icon's color change as well maybe?

KaelWD:
Yes, but this was intentional so inactive items aren't highlighted (see https://vuetify-lzjyu0p1p-vuetifyjs1.vercel.app/en/getting-started/installation/)
image
The text here is supposed to be black unless it's an active route.

Tofandel:
Okay let me see if I can fix that without an important keyword

Tofandel:
I think to solve this it would make sense to have a `color` prop and an `active-color` prop or to avoid a breaking change a `base-color` prop and `color` prop (which only applies when the list item is active)

Tofandel:
I removed the color class when the item is not active, which fixes the demonstrated issue

Also took the liberty of adding the base-color prop which is the prop that controls the color when it's not active (and the color prop defaults to it if its not set)

KaelWD:
This also changes icons to black when inactive:
image

image

Tofandel:
I changed it to only inherit icon color if the base color is set

I also added a colorful example in the doc
image

KaelWD (approved):

KaelWD (approved):

johnleider (changes requested):
I think the naming is confusing. If we have a visual change based upon the state of an item, it should have a name that reflects that.

Personally, I'd rather have the parent dictate a **color** and **active-color** and leave `v-list-item` alone.

I'm actively working on v3 lists and I'll chime in more as I experiment more.

Tofandel:
I agree that adding a `color` and `active-color` prop to the v-list would be a good idea to apply it to all the v-list-items

But it's a must to also be able to change the color of some of the list item individually, where you can override this

And sadly on the v-list-item the naming can't be changed, it would be a big breaking change

KaelWD:
`active-color` can't be done in v2 without being a breaking change, I'm ok with this as a stopgap then change it again in v3. 

> I'd rather have the parent dictate a color and active-color and leave v-list-item alone.

There does need to be some way to change the colour of an individual item imo

KaelWD:
Hey @Tofandel, we discussed this in today's meeting and decided not to implement the new prop in v2 as it would be immediately deprecated. To fix #9285 in v2 we can accept removing the active class when not active though.

Tofandel:
So just the fix without the base-color prop?

packages/vuetify/src/components/VList/VListItem.ts
```diff
@@ -35,6 +35,8 @@ interface options extends ExtractVue<typeof baseMixins> {
   isInList: boolean
   isInMenu: boolean
   isInNav: boolean
+  genAttrs (): object
+  click (e: MouseEvent | KeyboardEvent): void
```
KaelWD, line 39: This shouldn't be necessary.

johnleider (approved):

johnleider:
Can you take a look at `v-list` as of v3-alpha.5 and compare the styling. We went with the prop name **active-color**.

KaelWD (approved):

Tofandel:
@johnleider The doc site doesn't seem to include it there https://next.vuetifyjs.com/en/components/application/ any other way to check except building it locally? Did you also add a similar snippet in the doc like: https://github.com/vuetifyjs/vuetify/pull/13719/commits/54f8d4f26e92175d6f337d188c4fbf79edddcc78 ?

KaelWD:
@Tofandel I'm going to have to revert this, `color` no longer affects active route links: #14186

Feel free to open a new PR if you have a solution that works for this case as well.

Tofandel:
@KaelWD that seems weird, isn't `this.isActive` data true for router links? Because I swear this was working

Tofandel:
So yeah this.isActive is false even if the link is current is this normal? Isn't there something missing in https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/routable/index.ts

Tofandel:
One workaround is to check for the activeClass on the element instead of the isActive on every route change (or to set the isActive to true) because there doesn't seem to be any event or data available to save this externally in vue router..

Tofandel:
Well there you go.. It's already implemented https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/routable/index.ts#L153 The toggle function is empty and should be `this.isActive = true`

Tofandel:
@KaelWD So if you can do it quickly you can rerevert and change the line in the routable mixin, otherwise I'll make a new PR

Tofandel:
Indeed made a new PR, the onRouteChange should be called on mounted, as for the activeClass to set it back to false no need, you actually just need to call toggle, only if the class boolean is different than the isActive value, effectively trully reflecting isActive on every route change
The toggle also needs to be redefined in the VListItem because it's overriden by the groupable mixin

PR #13812, opened 2021-06-17
fix(types): add missing `goTo` types

jacekkarczmarczyk:
fixes #9648

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [x] My code follows the code style of this project.
- [x] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #13838, opened 2021-06-23
fix(types): declare vuetify/lib/services/goto module

jacekkarczmarczyk:
## Motivation and Context
#13812 was not enough

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/types/lib.d.ts
+++ b/packages/vuetify/types/lib.d.ts
@@ -14,7 +14,7 @@ declare module 'vuetify/lib' {
 }
 
 declare module 'vuetify/lib/components' {
-  import { VueConstructor, DirectiveOptions } from 'vue'
+  import { VueConstructor } from 'vue'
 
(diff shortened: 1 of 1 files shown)


KaelWD:
Why would you need to import this separately? It's already on the vuetify instance.

jacekkarczmarczyk:
So that I don't need to import vuetify instance, mainly for composition api
Anyway feel free to close if using $vuetify.goTo is obligatory

zoton2:
Another point to make is that the documentation example for using "use with router" also won't resolve types with the current setup either: https://vuetifyjs.com/en/features/scrolling/#use-with-router

KaelWD:
OK I wasn't aware we were documenting it that way, if that's the case then this should probably go in. 

`/es5` is legacy, we do have some type definitions for it but it was mostly kept to ease the 2.0 upgrade slightly.

KaelWD (approved):

PR #13905, opened 2021-07-08
feat: use APCA to determine theme text colors

KaelWD:
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <div class="d-flex flex-column align-start">
        <template v-for="value, color in colors" :key="color">
          <v-btn :color="color" class="my-2">{{ color }}, {{ value }}</v-btn>
        </template>
      </div>
    </v-container>
  </v-app>
</template>

<script>
  import { useTheme } from '../src/composables/theme'
  import { computed } from 'vue'

  export default {
    setup (props) {
      const theme = useTheme(props)
      const colors = computed(() => {
        return Object.fromEntries(
          Object.entries(theme.getTheme(theme.current.value).colors)
            .filter(([name]) => !name.startsWith('on'))
        )
      })

      return {
        colors,
      }
    },
  }
</script>
```


johnleider (approved):
Unless the TODO is for future and not part of this PR, lgtm.

Myndex:
Hello, just as an FYI, APCA is not available as an npm package

    npm i apca-w3

And there is also now **Bridge PCA** which is backwards compatible with WCAG 2 contrast, but using APCA technology.

    npm i bridge-pca

For questions of comments, please visit[ the GitHub repo.](https://github.com/Myndex/SAPC-APCA/discussions)

Issue #9376, opened 2019-10-16
[Bug Report] VTreeNode: Null or Undefined children causing trouble with event bubbling


GaetanWi:
### Environment
**Vuetify Version:** 2.1.4
**Vue Version:** 2.6.10
**Browsers:** Chrome 77.0.3865.120
**OS:** Windows 10

### Steps to reproduce
Open console on the codepen. The exception is happening at `const children = getObjectValueByPath(item, this.itemChildren, [])` (line 200) in the `buildTree` method. 

Messing up the behaviour of the item with `null` or `undefined` children value

### Expected Behavior
Should not crash since it's preventing event bubbling

### Actual Behavior
Crashing causing event bubbling issues since properties are not added to the cache

### Reproduction Link
<a href="https://codepen.io/GaetanWcz/pen/MWWeEGM?editors=1111" target="_blank">https://codepen.io/GaetanWcz/pen/MWWeEGM?editors=1111</a>

### Other comments
Was broken before my modification https://github.com/vuetifyjs/vuetify/pull/9082 and still is. (The intended fix works fine, when every one has no children or an empty children it works fine)

GaetanWi:
Not really sure how an array could be undefined with good backend architecture though.

jacekkarczmarczyk:
Happens only with `null`, `undefined` works fine

johnleider:
It works because undefined is an absence of a reference whereas null is a reference to nothing. That's why the method returns the null which attempts to get iterated.

Use undefined, an array or don't define it.

If you have any additional questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

JasonLandbridge:
@johnleider, thanks for the tip!

I'm using an object that has itself as children, but the most deeply nested children would show to be expandable. Even though they shouldn't be and looping over those instances and setting those to undefined fixed those nodes. 

```Typescript
export interface FooObject{
  id: number;
  key: number;
  children: FooObject[];
}

// Api call
response.children.forEach((child) => {
	child.children.forEach((grandChild) => {
		grandChild.children = undefined;
	});
});

```

Any way this "setting to undefined" could be added to: [v-treeview/#load-children](https://vuetifyjs.com/en/api/v-treeview/#load-children)?

PR #13913, opened 2021-07-10
fix(VTreeView): Null check when building tree from children

JasonLandbridge:
## Description

When using VTreeView and the [item-children](https://vuetifyjs.com/en/api/v-treeview/#props-item-children) is null or undefined, then a `children.map` in `VTreeview.ts` will cause an error. 

image

## Motivation and Context
Adding a simple null/undefined check and then setting it to `[]` fixes this problem. 
[[Bug Report] VTreeNode: Null or Undefined children causing trouble with event bubbling #9376 (comment)](https://github.com/vuetifyjs/vuetify/issues/9376#issuecomment-572765655) 

## How Has This Been Tested?
Tested in my current project by adding the same code and this fixes the problem. Children which are undefined are not expandable. 

## Markup:

<details>

```vue
<template>
  <v-treeview :items="items" />
</template>

<script>
  export default {
    data: () => ({
      items: [
        {
          id: 'a',
          name: 'This one is  active with no exception',
        },
        {
          id: 'b',
          name: 'So it this one, you can see it as active',
          children: [
            {
              id: 'a',
              name: 'This one is  active with no exception',
            },
            {
              id: 'b',
              name: 'So it this one, you can see it as active',
              children: [],
            },
            {
              id: 'c',
              name: 'But this one is crashing during the build tree so its messing up',
              children: null,
            },
            {
              id: 'd',
              name: 'and so is tzhis one',
              children: undefined,
            },
          ],
        },
        {
          id: 'c',
          name: 'But this one is crashing during the build tree so its messing up',
          children: null,
        },
        {
          id: 'd',
          name: 'and so is tzhis one',
          children: undefined,
        },
      ],
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VTreeview/VTreeview.ts
+++ b/packages/vuetify/src/components/VTreeview/VTreeview.ts
@@ -201,14 +201,10 @@ export default mixins(
       for (let i = 0; i < items.length; i++) {
         const item = items[i]
         const key = getObjectValueByPath(item, this.itemKey)
-        const children = getObjectValueByPath(item, this.itemChildren, [])
+        const children = getObjectValueByPath(item, this.itemChildren, []) ?? []
         const oldNode = this.nodes.hasOwnProperty(key) ? this.nodes[key] : {
           isSelected: false, isIndeterminate: false, isActive: false, isOpen: false, vnode: null,
         } as NodeState
-
-        if(!children){
-          children = []
-        }
 
         const node: any = {
           vnode: oldNode.vnode,
--- a/packages/vuetify/src/components/VTreeview/VTreeview.ts
+++ b/packages/vuetify/src/components/VTreeview/VTreeview.ts
@@ -201,7 +201,7 @@ export default mixins(
       for (let i = 0; i < items.length; i++) {
         const item = items[i]
         const key = getObjectValueByPath(item, this.itemKey)
-        const children = getObjectValueByPath(item, this.itemChildren, [])
+        const children = getObjectValueByPath(item, this.itemChildren, []) ?? []
         const oldNode = this.nodes.hasOwnProperty(key) ? this.nodes[key] : {
           isSelected: false, isIndeterminate: false, isActive: false, isOpen: false, vnode: null,
         } as NodeState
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
1. this will check also for `0`, `false`, `''`, such cases i'd rather treat as an invalid value and throw the error. I'd suggest `const children = getObjectValueByPath(item, this.itemChildren) ?? []`
2. https://github.com/vuetifyjs/vuetify/issues/9376#issuecomment-572765655 - however I personally don't agree with it and think that `null` should also mean "no children"

JasonLandbridge:
> 1. this will check also for `0`, `false`, `''`, such cases i'd rather treat as an invalid value and throw the error. I'd suggest `const children = getObjectValueByPath(item, this.itemChildren) ?? []`
> 2. [[Bug Report] VTreeNode: Null or Undefined children causing trouble with event bubbling #9376 (comment)](https://github.com/vuetifyjs/vuetify/issues/9376#issuecomment-572765655) - however I personally don't agree with it and think that `null` should also mean "no children"

Yes that is what I had initally but I wasn't sure if the null coalesce operator would work here.

Changed it!

packages/vuetify/src/components/VTreeview/VTreeview.ts
```diff
@@ -201,7 +201,7 @@ export default mixins(
       for (let i = 0; i < items.length; i++) {
         const item = items[i]
         const key = getObjectValueByPath(item, this.itemKey)
-        const children = getObjectValueByPath(item, this.itemChildren, [])
+        const children = getObjectValueByPath(item, this.itemChildren, []) ?? []
```
jacekkarczmarczyk, line 204:
```suggestion
        const children = getObjectValueByPath(item, this.itemChildren) ?? []
```

JasonLandbridge:
@jacekkarczmarczyk , thanks for reviewing and helping out.! Is there something that needs to happen before this can be approved and merged?

JasonLandbridge:
Hey @johnleider, Is there anything that needs to change in order for this pull request to get merged?

johnleider (approved):

JasonLandbridge:
@johnleider, Thanks for taking a look! Is it possible this can be merged in Vuetify V2 since its very minor?

I'm asking because I now see the v3.0.0 milestone added.

johnleider:
Good catch, just a mistake. Don't be alarmed 🔔

JasonLandbridge:
> Good catch, just a mistake. Don't be alarmed 🔔

Awesome, thanks you very much!

jacekkarczmarczyk (commented):
lgtm, I've added the playground

PR #13996, opened 2021-07-27
feat: update VTextField to v3

KaelWD:
## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const value = useProxiedModel(props, 'modelValue')
+
+    const internalDirty = ref(false)
+    const isDirty = computed(() => {
+      return internalDirty.value || !!value.value || dirtyTypes.includes(props.type)
+    })
+
+    const { intersectionRef, isIntersecting } = useIntersectionObserver((_, observer) => {
+      if (props.autofocus && isIntersecting) {
+        const el = isComponentInstance(intersectionRef) ? intersectionRef.$el : intersectionRef
```
nekosaur, line 51: not sure you need this check here? we are very definitely setting the `intersectionRef` value to an instance of HTMLInputElement, so it will never be a component instance.

packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -0,0 +1,89 @@
...
+  setup (props, { attrs, slots }) {
+    const value = useProxiedModel(props, 'modelValue')
+
+    const internalDirty = ref(false)
+    const isDirty = computed(() => {
+      return internalDirty.value || !!value.value || dirtyTypes.includes(props.type)
+    })
+
+    const { intersectionRef, isIntersecting } = useIntersectionObserver((_, observer) => {
+      if (props.autofocus && isIntersecting) {
```
nekosaur, line 50: 1. `isIntersecting` is a ref so should be accessed with `.value`
2. the `useIntersectionObserver` API is maybe a bit sloppy here. isIntersecting is only extracted from the return to be accessed inside the callback where it could also be available. perhaps callback signature should be `(isIntersecting: boolean, entries: ..., observer: ...) => void`

PR #14026, opened 2021-08-04
feat: create validation composable

### LINKED ISSUE #3530: [Feature Request] Don't show validation errors until validate() is called
### New Functionality
It will allow form validations happen only when this.$ref.form.validate() is called.
This avoids form errors be displayed when a field loses focus.
A example like this can be seen on google contacts. No errors is displayed when a field loses focus, but when save button is clicked

### Improvements
Forms will be more clean if no errors are displayed.

It avoids user to be worried about errors during filling in the forms

### LINKED ISSUE #4573: [Enhancement] Emit form validation errors
In 1.0 it was possible to get all validation errors from a form by doing something like this.

```
const errors = []
this.$refs.form.inputs.forEach(e => if(e.hasError) errors.push(e.errorBucket))
```

This is not possible in 1.1 as registered inputs do not provide hasError or errorBucket.

Wanting all validation errors from a form seems like a common enough issue that we should support it natively.

Emit an event with all errors

### LINKED ISSUE #8698: [Feature Request] Validate a single field
### Problem to solve
I need to just validate one of the fields in the form to do some processing like `.validate('email')` to used to send a E-mail verification code in the register form.
According to the documentation, I used `vee-validate`.But I encountered another problem: https://github.com/baianat/vee-validate/issues/2219 
If vuetify has this feature, I don't need to use other authentication components.It It took me too much time.

### Proposed solution
```html
<v-form ref="form">
  <v-text-field label="E-mail" name="email">
  </v-text-field>
  <v-text-field label="Mobile" name="mobile">
  </v-text-field>
  <v-text-field label="Addr." name="address">
  </v-text-field>
</v-form>
```
```javascript
// single
$refs.form.validate('email')
// multiple
$refs.form.validate(['email', 'mobile'])
// All
$refs.form.validate() // existing
```

### LINKED ISSUE #11163: [Feature Request] v-rating Rules Validation
### Problem to solve
Unable to validate whether the v-rating question has been answered by the user

### Proposed solution
Accepts an array of functions that take an input value as an argument and return either true / false or a string with an error message

johnleider:
fixes #3530
fixes #4573 
partially resolves #8698
fixes #11163

**TODO:**

* [x] unit tests (need help with this one)
* [x] publicly expose register/unregister
* [x] implement remaining props, **lazyValidation**, **readonly**, **disabled**, etc
* [x] publicly expose form functions, submit, reset, clear

## Description
* ported **v-form** to v3
* ported **validatable** to v3
* added new component **v-validation**
* added new composable **form**
* added new utility **wrapInPromise**
* implemented **validation** composable into **v-field**

### Composables
* **validation:** (previously validatable)
  * added new state variable **isPristine** which reflects whether **validate** has been invoked. (reverts to true when **reset** or **clear** is invoked)
  * added new state variable **isValidating** that denotes validation is currently occurring

* **form:**
  * new composable for propagating form registration/unregistration

### Components
* **v-form:**
  * added new **fastFail** prop that stops the validation chain when any rules fail
  * added support for promise based rules (validated synchronously)

* **v-validation:**
  * new component that uses the **validation** composable to enable inline template validation

## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="py-16 px-16 my-16">
      <!-- <v-btn @click="loading = !loading">Click me</v-btn>
      <v-btn @click="error = !error">Click me valid</v-btn> -->

      <!-- <v-messages :model-value="messages" /> -->

      <!-- <v-field label="Foobar" hint="Foobars">
        <template #default="{ props }">
          <input
            v-model="model"
            type="text"
            v-bind="props"
          >
        </template>
      </v-field> -->

      <!-- <v-field loading color="primary" /> -->

      <!-- <v-progress-linear indeterminate color="primary" /> -->

      <div class="my-12" />

      <v-form v-model="valid">
        <template #default="{ isValidating, clear }">
          <v-text-field
            v-model="model"
            :rules="rules"
            hint="Foo"
            label="Text field"
            :loading="isValidating.value || loading"
            :persistent-hint="hint"
            clearable
            counter="25"
            :persistent-clear="clears"
            :persistent-counter="counter"
            :append-inner-icon="valid ? 'mdi-check-circle' : undefined"
          />
          <!-- <v-validation
            v-model="model"
            :rules="rules"
            max-errors="4"
          >
            <template
              #default="{
                errorMessages,
                isValidating,
                isValid,
                reset,
              }"
            >
              <v-field
                label="Foobar"
                :loading="isValidating.value"
                :dirty="!!model"
                :error="loading || isValid.value === false"
                :readonly="isValidating.value"
                @click:append="reset"
              >
                <template #default="{ props }">
                  <input
                    v-model="model"
                    type="text"
                    v-bind="props"
                  >
                </template>

                <template #details>
                  {{ errorMessages }}
                </template>
              </v-field>
            </template>
          </v-validation> -->
          <v-btn @click="loading = !loading">Loading</v-btn>
          <v-btn @click="hint = !hint">Hint</v-btn>
          <v-btn @click="counter = !counter">Counter</v-btn>
          <v-btn @click="clears = !clears">Clearable</v-btn>
          <div class="my-6" />

          <v-btn type="input">Submit</v-btn>
          <v-btn type="reset">Reset</v-btn>
          <v-btn @click="clear">Clear</v-btn>
        </template>
      </v-form>

      Model value and valid state
      <pre>
        Model: {{ model }}
        Valid: {{ valid === null ? 'indeterminate' : valid }}
      </pre>
    </div>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  let rule = null

  export default {
    setup () {
      const model = ref('Text field model')
      const clears = ref(false)
      const counter = ref(false)
      const hint = ref(false)
      const messages = ref(['Foo', 'Bar'])
      const messages1 = ['Fizz']
      const messages2 = ['Buzz', 'Buzzer', 'Feet']
      const error = ref(false)
      const valid = ref(null)
      const loading = ref(false)
      const rules = [
        v => !!v || 'Required field 2',
        v => !!v || 'Required field 3',
        v => new Promise(resolve => {
          clearTimeout(rule)

          rule = setTimeout(() => {
            resolve((!!v && v.length > 12) || 'Promise required')
          }, 5000)
        }),
        v => !!v || 'Required field 4',
      ]

      function onMsgClick () {
        messages.value = messages.value.length === messages1.length
          ? messages2
          : messages1
      }

      return {
        model,
        loading,
        rules,
        valid,
        error,
        clears,
        counter,
        hint,
        messages,
        onMsgClick,
        onSubmit (e) {
          // eslint-disable-next-line no-console
          console.log('submitting')
        },
      }
    },
  }
</script>

```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/composables/__tests__/validation.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    errorCount: number | null,
+    expected: number,
+  ) => {
+    const model = ref('')
+    const props = {
+      errorCount,
+      rules: ['foo', 'bar', 'fizz', 'buzz'],
+    } as ValidationProps
+
+    const { errorMessages, validate } = useValidation(props, model)
```
KaelWD, line 52:
```suggestion
    const { errorMessages, validate } = useValidation({
      errorCount,
      rules: ['foo', 'bar', 'fizz', 'buzz'],
    }, ref(''))
```

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+// Types
+import type { MaybeRef } from '@/util/helpers'
+import type { PropType } from 'vue'
+
+export type ValidationRule = string | ((value: any) => string | boolean)
+export type ValidationRules = ValidationRule[]
+export type ValidationResult = string | boolean | Promise<string | boolean>
+
+export interface ValidationProps {
+  errorCount: string | number
```
nekosaur, line 14: I found `errorCount` a big ambiguous in what it does, had to look at the code. Perhaps `maxErrors` or something similar is clearer?

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
+// Utilities
+import { propsFactory, wrapInPromise } from '@/util'
+import { ref, unref, watch } from 'vue'
+
+// Types
+import type { MaybeRef } from '@/util/helpers'
+import type { PropType } from 'vue'
+
+export type ValidationRule = string | ((value: any) => string | boolean)
+export type ValidationRules = ValidationRule[]
```
nekosaur, line 10: This type is a bit superfluous

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+  },
+  rules: {
+    type: Array as PropType<ValidationRules>,
+    default: () => ([]),
+  },
+})
+
+export function useValidation (props: ValidationProps, value: MaybeRef<any>) {
+  const errorMessages = ref<string[]>([])
+  const isPristine = ref(true)
```
nekosaur, line 31: Is pristine related to validation? I see that it flips on `isValid`, but shouldn't it flip as soon as input value changes, regardless of if it's valid or not? Or perhaps I'm misunderstanding what it represents.

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+  const isValid = ref<null | boolean>(null)
```
nekosaur, line 32: This should maybe be `computed(() => errorMessagages.value.length === 0)`? For example, right now manually adding a message to `errorMessages` array will not flip `isValid` to false.

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+  const isValidating = ref(false)
+  const errorCount = ref(0)
+  const unwrefed = unref(value)
```
nekosaur, line 35: I can't really make sense of why this is done, and why in function argument value is `MaybeRef`?

I think the input value should always be a `Ref`.

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+        break
+      }
+    }
+    isValid.value = errorCount.value > 0
+    isValidating.value = false
+
+    isValid.value = errorMessages.value.length === 0
+  }
+
+  async function reset () {
```
nekosaur, line 65: No reason for method to be async

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      if (typeof valid === 'string') {
+        errorMessages.value.push(valid)
+      }
+
+      if (errorCount.value >= (props.errorCount || 1)) {
+        break
+      }
+    }
+    isValid.value = errorCount.value > 0
```
nekosaur, line 59: ?

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      if (valid === true) continue
+
+      errorCount.value += 1
+
+      if (typeof valid === 'string') {
+        errorMessages.value.push(valid)
+      }
+
+      if (errorCount.value >= (props.errorCount || 1)) {
```
nekosaur, line 55: Why separate `errorCount` variable? Can just be `errorMessages.value.length`, or am I missing some logic?

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+    isValidating.value = true
+    for (const rule of props.rules) {
+      const handler = typeof rule === 'function' ? rule : () => rule
+      const valid = await wrapInPromise<ValidationResult>(handler(unwrefed.value))
+
+      if (valid === true) continue
+
+      errorCount.value += 1
+
+      if (typeof valid === 'string') {
```
nekosaur, line 51: If `valid` is something other than string, the error will get lost/ignored. E.g. `v => !!v || 123456789`. Should probably print error/warning if it's something other than string.

packages/vuetify/src/composables/__tests__/validation.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe('validation.ts', () => {
+  it.each([
+    [[], []],
+    [['foo'], ['foo']],
+    [[() => 'bar'], ['bar']],
+    [[() => true], []],
+    [[() => false], []],
+    [[() => Promise.resolve('fizz')], ['fizz']],
+    [[() => new Promise(resolve => resolve('buzz'))], ['buzz']],
```
nekosaur, line 19: You should be testing with proper rule functions.

packages/vuetify/src/composables/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  watch(isValid, () => isPristine.value = false)
+
+  async function validate () {
+    await reset()
+
+    isValidating.value = true
+    for (const rule of props.rules) {
+      const handler = typeof rule === 'function' ? rule : () => rule
+      const valid = await wrapInPromise<ValidationResult>(handler(unwrefed.value))
```
nekosaur, line 45: See comment above about why unref is taking place. But assuming an unref, then it's the raw value, so `.value` should not be used.

nekosaur (changes requested):

packages/vuetify/src/composables/validation.ts
johnleider: I added it so that users can determine if an input has been validated at least once. We did something similar in v2 with focusing and validation (I believe). Not committed to it.

packages/vuetify/src/composables/validation.ts
johnleider: Left there from testing

packages/vuetify/src/composables/validation.ts
johnleider: errorCount !== props.errorCount. Not all errors get pushed into messages so every error iterates the total number of errors (which is used for determining the rules validation loop breaks). This will be less confusing with the suggested prop name change you made.

packages/vuetify/src/composables/validation.ts
johnleider: This was done under the pretense that we would accept anything, hence MaybeRef.

packages/vuetify/src/composables/validation.ts
johnleider: explained here https://github.com/vuetifyjs/vuetify/pull/14026/files#r683731887

packages/vuetify/src/composables/validation.ts
nekosaur: Why should the composable accept something that is not a ref? I get that the actual input value could be anything, but that's `Ref<any / unknown>`.

packages/vuetify/src/composables/validation.ts
nekosaur: Then why set `isValid.value` again two lines down?

Anyway, now I understand the purpose of `errorCount` variable, but I don't think I agree with it. As I mentioned here https://github.com/vuetifyjs/vuetify/pull/14026/files#r683723806 I think it is a bad idea to silently ignore rules that don't have a string.

IMO we should be nice to developers here (and by extension end users) by warning that a rule is missing a string message. If developer really really want to fail an input without a visible message, they can do `v => !!v || ''`

packages/vuetify/src/composables/validation.ts
johnleider: Not disagreeing, I hadn't removed it yet.

packages/vuetify/src/composables/validation.ts
johnleider: The second isValid comparison was accidentally left in.

KaelWD (approved):

PR #14035, opened 2021-08-07
feat(filter.ts): add new composable

johnleider:
## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="ma-12 pa-12">
      <input v-model="query" type="text">

      <div />

      <v-btn @click="onClick">Click me</v-btn>

      <div class="my-6" />

      <strong>Items:</strong> {{ items.length }}

      <div />
      <strong>Filtered:</strong> {{ filtered.length }}

      <div class="my-6" />

      <v-btn v-for="(v, i) in filtered" :key="i">
        Button {{ v }}
      </v-btn>
    </div>
  </v-app>
</template>

<script>
  import { ref } from 'vue'
  import { useFilter } from '../src/composables/filter'

  export default {
    setup (props) {
      const items = ref(Array.from({ length: 100 * 10 }, (v, k) => `Foo-${k}`))
      const query = ref('1')
      const { filtered } = useFilter(props, items, query)

      function onClick () {
        items.value.push('fizzbuzz 1')
      }

      return {
        filtered,
        items,
        query,
        onClick,
      }
    },
  }
</script>
X

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/filter.ts
+++ b/packages/vuetify/src/composables/filter.ts
@@ -1,26 +1,60 @@
 // Utilities
 import { computed } from 'vue'
+import { consoleWarn, propsFactory } from '@/util'
 
 // Types
-import type { Ref } from 'vue'
+import type { PropType, Ref } from 'vue'
 
-export interface FilterProps {
-  filter?: (text: string, query: any) => boolean
+export function defaultFilter (text: string, query?: string) {
+  if (typeof query !== 'string') return true
+  if (typeof text !== 'string') return false
+
+  return text.toLocaleLowerCase().includes(query.toLocaleLowerCase())
 }
 
-export function useFilter (props: FilterProps, items: Ref<any[]>, query: Ref<any>) {
-  function filter (text: string, query: any) {
-    if (typeof props.filter === 'function') return props.filter(text, query)
+export interface FilterProps {
+  filterFn?: (item: any, query?: string) => boolean
+}
 
-    if (query == null || query === false) return true
+// Composables
+export const makeFilterProps = propsFactory({
+  filterFn: Function as unknown as PropType<FilterProps['filterFn']>,
+}, 'filter')
 
-    return text.toLocaleLowerCase().includes(query.value.toLocaleLowerCase())
-  }
+export function useFilter (
+  props: FilterProps,
+  items: Ref<unknown[]>,
+  query?: Ref<string | number>
+) {
+  const strQuery = computed(() => (
+    typeof query?.value !== 'string' &&
+    typeof query?.value !== 'number'
+  ) ? '' : String(query.value))
 
-  const filtered = computed(() => items.value.filter((item: any) => filter(item, query)))
+  const filtered = computed(() => {
+    if (!strQuery.value) return items.value
 
-  return {
-    filtered,
-    filter,
-  }
+    const array: (typeof items.value) = []
+
+    for (const item of items.value) {
+      if (props.filterFn) {
+        if (props.filterFn(item, strQuery.value)) {
+          array.push(item)
+        }
+      } else if (
+        typeof item === 'string' ||
+        typeof item === 'number'
+      ) {
+        if (defaultFilter(String(item), strQuery.value)) {
+          array.push(item)
+        }
+      } else {
+        consoleWarn(`The default filter function expects a string | number, received ${item}. Update the item value or use a custom filter.`)
+      }
+    }
+
+    return array
+  })
+
+  return { filtered }
 }
--- a/packages/vuetify/src/composables/filter.ts
+++ b/packages/vuetify/src/composables/filter.ts
@@ -1,0 +1,26 @@
+// Utilities
+import { computed } from 'vue'
+
+// Types
+import type { Ref } from 'vue'
+
+export interface FilterProps {
+  filter?: (text: string, query: any) => boolean
+}
+
+export function useFilter (props: FilterProps, items: Ref<any[]>, query: Ref<any>) {
+  function filter (text: string, query: any) {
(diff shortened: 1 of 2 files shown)


nekosaur:
Haven't taken a close look, but shouldn't this support objects by default? It's a pretty common use-case for our components as well

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,63 @@
...
+  },
+}, 'filter')
+
+export function useFilter (
+  props: FilterProps,
+  items: Ref<any[]>,
+  query?: Ref<string | number>,
+  filterKeys?: (string | string[]) | (FilterFunction | FilterFunction[]),
+) {
+  const keys = computed(() => wrapInArray(filterKeys))
```
nekosaur, line 36: making this `computed` here has no effect. if you want `filterKeys` to be reactive then the function needs to take `Ref<(string | string[]) ...>`

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,86 @@
...
+}, 'filter')
+
+export function filterItems (
+  items: Record<string, any>[],
+  keys: string[],
+  query?: string,
+  filter?: {
+    default?: FilterFunction
+    mode?: FilterMode
+    keyFilters?: Record<string, FilterFunction>
```
nekosaur, line 40: Thinking of how we could support both complete custom filter fn and custom default key filter fn with a decent API, I wonder if we could do something like

```
// props
filterFn: {
  type: Function,
  default: defaultFilterFn
},
keyFilters: {
  type: Object as Prop<Record<string | Symbol, Function>,
}
```

```
const defaultKeyFilters = {
  [Symbol.for('some-key')]: defaultKeyFilterFn
}
```

this gets merged with the user provided one from `keyFilters`. so that users can both add their own key filter fns and replace the default one through a single prop.

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,86 @@
...
+export type FilterFunction = typeof defaultFilter
+export type FilterMode = 'intersection' | 'union'
+
+export interface FilterProps {
+  filterFn?: FilterFunction
+  filterMode?: FilterMode
+}
+
+// Composables
+export function defaultFilter (text: string, query?: string) {
```
nekosaur, line 18: we should probably pass along the entire item, if users need it for whatever reason

packages/vuetify/src/composables/filter.ts
johnleider: Do we want to make keyFilters part of public props?

nekosaur: I just used keyFilters cause that's what you're calling it atm. I think it would make sense to call the prop the same across components if it does the same thing.

johnleider: Could you take a stab at implementing the functionality that you're describing?

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export type FilterFunction = typeof defaultFilter
+export type FilterMode = 'intersection' | 'union'
+export type FilterKeys = string | string[]
+export type FilterKeysFn = Record<string, FilterFunction>
+
+export interface FilterProps {
+  filterFn?: FilterFunction
+  filterKeys?: FilterKeys
+  filterKeysFn?: FilterKeysFn
```
nekosaur, line 17: Personally I feel like these are a bit confusing and will probably lead to a lot of repeated questions. The following are all hypothetical/devil's advocate questions, not me wanting answers.

What's the difference between `filterFn` and `filterKeysFn`?
Why is `filterFn` a function and `filterKeysFn` an object when they both end in `Fn`?
What's the relationship between `filterKeys` and `filterKeysFn`?

In this current setup, could you not combine filterKeys and filterKeysFn?

```
[
  'name',
  {
    key: 'title',
    filterFn: () => ...
  }
]
```

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,99 @@
...
+// Composables
+export function defaultFilter (text: string, query?: string, item?: any) {
+  if (typeof query !== 'string') return true
+  if (typeof text !== 'string') return false
+
+  return text.toLocaleLowerCase().includes(query.toLocaleLowerCase())
+}
+
+export const makeFilterProps = propsFactory({
+  filterFn: Function as PropType<FilterFunction>,
```
nekosaur, line 30: defaultFilter could be default value

packages/vuetify/src/composables/filter.ts
```diff
@@ -0,0 +1,99 @@
...
+    type: String as PropType<FilterMode>,
+    default: 'intersection',
+  },
+}, 'filter')
+
+export function filterItems (
+  items: (Record<string, any> | string)[],
+  query: string,
+  options?: {
+    default?: FilterFunction
```
nekosaur, line 43: Why not call it `filterFn` here as well?

KaelWD:
> Using the union option checks that the provided query has a match for every key

This is backwards, a union of A and B is the set of elements which are in A, in B, or in both A and B

Matching every key doesn't seem super useful either, in v2 it was only all custom filters, or at least one default filter.

KaelWD:
I could even see four modes for this
 - `some` - any custom filter, or any string match. Basically the current `intersection`
 - `union` - all custom filters, or any string match. This was the behaviour between v2.2.27 and v2.2.33
 - `intersection` - all custom filters and any string match. This is the current behaviour in v2
 - `all` - all custom filters and all remaining string matches. Currently `union`

PR #14038, opened 2021-08-08
fix(VIcon): allow colons to be used in icon names

### LINKED ISSUE #13975: [Bug Report][3.0.0-alpha.9] It's not possible to use an icon with a colon in its name/path/data
### Environment
**Vuetify Version:** 3.0.0-alpha.9
**Vue Version:** 3.1.5
**Browsers:** Chrome 92.0.4515.107
**OS:** Windows 10

### Steps to reproduce
See reproduction.

While working on the reproduction for #13974, I noticed that if you have a custom icon set that accepts image paths (for example), then it'll fail if the path contains a colon (f.e. in the scheme of a URL).

```html
<v-icon icon="img:https://example.com/image.svg">
```

With this code, the set would be correctly extracted as "img", but the icon would be set to just "https".

A simple solution would be to just extract from the first colon to the end of the string, however that will fail for default icon sets:

```html
<v-icon icon="https://example.com/image.svg">
```

Named icon set ("https") or default icon set?

Dealing with formatted strings is often tricky, so I'm wondering if it wouldn't make more sense to use a prop for the set, like:

```html
<v-icon set="img" icon="https://example.com/image.svg">
```

### Expected Behavior
Ability to use colons in icon name/path/data.

### Actual Behavior
Not possible.

### Reproduction Link
<a href="https://codepen.io/glen-84/pen/KKmVOmB" target="_blank">https://codepen.io/glen-84/pen/KKmVOmB</a> (from #13974, where I was forced to use a root-relative URL)

glen-84:
## Description

Implements the [suggestion](https://github.com/vuetifyjs/vuetify/issues/13975#issuecomment-893723888) by @nekosaur in #13975 – match the start of an icon string with the registered icon sets, instead of assuming that any string containing a colon includes an icon set name.

Fixes #13975. (_hopefully_ – not yet tested [see below])

## Motivation and Context

See #13975. It allows icons to be URLs.

## How Has This Been Tested?

There is no test file for the icon composable. I tried to add one, but I got a bit stuck:

```ts
// Composables
import { useIcon } from '../icons'

describe('icons.ts', () => {
  it.each([
    // Default icon set.
    ["success", {component: ???, icon: "success"}],
    ["https://example.com/icon.svg", {component: ???, icon: "https://example.com/icon.svg"}],
    // MDI icon set.
    ["mdi:success", {component: ???, icon: "success"}],
    // Custom icon set.
    ["custom:https://example.com/icon.png", {component: ???, icon: "https://example.com/icon.png"}],
  ])('should return the correct icon set and name', (props, expected) => {
    const { iconData } = useIcon(props)

    expect(iconData.value.component).toEqual(expected.component)
    expect(iconData.value.icon).toEqual(expected.icon)
  })
})
```

I'm not sure how to configure/reference the icon set components.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/icons.tsx
+++ b/packages/vuetify/src/composables/icons.tsx
@@ -182,23 +182,16 @@ export const useIcon = (props: Ref<string | undefined> | { icon?: IconValue }) =
       }
     }
 
-    let iconSet: IconSet | undefined
-    let iconName: string | undefined
+    const iconSetName = Object.keys(icons.sets).find(
+      setName => typeof icon === 'string' && icon.startsWith(`${setName}:`)
+    )
 
-    for (const iconSetName of Object.keys(icons.sets)) {
-      const iconSetPrefix = `${iconSetName}:`
-
-      if (icon.startsWith(iconSetPrefix)) {
-        iconSet = icons.sets[iconSetName]
-        iconName = icon.slice(iconSetPrefix.length)
-
-        break
-      }
-    }
+    const iconName = iconSetName ? icon.slice(iconSetName.length + 1) : icon
+    const iconSet = icons.sets[iconSetName ?? icons.defaultSet]
 
     return {
-      component: (iconSet ?? icons.sets[icons.defaultSet]).component,
-      icon: iconName ?? icon,
+      component: iconSet.component,
+      icon: iconName,
     }
   })
 
--- a/packages/vuetify/src/composables/icons.tsx
+++ b/packages/vuetify/src/composables/icons.tsx
@@ -182,18 +182,23 @@ export const useIcon = (props: Ref<string | undefined> | { icon?: IconValue }) =
       }
     }
 
-    const hasSet = icon.includes(':')
-    const setName = hasSet ? icon.split(':')[0] : icons.defaultSet
-    const iconName = hasSet ? icon.split(':')[1] : icon
-    const set = icons.sets[setName ?? icons.defaultSet]
+    let iconSet: IconSet | undefined
+    let iconName: string | undefined
 
-    if (!set) {
-      throw new Error(`Could not find icon set "${setName}"`)
+    for (const iconSetName of Object.keys(icons.sets)) {
+      const iconSetPrefix = `${iconSetName}:`
+
+      if (icon.startsWith(iconSetPrefix)) {
+        iconSet = icons.sets[iconSetName]
+        iconName = icon.slice(iconSetPrefix.length)
+
+        break
+      }
     }
 
     return {
-      component: set.component,
-      icon: iconName,
+      component: (iconSet ?? icons.sets[icons.defaultSet]).component,
+      icon: iconName ?? icon,
     }
   })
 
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/composables/icons.tsx
```diff
@@ -182,18 +182,23 @@ export const useIcon = (props: Ref<string | undefined> | { icon?: IconValue }) =
...
-    const hasSet = icon.includes(':')
-    const setName = hasSet ? icon.split(':')[0] : icons.defaultSet
-    const iconName = hasSet ? icon.split(':')[1] : icon
-    const set = icons.sets[setName ?? icons.defaultSet]
+    let iconSet: IconSet | undefined
+    let iconName: string | undefined
 
-    if (!set) {
-      throw new Error(`Could not find icon set "${setName}"`)
+    for (const iconSetName of Object.keys(icons.sets)) {
```
nekosaur, line 188: I think I would prefer `Object.keys(icons.sets).find(...)` here instead of a for loop

nekosaur (changes requested):
Here's how you could test

```ts
// Composables
import { defineComponent } from 'vue'
import { mount } from '@vue/test-utils'
import { createVuetify } from '@/framework'
import { useIcon } from '../icons'

declare const expect: jest.Expect
declare const it: jest.It

describe('icons', () => {
  const Component = defineComponent({
    props: {
      icon: String,
    },
    setup (props) {
      const { iconData } = useIcon(props)

      return () => iconData.value.icon
    },
  })

  const vuetify = createVuetify({
    icons: {
      defaultSet: 'mdi',
      sets: {
        custom: {
          component: () => null,
        },
      },
    },
  })

  it.each([
    // Default icon set.
    ['success', 'success'],
    ['https://example.com/icon.svg', 'https://example.com/icon.svg'],
    // MDI icon set.
    ['mdi:success', 'success'],
    // Custom icon set.
    ['custom:https://example.com/icon.png', 'https://example.com/icon.png'],
  ])('should return the correct icon name', (icon, expected) => {
    const wrapper = mount(Component, {
      props: {
        icon,
      },
      global: {
        plugins: [vuetify],
      },
    })

    expect(wrapper.text()).toEqual(expected)
  })
})
```

packages/vuetify/src/composables/icons.tsx
glen-84: Like this?

```ts
const iconSetName = Object.keys(icons.sets).find(
  setName => typeof icon === 'string' && icon.startsWith(`${setName}:`)
)

const iconName = iconSetName ? icon.slice(iconSetName.length + 1) : icon
const iconSet = icons.sets[iconSetName ?? icons.defaultSet]

return {
  component: iconSet.component,
  icon: iconName,
}
```

nekosaur: Yeah 👍 

Although is the typeof string necessary? at that point in the logic it should always be string I think?

glen-84: Within a callback context, TS cannot know whether the value/type of `icon` has changed, so its type is `IconValue | undefined`.

We could also do `(icon as string)`, but I try to avoid type assertions when possible.

glen-84: Should I push this change with the `typeof`, or do you have another preference?

nekosaur: no it's fine as is

PR #14076, opened 2021-08-23
feat(VNoSsr): create new component

blalan05:
## Description
This creates a new component that wraps content that doesn't get rendered by Server-side generation.

## Markup:
<details>

```vue
<template>
  <div>Server-side; This should display</div>
  <v-no-ssr> This is only supposed to be on the client</v-no-ssr>
</template>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VNoSsr/VNoSsr.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  setup (props, { attrs, slots }) {
+    const show = ref(false)
+
+    onMounted(() => {
+      show.value = true
+    })
+
+    return () => {
+      return show.value && (
```
johnleider, line 23: Does this have to return a root tag or can it be optional? Would like @KaelWD to chime in as well please.

johnleider (changes requested):
You don't have to populate it, but can you put in a placeholder page to the docs? Thank you.

packages/vuetify/src/components/VNoSsr/VNoSsr.tsx
KaelWD: don't think so

KaelWD: it doesn't need a root element though, just return whatever's in the default slot

blalan05: Now returning just the slot

PR #14094, opened 2021-08-26
feat(styles): use the sass module system instead of @import

### LINKED ISSUE #13694: [Bug Report][2.5.0] New Sass Deprecation Recommended
### Environment
**Vuetify Version:** 2.5.0
**Vue Version:** 2.6.12
**Browsers:** Chrome 90.0.4430.212
**OS:** Windows 10

### Steps to reproduce
npm run build 

### Expected Behavior
Should not give warnings about sass

### Actual Behavior
gives a lot of recommendations warnings of near future deprecated syntax

### Reproduction Link
<a href="https://codesandbox.io/s/nuxtjs-vuetify-forked-nrfuw" target="_blank">https://codesandbox.io/s/nuxtjs-vuetify-forked-nrfuw</a>

KaelWD:
https://sass-lang.com/documentation/at-rules/use
https://sass-lang.com/documentation/at-rules/forward

Fixes #13694

Also adds a `$reset` variable and allows `$utilities` to be set to false.


packages/vuetify/src/components/VBtn/_index.scss
```diff
@@ -1,3 +1,3 @@
-@import '../../styles/styles.sass';
-@import './mixins';
-@import './variables';
+@use '../../styles/styles.sass';
```
johnleider, line 1: remove file

johnleider (changes requested):
Looks good. Would it have been possible to keep the index sass files or was there a specific reason not to? Reference: <https://sass-lang.com/documentation/at-rules/use#index-files>

KaelWD:
Could but it would be pretty pointless because all it would be is `@forward './variables'`, the global stuff still needs to be imported in the component file itself. You have to be explicit  about dependencies now, it isn't all just thrown into one namespace anymore.

PR #14099, opened 2021-08-30
feat: define package exports

KaelWD:
```js
import 'vuetify/styles'
import { createApp } from 'vue'
import { createVuetify } from 'vuetify' // no more components in here
// import { createVuetify } from 'vuetify/framework' - also works
import { VApp } from 'vuetify/components'
import { Ripple } from 'vuetify/directives'
import { useDisplay } from 'vuetify/composables'
import { aliases, mdi } from 'vuetify/iconsets/mdi'
import { en } from 'vuetify/locale'
```

### DIFF
--- a/packages/docs/src/pages/en/features/display-and-platform.md
+++ b/packages/docs/src/pages/en/features/display-and-platform.md
@@ -24,7 +24,7 @@ The following shows how to access the application's display information:
 ```html
 <script>
   // Composables
-  import { useDisplay } from 'vuetify/lib/composables/display'
+  import { useDisplay } from 'vuetify/composables'
 
   export default {
     mounted () {
@@ -87,7 +87,7 @@ In the following example, we use a switch statement and the current breakpoint n
 
 <script>
   // Composables
-  import { useDisplay } from 'vuetify/lib/composables/display'
+  import { useDisplay } from 'vuetify/composables'
 
   export default {
     computed: {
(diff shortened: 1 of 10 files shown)


johnleider (approved):

glen-84 (commented):
Looks good, some random stuff:

1. Did you decide not to go with api-extractor? More bugs?
2. I'm a bit confused about package exports and TypeScript. Is [this issue](https://github.com/microsoft/TypeScript/issues/33079) going to cause any problems? Or are the problems worked around by using `typesVersions`?
3. Would it be possible to expose the `DisplayBreakpoint` type from `display`?
4. Should the docs be updated to reference the new import paths, or will that be done later?

I'll close my PR. Thanks for working on this!

KaelWD:
> api-extractor

Haven't tried again since it was originally broken, but I think it only does a single file anyway so wouldn't really work here. 

> TypeScript

Yeah `typesVersions` should work around that :crossed_fingers: 

> `DisplayBreakpoint`

What do you need this for? It is available on `ReturnType<typeof useDisplay>['name']` if you really need it. We could expose more interfaces from here if we need to, just keeping it tidy for now. 

> docs

done

glen-84:
> What do you use this for? It is available on `ReturnType<typeof useDisplay>['name']` if you really need it. We could expose more interfaces from here if we need to, just keeping it tidy for now.

I have code like:

```ts
const logoWidthMap = new Map<DisplayBreakpoint, number>([
    ["xs", 100],
    ["sm", 200]
]);
```

> Maybe it should just be `export type * from './display'`?

That might over-expose types that shouldn't be public API? Things like `InternalDisplayOptions` and `DisplayInstance`. IDK – might be best to ask John what the intended public API is for this composable?

PR #14103, opened 2021-08-31
feat(hydration): create new composable

blalan05:
## Description
Hydration composable checks if app/component is being generated/rendered on the server and waits to execute until on the client, else it executes right away.

## Motivation and Context
This composable supports the VNoSsr component and other components that need to only execute functions on the client.

## How Has This Been Tested?
Playground and dev:ssr

## Markup:
<details>

```vue
<template>
  <div>Server-side; This should display</div>
  <v-no-ssr> This is only supposed to be on the client</v-no-ssr>
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/hydration.ts
+++ b/packages/vuetify/src/composables/hydration.ts
@@ -1,0 +1,10 @@
+import { IN_BROWSER } from '@/util'
+import { getCurrentInstance, onMounted } from 'vue'
+
+export function useHydration (callback: () => void) {
+  if (!IN_BROWSER) return
+  const vm = getCurrentInstance()
+  const elId = vm?.root?.appContext?.app?._container?.id
+  // @ts-expect-error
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/composables/hydration.ts
```diff
@@ -0,0 +1,10 @@
+import { IN_BROWSER } from '@/util'
+import { getCurrentInstance, onMounted } from 'vue'
+
+export function useHydration (callback: () => void) {
+  if (!IN_BROWSER) return
+  const vm = getCurrentInstance()
+  const elId = vm?.root?.appContext?.app?._container?.id
+  // @ts-expect-error
+  return document.querySelector(`#${elId}`)?.__vue_app__ ? callback() : onMounted(callback)
```
KaelWD, line 9: Isn't _container already the element we want?

blalan05: doh! Yes... yes it is... Fixed.

johnleider (approved):

PR #14126, opened 2021-09-04
fix: prevent inputs from emitting native events twice

KaelWD:
## Description

VField passes attrs down, so we want to avoid binding events to it if we're also binding them to an input element. This kinda sucks though, anyone got a better idea?

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <VTextField
        v-model="model"
        class="foobar"
        label="foo"
        data-test="input"
        @input="input"
        @change="change"
        @blur="blur"
        @update:modelValue="update"
      />
    </v-container>
  </v-app>
</template>

<script>

  export default {
    setup () {
      const _model = ref('')
      const model = computed({
        get () {
          return ref.value
        },
        set (val) {
          console.log('set')
          ref.value = val
        },
      })
      return {
        input: console.log.bind(console, 'input'),
        change: console.log.bind(console, 'change'),
        blur: console.log.bind(console, 'blur'),
        update: (...args) => {
          console.log('update:modelValue', ...args)
        },
        model,
      }
    },
  }
</script>
```
</details>


nekosaur:
Could VField be responsible for filtering out event handlers?

KaelWD:
We're also duplicating attributes so not really
Screenshot_20210910_014825

packages/vuetify/src/components/VFileInput/VFileInput.tsx
```diff
@@ -105,15 +106,13 @@ export const VFileInput = defineComponent({
 
     useRender(() => {
       const hasCounter = !!(slots.counter || props.counter || counterValue.value)
-      const [_, restAttrs] = pick(attrs, ['class'])
+      const [rootAttrs, inputAttrs] = filterInputAttrs(attrs)
+      const [fieldProps, _] = filterFieldProps(props)
```
KaelWD, line 110: This kinda sucks performance-wise, it would be simpler to just remove `makeVFieldProps()` from these components and use fallthrough instead. The downsides are that these props would no longer be automatically added to the type definitions or docs.

packages/vuetify/src/components/VCheckbox/VCheckbox.tsx
```diff
@@ -48,6 +49,10 @@ export const VCheckbox = defineComponent({
     ...makeValidationProps(),
   },
 
+  emits: {
+    'update:modelValue': (val: any) => true,
```
johnleider, line 53: is there a reason for using `val: any` instead of `val: unknown`?

packages/vuetify/src/components/VList/VList.tsx
```diff
@@ -108,6 +108,12 @@ export const VList = genericComponent<new <T>() => {
     ...makeThemeProps(),
   },
 
+  emits: {
+    'update:selected': (val: string[]) => true,
+    'update:opened': (val: string[]) => true,
+    'update:active': (val: string[]) => true,
```
johnleider, line 114:
```suggestion
    'update:active': (val: string[]) => true,
    'update:opened': (val: string[]) => true,
    'update:selected': (val: string[]) => true,
```

johnleider (changes requested):

packages/vuetify/src/components/VList/VList.tsx
KaelWD: This is the same order as in the original interface.

packages/vuetify/src/components/VCheckbox/VCheckbox.tsx
KaelWD: Had some problems with unknown here in another component.

PR #14157, opened 2021-09-14
docs(VWindow): update onboarding example

rohan220217:
I have added text-center property to align it to center

### DIFF
--- a/packages/docs/src/examples/v-window/misc-onboarding.vue
+++ b/packages/docs/src/examples/v-window/misc-onboarding.vue
@@ -19,9 +19,7 @@
             align="center"
             justify="center"
           >
-            <v-card-text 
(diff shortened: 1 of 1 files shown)


packages/docs/src/examples/v-window/misc-onboarding.vue
```diff
@@ -19,7 +19,9 @@
             align="center"
             justify="center"
           >
-            <v-card-text>
+            <v-card-text 
+             class="text-center"
+            >
```
johnleider, line 24:
```suggestion
            <v-card-text class="text-center">
```

johnleider (changes requested):

johnleider (approved):

PR #14161, opened 2021-09-15
feat(active): create new composable

johnleider:
## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VMenu/VMenu.tsx
+++ b/packages/vuetify/src/components/VMenu/VMenu.tsx
@@ -6,8 +6,8 @@ import { VOverlay } from '@/components/VOverlay'
 import { VDialogTransition } from '@/components/transitions'
 
 // Composables
+import { makeActiveProps, useActive } from '@/composables/active'
 import { makeTransitionProps } from '@/composables/transition'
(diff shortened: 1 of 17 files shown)


packages/vuetify/src/components/VMenu/VMenu.tsx
```diff
@@ -29,28 +29,28 @@ export default defineComponent({
...
+    const { isActive } = useActive(props, 'v-menu')
 
     const uid = getUid()
     const id = computed(() => props.id || `v-menu-${uid}`)
 
     return () => {
       return (
         <VOverlay
-          v-model={ isActive.value }
+          v-model:active={ isActive.value }
```
KaelWD, line 53: This is invalid: https://github.com/vuejs/jsx-next/issues/325

KaelWD:
I'm not really a fan of this change. Most of these components don't really need a separate active class because the inactive state is "not visible". It might make sense for some components like VListItem which could have several equally-valid uses for modelValue, but for others like VOverlay it means you have to always use `v-model:active` when there is only one state that could be set.

PR #14253, opened 2021-10-10
fix(VSlideGroup): recalculate widths when it shows

### LINKED ISSUE #14195: [Bug Report][2.5.8] VSlideGroup: overflow arrows do not show up if the slide group is initially invisible
### Environment
**Vuetify Version:** 2.5.8
**Vue Version:** 2.6.11
**Browsers:** Chrome 94.0.4606.61
**OS:** Mac OS 10.15.7, Android

### Steps to reproduce
1. click the 'Show bug' button.

### Expected Behavior
The overflow arrow indicators should show up with the slide group as it overflows its container `v-sheet`.

### Actual Behavior
The arrows do not show up until click on an item.

### Reproduction Link
<a href="https://codepen.io/Kohn1/pen/KKqGjme" target="_blank">https://codepen.io/Kohn1/pen/KKqGjme</a>

### Other comments
https://github.com/vuetifyjs/vuetify/blob/4c3e561af7042fc5563b8ff49ff873d23ec211c8/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts#L429-L437

A possible cause might be that `this.isOverflowing` is not correctly calculated when the element is hidden, in which case `content.clientWidth` would be 0. A solution I can come up with is to watch the visibility changes and re-calculate `this.isOverflowing` when there is a change (with Intersection Observer API or something; might sound a little heavy).

A workaround: invoke `.onResize()` or `.setWidths()` after changing visibility.
```html
<v-btn @click="show = true; $refs.group.setWidths()">
```

beicause:
## Description

fixes #14195
## Markup:

<details>

```vue
<template>
  <v-container>
    <v-btn @click="show = !show">
      Show bug
    </v-btn>
    <v-sheet
      elevation="5"
      class="ma-5 pa-5"
      max-width="800"
    >

      <v-slide-group
        v-show="show"
        show-arrows
      >
        <v-slide-item
          v-for="n in list"
          :key="n"
          v-slot="{ active, toggle }"
        >
          <v-btn
            class="mx-2"
            :input-value="active"
            active-class="primary white--text"
            depressed
            rounded
            @click="toggle"
          >
            Options {{ n }}
          </v-btn>
        </v-slide-item>
      </v-slide-group>
    </v-sheet>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      show: false,
      list: Array.from(Array(10).keys()),
    }),
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
+++ b/packages/vuetify/src/components/VSlideGroup/VSlideGroup.ts
@@ -41,6 +41,12 @@ interface options extends Vue {
   }
 }
 
+function bias (val: number) {
+  const c = 0.501
+  const x = Math.abs(val)
+  return Math.sign(val) * (x / ((1 / c - 2) * (1 - x) + 1))
+}
+
 export const BaseSlideGroup = mixins<options &
 /* eslint-disable indent */
   ExtractVue<[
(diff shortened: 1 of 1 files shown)


Goshin:
Would this solution still work if we toggle the parent's display property instead of that of the VSlideGroup itself?

beicause:
> Would this solution still work if we toggle the parent's display property instead of that of the VSlideGroup itself?

it's really a problem😅

PR #14413, opened 2021-11-17
feat(VChipGroup): port to v3

blalan05:
## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
     <v-chip>Not in Group</v-chip>
      <v-chip-group column v-model="group">
        <v-chip>One</v-chip>
        <v-chip disabled>Two</v-chip>
        <v-chip>Three</v-chip>
        <v-chip>Four</v-chip>
        <v-chip value="five">Five</v-chip>
      </v-chip-group>

      {{ `group value is ${group}` }}
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const group = ref(null)
      return {
        group
      }
    },
  }
</script>
```
</details>


packages/vuetify/src/components/VChipGroup/VChipGroup.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const VChipGroup = defineComponent({
+  name: 'VChipGroup',
+
+  props: {
+    column: Boolean,
+    ...makeGroupProps({
+      selectedClass: 'v-chip--selected',
+    }),
+    ...makeTagProps({ tag: 'div' }),
```
johnleider, line 30: the **tag** prop uses `div` by default.

packages/vuetify/src/components/VChip/VChip.tsx
```diff
@@ -87,6 +89,7 @@ export const VChip = defineComponent({
     const { borderClasses } = useBorder(props)
     const { colorClasses, colorStyles, variantClasses } = useVariant(props)
     const { elevationClasses } = useElevation(props)
+    const group = useGroupItem(props, VChipGroupSymbol, false)
```
johnleider, line 92: could probably destructure this as you did in VChipGroup.

johnleider (changes requested):
Looking good. Needs work on focus/hover/selected styles as well as some refactoring for components that provide slots using **genericComponent** so that we get proper type checking.

johnleider (approved):

PR #14449, opened 2021-11-30
fix(VOverlay): account for max-width from props

### LINKED ISSUE #14420: [Bug Report][3.0.0-alpha.11] Prop `max-width` of VTooltip component has no effect
### Environment
**Vuetify Version:** 3.0.0-alpha.11
**Vue Version:** 3.2.21
**Browsers:** Chrome 96.0.4664.45
**OS:** Windows 10

### Steps to reproduce
See repro. Hover over the button.

### Expected Behavior
The tooltip width should be limited to 300px.

### Actual Behavior
It's not.

### Reproduction Link
<a href="https://codepen.io/glen-84/pen/NWgvKOw" target="_blank">https://codepen.io/glen-84/pen/NWgvKOw</a>

### Other comments
From Kael on Discord:

> looks like a bug, there is a max-width prop but it's overridden by the computed width available on that side

yuwu9145:
fixes #14420

## Description

fixes #14420

## How Has This Been Tested?

visually | unit 
## Markup:

<details>

```vue
<template>
  <v-container>
    <v-btn>
        <v-tooltip
            :text="longTooltip"
            activator="parent"
            max-width="300" />
        Example
    </v-btn>
  </v-container>
</template>

<script>
export default {
  data: () => ({
    longTooltip: "x".repeat(200)
  }),
};
</script>


```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VOverlay/positionStrategies.ts
+++ b/packages/vuetify/src/components/VOverlay/positionStrategies.ts
@@ -208,10 +208,15 @@ function connectedPositionStrategy (data: PositionStrategyData, props: StrategyP
 
     const canFill = doesOverlap.value || ['center', 'top', 'bottom'].includes(anchor.side)
 
+    const maxWidthFromProps: number | undefined = typeof props.maxWidth === 'string' ? parseInt(props.maxWidth, 10) : props.maxWidth
+
+    const getEndStartMaxWidth = (lrSpace: number) => maxWidthFromProps ? Math.min(maxWidthFromProps, lrSpace) : lrSpace
+
     const maxWidth = canFill ? Math.min(viewportWidth, Math.max(targetBox.width, viewportWidth - viewportMargin * 2))
-      : anchor.side === 'end' ? freeSpace.right
-      : anchor.side === 'start' ? freeSpace.left
+      : anchor.side === 'end' ? getEndStartMaxWidth(freeSpace.right)
+      : anchor.side === 'start' ? getEndStartMaxWidth(freeSpace.left)
       : null
+
     const minWidth = Math.min(configuredMinWidth.value, maxWidth!, targetBox.width)
     const maxHeight = fitsY ? configuredMaxHeight.value : Math.min(
       configuredMaxHeight.value,
--- a/packages/vuetify/src/components/VOverlay/positionStrategies.ts
+++ b/packages/vuetify/src/components/VOverlay/positionStrategies.ts
@@ -189,12 +189,15 @@ function connectedPositionStrategy (data: PositionStrategyData, props: StrategyP
 
     const contentHeight = Math.min(configuredMaxHeight.value, contentBox.height)
 
+    // Regard undefined maxWidth as maximally occupying whole remaining space by default
+    const maxFreeSpaceWidth = props.maxWidth === undefined ? Number.MAX_VALUE : parseInt(props.maxWidth ?? 0, 10)
+
     const viewportMargin = 12
     const freeSpace = {
       top: targetBox.top - viewportMargin,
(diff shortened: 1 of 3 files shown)


KaelWD:
This probably needs to be changed in the position strategy instead - if there's only 250px of space available but I've set `max-width="300"`, what should the width be?

yuwu9145:
Makes a lot of sense. Changes are switched to using position strategy

packages/vuetify/src/components/VOverlay/positionStrategies.ts
```diff
@@ -208,10 +208,15 @@ function connectedPositionStrategy (data: PositionStrategyData, props: StrategyP
 
     const canFill = doesOverlap.value || ['center', 'top', 'bottom'].includes(anchor.side)
 
+    const maxWidthFromProps: number = !props.maxWidth ? Number.MAX_SAFE_INTEGER // use this constant to indicate no max width limit
```
johnleider, line 211: Is it necessary to use this? I don't see a specific benefit to it.

yuwu9145: Without this, when `maxWidth` is not set in props, `maxWidthFromProps` will be `undefined` which can cause `Math.min(undefined, 1)` to return `NaN` in line 216 & 217.

yuwu9145: Probably better to allow `undefined` here and do Falsy conditional check in 216 & 217.

Leave with me and I will optimise

yuwu9145:
@johnleider @KaelWD is there any objection to upgrade `@vue/test-utils` to version `rc.17` ? 

It will make writing tests easier with components using `teleport` like `VTooltip` & `VOverlay`. Here is the [detail](https://github.com/vuejs/test-utils/issues/1266)

yuwu9145:
could I please ask for a feedback for this PR? @KaelWD

johnleider (changes requested):
I think this implementation is more complex than it needs to be. Only needs to be 3 new lines:

image

Also, please remove the VTooltip test changes. Thank you!

yuwu9145:
> I think this implementation is more complex than it needs to be. The following accomplishes the same thing.
> 
> image
> 
> Also, please remove the VTooltip test changes. Thank you!

Hi @johnleider , thanks for your tip. But it does not cover the senario when `props.maxWidth` is `undefined`.

I think we should regard undefined maxWidth as maximally occupying whole remaining space by default. Pull request has been updated to achieve this on top of your changes.

Thank you!

johnleider:
> Hi @johnleider , thanks for your tip. But it does not cover the senario when props.maxWidth is undefined.

Then add that one use-case. The general point I was trying to make was that I did not like your implementation and wanted you to approach it differently.

yuwu9145:
> > Hi @johnleider , thanks for your tip. But it does not cover the senario when props.maxWidth is undefined.
> 
> 
> 
> Then add that one use-case. The general point I was trying to make was that I did not like your implementation and wanted you to approach it differently.

No worries, understand. Pull request has been updated based on your advice. Feel free to leave any further comments

packages/vuetify/src/components/VOverlay/positionStrategies.ts
```diff
@@ -212,6 +215,7 @@ function connectedPositionStrategy (data: PositionStrategyData, props: StrategyP
       : anchor.side === 'end' ? freeSpace.right
       : anchor.side === 'start' ? freeSpace.left
       : null
+
```
johnleider, line 218:
```suggestion
```

johnleider (changes requested):

johnleider (approved):
Looks good. Just need review from @KaelWD

PR #14486, opened 2021-12-09
fix(VField): don't show gap when missing label

LucianHij:
## Description
Handling the case when label is missing or it has falsy value
I am not aware of any open issue that this PR resolves. It resolves an issue that we discussed on the discord chat. 

## Motivation and Context
The layout is broken when using outlined variant. The notch is present and the border is interrupted with white space.

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        v-model="something"
        variant="outlined"
        label="Something"
      />
      <v-text-field
        v-model="something"
        variant="outlined"
        label=""
      />
    </v-container>
  </v-app>
</template>

<script>
export default {
  name: "Playground",
  setup() {
    return {
      something: "",
    };
  },
};
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VField/VField.tsx
+++ b/packages/vuetify/src/components/VField/VField.tsx
@@ -213,6 +213,7 @@ export const VField = genericComponent<new <T>() => {
         })
         : props.label
       const [inputProps, _] = filterInputProps(props)
+
       return (
         <VInput
           class={[
--- a/packages/vuetify/src/components/VField/VField.tsx
+++ b/packages/vuetify/src/components/VField/VField.tsx
@@ -140,7 +140,7 @@ export const VField = genericComponent<new <T>() => {
     }))
 
     watch(isActive, val => {
-      if (!props.singleLine) {
+      if (!props.singleLine && props.label) {
         const el: HTMLElement = labelRef.value!.$el
         const targetEl: HTMLElement = floatingLabelRef.value!.$el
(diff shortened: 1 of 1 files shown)


johnleider (changes requested):
Since it's reused so much, create a computed property:

```ts
const hasLabel = computed(() => !props.singleLine && !!label)
```

and then use `hasLabel.value` throughout.

LucianHij:
> Since it's reused so much, create a computed property:
> 
> 
> 
> ```ts
> 
> const hasLabel = computed(() => !props.singleLine && !!label)
> 
> ```
> 
> 
> 
> and then use `hasLabel.value` throughout.

I will address it tomorrow

packages/vuetify/src/components/VField/VField.tsx
```diff
@@ -126,6 +126,7 @@ export const VField = genericComponent<new <T>() => {
     const inputRef = ref<HTMLInputElement>()
     const isFocused = ref(false)
     const id = computed(() => props.id || `input-${uid}`)
+    const hasLabel = computed(() => !props.singleLine && !!props.label)
```
KaelWD, line 129:
```suggestion
    const hasLabel = computed(() => !props.singleLine && !!(props.label || slots.label))
```

johnleider (approved):

KaelWD (approved):

PR #14504, opened 2021-12-13
feat(VSelect): port to v3

johnleider:
## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-select label="Foobar" />
    </div>
  </v-app>
</template>

<script>
  export default {

  }
</script>

```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+//     const val = this.getValue(item)
+
+//     // TODO: comparator
+//     !uniqueValues.has(val) && uniqueValues.set(val, item)
+//   }
+//   return Array.from(uniqueValues.values())
+// }
+
+export const VSelect = genericComponent<new <T>() => {
```
nekosaur, line 39: generic `T` is not used

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -0,0 +1,122 @@
...
+                  <VList
+                    elevation="4"
+                    rounded
+                    v-model:active={ activated.value }
+                    items={ items.value }
+                    activeStrategy={ props.multiple ? 'multiple' : 'single' }
+                  />
+                </VMenu>
+
+                { slots.default?.(slotProps) }
```
nekosaur, line 110: I might be missing some use case but I we should render either the VList or the default slot, not both? My first thought was making a treeview select by rendering a custom VList or VTreeview

johnleider: I typically lean towards leaving both in but this instance may be different.

jacekkarczmarczyk:
1. Should `@click:prepend` work at this stage? Because it doesn't form me

<details>

```vue
<template>
  <v-app>
    <v-main>
      <v-select :items="['a', 'b', 'c']" prepend-icon="mdi-vuetify" clearable @click:prepend="c" />
    </v-main>
  </v-app>
</template>

<script setup>
  const c = e => console.log(e)
</script>
```

</details>

2. Not a fan of current transition for items menu, not only visually but also quick clicking on the option after opening the menu hides the menu without seleting the option. There's no `menuProps` any more to turn off transtiion, is there another way?

3. ~~Clearable icon shows only on hover, not sure if that's a good idea.~~ Found the `persistent-clear` prop, I think it should be default `true` 

4. Keyboard navigation doesn't work (arrows and typing letters, only enter/space works for opening the menu)

johnleider:
> 1. Should `@click:prepend` work at this stage? Because it doesn't form me
> 
> 2. Not a fan of current transition for items menu, not only visually but also quick clicking on the option after opening the menu hides the menu without seleting the option. There's no `menuProps` any more to turn off transtiion, is there another way?
> 3. ~Clearable icon shows only on hover, not sure if that's a good idea.~ Found the `persistent-clear` prop, I think it should be default `true`
> 4. Keyboard navigation doesn't work (arrows and typing letters, only enter/space works for opening the menu)

1. Yes, I'll look into it
2. This requires changes to how `v-menu` works, will be separate from this PR
3. This matches the default text input functionality and I don't like boolean true default properties
4. This is the responsibility of `v-menu` and `v-list`. It will come as a separate composable that's integrated into **nested**.

jacekkarczmarczyk:
> 3. This matches the default text input functionality and I don't like boolean true default properties

Could be done with different property name, ex. `show-clear-only-on-hover` (not that exact name but you get the idea). I agree that that's not the scope of this PR though

packages/vuetify/src/components/VPagination/__tests__/VPagination.spec.cy.tsx
```diff
@@ -86,7 +86,7 @@ describe('VPagination', () => {
       <VPagination length="100" />
     ))
 
-    cy.get('.v-pagination__item').should('have.length', 8)
+    cy.get('.v-pagination__item').should('have.length', 6)
```
KaelWD, line 89: ?

packages/vuetify/src/components/VTextarea/__tests__/VTextarea.spec.cy.tsx
```diff
@@ -3,7 +3,7 @@
 import { VTextarea } from '../VTextarea'
 import { ref } from 'vue'
 
-describe('VTextarea', () => {
+describe.skip('VTextarea', () => {
```
KaelWD, line 6: This is fixed in #14683

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -0,0 +1,231 @@
...
+    const vTextFieldRef = ref()
+    const activator = ref()
+    const model = useProxiedModel(
+      props,
+      'modelValue',
+      [],
+      v => wrapInArray(v),
+      (v: any) => props.multiple ? v : v[0]
+    )
+    const { filteredItems } = useFilter(props, props.items)
```
KaelWD, line 76: I still don't understand why we're filtering on nothing here

packages/vuetify/src/components/VPagination/__tests__/VPagination.spec.cy.tsx
johnleider: This error isn't part of this branch, I simply fixed it to get running tests.

packages/vuetify/src/components/VSelect/VSelect.tsx
johnleider: So that we are working with the same data structure that other selects will have.

packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -85,28 +91,48 @@ export const VTextField = defineComponent({
...
+        ? props.messages
+        : (isActive.value || props.persistentHint) ? props.hint : ''
+    })
+    function onFocus () {
+      if (inputRef.value !== document.activeElement) {
+        inputRef.value?.focus()
+      }
+
+      if (!isFocused.value) isFocused.value = true
     }
```
tongxuanbao, line 114: -

PR #14527, opened 2021-12-17
fix(VList): improve nested list padding

nekosaur:
## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="d-flex flex-row">
      <template v-for="d in density">
        <div style="width: 500px">
          <v-list :items="items" :density="d" />
          <div style="height: 40px; background: red"></div>
          <v-list :items="itemsWithIcons" :density="d" />
        </div>
      </template>
    </div>
  </v-app>
</template>

<script lang="ts">
  import { defineComponent } from 'vue'

    data: () => ({
      density: ['default', 'comfortable', 'compact'],
      items: [
        {
          title: 'Title',
          subtitle: 'Subtitle',
          prependIcon: 'mdi-home',
          children: [
            {
              title: 'Child',
              prependIcon: 'mdi-home',
            },
            {
              title: 'Foo',
            },
          ],
        },
      ],
      itemsWithIcons: [
        {
          title: 'Title',
          subtitle: 'Subtitle',
          prependIcon: 'mdi-home',
          children: [
            {
              title: 'Child',
            },
          ],
        },
        {
          title: 'Foo',
          subtitle: 'Bar',
        },
      ],
    }),
  })
</script>


```
</details>

## Types of changes

## Checklist:


KaelWD:
Padding needs to decrease in dense lists to account for the icon size change:
Screenshot_20211218_175343

nekosaur:
I thought @johnleider said that density should only affect verticality? Which is why I removed the complicated density stuff

KaelWD:
It's a nice idea but already doesn't apply to square components like icon buttons. If that's the case then we still need to use density to add padding to the ones with icons. Also consider if a dense list looks good if it still has huge indents. Dense lists in v2 also had a reduced font size which is something else we might want to consider.

nekosaur:
ugh, yeah..

johnleider:
> It's a nice idea but already doesn't apply to square components like icon buttons. If that's the case then we still need to use density to add padding to the ones with icons. Also, consider if a dense list looks good if it still has huge indents. Dense lists in v2 also had a reduced font size which is something else we might want to consider.

We have a **nav** property that does this but I'm not sure it's the best way to go. If density needs to also take horizontal size into consideration, it might require a larger discussion.

nekosaur:
@johnleider have you looked at the results of this PR? I think I agree with kael that the dense nested indents are too big. I have updated the playground.

johnleider:
> It's a nice idea but already doesn't apply to square components like icon buttons. If that's the case then we still need to use density to add padding to the ones with icons. Also consider if a dense list looks good if it still has huge indents. Dense lists in v2 also had a reduced font size which is something else we might want to consider.

I can get on board with regards to padding changes per density level. We can keep the **dense** prop or rename it if it's potentially confusing.

PR #14528, opened 2021-12-17
feat(theme): support partial custom themes

nekosaur:
- useTheme is now a public composable that takes no arguments
- old useTheme renamed to provideTheme
- VApp uses useTheme instead of provideTheme, so that it uses root context without creating a new one
- you can pass partial custom themes and they will be filled out with default light/dark values
- classes are generated for all colors across all defined themes, not just the first one found

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VApp/VApp.tsx
```diff
@@ -18,10 +19,12 @@ export const VApp = defineComponent({
   },
 
   setup (props, { slots }) {
-    const { themeClasses } = useTheme(props)
+    const { themeClasses, current } = useTheme()
     const { layoutClasses, getLayoutItem, items } = createLayout(props)
     const { rtlClasses } = useRtl()
 
+    watch(() => props.theme, value => value && (current.value = value), { immediate: true })
```
KaelWD, line 26: VApp isn't a singleton anymore, you can have more than one which could cause some problems with this watcher. Any reason not to just use `provideTheme` instead?

nekosaur: Isn't it still kind of? What's an example of where you would have multiple ones?

The reason for this change is so that you can call `useTheme` where you are rendering VApp and it will be the same theme context. Perhaps a pretty niche thing so I wouldn't be too sad to revert it 🤷

KaelWD: Yeah seems pretty niche. You could also expose `this.$refs.app.theme` for the same effect.

packages/vuetify/src/composables/theme.ts
```diff
@@ -171,9 +165,14 @@ const defaultThemeOptions: ThemeOptions = {
 const parseThemeOptions = (options: ThemeOptions = defaultThemeOptions): InternalThemeOptions => {
   if (!options) return { ...defaultThemeOptions, isDisabled: true } as InternalThemeOptions
 
+  const themes = Object.entries(options.themes ?? {}).reduce((obj, [key, theme]) => {
+    const defaultTheme = theme.dark ? defaultThemeOptions.themes?.dark : defaultThemeOptions.themes?.light
+    return { ...obj, [key]: mergeDeep(defaultTheme, theme) }
```
KaelWD, line 170:
```suggestion
    obj[key] = mergeDeep(defaultTheme, theme)
    return obj
```

KaelWD (approved):

PR #14537, opened 2021-12-21
refactor(defaults): simplify provideDefaults interface

johnleider:
## Description
Reduces markup needed to provide defaults and removes the need to use the **reactive**, **computed**, and **toRef** helpers.

```ts
// Before
provideDefaults(reactive({
  defaults: {
    VBtn: {
      color: toRef(props, 'color'),
      flat: true,
    },
  },
  scoped: true,
}))

// After
provideDefaults({
  VBtn: {
    color: toRef(props, 'color'),
    flat: true,
  },
}, { scoped: true })
```

## Motivation and Context
Improve the interface and make it easier to provide defaults from within a component.

## Markup:

<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-switch
        v-model="color"
        label="color"
        true-value="primary"
        false-value=""
      />

      <v-switch
        v-model="variant"
        label="text"
        true-value="text"
        false-value="contained-text"
      />

      <v-btn-group
        :color="color"
        :variant="variant"
        class="flex-wrap"
      >
        <v-btn
          v-for="n in 200"
          :key="n"
          height="32"
        >Button {{ n }}</v-btn>
      </v-btn-group>
    </div>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
      color: '',
      variant: 'text',
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/defaults.ts
+++ b/packages/vuetify/src/composables/defaults.ts
@@ -1,9 +1,10 @@
 // Utilities
 import { computed, inject, provide, ref } from 'vue'
-import { mergeDeep, wrapInRef } from '@/util'
+import { mergeDeep } from '@/util'
 
 // Types
 import type { ComputedRef, InjectionKey, Ref } from 'vue'
+import type { MaybeRef } from '@/util'
 
 export interface DefaultsInstance {
   [key: string]: undefined | Record<string, unknown>
@@ -27,7 +28,7 @@ export function useDefaults () {
 }
 
 export function provideDefaults (
-  defaults?: Ref<DefaultsInstance | undefined> | DefaultsInstance,
+  defaults?: MaybeRef<DefaultsInstance | undefined>,
   options?: {
     reset?: Ref<number | string | undefined>
     root?: Ref<boolean | undefined>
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/composables/defaults.ts
```diff
@@ -26,21 +26,24 @@ export function useDefaults () {
...
+export function provideDefaults (
+  defaults?: Ref<DefaultsInstance | undefined> | DefaultsInstance,
+  options?: {
+    reset?: Ref<number | string | undefined>
+    root?: Ref<boolean | undefined>
+    scoped?: Ref<boolean | undefined>
+  }
+) {
+  const injectedDefaults = useDefaults()
+  const providedDefaults = wrapInRef(defaults)
```
KaelWD, line 38: What is `wrapInRef` for? `ref(val)` already does nothing if `val` is already a Ref.

packages/vuetify/src/composables/defaults.ts
```diff
@@ -26,21 +26,24 @@ export function useDefaults () {
...
 
-export function provideDefaults (props?: {
-  defaults?: DefaultsInstance
-  reset?: number | string
-  root?: boolean
-  scoped?: boolean
-}) {
-  const defaults = useDefaults()
+export function provideDefaults (
+  defaults?: Ref<DefaultsInstance | undefined> | DefaultsInstance,
```
KaelWD, line 30:
```suggestion
  defaults?: MaybeRef<DefaultsInstance | undefined>,
```

KaelWD (approved):

PR #14632, opened 2022-01-19
feat(VTimePicker): add activePicker prop

### LINKED ISSUE #14629: [Feature Request] v-time-picker needs active-picker prop like v-date-picker
### Problem to solve
When trying to create a combined v-date-time-picker, I need to have control over which picker is shown in a v-time-picker. I need to allow the user to "go back to" the hour picker once they have picked a time, in case they picked the wrong hour. Unfortunately, there seems to be no way to control the hour vs minute picker from outside the component. Since the combined picker is not using the header of the v-time-picker there is no way for the user to get back to the hour.

### Proposed solution
I would like to see symmetry between the v-time-picker and the v-date-picker with an active-picker being added to the v-time-picker. This would allow the combined picker to tell v-time-picker to cycle back to hour when needed.

It would be something like this:
Name: active-picker
type: string
default: undefined
Description: Determines which picker in the hour or minute picker is being displayed. Allowed values: "HOUR", "MINUTE"

I would use it like this:

  name: 'VDateTimePicker',
  setup() {
  timePicker = ref('HOUR');
  return { timePicker }
}

<v-time-picker :active-picker="timePicker" />

timePicker:

gregveres:
## Description
adds active-picker prop to be consistent with v-date-picker. active-picker
is used to have v-time-picker switch to the specified picker. With a
.sync option, the prop is kept current with visibile picker. Needed for
implementing a v-date-time-picker.

Closes #14629

## Motivation and Context
This feature request was sparked by the lack of a combined date / time picker. When I started to build my own date / time picker, I realized that the date picker had a prop called active-picker and I needed to use that prop to allow my header to control which date picker was visible. Unfortunately, the time picker didn't have the same property, so clicking on components of the time wasn't able to control the time picker. And, even worse is that once the user selects the hour, there is no way to let the user go back to change their mind on which hour to select because the v-time-picker does not have a way from the outside to programmatically get back to the hour picker (when the header is hidden).

## How Has This Been Tested?
I have added 4 new unit tests to v-time-picker.
I have run all the existing unit tests to make sure that they still pass.
I also used this playground setup to ensure that the proper messages were emitted and that the time picker behaved correctly to changing prop values:

## Markup:
```
<template>
  <v-container>
    <v-time-picker
      v-model="time"
      :active-picker.sync="active"
      use-seconds
    />
    {{ active }}
    <v-btn @click="active = 'HOUR'">HOUR</v-btn>
    <v-btn @click="active = 'MINUTE'">MINUTE</v-btn>
    <v-btn @click="active = 'SECOND'">SECOND</v-btn>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      time: '10:30:15',
      active: '',
    }),
  }
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VTimePicker/VTimePicker.ts
```diff
@@ -197,6 +201,17 @@ export default mixins(
...
       if (value !== null) this.$emit('input', value)
     },
+    emitPicker (value: SelectingTimes) {
+      let val = 'HOUR'
+      if (value === SelectingTimes.Minute) val = 'MINUTE'
+      else if (value === SelectingTimes.Second) val = 'SECOND'
+      this.$emit('update:active-picker', val)
+    },
+    setPicker (picker: ActivePicker) {
+      if (picker === 'HOUR') this.selecting = SelectingTimes.Hour
```
jacekkarczmarczyk, line 211: please add brackets (also in `emitPicker`)
```js
if (...) {
  ...
} else if (...) {
  ...
}
```

packages/vuetify/src/components/VTimePicker/VTimePicker.ts
```diff
@@ -197,6 +201,17 @@ export default mixins(
       const value = this.genValue()
       if (value !== null) this.$emit('input', value)
     },
+    emitPicker (value: SelectingTimes) {
+      let val = 'HOUR'
```
jacekkarczmarczyk, line 205:
```suggestion
      let picker = 'HOUR'
```
or maybe even `activePicker`

jacekkarczmarczyk (changes requested):
lgtm, this is a feature though so please rebase to dev

packages/vuetify/src/components/VTimePicker/VTimePicker.ts
gregveres: changed to activePicker

packages/vuetify/src/components/VTimePicker/VTimePicker.ts
gregveres: done

gregveres:
> lgtm, this is a feature though so please rebase to dev

I believe I have rebased to dev now.
I have also updated the change based on your review comments. 

Thx

KaelWD (approved):

Issue #9720, opened 2019-11-18
[Bug Report] DataTable click:row.ctrl Modifier Not Working


jim-brown-fst:
### Environment
**Vuetify Version:** 2.1.10
**Vue Version:** 2.6.10
**Browsers:** Chrome 78.0.3904.97
**OS:** Windows 10

### Steps to reproduce
```
<v-data-table ....... @click:row.ctrl="rowCtrlClicked"> ........ </v-data-table>
```

### Expected Behavior
Event handler fires

### Actual Behavior
Event handler does not fire

### Reproduction Link
<a href="https://codepen.io/jim55063/pen/PooXgWY?&editable=true&editors=101#anon-signup" target="_blank">https://codepen.io/jim55063/pen/PooXgWY?&editable=true&editors=101#anon-signup</a>

### Other comments
I am replacing an existing data table in our application which allows the user to ctrl+click a row in the table and open that item in a new window (tab).

```@click:row.once``` works, since it doesn't require the event object to be passed to the ```.once``` event modifier. But for the ```.ctrl``` modifier, since you aren't passing the event to the modifier, it can't check to see if the control key is pressed and filters the event out and the handler isn't called.

Also, since the event isn't passed to the handler, there is no way to know if the user had the ctrl key held down when the row was clicked.

jim-brown-fst:
Other modifiers like ```.prevent``` and ```.stop``` and probably most others don't work either due to not passing the event object.

jacekkarczmarczyk:
`click:row` is not a DOM event, so these modifiers won't work by default, changing the issue to feature request

jim-brown-fst:
It starts out as a click event and then the event gets hidden. This happens in a lot of controls and events actually. Can you just add the event as a second parameter in the handlers so we can use them in our code? Passing along the events allows for more use cases (like the one in this issue) and I can't see any downside really.

nekosaur:
For the `.ctrl` modifier to work we'd have to pass the native event as the first argument. That also has the added drawback of just giving the method name in template `@click.ctrl="nameOfMethod"` not working since that only forwards the first argument to the handler. So one would have to do `@click.ctrl="(e, foo) => nameOfMethod(foo)"` or `@click.ctrl="nameOfMethod(...arguments)"`.

Passing the native event as a second argument is certainly possible, but that means any modifiers will have to be replicated manually.

jim-brown-fst:
[...]

PR #14659, opened 2022-01-27
feat(VDataTable): add missing event as 3rd argument

### LINKED ISSUE #10302: [Feature Request] v-datatable click:row event should include the event as second parameter
### Problem to solve
So we can do DOM manipulation with the event.target

### Proposed solution
Probably a small changing cascading the event target in packages/vuetify/src/components/VDataTable/VDataTable.ts

Leinad4Mind:
## Description
Add event as 3rd argument to click:row
fixes #10302, #9720

## Motivation and Context
Since adding event as 1st argument would be an breaking change, we could at least add it as 3rd argument so we can at least make use of it in vuetify 2. The alternatives to make a click on each row have an huge impact on developing, since we need to add on all tables the item slot instead and add all the td's manually.

## How Has This Been Tested?
There is already tests for this, same as dblclick event.

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-generator/src/locale/en/v-data-table.json
+++ b/packages/api-generator/src/locale/en/v-data-table.json
@@ -62,7 +62,7 @@
     "top": "Slot to add content above the table"
   },
   "events": {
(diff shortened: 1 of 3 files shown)


KaelWD:
Missed this somehow, duplicated by #15864

Leinad4Mind:
No problem. My code accepts keyboard event too and not only MouseEvents according the duplicated. Beaware of that.

PR #14664, opened 2022-01-28
fix(VInput/VField): input and label location with density

johnleider:
## Description
Change how VField and VInput handle control sizing.

## Motivation and Context
The original implementation assumed a fixed height and used center alignment. Components like VTextarea and VSelect have a variable control height and require this change.

## How Has This Been Tested?
visually

## Markup:

<details>
https://www.toptal.com/developers/hastebin/xurekiqopa.kotlin
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VField/VField.sass
+++ b/packages/vuetify/src/components/VField/VField.sass
@@ -79,6 +79,7 @@
       @include tools.density('v-input', $input-density) using ($modifier)
         @at-root #{selector.nest(&, $root)}
           --v-input-control-height: #{$field-control-height + $modifier}
+          --v-field-padding-bottom: #{$field-control-padding-bottom + $modifier * .5}
 
   &--variant-underlined
     $root: &
@@ -97,7 +98,6 @@
   padding-inline-start: var(--v-field-padding-start)
   padding-inline-end: var(--v-field-padding-end)
   padding-top: var(--v-field-padding-top, $field-control-padding-top)
-  padding-bottom: var(--v-field-padding-bottom, $field-control-padding-bottom)
   width: 100%
 
   &:focus,
@@ -112,6 +112,7 @@
   flex: 1 0
   min-height: var(--v-input-control-height, $input-control-height)
   grid-area: field
+  padding-bottom: var(--v-field-padding-bottom, $field-control-padding-bottom)
   position: relative
 
 /* endregion */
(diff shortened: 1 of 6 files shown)


nekosaur:
I don't feel well-versed enough in the vfield/vinput code to comment on the changes, but the playground looks good. The only thing I would say is to put a min height on textarea

KaelWD:
Auto-grow no longer works. Not sure why this is a separate PR, how does it go with the changes in #14504?

packages/vuetify/src/components/VTextarea/VTextarea.sass
```diff
@@ -8,18 +8,8 @@
...
-    flex: 1 1 auto
-    outline: none
     mask-image: linear-gradient(to bottom, transparent, transparent calc(var(--v-field-padding-top, 0) / 1.25), black var(--v-field-padding-top, 0))
 
-    $root: &
-
-    @at-root
-      @include tools.density('v-input', $input-density) using ($modifier)
-        @at-root #{selector.append(&, $root)}
-          line-height: $textarea-line-height + math.div($modifier * .125, settings.$spacer) * 1.5
```
KaelWD, line 21: This shouldn't really be removed, without it density doesn't really do much to multi-line inputs.

packages/vuetify/src/components/VField/VField.sass
```diff
@@ -102,46 +95,35 @@
...
+  grid-area: append-inner
+
+.v-field__field,
 .v-field__prepend-inner,
 .v-field__append-inner,
 .v-field__clearable
-  align-items: center
+  align-items: flex-start
   display: flex
+  padding-top: var(--v-input-padding-top, $field-control-padding)
```
KaelWD, line 120: This needs to be on the input element

johnleider:
> Auto-grow no longer works. Not sure why this is a separate PR, how does it go with the changes in #14504?

Separate PR because v3-select is already massive and I wanted to make these changes in an isolated environment.

packages/vuetify/src/components/VField/VField.sass
johnleider: What does making this change solve that's broken in this PR?

packages/vuetify/src/components/VTextarea/VTextarea.sass
johnleider: I don't think textarea line-height should change with density.

johnleider:
> Auto-grow no longer works. Not sure why this is a separate PR, how does it go with the changes in #14504?

This does break auto-grow but fixes just about everything else. I'm opting for proceeding with it as a known bug to be fixed later since this is blocking #14504

PR #14769, opened 2022-02-28
fix(vprogresscircular): handle dom width change

### LINKED ISSUE #14767: [Bug Report][3.0.0-alpha.13] VProgressCircular not shown on v-show
### Environment
**Vuetify Version:** 3.0.0-alpha.13
**Last working version:** 2.6.3
**Vue Version:** 3.2.31
**Browsers:** Chrome 98.0.4758.102
**OS:** Windows 10

### Steps to reproduce
1. Open https://codepen.io/andrewspy/pen/qBVMpZo
2. Click INC button to cycle through VProgressCircular

### Expected Behavior
All 3 VProgressCircular should be shown accordingly

### Actual Behavior
Only the first VProgressCircular is shown

### Reproduction Link
<a href="https://codepen.io/andrewspy/pen/qBVMpZo" target="_blank">https://codepen.io/andrewspy/pen/qBVMpZo</a>

### Other comments
VProgressLinear will work as expected.

luxaritas:
## Description

The `size` property depends on the element's DOM width as calculated via `getBoundingClientRect`. As this method is not reactive, it means that the computed property is not recalculated when the element width changes in the DOM, namely when v-show is changed from false to true.

## Motivation and Context
When v-show is initially false, the initial value of size is 0 because it is not rendered in the
DOM.

resolves #14767

I'm not super happy with the solution of "re-set the ref every time onUpdated is called", but as far as I can tell 1) we need to actually use the DOM width, so that we're not hardcoding values from the SASS theme (this call is used to determine the width if the size is set to a keyword rather than a number) and 2) there's no other way to determine a change in v-show. I'm open to alternative suggestions (maybe there's some other utility I don't know of, as I'm new to the codebase!).

## How Has This Been Tested?

Manual testing with the use case given in #14767, copied below. I should be able to write a unit test for this given a little extra time.

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <!-- -->
      <div>
        Sample. <v-progress-circular :model-value="50" />
      </div>
      <v-btn @click="counter++">Inc</v-btn>
      <div v-show="counter % 3 == 0">
        0.
        <v-progress-circular :model-value="50" />
        <v-progress-linear :model-value="50" clickable />
      </div>
      <div v-show="counter % 3 == 1">
        1.
        <v-progress-circular :model-value="50" />
        <v-progress-linear :model-value="50" />
      </div>
      <div v-show="counter % 3 == 2">
        2.
        <v-progress-circular :model-value="50" />
        <v-progress-linear :model-value="50" />
      </div>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'
  export default {
    name: 'Playground',
    setup () {
      return {
        counter: ref(0),
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


KaelWD:
ResizeObserver would be preferable here, I think we have a composable for it already. Ideally this would all be done in CSS instead though, is it possible to use `calc()` and `var()` in `stroke-width`?

luxaritas:
Good call, knew there must be a better way :) I'll try it out tomorrow

luxaritas:
I don't think this can be done with CSS, as the diameter (which also uses size) is used in the viewBox parameter, which from what I can tell only supports an actual number

luxaritas:
Changed to use ResizeObserver. I didn't implement the unit test as the spec file is currently skipped, but if I should still add it just let me know

packages/vuetify/src/components/VProgressCircular/VProgressCircular.tsx
```diff
@@ -9,11 +9,12 @@ import { useIntersectionObserver } from '@/composables/intersectionObserver'
 import { useTextColor } from '@/composables/color'
 
 // Utilities
-import { computed, toRef } from 'vue'
+import { computed, ref, toRef, watchEffect } from 'vue'
 import { convertToUnit, defineComponent } from '@/util'
 
 // Types
 import type { PropType } from 'vue'
+import { useResizeObserver } from '@/composables/resizeObserver'
```
KaelWD, line 17: Move to line 9

luxaritas: Sorry for the delay, fixed

KaelWD (approved):

PR #14819, opened 2022-03-11
fix: Replace deprecated String.prototype.substr()

CommanderRoot:
## Description
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## How Has This Been Tested?
I have tested all functions individually to make sure they return the same results as before

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/components/doc/ColorPalette.vue
+++ b/packages/docs/src/components/doc/ColorPalette.vue
@@ -106,7 +106,7 @@
       },
       convertToClass (str) {
         const end = this.endStr(str)
-        const sub = str.substr(0, str.length - 1)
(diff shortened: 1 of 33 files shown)


KaelWD:
Thanks, but if it ain't broke don't fix it.

PR #14920, opened 2022-04-05
feat(VSnackbar): re-implement using VOverlay

### LINKED ISSUE #14909: [Bug Report][3.0.0-beta.0] VSnackbar is cropped when inside toolbar
### Environment
**Vue Version:** 3.2.31
**Browsers:** Chrome 98.0.4758.136
**OS:** Linux x86_64

### Steps to reproduce
I don't know whether this is a bug or "not supported/as-designed":
- open the codepen
- observe how the left snackbar looks great, but the right snackbar is cropped by the app
- the left snackbar is inside the top-level VApp while the right one is inside VToolBar
- my use-case has the snackbar nested inside a status icon that sits in the toolbar, so it's not that nice or elegant to move it to the app top-level

### Expected Behavior
I expected the snackbar to be lifted to the top of the page so it can always sit on top of everything else

### Actual Behavior
The snackbar is cropped.

### Reproduction Link
<a href="https://codepen.io/codesurfr/pen/mdpMpKR" target="_blank">https://codepen.io/codesurfr/pen/mdpMpKR</a>

KaelWD:
resolves #14909

```vue
<template>
  <v-app>
    <v-container>
      <v-snackbar timeout="-1">
          <v-btn v-bind="props">Snackbar</v-btn>
        </template>
        Snackbar
      </v-snackbar>

      <v-dialog>
          <v-btn v-bind="props">Dialog</v-btn>
        </template>
        <v-card title="Dialog" />
      </v-dialog>
    </v-container>
  </v-app>
</template>
```


nekosaur:
Should centered/top/left etc be condensed into single prop?

KaelWD:
Yeha probably, any ideas?

nekosaur:
Seems like we're not terribly consistent. There's already `anchor` from overlay. It's used similarly in tooltip. But in nav drawer we use `position`. And in badge it's called `location`.

PR #14929, opened 2022-04-06
fix: update router compatible components when route changes

### LINKED ISSUE #14914: [Bug Report][3.0.0-beta.0] VTab not correctly selected on initial load when using vue-router
### Environment
**Vue Version:** 3.2.31
**Browsers:** Chrome 99.0.4844.84
**OS:** Mac OS 10.15.7

### Steps to reproduce
I believe I'm encountering a bug on the latest beta. On initial page load it doesn't seem to be set to the correct tab if using the 'to' prop (there's an issue with code sandbox loading modules, just click the refresh button in the browser to get rid of the 'cannot use import statement' message). Click the 'Foo' or 'Bar' tab item then refresh the code sandbox browser again to see the issue (if you see the import error again just refresh, it will go away). You can see the selected tab will be 'Home' even though the url is using '/foo' or '/bar' depending on what was chosen.

### Expected Behavior
Selecting the 'Foo' tab then refreshing, the url should contain '/foo' and the 'Foo' tab should be selected.

### Actual Behavior
Selecting the 'Foo' tab then refreshing, the url does contain '/foo' but the 'Home' tab is selected instead.

### Reproduction Link
<a href="https://codesandbox.io/s/inspiring-wilson-6wbbn9?file=/src/App.vue" target="_blank">https://codesandbox.io/s/inspiring-wilson-6wbbn9?file=/src/App.vue</a>

nekosaur:
## Description

maybe something like this?

fixes #14914

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-navigation-drawer :model-value="true" permanent>
      <v-list :items="items" v-model:selected="bar"></v-list>
      <pre>{{ bar }}</pre>
    </v-navigation-drawer>
    <v-main>
      <v-tabs v-model="foo">
        <v-tab v-for="item in items" :key="item.name" :to="item.to" color="primary">
          {{ item.title }}
        </v-tab>
      </v-tabs>
      <pre>{{ foo }}</pre>
      <v-btn to="/">asd</v-btn>
      <v-tabs
        v-model="tab"
        background-color="primary"
      >
        <v-tab value="one">Item One</v-tab>
        <v-tab value="two">Item Two</v-tab>
        <v-tab value="three">Item Three</v-tab>
      </v-tabs>
      Component loads below:
      <router-view></router-view>
    </v-main>
  </v-app>
</template>

<script>
  export default {
    name: 'App',
    data: () => ({
      foo: null,
      bar: null,
      tab: null,
      items: [
        {
          title: 'Home',
          to: { name: 'home' },
        },
        {
          title: 'Foo',
          to: { name: 'page1' },
        },
        {
          title: 'Bar',
          to: undefined,
        },
      ],
    }),
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/dev/router.js
+++ b/packages/vuetify/dev/router.js
@@ -4,32 +4,32 @@ const home = {
   setup: () => () => h('div', 'hello'),
 }
 const page1 = {
-  setup: () => () => h('div', 'hello'),
+  setup: () => () => h('div', 'page1'),
 }
 const page2 = {
-  setup: () => () => h('div', 'hello'),
+  setup: () => () => h('div', 'page2'),
 }
 const nested1 = {
-  setup: () => () => h('div', 'hello'),
+  setup: () => () => h('div', 'nested1'),
 }
 const nested2 = {
-  setup: () => () => h('div', 'hello'),
+  setup: () => () => h('div', 'nested2'),
 }
 
 export const routes = [
   {
     path: '/',
-    name: 'Home',
+    name: 'home',
     component: home,
   },
   {
     path: '/page1',
-    name: 'Page 1',
+    name: 'page1',
     component: page1,
   },
   {
     path: '/page2',
-    name: 'Page 2',
+    name: 'page2',
     component: page2,
   },
   {
--- a/packages/vuetify/src/components/VBtn/VBtn.tsx
+++ b/packages/vuetify/src/components/VBtn/VBtn.tsx
@@ -18,6 +18,7 @@ import { makeSizeProps, useSize } from '@/composables/size'
 import { makeTagProps } from '@/composables/tag'
 import { makeThemeProps, provideTheme } from '@/composables/theme'
 import { genOverlays, makeVariantProps, useVariant } from '@/composables/variant'
+import { useSelectLink } from '@/composables/selectLink'
 
 // Directives
 import { Ripple } from '@/directives/ripple'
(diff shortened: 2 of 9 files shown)


nekosaur:
When present, this line https://github.com/vuetifyjs/vuetify/pull/14929/files#diff-dc31fd609311b4cc3871797d3e46fc36d477238237b4f3ff45151c5edf2581ecL318 was causing recursive updates

```
[Vue warn]: Maximum recursive updates exceeded in component <VSlideGroup>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.
```

Removing the reference to `group.selected.value` made it go away. I'm not exactly sure why it's needed in the first place? @johnleider

KaelWD:
For keyboard support, the first active tab is focused then you use arrow keys to switch between them: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html

nekosaur:
hm. it seems to work fine with static `-1`. or I'm just not doing the right steps

KaelWD:
Ah right I had it confused with the one on VTab which you kept. This one is so you can still focus something if no items are selected (`:mandatory="false"`).

nekosaur:
It didn't always happen. Try loading playground, then saving VBtn.tsx. That seemed to be pretty consistent for me

KaelWD:
Does this replace #14868?

I'm putting the tabindex back, that hmr bug is acceptable for now.

KaelWD (approved):

nekosaur:
Yes

PR #14943, opened 2022-04-09
fix(validation): respect number 0 when calculating isDirty

### LINKED ISSUE #14942: [Bug Report][3.0.0-beta.0] VTextField thinks 0 is an empty value
### Environment
**Vue Version:** 3.2.19
**Browsers:** Firefox 99.0
**OS:** Windows 10

### Steps to reproduce
Set value to 2, click outside the field
Set value to 0, click outside the field

### Expected Behavior
0 should show even when focus is lost since the field isn't empty

### Actual Behavior
the label hides the number

### Reproduction Link
<a href="https://codepen.io/SuperCuber2/pen/VwyxQmV" target="_blank">https://codepen.io/SuperCuber2/pen/VwyxQmV</a>

yuwu9145:
fixes #14942

## Description

fixes #14942

## Markup:

<details>

```vue
<template>
  <v-app>
   <v-container>
      Note how when num=0 the number gets hidden on lose focus but when num>0 it works fine
      <v-text-field label="works" type="number" />
    </v-container> 
  </v-app> 
</template>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/validation.ts
+++ b/packages/vuetify/src/composables/validation.ts
@@ -56,7 +56,7 @@ export function useValidation (
   const form = useForm()
   const internalErrorMessages = ref<string[]>([])
   const isPristine = ref(true)
-  const isDirty = computed(() => wrapInArray(model.value === 0 ? 0 : (model.value || [])).length > 0)
+  const isDirty = computed(() => wrapInArray(model.value).length > 0)
   const isDisabled = computed(() => !!(props.disabled || form?.isDisabled.value))
   const isReadonly = computed(() => !!(props.readonly || form?.isReadonly.value))
   const errorMessages = computed(() => {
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/composables/validation.ts
```diff
@@ -56,7 +56,7 @@ export function useValidation (
   const form = useForm()
   const internalErrorMessages = ref<string[]>([])
   const isPristine = ref(true)
-  const isDirty = computed(() => wrapInArray(model.value || []).length > 0)
```
yuwu9145, line 59: old code seems try to convert falsy values to `[]`, however, `0` should  not be treated as falsy in this case. Here is summary of what should happen:

- for `null` & `undefined`, `wrapInArray` should return `[]` which has been implemented in `wrapInArray`
- for `0`, it should be passed as `0` and `wrapInArray` will return `[0]` as expected
- for `''` however, it should be converted to `null` ,  `wrapInArray` will return `[]` which makes `isDirty` as `false`

Therefore, in terms of `isDirty` calculation, only empty string `''` should be specifically addressed

KaelWD (approved):

PR #15012, opened 2022-04-28
Update ca translation

jaumesala:


KaelWD:
This should be rebased to the master branch if it is relevant for v2 also

PR #15079, opened 2022-05-10
feat(items): add new composable

johnleider:
## Description
New composable for handling items functionality in components. Adds support for **item-text**, **item-value**, and **item-children**. Added new **item-props** prop that accepts a function that returns an object that is merged with the default **text, value, and children** properties.

```ts
function itemProps (item: any) {
  return {
    // these properties + values are added to the generated items object
    subtitle: 'lorem ipsum',
    color: 'blue',
    active: true,
  }
}
```

**TODO:**
* improve typings
* add unit tests

## Motivation and Context
Normalize how we derive item object property values in components such as `v-list` and `v-select`

## How Has This Been Tested?
**TODO**

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-main>
      <div class="ma-4 pa-4">
        <v-list :items="items" :item-props="itemProps" />
      </div>
    </v-main>
  </v-app>
</template>

<script setup>
  // import { computed, ref } from 'vue'

  const items = [
    {
      title: 'Foo',
      value: 'foo',
      id: 0,
      what: 'is',
    },
  ]

  const itemProps = item => {
    return {
      subtitle: 'lorem ipsum',
      id: item.id,
      color: 'blue',
      active: true,
    }
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/items.ts
+++ b/packages/vuetify/src/composables/items.ts
@@ -1,0 +1,56 @@
+// Utilities
+import { computed } from 'vue'
+import { getPropertyFromItem, propsFactory } from '@/util'
+
+// Types
+export interface ItemProps {
+  [key: string]: any
+}
+
+// Composables
+export const makeItemsProps = propsFactory({
+  items: {
+    type: Array,
+    default: () => ([]),
+  },
+  itemText: {
+    type: String,
+    default: 'title',
+  },
+  itemValue: {
+    type: String,
+    default: 'value',
+  },
+  itemChildren: {
+    type: String,
+    default: 'children',
+  },
+  itemProps: {
+    type: Function,
+    default: (item: any) => ({}),
+  },
+}, 'item')
+
+export function useItems (props: ItemProps) {
+  const items = computed(() => {
+    const array = []
+
+    for (const item of props.items) {
+      const children = item[props.itemChildren]
+
+      const newItem = {
+        [`${props.itemText ?? 'text'}`]: item[props.itemText],
+        value: getPropertyFromItem(item, props.itemValue, item),
+        ...props.itemProps(item),
+      }
+
+      if (children) newItem.children = children
+
+      array.push(newItem)
+    }
+
+    return array
+  })
+
+  return { items }
+}
--- a/packages/vuetify/src/components/VList/VList.tsx
+++ b/packages/vuetify/src/components/VList/VList.tsx
@@ -26,6 +26,7 @@ import { genericComponent, useRender } from '@/util'
 import type { Prop, PropType } from 'vue'
 import type { MakeSlots } from '@/util'
 import type { ListGroupActivatorSlot } from './VListGroup'
+import { makeItemsProps, useItems } from '@/composables/items'
 
 export type ListItem = {
   [key: string]: any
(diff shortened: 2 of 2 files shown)


packages/vuetify/src/composables/items.ts
```diff
@@ -0,0 +1,56 @@
...
+
+// Composables
+export const makeItemsProps = propsFactory({
+  items: {
+    type: Array,
+    default: () => ([]),
+  },
+  itemText: {
+    type: String,
+    default: 'title',
```
nekosaur, line 18: Why not align name and default value by changing this to `itemTitle`, the way the others are?

packages/vuetify/src/composables/items.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function useItems (props: ItemProps) {
+  const items = computed(() => {
+    const array = []
+
+    for (const item of props.items) {
+      const children = item[props.itemChildren]
+
+      const newItem = {
+        [`${props.itemText ?? 'text'}`]: item[props.itemText],
```
nekosaur, line 42: In v2 we supported dot notation for deeply nested properties

nekosaur:
Is return-object in scope for this PR?

packages/vuetify/src/composables/items.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Composables
+export const makeItemsProps = propsFactory({
+  items: {
+    type: Array as PropType<ItemProps['items']>,
+    default: () => ([]),
+  },
+  itemTitle: {
+    type: String,
```
KaelWD, line 30: `getPropertyFromItem` supports `SelectItemKey`

packages/vuetify/src/composables/items.ts
```diff
@@ -0,0 +1,77 @@
...
+  itemProps: {
+    type: Function as PropType<ItemProps['itemProps']>,
+    default: (item: any) => ({}),
+  },
+}, 'item')
+
+export function transformItem (props: ItemProps, item: any) {
+  const title = getPropertyFromItem(item, props.itemTitle, item)
+  const value = getPropertyFromItem(item, props.itemValue, title)
+  const children = getObjectValueByPath(item, props.itemChildren, [])
```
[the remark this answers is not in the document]
KaelWD: Is there a case were you'd want `children: []` in the original item to be kept?

packages/vuetify/src/composables/items.ts
johnleider: Would the same logic apply to **itemValue**? Should **itemChildren** be updated in some form since it uses **getObjectValueByPath**?

KaelWD: - yes
- yeah we were using `getPropertyFromItem` before but it returns the item if the property is undefined

packages/vuetify/src/composables/items.ts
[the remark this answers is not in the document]
johnleider: Can we differentiate between **undefined** from an empty **array**?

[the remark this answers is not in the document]
nekosaur: The current logic in treeview (and I think it's the same way in v2) is that no children property indicates leaf node. If the item has an empty children array, then it's still a branch node because it might contain lazy loaded items.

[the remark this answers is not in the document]
johnleider: What is the expected behavior if an item has **children** and that value is **undefined**? Can we use that in some way?

PR #15082, opened 2022-05-11
Update VDatePicker.ts

doox911:
Optimized work. Extra calls .split()

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.ts
@@ -352,8 +352,11 @@ export default mixins(
       }
     },
     monthClick (value: string) {
-      this.inputYear = parseInt(value.split('-')[0], 10)
-      this.inputMonth = parseInt(value.split('-')[1], 10) - 1
+      const [year, month] = value.split('-')
+
+      this.inputYear = parseInt(year, 10)
+      this.inputMonth = parseInt(month, 10) - 1
+      
       if (this.type === 'date') {
         if (this.inputDay) {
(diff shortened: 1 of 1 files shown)


KaelWD (dismissed):
split is fast and these are only called on click so the performance difference is practically nothing, it is a bit cleaner though

adamkasper (approved):

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
```diff
@@ -352,8 +352,11 @@ export default mixins(
       }
     },
     monthClick (value: string) {
-      this.inputYear = parseInt(value.split('-')[0], 10)
-      this.inputMonth = parseInt(value.split('-')[1], 10) - 1
+      const [year, month] = value.split('-')
+
+      this.inputYear = parseInt(year, 10)
```
adamkasper, line 357: what better to use `+year` instead of `parseInt()`?

doox911: Becouse: https://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which/17106702.

We need integer.

KaelWD: Other places use `value.split('-').map(Number)` but this is fine

KaelWD:
You still have lint errors

packages/vuetify/src/components/VDatePicker/VDatePicker.ts
doox911: @KaelWD When to release?)

doox911:
@KaelWD what happend?)) I just removed the spaces, and I don't see the errors of lint or test

PR #15103, opened 2022-05-17
feat: Logical border utilities (RTL support)

jd-solanki:
## Description
This PR will make border related utilities like `rounded-tl` and `border-l`  RTL compatible by changing the property and adjusting the class name according to its property.

This PR aligns with vuetify 3 preference (This came to mind thanks to @KaelWD) of renaming `left` => `start` and `right` => `end`. It preserves the top and bottom shorthand. Hence, `rounded-tl` is `rounded-ts`.

### Why `rounded-tl` !== `rounded-ss`

I renamed class `rounded-tl` to `rounded-ts` and not `rounded-ss`. This is because if we try to align with logical property can rename it to `rounded-ss`, it confuses dev who don't have experience with logical properties.

Moreover, `rounded-ts` can be easily identified as top left.

## Motivation and Context
Vuetify 3 RTL support for border related utilities wasn't complete.

## How Has This Been Tested?
I updated the tests but I didn't run them because I don't have knowledge of tests in Vue.

## Markup:
-

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/examples/border-radius/misc-removing-border-radius.vue
+++ b/packages/docs/src/examples/border-radius/misc-removing-border-radius.vue
@@ -5,8 +5,8 @@
       .rounded-0
     </div>
 
-    <div class="pa-4 text-center bg-secondary rounded-xl rounded-br-0">
-      .rounded-xl .rounded-br-0
+    <div class="pa-4 text-center bg-secondary rounded-xl rounded-be-0">
+      .rounded-xl .rounded-be-0
     </div>
   </div>
 </template>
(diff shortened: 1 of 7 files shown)


KaelWD:
These are only supported since Safari 15, we're targeting 13.1 in v3. You can use class-based rtl instead which works in all browsers: https://github.com/vuetifyjs/vuetify/blob/13ab192c573a8401c5f4c52d4b0d3b1bf5f3c349/packages/vuetify/src/styles/settings/_utilities.scss#L203-L214

PR #15132, opened 2022-05-21
fix(VBtn): loading prop

### LINKED ISSUE #15056: [Feature Request] The button component has no loading state
### Problem to solve
Often the user interface needs to display the loading state of the button, which is not currently the case

### Proposed solution
Use the same API that was provided in [version 2.6.5](https://vuetifyjs.com/en/components/buttons/#loaders)

zorn-v:
Fixes #15056

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.tsx
+++ b/packages/vuetify/src/components/VBtn/VBtn.tsx
@@ -4,6 +4,7 @@ import './VBtn.sass'
 // Components
 import { VBtnToggleSymbol } from '@/components/VBtnToggle/VBtnToggle'
 import { VIcon } from '@/components/VIcon'
+import { VProgressCircular } from '@/components/VProgressCircular'
 
 // Composables
 import { makeBorderProps, useBorder } from '@/composables/border'
@@ -45,6 +46,8 @@ export const VBtn = defineComponent({
 
     block: Boolean,
     stacked: Boolean,
+
+    loading: Boolean,
 
     ripple: {
       type: Boolean,
@@ -102,6 +105,7 @@ export const VBtn = defineComponent({
               'v-btn--flat': props.flat,
               'v-btn--icon': !!props.icon,
               'v-btn--stacked': props.stacked,
+              'v-btn--loading': props.loading,
             },
             themeClasses.value,
             borderClasses.value,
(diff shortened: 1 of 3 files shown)


zorn-v:
By the way it also will be possible to fix text overflow issue because added `.v-btn__content` wrapper.

Before
изображение


After (with added styles)
```css
.v-btn__content {
  width: 100%;
  overflow: hidden;
  text-overflow: ellipsis;
}
```
изображение

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -153,14 +157,27 @@ export const VBtn = defineComponent({
...
+          { props.loading && (
+            <span class="v-btn__loader">{
+              slots.loader ? slots.loader() : (
+                <VProgressCircular
+                  indeterminate
+                  size="23"
+                  width="2"
+                />
+              )
+            }</span>
```
KaelWD, line 179: We have a component for this, see VSwitch for example usage

johnleider:
Buttons should not ellipsis text by default.

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
```diff
@@ -125,6 +125,29 @@ describe('VBtn', () => {
...
+    it('when loading is true, show the progress indicator', () => {
+      cy.mount(<VBtn loading>{ loadingText }</VBtn>)
+        .get('button')
+        .should('contain.text', loadingText)
+        .get('.v-btn__content')
+        .should('not.be.visible')
+        .get('[role="progressbar"]')
+        .should('be.visible')
+    })
+  })
```
KaelWD, line 149: Kinda pointless really, just the visual test is fine

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -153,14 +157,27 @@ export const VBtn = defineComponent({
...
+
           ) }
```
KaelWD, line 181:
```suggestion
          { props.loading && (
            <span class="v-btn__loader">
              { slots.loader ? slots.loader() : (
                <VProgressCircular
                  indeterminate
                  size="23"
                  width="2"
                />
              )}
            </span>
          )}
```

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
zorn-v: What you mean ? I just get existing disabled test and add some fixes.

zorn-v:
> Buttons should not ellipsis text by default.

It just for example. It should not be in core indeed.
In your project you may want something like
```css
mask-image: linear-gradient(to left, transparent 10px, black 30px);
```

packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
KaelWD: I mean these can be deleted, it's already covered on line 29

zorn-v: In which case it cover "default progress does not exist with slot but slot is exist", "btn text does not visible when in loading state" and "loading progress bar should be visible" ?

zorn-v: By the way - `btn__content` should be around ALL content including pre/post icons/slots

KaelWD:
There was no disabled style by default with loading in v2, the examples have `:disabled="isLoading"`

I'm not happy with the LoaderSlot component for this, it was made for a linear progress on inputs so needs a big ugly slot to use a circular progress instead.

PR #15140, opened 2022-05-23
feat(V-Form): add reactive errorMessages

Surof1n:
## Description

This PR adds a reactivity for errors. This change helps to make error messages more flexible.

## Motivation and Context

This will help solve the problem if the developer needs to render errors or use them at the time of the actual occurrence, and not at the time of full form validation.

## How Has This Been Tested?

add e2e test

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-form ref="form" v-model="isValid">
        <template #default="{ errorMessages }">
          <v-text-field
            v-model="name1"
            :max-errors="2"
            required
            :rules="[(v) => !!v || 'Error 1']"
            class="mb-2"
            label="Label 3"
            variant="underlined"
            single-line
          />
          <v-text-field
            v-model="name2"
            :max-errors="2"
            required
            :rules="[(v) => !!v || 'Error 2']"
            class="mb-2"
            label="Label 3"
            variant="underlined"
            single-line
          />
          <v-text-field
            v-model="name3"
            :max-errors="2"
            required
            :rules="[(v) => !!v || 'Error 3']"
            class="mb-2"
            label="Label 3"
            variant="underlined"
            single-line
          />
          <div class="errors">
            {{ errorMessages }}
          </div>
          <div>valid status: {{isValid}}</div>
        </template>
      </v-form>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    data () {
      return {
        isValid: false,
        name1: '',
        name2: '',
        name3: '',
        menu: false,
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


Surof1n:
@KaelWD Sorry, maybe run a PR workflow, thanks!

Surof1n:
@KaelWD sorry, tell me how I can conduct PR in the next branch

KaelWD:
Please don't reformat files like that

packages/vuetify/src/composables/form.ts
```diff
@@ -56,22 +63,33 @@ export const makeFormProps = propsFactory({
...
+    default: false,
+  },
 })
 
 export function createForm (props: FormProps) {
   const model = useProxiedModel(props, 'modelValue')
 
   const isDisabled = computed(() => props.disabled)
   const isReadonly = computed(() => props.readonly)
+  const isReactiveErrors = computed(() => props.isReactiveErrors)
```
KaelWD, line 77: This computed property doesn't need to exist

KaelWD:
Is this supposed to be like v2's lazyValidation, which seems to not be implemented yet?

Surof1n:
> Please don't reformat files like that

ok, this need change?

Surof1n:
> Is this supposed to be like v2's lazyValidation, which seems to not be implemented yet?

I read the lazyValidation functionality, my changes are not very similar to this, I described in the motivation for what I needed it.

packages/vuetify/src/composables/form.ts
Surof1n: i fixed it

nekosaur:
Do we not want `errorMessages` to always reflect current state?

Surof1n:
> Do we not want `errorMessages` to always reflect current state?

When I was developing, I ran into a problem that I only receive error messages at the time of form validation, I decided that it would be worthwhile to leave the old behavior and add some prop for it and create a situation where errorMessages appeared after the form elements were validated.

PR #15152, opened 2022-05-25
feat: consolidate all position and location props

### LINKED ISSUE #14952: Normalise "direction" props
https://github.com/vuetifyjs/vuetify/pull/14920#issuecomment-1093790709

- VOverlay - `anchor`/`origin` as `'top' | 'bottom' | 'start' | 'end' | 'center' | 'top start' | 'end center'` etc.
- VNavigationDrawer - `position` as `'left' | 'right' | 'bottom'`
- VBadge - `location` as `${'top' | 'bottom'}-${'start' | 'end'}`
- VSnackbar - `centered` + `top`/`bottom`/`left`/`right`

`makePositionProps` used by VSheet, VAlert, VBanner, VBtn, VCard, VFooter, and VSnackbar also includes
 - `top`/`bottom`/`left`/`right`
 - `absolute`/`fixed`
 - `position` as `'static' | 'relative' | 'fixed' | 'absolute' | 'sticky'` (uses utility classes)

KaelWD:
resolves #14952

- `anchor`/`position`/`location`/`top`/`bottom`/`left`/`right` combined into a single `location` prop
- `position`/`absolute`/`fixed` combined into `position`, now uses per-component classes instead of utility classes
- layout components still use `absolute` because they're fixed by default and I'm lazy
  - They also still use left/right instead of start/end
  - I didn't fix this but I noticed navigation drawers are inconsistent with rtl. `location="left"` opens on the left side, but applies a `--start` class which is rtl-aware for the border

### DIFF
--- a/packages/docs/src/components/about/TeamMember.vue
+++ b/packages/docs/src/components/about/TeamMember.vue
@@ -16,7 +16,7 @@
             <v-tooltip
               v-if="link.href || link.copyText"
               :key="link.href || link.copyText"
-              anchor="bottom"
+              location="bottom"
             >
               <template #activator="{ props }">
                 <a
--- a/packages/docs/src/components/app/Markup.vue
+++ b/packages/docs/src/components/app/Markup.vue
@@ -27,7 +27,7 @@
 
       <v-spacer />
 
-      <v-tooltip anchor="bottom">
+      <v-tooltip location="bottom">
         <template #activator="{ props }">
           <v-btn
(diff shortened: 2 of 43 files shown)


Vincebe:
A lot more consistent.
Thank you for the awesome work!

Most likely WIP but FYI with nightly 3.0.0-next-20220529.0 `v-snackbar` position `top right` doesn't work (`top left` is OK).

PR #15188, opened 2022-06-02
fix(form): errorMessages should be reactive

nekosaur:
## Description

Alternative solution to https://github.com/vuetifyjs/vuetify/pull/15140

Here `errorMessages` is always reactive. Not sure if there are scenarios where you'd want the previous behaviour?

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


KaelWD:
If you have an input without rules then form.valid is always null:
```html
<template>
  <v-app>
    <v-container>
      <v-form ref="form" v-slot="{ errorMessages }" v-model="valid" @submit.prevent>
        <v-text-field :rules="rules" />
        <v-text-field :rules="rules" />
        <v-text-field />
        
        <pre>{{ { valid, errorMessages } }}</pre>

        <v-btn type="submit">Validate</v-btn>
      </v-form>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'
  import { VForm } from '../src/components'

  const rules = [
    v => !!v || 'Required',
    v => v.length < 5 || 'More than 5',
  ]

  const form = ref<VForm>()
  const valid = ref<boolean>(false)
</script>
```

nekosaur:
Not sure how we want form to work. With your example, form should of course validate to true if first two inputs have valid inputs. But without writing anything in third field, should it do it "on the fly", or just when explicitly calling validate?

For immediate valid state we'd probably need something like https://github.com/vuetifyjs/vuetify/pull/15191

Surof1n:
You don't always use reactive error messages, you still get errors under form elements.

KaelWD:
No reason it shouldn't always be available though. 

> without writing anything in third field, should it do it "on the fly"?

Yeah if there's no rules (or error-messages) it's always valid.

johnleider (approved):

KaelWD (approved):

PR #15191, opened 2022-06-02
feat(validation): Add 'optional' prop

carlesalbasboix:
## Description
Added a boolean prop called 'optional' to the validation composable. If the model value is empty (undefined, empty string or empty array) and optional is set, then the validation ignores the rules and returns true.

Useful for:
- Validation rules don't need to take into account an empty input.
- An optional field will be automatically valid on render if the initial value is empty.

## Motivation and Context
Right now when writing validation rules you must take into account whether you accept an empty string. Furthermore, an empty optional field will not be valid by default until the user changes it's value. This solves the problem of having optional fields that still require rules for when they are set.

## How Has This Been Tested?
There are no tests yet. I want to wait and see whether the functionality is approved before writing tests.

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      Hola
      <v-form v-model="formValid">
        <v-text-field v-model="inputOne" optional :rules="[v => 'Rule fails']" />
      </v-form>
      {{ formValid ? "valid" : "invalid" }}
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const formValid = ref(false)

  const inputOne = ref('')
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/composables/validation.ts
+++ b/packages/vuetify/src/composables/validation.ts
@@ -27,6 +27,7 @@ export interface ValidationProps {
   maxErrors: string | number
   name: string | undefined
   label: string | undefined
+  optional: boolean
   readonly: boolean
   rules: ValidationRule[]
   modelValue: any
@@ -48,6 +49,7 @@ export const makeValidationProps = propsFactory({
   },
   name: String,
   label: String,
+  optional: Boolean,
   readonly: Boolean,
   rules: {
     type: Array as PropType<ValidationRule[]>,
@@ -81,7 +83,12 @@ export function useValidation (
       ? wrapInArray(props.errorMessages).slice(0, Math.max(0, +props.maxErrors))
       : internalErrorMessages.value
   })
+  const isEmpty = computed(() => {
+    return model.value === undefined || model.value === '' || model.value === null ||
+      (Array.isArray(model.value) && model.value.length === 0)
+  })
   const isValid = computed(() => {
+    if (props.optional && isEmpty.value) return true
     if (props.error || errorMessages.value.length) return false
     if (!props.rules.length) return true
 
@@ -156,24 +163,26 @@ export function useValidation (
 
     isValidating.value = true
 
-    for (const rule of props.rules) {
-      if (results.length >= +(props.maxErrors ?? 1)) {
-        break
+    if (!props.optional || !isEmpty.value) {
+      for (const rule of props.rules) {
+        if (results.length >= +(props.maxErrors ?? 1)) {
+          break
+        }
+
+        const handler = typeof rule === 'function' ? rule : () => rule
+        const result = await handler(validationModel.value)
+
+        if (result === true) continue
+
+        if (typeof result !== 'string') {
+          // eslint-disable-next-line no-console
+          console.warn(`${result} is not a valid value. Rule functions must return boolean true or a string.`)
+
+          continue
+        }
+
+        results.push(result)
       }
-
-      const handler = typeof rule === 'function' ? rule : () => rule
-      const result = await handler(validationModel.value)
-
-      if (result === true) continue
-
-      if (typeof result !== 'string') {
-        // eslint-disable-next-line no-console
-        console.warn(`${result} is not a valid value. Rule functions must return boolean true or a string.`)
-
-        continue
-      }
-
-      results.push(result)
     }
 
     internalErrorMessages.value = results


johnleider:
Could you provide markup for your use-case?

carlesalbasboix:
I have a form with an optional email. I want the validation to evaluate to true either if the email is valid or the input is left blank. Hence this "optional" prop that automatically evaluates the validation to true if the input is blank.

carlesalbasboix:
Ah, ok, I resolved the branch conflicts and provided the markup.

johnleider:
@KaelWD how does this fit in with our conversation regarding lazy-validation last week?

Fadiabb:
Hi,
is there a plan when this will be merged and included in a coming release ?

johnleider:
Maybe 3.1. We are pretty much locked for v3 release at this point.

carlesalbasboix:
@johnleider Can we consider it now that v3 has been released?

carlesalbasboix:
Any news?

johnleider:
@KaelWD following up on https://github.com/vuetifyjs/vuetify/pull/15191#issuecomment-1224209600

packages/vuetify/src/composables/validation.ts
```diff
@@ -81,7 +83,11 @@ export function useValidation (
       ? wrapInArray(props.errorMessages).slice(0, Math.max(0, +props.maxErrors))
       : internalErrorMessages.value
   })
+  const isEmpty = computed(() => {
```
nekosaur, line 86: no need for this computed I think. should be fine to just use `if (model.value)`

carlesalbasboix: I did it like that because originally I was also checking for an empty array, but I removed it I don't remember why. I could add it again.

carlesalbasboix: For example,
```suggestion
  const isEmpty = computed(() => !props.modelValue || (Array.isArray(props.modelValue) && !props.modelValue.length))
```

carlesalbasboix: And I'm not sure doing if(model.value) is good, because we would count 0 as an empty value which I don't think it should be.

carlesalbasboix: I have updated isEmpty

nekosaur: the model value is always a string though, so 0 shouldn't be an issue

KaelWD: Yeah there isn't really a consistent definition for "empty". Text fields would be `''` or `null`, but both are valid select and radio values. `[]` could be empty for multi select but not single select or radio. Really the only empty value everywhere is `undefined` 

[Playground](https://play.vuetifyjs.com/#eNqlkj1v2zAQhv8KwUUJYJIOPNVQjHTt0A4dLQ+MdE7Y8gskpaIQ9N9zlBTHsYNAQCaJx+fe9z647+l373nXAt3SMoHxWibYVZaQsmPS+/F3PNTOJqkshDk0BoNslGNPwbWedMy4BvR9RSNoqBM0FT2xbzTZdlK3gJhttUYkf0ox336acJQ6AmaM32UpRYF8USyD18iul6H7A7L7wzV8ikxTwfjpxgfY9T358fvXTx5TUPZJHf/fvA7rlgxDKTIzjxyF3s88R6aVlOJsVXiMdVA+kQhpdCREGe9CIj0JcCQDOQZnSIFbLqZyUDdmfHIm9xm7uc26k9KOrugkwYz0/E90Ft9Hn1Or+SJWdEvGSI6hdD5X9DklH7dC1I3FNHwOqgvcQhLWG/GAmAitTcoAa5x52PANXzOp/bPk30SjYjq/5hANewzuX4SAYhVdndklnN0Cy5lEpzt+t5ks5hjT8jFmjyttgYYdBBbANhAgLG3tIu2j9i6QqxZzFUNlB1yAVvZvvJh9Hce571+L/Vrvo1qWOmTHYUXfFYwlTJn08ALHJVx8)

carlesalbasboix: What is your proposal then? Can we make it component-specific? Each component that uses validation could have a list of 'empty values'.

johnleider:
Pulse check on this.

cintek:
@johnleider may I ask you why you closed this PR? We are waiting for such a prop but now it's unclear if there will be such a feature or not.

johnleider:
> @johnleider may I ask you why you closed this PR? We are waiting for such a prop but now it's unclear if there will be such a feature or not.

Because there was no response from the author for over a year.

PR #15211, opened 2022-06-05
feat(VInput): Show error messages with hide-details

zorn-v:
## Description

изображение

## Motivation and Context

If you have input with hide-details you can not see why it is in error state.

## How Has This Been Tested?

visually

## Markup:

<details>

```vue
<v-text-field hide-details :rules="[true]" :error-messages="['ERROR 1', 'ERROR 2']"/>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VInput/VInput.tsx
+++ b/packages/vuetify/src/components/VInput/VInput.tsx
@@ -171,7 +171,7 @@ export const VInput = genericComponent<new <T>() => {
                 activator="parent"
               >
                 <VMessages
-                  active={ true }
+                  active
                   messages={ errorMessages.value }
                 />
               </VTooltip>
--- a/packages/vuetify/src/components/VInput/VInput.tsx
+++ b/packages/vuetify/src/components/VInput/VInput.tsx
@@ -158,7 +158,7 @@ export const VInput = genericComponent<new <T>() => {
               ) }
             </div>
           ) }
-          
+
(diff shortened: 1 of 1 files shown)


KaelWD:
You can already do this through the append slot, or in a separate element with v-form's slot.

zorn-v:
Yes, but I though that will be good to have such behavior as default.

zorn-v:
More context - if you use `rules` or `error-messages` you expect that messages will be visible in such way.
`hide-details` is broke this expectations. By the way I have hide-details by default because huge margins.

KaelWD:
You can use `hide-details="auto"`. 
hide-details might also be used if you're showing the validation errors externally like in an alert.

zorn-v:
> You can use hide-details="auto"
"Dancing" markup is no good.

zorn-v:
I found a solution :smile:
```vue
<script setup>
const props = defineProps({
  errorText: { type: String },
})
</script>

<template>
  <v-text-field v-bind="$attrs">
    <template #append v-if="errorText">
      <v-tooltip :text="errorText" location="bottom">
          <v-icon v-bind="props" role="button">mdi-alert-circle-outline</v-icon>
        </template>
      </v-tooltip>
    </template>
  </v-text-field>
</template>
```

And use function in `error-text` prop which do all magic to return client/server messages or false.
Moreover it is possible to use `append-inner` slot (which is looks more better in this case in my opinion), as no slot props is not required now.

PR #15311, opened 2022-06-16
feat: rename variants

KaelWD:
Match https://m3.material.io/components/all-buttons better, and remove hyphenated variant names. 

Sheets:
 - `contained` -> `elevated`
 - `contained-flat` -> `flat`
 - `contained-text` -> `tonal`

Fields:
 - `contained` -> `solo` (same as v2)

Screenshot_20220617_032703


johnleider (approved):
Everything looks good from a code and terminology standpoint.

PR #15459, opened 2022-07-18
fix(VCarousel): add keys to delimiter buttons

Tofandel:
## Motivation and Context
The missing key would create issues with SSR, the controls would in some rare random cases get duplicated
image

The vue doc does say it's required on templated v-for custom components, I don't know how much of that applies to render components but from some reading through vue's code and some debugging, I'm confident the same applies
https://vuejs.org/style-guide/rules-essential.html#use-keyed-v-for

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VCarousel/VCarousel.ts
+++ b/packages/vuetify/src/components/VCarousel/VCarousel.ts
@@ -139,6 +139,7 @@ export default VWindow.extend({
       const children = []
 
       for (let i = 0; i < length; i++) {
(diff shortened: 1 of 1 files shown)


Tofandel:
Ideally I'd use something unique in `this.items[i]`, but except the internal `_uid` of the component there doesn't seem to be anything else to identify it, but given that this is not using transitions and that if the items change the render gets invalidated, using the loop index is fine

KaelWD (approved):

PR #15503, opened 2022-07-22
fix(VListItem): allow explicit false declaration for active prop

### LINKED ISSUE #15302: [Bug Report][3.0.0-beta.3] v-list-item cannot always be made inactive via the "active" prop
### Environment
**Vuetify Version:** 3.0.0-beta.3
**Vue Version:** 3.2.37
**Browsers:** Chrome 102.0.5005.61
**OS:** Mac OS 10.15.7

### Steps to reproduce
If you have a list item and specify the "active" property you can set the active state of the list item by changing that property (as per the docs).  If, however, you subsequently make the list item active by simply clicking on it, then changing the value of the active property to false afterwards does it deactivate the list item.  The property is false, but the list item retains its active appearance (the classes defining its active appearance are not removed).

### Expected Behavior
Clicking the list item or changing the active property should always behave in the same way.

### Actual Behavior
If you make an item active by clicking it, the active property ceases to have the desired effect.

### Reproduction Link
<a href="https://codepen.io/kpturner/pen/vYdbzXY" target="_blank">https://codepen.io/kpturner/pen/vYdbzXY</a>

### Other comments
There are documentation issues too.  The default slot exposes an "isActive" property but in the Vuetify 3 docs it is defined as an "active" property:
https://next.vuetifyjs.com/en/api/v-list-item/#slots-default

johnleider:
## Description
if active === true / forced active
if active === undefined / inferred by route or group
if active === false / forced inactive

## Motivation and Context
fixes #15302

## How Has This Been Tested?
visually

## Markup:

<details>

```vue
<template>
  <div class="ma-4 pa-4">
    <v-container>
      <v-list>
        <v-list-subheader>
          The list item below can be made active by clicking it or by using the checkbox. If you simply use the checkbox the active state of the list item (color etc) will change as expected. If you make the list item active by clicking the list item itself (rather than the checkbox) then the checkbox will subsequently not deactivate the list item (it retains the active class even though the "active" property is false )
          <v-checkbox
            v-model="active"
            hide-details
          />
        </v-list-subheader>
        <v-list-item
          value="notifications"
          :active="active"
          active-class="active-class"
        >
          <template #default>
            <v-list-item-header>
              <v-list-item-title>I am a list item</v-list-item-title>
              <v-list-item-subtitle>
                <span v-if="active">I am active</span>
              </v-list-item-subtitle>
            </v-list-item-header>
          </template>
        </v-list-item>
      </v-list>

    </v-container>
  </div>
</template>

<script>
  export default {
    data () {
      return {
        active: false,
      }
    },
  }
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -62,7 +62,10 @@ export const VListItem = genericComponent<new () => {
   directives: { Ripple },
 
   props: {
-    active: Boolean,
+    active: {
+      type: Boolean,
+      default: undefined,
+    },
     activeColor: String,
     activeClass: String,
     appendAvatar: String,
@@ -94,6 +97,8 @@ export const VListItem = genericComponent<new () => {
     const { select, isSelected, isIndeterminate, isGroupActivator, root, parent } = useNestedItem(id, false)
     const list = useList()
     const isActive = computed(() => {
+      if (props.active === false) return false
+
       return props.active || link.isExactActive?.value || isSelected.value
     })
     const roundedProps = computed(() => props.rounded || props.nav)


KaelWD:
Duplicate of #15390

PR #15590, opened 2022-08-07
feat(VInfiniteScroll): initial implementation

### LINKED ISSUE #3538: [Feature Request] Infinite scrolling in lists / tables
# New Functionality
Currently there is no support for infinite scrolling, only normal pagination (by pages).
Infinite scrolling is being used in many apps / websites / supported by many frameworks and gaining popularity pretty fast.
Scroll in a list / table until you reach the bottom of the list / table and then load another bulk of items instead of splitting it into pages.

# Improvements
* No need to mess with pagination, pages, filters, etc.
* Can avoid bugs related to pages / filters.
* High popularity.
* Cleaner view.

# Bugs or Edge Cases it Helps Avoid
I think it can reduce the amount of bugs related to traditional pagination, but i don't think it will produce more bugs.

nekosaur:
closes #3538

## Description

infinite scroll implementation

## Motivation and Context

requested feature. also needed internally by date picker

## How Has This Been Tested?

playground and cypress

## Markup:

<details>

```vue
<template>
  <v-infinite-scroll
    :load="load"
    height="400"
  >
    <template v-for="(item, index) in items" :key="item">
      <div :class="['pa-2', index % 2 === 0 ? 'bg-grey-lighten-2' : '']">
        Item #{{ item }}
      </div>
    </template>
  </v-infinite-scroll>
</template>

<script>
  export default {
    data: () => ({
      items: Array.from({ length: 100 }, (k, v) => v + 1),
    }),

    methods: {
      load () {
        return new Promise(resolve => {
          setTimeout(() => {
            this.items.push(...Array.from({ length: 10 }, (k, v) => v + this.items.at(-1) + 1))

            resolve('ok')
          }, 1000)
        })
      },
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -0,0 +1,257 @@
...
+    ))
+  },
+})
+
+export const VInfiniteScroll = defineComponent({
+  name: 'VInfiniteScroll',
+
+  props: {
+    color: String,
+    load: {
```
johnleider, line 61: What if we made this optional and instead provide the user with 2 more pieces of functionality?

1. emit the load event
2. add a loading prop

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -0,0 +1,257 @@
...
+      if (val) emit('intersect', props.side)
+    })
+
+    useRender(() => (
+      <div class="v-infinite-scroll-intersect" ref={ intersectionRef }>&nbsp;</div>
+    ))
+  },
+})
+
+export const VInfiniteScroll = defineComponent({
```
johnleider, line 56: will need to be updated with the new genericComponent fn so we can add slots typing. example https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VBtn/VBtn.tsx#L84

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+type VInfiniteScrollSlots = {
+  default: [InfiniteScrollSlot]
+  loading: [InfiniteScrollSlot]
+  error: [InfiniteScrollSlot]
+  empty: [InfiniteScrollSlot]
+  'load-more': [InfiniteScrollSlot]
+}
+
+export const VInfiniteScrollIntersect = genericComponent()({
```
johnleider, line 39: @KaelWD since this doesn't have a slot should I leave it as **defineComponent**?

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
nekosaur: I'm not very keen on having multiple ways of interacting with the component. But I'm not against finding an alternative to the current solution. However just having load event and loading prop would not be enough to replace what is there now.

Both sides of the infinite list have independent statuses, which include not only `loading` but `error` and `empty` as well.

johnleider (approved):

johnleider (approved):

PR #15602, opened 2022-08-09
fix(VAlert): use v-btn for closable functionality

### LINKED ISSUE #15589: [Feature Request] [3.0.0-beta.6] VAlert should have an accessible close action
### Problem to solve
Currently, the VAlert component close action is not accessible: it is not keyboard focusable and does not have any textual alternative.

### Proposed solution
I think there are two available solutions to solve this:

- Add a `role="button"`, `tabindex="1"` and `aria-label="Close"` to the `div.v-alert__close`
- Use a `v-btn` with appropriate settings to match alert style (probably better IMO)

I can create a PR if needed, but I would like to know which solution is preferred by the core team.

johnleider:
## Motivation and Context
fixes #15589

## Markup:

<details>

```vue
<template>
  <div class="ma-4 pa-4">
    <v-alert closable>
      Lorem ipsum dolor sit amet consectetur adipisicing elit. Commodi, ratione debitis quis est labore voluptatibus! Eaque cupiditate minima, at placeat totam, magni doloremque veniam neque porro libero rerum unde voluptatem!

      <template #close="props">
        <v-btn size="large" v-bind="props" />
      </template>
    </v-alert>
  </div>
</template>

<script>
  export default {

  }
</script>

```
</details>

### DIFF
--- a/packages/vuetify/src/components/VAlert/VAlert.tsx
+++ b/packages/vuetify/src/components/VAlert/VAlert.tsx
@@ -218,7 +218,7 @@ export const VAlert = defineComponent({
               }}
             >
               <div class="v-alert__close">
-                { slots.close?.(closeProps.value) ?? <VBtn { ...closeProps.value } /> }
+                { slots.close?.({ props: closeProps.value }) ?? <VBtn { ...closeProps.value } /> }
               </div>
             </VDefaultsProvider>
           ) }
--- a/packages/vuetify/src/components/VAlert/VAlert.tsx
+++ b/packages/vuetify/src/components/VAlert/VAlert.tsx
@@ -210,7 +210,6 @@ export const VAlert = defineComponent({
               key="close"
               defaults={{
                 VBtn: {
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VAlert/VAlert.tsx
```diff
@@ -203,20 +209,16 @@ export const VAlert = defineComponent({
...
-              <div
-                class="v-alert__close"
-                onClick={ onCloseClick }
-              >
-                { slots.close
-                  ? slots.close()
-                  : (<VIcon />)
-                }
+              <div class="v-alert__close">
+                { slots.close?.(closeProps.value) ?? <VBtn { ...closeProps.value } /> }
```
KaelWD, line 221:
```suggestion
                { slots.close?.({ props: closeProps.value }) ?? <VBtn { ...closeProps.value } /> }
```

KaelWD (approved):

PR #15646, opened 2022-08-16
feat: add blueprints to package exports

johnleider:
## Motivation and Context

Allow user to import blueprints without having to use `lib`.

```javascript
// before
import { md1 } from 'vuetify/lib/blueprints'
import { md2 } from 'vuetify/lib/blueprints/md2'

// after
import { md1 } from 'vuetify/blueprints'
import { md2 } from 'vuetify/blueprints/md2'
```

### DIFF
--- a/packages/vuetify/package.json
+++ b/packages/vuetify/package.json
@@ -55,6 +55,8 @@
     "./styles": "./lib/styles/main.css",
     "./styles/*": "./lib/styles/*",
     "./framework": "./lib/framework.mjs",
+    "./blueprints": "./lib/blueprints/index.mjs",
+    "./blueprints/*": "./lib/blueprints/*/index.mjs",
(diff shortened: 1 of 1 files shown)


packages/vuetify/package.json
```diff
@@ -55,6 +55,8 @@
     "./styles": "./lib/styles/main.css",
     "./styles/*": "./lib/styles/*",
     "./framework": "./lib/framework.mjs",
+    "./blueprints": "./lib/blueprints/index.mjs",
+    "./blueprints/*": "./lib/blueprints/*/index.mjs",
```
KaelWD, line 59:
```suggestion
    "./blueprints/*": "./lib/blueprints/*.mjs",
```

KaelWD (approved):

PR #15660, opened 2022-08-21
docs: reduce sidebar ad padding

KaelWD:


packages/docs/src/components/promoted/Script.vue
```diff
@@ -4,7 +4,7 @@
 
 <script lang="ts">
   // Utilities
-  import { wait } from '@/util/helpers'
+  // import { wait } from '@/util/helpers'
```
johnleider, line 7: dev code?

johnleider (changes requested):

packages/docs/src/components/promoted/Script.vue
KaelWD: It breaks the ad if it takes more than 1000ms to load, not sure what it's actually for so I left it

johnleider (approved):

PR #15664, opened 2022-08-22
docs: remove locale from route name

johnleider:
## Motivation and Context
Remove the locale from the generated route name


packages/docs/src/components/getting-started/WireframeExamples.vue
```diff
@@ -0,0 +1,50 @@
...
+      :key="i"
+      cols="12"
+      md="6"
+      lg="4"
+    >
+      <router-link
+        :to="{
+          name: `wireframes-${wireframe}`
+        }"
+        class="text--primary text-decoration-none"
```
KaelWD, line 14: text--primary doesn't exist

packages/docs/vite.config.ts
```diff
@@ -87,31 +87,48 @@ export default defineConfig(({ command, mode }) => {
...
-        pagesDir: [
-          { dir: 'src/pages', baseRoute: '' },
+        dirs: [
+          { dir: 'src/pages', baseRoute: 'pages' },
           { dir: 'src/api', baseRoute: 'api' },
+          { dir: 'src/wireframes', baseRoute: 'wireframes' },
         ],
         extendRoute (route) {
-          if (['index', 'all'].includes(route.name)) {
+          if (['index', 'all'].includes(route.component)) {
```
KaelWD, line 96: This never matches, and probably didn't before either.

KaelWD:
Wireframes were at `examples/wireframes/:wireframe/` before

packages/docs/vite.config.ts
johnleider: double check. It didn't work with route.name, but route.component uses the actual index.md filename.

KaelWD: yeah but it's never just `'index'` or `'all'`

PR #15859, opened 2022-10-09
fix(VContainer): fill-height prop

stratdev3:
apply to a v-container, fill-height property have to set display flex and default align-items center to vertical align container (backport from v2.x)

## Types of changes

### DIFF
--- a/packages/vuetify/src/components/VGrid/VGrid.sass
+++ b/packages/vuetify/src/components/VGrid/VGrid.sass
@@ -7,6 +7,11 @@
 
   &--fluid
     max-width: 100%
+    
+  &.fill-height
+    align-items: center
+    display: flex
+    flex-wrap: wrap
 
 // Row
 //


johnleider:
@KaelWD thoughts?

johnleider:
You say this is a backport, but I'm not seeing where we had this in v2.

ferferga:
@johnleider It's not documented in the API section of ``v-container``, but I can indeed confirm you that prop worked with vuetify 2.6.0
It's also documented here: https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/pages/en/getting-started/frequently-asked-questions.md?plain=1#L160

johnleider:
I think that has to do with **fill-height** being a valid utility class in v2. The class now is `h-100`.

ferferga:
@johnleider It's still not the same, and I even tested it, since flex is not applied. There's something else changed from V2 to V3, because my `v-container`s in V2.6.0 are perfectly centered and in V3. 1.0 they aren't.

There doesn't seem to be a migration path for this. If there it's another way of that now, it should be mentioned in the upgrade guide at least.

johnleider:
Fair enough.

stratdev3:
thanks for the merge

hanzoz:
For my case, this changes cause the layout to run, please refer to the screenshot bottom

image

if I take out the style applied to .fill-height from .v-container then it works fine

image

not sure what causing this, any thoughts?

johnleider:
> For my case, this changes cause the layout to run, please refer to the screenshot bottom
> 
> image
> 
> if I take out the style applied to .fill-height from .v-container then it works fine
> 
> image
> 
> not sure what causing this, any thoughts?

Please create a new issue

PR #15866, opened 2022-10-11
docs: change global configuration order wording

ricardovanlaarhoven:
## Description
The order of global-configuration was wrong. When i read it, it seemed odd to me since you'd think that the component specific would overwrite the global section. After testing it i saw that it indeed had a logical order and that's why the update to the docs is needed.

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/pages/en/features/global-configuration.md
+++ b/packages/docs/src/pages/en/features/global-configuration.md
@@ -39,8 +39,9 @@ export default createVuetify({
 
 When creating and mounting a component, Vuetify uses the following priority in determining which default prop value to use:
 
-1. Value defined in global section of `defaults` configuration object
-2. Value defined in component specific section of `defaults` configuration object
-3. Value defined in the component definition
+1. Value set as prop value to the component itself
+2. Value defined in component specific section of defaults configuration object
+3. Value defined in global section of defaults configuration object
+4. Value defined in the component definition by Vuetify
 
 <backmatter />
--- a/packages/docs/src/pages/en/features/global-configuration.md
+++ b/packages/docs/src/pages/en/features/global-configuration.md
@@ -37,11 +37,11 @@ export default createVuetify({
 
 ## Priority
 
-When creating and mounting a component, Vuetify uses the following priority in determining which default prop value to use:
+When creating and mounting a component, Vuetify uses the following priority in determining which prop value to use:
 
 1. Value set as prop value to the component itself
 2. Value defined in component specific section of defaults configuration object
 3. Value defined in global section of defaults configuration object
-4. Value defined in the component definition by Vuetify
+4. Value defined in the prop definition of the Vueitfy component itself.
 
 <backmatter />
--- a/packages/docs/src/pages/en/features/global-configuration.md
+++ b/packages/docs/src/pages/en/features/global-configuration.md
@@ -43,5 +43,4 @@ When creating and mounting a component, Vuetify uses the following priority in d
 2. Value defined in component specific section of `defaults` configuration object
 3. Value defined in the component definition
 
-
 <backmatter />
--- a/packages/docs/src/pages/en/features/global-configuration.md
+++ b/packages/docs/src/pages/en/features/global-configuration.md
@@ -39,8 +39,9 @@ export default createVuetify({
 
 When creating and mounting a component, Vuetify uses the following priority in determining which default prop value to use:
 
-1. Value defined in component specific section of `defaults` configuration object
-2. Value defined in global section of `defaults` configuration object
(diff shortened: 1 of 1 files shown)


KaelWD:
This was already correct. Buttons will be green with the following:

```
defaults: {
  global: { color: 'red' },
  VBtn: { color: 'green' }
}
```

ricardovanlaarhoven:
> This was already correct. Buttons will be green with the following:
> 
> ```
> defaults: {
>   global: { color: 'red' },
>   VBtn: { color: 'green' }
> }
> ```


The old doc order says:

Value defined in component specific section of defaults configuration object
Value defined in global section of defaults configuration object
Value defined in the component definition

Which i read as:
it first uses the components default configuration
Which is overwritten by the global configuration
Which is overwritten by the component definition

But it first checks the global configuration and that is overwritten by the component's default configuration. Not the other way around.

Or am I missing something?

KaelWD:
Yeah it's more like  `prop ?? defaults.component ?? defaults.global ?? component.props.default`

ricardovanlaarhoven:
> Yeah it's more like `prop ?? defaults.component ?? defaults.global ?? component.props.default`

But what's described in the documentation is
`prop ?? default.global ?? defaults.component`
And i changed it to
`prop ?? defaults.component ?? default.global`

KaelWD:
You might be reading it backwards? It isn't 1 overrides 2 overrides 3, it's 1 otherwise 2 otherwise 3
If you think you can clarify the wording a bit go ahead, but the order is already correct

jacekkarczmarczyk:
IMHO current order is ok - first check for component's specific defaults, if not provided the globals, if not provided the default value from prop definition

BTW - shouldn't it include also default values provided by `VDefaultsProvider`?

KaelWD:
Yeah that's kinda complicated because VDefaultsProvider merges then replaces the global configuration, so it'd be
`prop ?? defaultsProvider.component ?? defaultsProvider.global ?? globalDefaults.component ?? globalDefaults.global ?? component.props.default`

ricardovanlaarhoven:
Ah now I know why I was confused. I thought you meant the creation of the component `<VBtn color="green"/>` with " component definition". But it's the `default` inside a `prop` inside the `VBtn`

packages/docs/src/pages/en/features/global-configuration.md
```diff
@@ -39,8 +39,8 @@ export default createVuetify({
 
 When creating and mounting a component, Vuetify uses the following priority in determining which default prop value to use:
 
-1. Value defined in component specific section of `defaults` configuration object
-2. Value defined in global section of `defaults` configuration object
+1. Value defined in global section of `defaults` configuration object
+2. Value defined in component specific section of `defaults` configuration object
 3. Value defined in the component definition
```
ricardovanlaarhoven, line 44:
```suggestion
1. Value set as prop value to the component itself
2. Value defined in component specific section of defaults configuration object
3. Value defined in global section of defaults configuration object
4. Value defined in the component definition by Vuetify
```

jacekkarczmarczyk:
> which default prop value to use:

So either "value set as a prop ..." shouldn't be on the list or the bit above should be changed

packages/docs/src/pages/en/features/global-configuration.md
```diff
@@ -39,8 +39,9 @@ export default createVuetify({
 
 When creating and mounting a component, Vuetify uses the following priority in determining which default prop value to use:
```
ricardovanlaarhoven, line 40:
```suggestion
When creating and mounting a component, Vuetify uses the following priority in determining which prop value to use:
```

packages/docs/src/pages/en/features/global-configuration.md
```diff
@@ -39,8 +39,9 @@ export default createVuetify({
...
 
-1. Value defined in component specific section of `defaults` configuration object
-2. Value defined in global section of `defaults` configuration object
-3. Value defined in the component definition
+1. Value set as prop value to the component itself
+2. Value defined in component specific section of defaults configuration object
+3. Value defined in global section of defaults configuration object
+4. Value defined in the component definition by Vuetify
```
ricardovanlaarhoven, line 45:
```suggestion
1. Value set as prop value to the component itself
2. Value defined in component specific section of defaults configuration object
3. Value defined in global section of defaults configuration object
4. Value defined in the prop definition of the Vueitfy component itself.
```

ricardovanlaarhoven:
If you guys think it's clear enough you can close this PR. Me and some coworkers just didn't get it and i hope this prevents some questions about this.

PR #15861, opened 2022-10-10
fix(VCombobox): close first chip correctly

7nohe:
## Description
fixes #15824 

## Motivation and Context
`VChip` uses `useProxiedModel`and manages `modelValue` internally, so when the first chip is closed, the remaining chips do not become Active properly and all disappear.
To solve this, specify `onUpdate:modelValue` for `VChip` in `VCombobox` to manage `modelValue` on `VCombobox` side.

## How Has This Been Tested?
visually (on Playground.vue)

## Markup:
<details>

```vue
<template>
  <v-container>
    <v-row>
      <v-col>
        <v-combobox
            v-model="select"
            :items="items"
            label="I use chips"
            multiple
            chips
            closable-chips
          ></v-combobox>
      </v-col>
    </v-row>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
      select: ['Vuetify', 'Programming'],
      items: [
        'Programming',
        'Design',
        'Vue',
        'Vuetify',
      ],
    }),
  }
</script>

```
</details>

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:
- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [x] My code follows the code style of this project.
- [x] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #15871, opened 2022-10-11
fix(VAutocomplete): wrong behavior on chip close

mlesin:
## Description
This change uses `item.value` instead of `index` for the Vue item key for items in multiple selection.

## Motivation and Context
Because of the use of indexes as a Vue key, if you use VAutocomplete with "chips, closable-chips, and multiple" properties set to true, after clicking the close button on a VChip when multiple chips were selected, the chip next to the deleted one would also disappear.

## How Has This Been Tested?
Visually

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        v-model="assignedUsers"
        :items="existingUsers"
        item-title="name"
        item-value="id"
        return-object
        clearable
        chips
        closable-chips
        multiple
      />
    </v-container>
  </v-app>
</template>

<script lang="ts">
  import { ref } from 'vue'

  interface IUser {
    id: string,
    name: string,
  }

  export default {
    name: 'Playground',
    setup () {
      const assignedUsers = ref<IUser[]>([{ id: '1', name: 'a' }, { id: '2', name: 'b' }, { id: '3', name: 'c' }])
      const existingUsers: IUser[] = [{ id: '1', name: 'a' }, { id: '2', name: 'b' }, { id: '3', name: 'c' }, { id: '4', name: 'd' }]

      return {
        assignedUsers, existingUsers,
      }
    },
  }
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -298,7 +298,7 @@ export const VAutocomplete = genericComponent<new <
                   }
 
                   return (
-                    <div key={ index } class="v-autocomplete__selection">
+                    <div key={ item.value } class="v-autocomplete__selection">
                       { hasChips ? (
                         <VDefaultsProvider
                           defaults={{


KaelWD:
Thanks for your contribution @mlesin, but both this and #15861 were needed so I merged them in 380044ca8

PR #15903, opened 2022-10-16
docs(api): specify types to reference by link

KaelWD:
before:
Screenshot_20221018_000157
Screenshot_20221018_000344

after:
Screenshot_20221018_000317
Screenshot_20221018_000358

### DIFF
--- a/packages/api-generator/src/index.ts
+++ b/packages/api-generator/src/index.ts
@@ -10,6 +10,14 @@ import mkdirp from 'mkdirp'
 import { createVeturApi } from './vetur'
 import rimraf from 'rimraf'
 import { createWebTypesApi } from './web-types'
+import inspector from 'inspector'
+import yargs from 'yargs'
+
+const argv = yargs(process.argv.slice(2))
+  .option('components', {
+    type: 'array',
+  })
+  .argv
 
 const run = async () => {
   const locales = ['en']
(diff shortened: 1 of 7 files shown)


nekosaur (approved):

packages/api-generator/src/index.ts
```diff
@@ -18,7 +26,7 @@ const run = async () => {
   const pool = new Piscina({
     filename: './src/worker.js',
     niceIncrement: 10,
-    maxThreads: Math.max(1, Math.floor(Math.min(os.cpus().length / 2, os.freemem() / (1.1 * 1024 ** 3)))),
+    maxThreads: inspector.url() ? 1 : Math.max(1, Math.floor(Math.min(os.cpus().length / 2, os.freemem() / (1.1 * 1024 ** 3)))),
```
johnleider, line 29: is there any value in a comment block that explains what this essentially equates to?

packages/api-generator/src/types.ts
```diff
@@ -290,6 +301,13 @@ function generateDefinition (node: Node<ts.Node>, recursed: string[], project: P
   const tc = project.getTypeChecker()
   type = type ?? node.getType()
 
+  if (type.getAliasSymbol()?.getName() === 'NonNullable') {
```
johnleider, line 304: `getAliasSymbol` is called at least 3 times in this file. Could probably be simplified.

johnleider (changes requested):

packages/api-generator/src/index.ts
KaelWD: What, you can't read that?

packages/api-generator/src/types.ts
KaelWD: Parsing is inherently complicated.

The second one could probably change to `symbol?.getName()`, although idk how that behaves if `type.getSymbol()` is hit.

Issue #10115, opened 2020-01-04
[Feature Request] Non-interactive label chips


mitar:
### Problem to solve
This is a followup to https://github.com/vuetifyjs/vuetify/issues/3817. Now, when I have a chip like:

```
<v-chip label>Foobar</v-chip>
```

If I hover with the mouse over it, it still changes the gray-level of the background. This makes it look like it is actionable. Still confusing.

### Proposed solution
If I do:

```
<v-chip label :input-value="true">Foobar</v-chip>
```

It makes it not change on hover the color. I think this should be the default for labels at least.

jacekkarczmarczyk:
It sounds more like https://github.com/vuetifyjs/vuetify/issues/4656, imho `readonly` prop would a good choice here (i wouldn't make it default though)

mitar:
#4656 was closed as a duplicate of #3817. And I think this one is a followup to #3817. :-)

johnleider:
Thank you for the Feature Request and interest in improving Vuetify. After careful consideration the team has decided that this is not functionality that we are looking to implement at this time.

If you have any additional questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

ikovac:
@mitar This solution:
`<v-chip label :input-value="true">Foobar</v-chip>`
actually changes chip color by applying hover state color on a chip. Color won't be changed on hover but the initial color will be changed. That may result in unexpected behavior and shouldn't be used by the default.

johnleider:
fixed by #13192

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

PR #15909, opened 2022-10-16
fix(VChip): highlight overlay on hover

xignp:
## Description

fixes vuetifyjs#15853
v-chip's backgroundcolor will be hightlight when mouse hover
## How Has This Been Tested?

Playground.vue
## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-chip>non-clickable</v-chip>
      <v-chip @click="chipClick">clickable</v-chip>
    </v-container>
  </v-app>
</template>

<script>
export default {
  name: "Playground",
  setup() {
    const chipClick = () => {
      console.log("clicked")
    }
    return {
      chipClick
    };
  },
};
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VChip/VChip.tsx
+++ b/packages/vuetify/src/components/VChip/VChip.tsx
@@ -146,7 +146,7 @@ export const VChip = defineComponent({
           v-ripple={ [isClickable && props.ripple, null] }
           onClick={ isClickable && onClickFunc }
         >
-          { genOverlays(isClickable, 'v-chip') }
+          { genOverlays(true, 'v-chip') }
 
           { hasFilter && (
             <VDefaultsProvider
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VChip/VChip.tsx
```diff
@@ -146,7 +146,7 @@ export const VChip = defineComponent({
           v-ripple={ [isClickable && props.ripple, null] }
           onClick={ isClickable && onClickFunc }
         >
-          { genOverlays(isClickable, 'v-chip') }
+          { genOverlays(true, 'v-chip') }
```
KaelWD, line 149: #10115

Should probably use the same logic as VListItem

xignp: Thank you for feedback.
I thought it would be highlighted even if it was unclickable.
I fixed to highlight only when clickable.

Issue #15784, opened 2022-09-14
[Bug Report][3.0.0-beta.10] VTooltip overlay interfering with VMenu Overlay Clickoutside behaviour


shaydoc:
### Environment
**Vuetify Version:** 3.0.0-beta.10
**Vue Version:** 3.2.39
**Browsers:** Chrome 105.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Click on the VMenus on the Navigation Drawers, when the tooltip is active it prevents the VMenu from closing

### Expected Behavior
VMenu should close when clicking on a new VMenu Element

### Actual Behavior
2 VMenus appear stacked on top of each other

### Reproduction Link
<a href="https://codepen.io/shaydoc-the-reactor/pen/ExLNpxa" target="_blank">https://codepen.io/shaydoc-the-reactor/pen/ExLNpxa</a>

KaelWD:
Overlays only close on click outside if they are the most recently opened one. I'm not really sure how to solve this because the current behaviour is useful for situations like a menu inside a dialog, where a click outside would otherwise close both instead of just the menu.

shaydoc:
I think we need to distinguish tooltips from other overlays perhaps, as they typically only show on hover and maybe are an exclusion to this rule ?

gkubisa:
@KaelWD Perhaps we could simply ignore clicks on overlays which work based on `mouseenter` and `mouseleave`? It should solve the issue, as tooltips are based on `mouseenter` and `mouseleave`, and they don't use `click` events.

shaydoc:
do you have any thoughts on this issue @johnleider @KaelWD ?

PR #15934, opened 2022-10-19
fix(VTooltip): tooltip overlay interfering with VMenu

LucasLeandro1204:
## Description
fixes https://github.com/vuetifyjs/vuetify/issues/15784
v-tooltip overlay should not be on global stack by default

## How Has This Been Tested?
Playground.vue

## Markup:
<details>

```vue<template>
  <v-app>
    <v-container>
      <div class="text-center">
        <v-menu>
          <template v-slot:activator="{ props: menu }">
            <v-tooltip location="top">
              <template v-slot:activator="{ props: tooltip }">
                <v-btn color="primary" v-bind="mergeProps(menu, tooltip)">
                  Left Dropdown
                </v-btn>
              </template>
              <span>Left Tooltip</span>
            </v-tooltip>
          </template>
          <v-list>
            <v-list-item v-for="(item, index) in items" :key="index">
            </v-list-item>
          </v-list>
        </v-menu>

        <span class="mr-4"></span>

        <v-menu>
          <template v-slot:activator="{ props: menu }">
            <v-tooltip location="top">
              <template v-slot:activator="{ props: tooltip }">
                <v-btn color="primary" v-bind="mergeProps(menu, tooltip)">
                  Right Dropdown
                </v-btn>
              </template>
              <span>Right Tooltip</span>
            </v-tooltip>
          </template>
          <v-list>
            <v-list-item v-for="(item, index) in items" :key="index">
            </v-list-item>
          </v-list>
        </v-menu>
      </div>
    </v-container>
  </v-app>
</template>

<script>
import { mergeProps } from "vue";

export default {
  name: "Playground",
  setup() {
    const items = [
      { title: "Click Me1" },
      { title: "Click Me2" },
      { title: "Click Me3" },
      { title: "Click Me4" },
    ];
    return {
      mergeProps,
      items,
    };
  },
};
</script>

```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/composables/stack.ts
```diff
@@ -14,8 +14,13 @@ interface StackProvide {
 
 const globalStack = reactive<[uid: number, zIndex: number][]>([])
 
-export function useStack (isActive: Readonly<Ref<boolean>>, zIndex: Readonly<Ref<string | number>>) {
+export function useStack (
+  isActive: Readonly<Ref<boolean>>,
+  zIndex: Readonly<Ref<string | number>>,
+  disableGlobalStack: Readonly<Ref<boolean>>
```
gkubisa, line 20: In order to guarantee correct behavior I think this should be just a boolean, not a ref containing a boolean.

packages/vuetify/src/composables/stack.ts
```diff
@@ -27,22 +32,33 @@ export function useStack (isActive: Readonly<Ref<boolean>>, zIndex: Readonly<Ref
...
-    const _isTop = globalStack.at(-1)?.[0] === vm.uid
-    setTimeout(() => globalTop.value = _isTop)
+    if (createStackEntry.value) {
+      const _isTop = globalStack.at(-1)?.[0] === vm.uid
+      setTimeout(() => globalTop.value = _isTop)
+
+      return
+    }
+
+    globalTop.value = true
```
gkubisa, line 61: If we turn `disableGlobalStack` into a `boolean` (not a ref containing a `boolean`), this line will not be needed because `true` is already the default value of `globalTop`.

gkubisa (commented):
Looks good to me. I suggested just a small change there.

packages/vuetify/src/composables/stack.ts
gkubisa: Just thinking about it... if we disable global stack for a particular overlay, should `globalTop` and `localTop` be `true`?

packages/vuetify/src/composables/stack.ts
johnleider: Could you elaborate on why behavior would be incorrect?

gkubisa: Well, it's an edge case but with the current implementation, if someone changed the `disableGlobalStack` prop on `VOverlay` from `false` to `true` during the lifetime of the component, then there is risk that the clean-up code () would not run - an orphan entry would remain on the `globalStack`. By "clean-up" code I mean only this bit:

```js
const idx = globalStack.findIndex(v => v[0] === vm.uid)
globalStack.splice(idx, 1)
```

LucasLeandro1204:
@johnleider I got the merge conflicts fixed 😄

KaelWD:
I have some ideas about how to do this in a more flexible way so I've made the prop private for now so we can remove it later if something better comes up.

KaelWD (approved):

PR #15958, opened 2022-10-24
fix(VSelectionControlGroup): provided props & defaults target

### LINKED ISSUE #15896: [Bug Report][3.0.0-beta.13] `valuecomparator` shows as an attribute on `v-radio-group`
### Environment
**Vuetify Version:** 3.0.0-beta.13
**Vue Version:** 3.2.41
**Browsers:** Chrome 106.0.0.0
**OS:** Windows 10

### Steps to reproduce
```html
    <v-radio-group>
      <v-radio value="a" />
    </v-radio-group>
```

### Expected Behavior
- no `valuecomparator` attribute
- should be possible to use value comparator with radio group

### Actual Behavior
- unexpected `valuecomparator` attribute
- not possible to use value comparator with radio group

image

### Reproduction Link
<a href="https://codepen.io/jkarczm/pen/wvjOpgV?editors=1010" target="_blank">https://codepen.io/jkarczm/pen/wvjOpgV?editors=1010</a>

## Additional comments

- `VRadioGroup` passes props (including `valueComparator`) to `VSelectionControlGroup`, but `VSelectionControlGroup` doesn't have this prop defined. 
- `VSelectionControl` takes its value from props, so it would work when you pass value comparator to `VRadio` directly
- `valueComparator` could be maybe added to `VSelectionGroupContext` and `VSelectionControl` could take the value either from the injection or from props, not sure if this is the best solution

johnleider:
## Motivation and Context

fixes #15896

## Markup:
<details>

```vue
<template>
  <div class="ma-4 pa-4">
    <v-radio-group v-model="model">
      <v-radio label="A" value="a" />
      <v-radio label="B" value="b" />
    </v-radio-group>
  </div>
</template>

<script setup>
  import { ref } from 'vue'
  const model = ref('a')
</script>

```
</details>

### DIFF
--- a/packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
+++ b/packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
@@ -3,47 +3,65 @@ import './VSelectionControlGroup.sass'
 
 // Composables
 import { IconValue } from '@/composables/icons'
+import { makeDensityProps } from '@/composables/density'
+import { makeThemeProps } from '@/composables/theme'
 import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utilities
 import { computed, provide, toRef } from 'vue'
-import { defineComponent, getUid, useRender } from '@/util'
+import { deepEqual, defineComponent, getUid, propsFactory, useRender } from '@/util'
 
 // Types
 import type { InjectionKey, PropType, Ref } from 'vue'
+import { provideDefaults } from '@/composables/defaults'
 
 export interface VSelectionGroupContext {
-  disabled: Ref<boolean>
-  inline: Ref<boolean>
-  name: Ref<string | undefined>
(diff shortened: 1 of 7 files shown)


KaelWD (approved):

packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
```diff
@@ -3,47 +3,65 @@ import './VSelectionControlGroup.sass'
...
-    trueIcon: IconValue,
-    multiple: {
-      type: Boolean as PropType<boolean | null>,
-      default: null,
-    },
-    readonly: Boolean,
-    type: String,
-    modelValue: null,
-  },
+  props: makeSelectionControlProps(),
```
KaelWD, line 64: Some of these don't make sense for the group root, and aren't being used anyway (value, trueValue, falseValue, label)

PR #15970, opened 2022-10-25
docs: removed unrelevant v-system-bar from examples

jacekkarczmarczyk:
## Description
removes unrelevant usages of `v-system-bar`


MajesticPotatoe (approved):
👍

PR #16017, opened 2022-11-02
docs: generate missing descriptions in locale files

jacekkarczmarczyk:
adds `foo: "MISSING DESCRIPTION (edit in github)"` to locale files to make finding items without descriptions easier

image

```
cd packages/api-generator
yarn build --missing-descriptions
```


nekosaur:
Does this take into account shared props through composables?

jacekkarczmarczyk:
yes

packages/api-generator/src/template.d.ts
```diff
@@ -1,5 +1,5 @@
 import type { AllowedComponentProps, ComponentPublicInstance, FunctionalComponent, RenderFunction, VNodeChild, VNodeProps } from 'vue'
-import type { __component__ } from '@/__name__'
+import type { __component__ } from '@/components'
```
jacekkarczmarczyk, line 2: @KaelWD can you double check your change? I've reverted it here as with this change it didn't generate anything for me, ex. `api-generator/dist/attributes.json` was empty (so nothing related to this PR)

packages/api-generator/src/index.ts
```diff
@@ -63,6 +73,67 @@ const run = async () => {
...
+          translations[filename] = data
+        } catch (e) {
+          translations[filename] = {}
+        }
+      }
+
+      return translations[filename]
+    }
+
+    for (const index in componentData) {
```
nekosaur, line 102: This is just personal preference so take it as a suggestion only (although I do wonder if we use in anywhere else), but I think I prefer a regular for loop or .forEach instead of for .. in for arrays. I think it's an unnecessary possible confusion of arrays and objects.

packages/api-generator/src/index.ts
```diff
@@ -63,6 +73,67 @@ const run = async () => {
...
+            console.warn(`Missing source for ${component.kebabName} ${type}: ${name}`)
+          }
+
+          const filename = type === 'props'
+            ? kebabCase(component[type][name].source ?? componentData[index].componentName)
+            : component.kebabName
+
+          for (const locale of locales) {
+            const sourceData = await readData(`./src/locale/${locale}/${filename}.json`)
+            const githubUrl = `https://github.com/vuetifyjs/vuetify/tree/next/packages/api-generator/src/locale/${locale}/${filename}.json`
```
nekosaur, line 117: Could we possibly do this dynamically so it uses branch that it's run on? Maybe overkill.

packages/api-generator/src/index.ts
```diff
@@ -63,6 +73,67 @@ const run = async () => {
...
+
+    for (const filename in translations) {
+      try {
+        await fs.writeFile(filename, JSON.stringify(translations[filename], null, 2) + '\n')
+      } catch (e: unknown) {
+        console.error(filename, e)
+      }
+    }
+
+    process.exit()
```
nekosaur, line 134: ?

jacekkarczmarczyk: this only updates locales, I didn't want it to change anything in git on build

packages/api-generator/src/index.ts
```diff
@@ -40,10 +49,11 @@ const run = async () => {
...
   for (const component in components) {
-    await fs.writeFile(`./src/tmp/${component}.d.ts`,
-      template.replaceAll('__component__', component)
-        .replaceAll('__name__', componentsInfo[component].from.replace('.mjs', '.js'))
-    )
+    await fs.writeFile(`./src/tmp/${component}.d.ts`, template.replaceAll('__component__', component))
+    // await fs.writeFile(`./src/tmp/${component}.d.ts`,
+    //   template.replaceAll('__component__', component)
+    //     .replaceAll('__name__', componentsInfo[component].from.replace('.mjs', '.js'))
+    // )
```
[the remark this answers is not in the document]
jacekkarczmarczyk: because https://github.com/vuetifyjs/vuetify/pull/16017#discussion_r1019616174

packages/api-generator/src/template.d.ts
KaelWD: Looks fine to me on `next`

johnleider (approved):

Issue #16008, opened 2022-11-01
[Bug Report][2.6.12] Combobox 'allow-overflow' is not working


MayaElf:
### Environment
**Vuetify Version:** 2.6.12
**Vue Version:** 2.6.11
**Browsers:** Chrome 106.0.0.0
**OS:** Windows 10

### Steps to reproduce
1. open Combobox
2. put 'allow-overflow' = false
3. and then for another combobox 'allow-overflow' = true

### Expected Behavior
Behaviour is different for 'allow-overflow' = false vs =true

### Actual Behavior
Behaviour is the same for 'allow-overflow' = false vs =true

### Reproduction Link
[https://codepen.io/MaiiaAfanaseva/pen/ZERQJXO?editors=1010](https://codepen.io/MaiiaAfanaseva/pen/ZERQJXO?editors=1010)

MayaElf:
@Z3rio In Vuetify you have implementation of 'allow-overflow' for v-menu. I added it to examples in codepen https://codepen.io/MaiiaAfanaseva/pen/ZERQJXO?editors=1010 
I believe the implementation for v-combobox should be the same as in v-menu.

KaelWD:
This was missed in f5a3f8296906fc63c87a60ca63c4ed9397bbe37c, and looks like it wasn't hooked up correctly even then. Since v1.2.0 the correct  way to pass additional props to v-menu is through the `menu-props` prop.

Z3rio:
Okay, I will take a look at this.

MayaElf:
@KaelWD Sorry, my issue was about v-combobox. So in api-documentation for v-combobox 'allow-overflow' will be deleted?

KaelWD:
Yes, it extends VAutocomplete.

PR #16021, opened 2022-11-02
fix(VComboBox): add "allow-overflow" prop

Z3rio:
## Description
Before this change, adding "allow-overflow" as a prop to v-combobox would do nothing.
But with this change, setting allow-overflow to true will allow overflow (scrolling), this is also the default if this prop isnt passed through.
Upon setting allow overflow to false, scrolling is turned off.

**Resolves issue: #16008** 

## Motivation and Context
This would be needed because alot of people would like to display the combo list in its entirety instead of requiring the user to scroll through the list to see everything.

## How Has This Been Tested?
I have tested this with both the prop set to false, true and also without the prop.
I also ran the tests included in vuetify.

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-generator/src/locale/en/generic.json
+++ b/packages/api-generator/src/locale/en/generic.json
@@ -11,7 +11,8 @@
     "tag": "Specify a custom tag used on the root element.",
     "textColor": "Applies a specified color to the control text",
     "trueValue": "Sets value for truthy state",
-    "width": "Sets the width for the component."
+    "width": "Sets the width for the component.",
+    "allow-overflow": "Turns on or off scrolling for the items list"
   },
   "slots": {
     "activator": "When used, will activate the component when clicked (or hover for specific components). This manually stops the event propagation. Without this slot, if you open the component through its model, you will need to manually stop the event propagation",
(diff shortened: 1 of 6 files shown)


KaelWD:
f16c2c506

VMenu already has allowOverflow throu the menuable mixin, and it can be passed through menuProps from the select components. It was supposed to be removed as a separate VAutocomplete prop in v2.0.0

MayaElf:
@KaelWD @Z3rio  Sorry, my issue was about v-combobox. So in api-documentation for v-combobox 'allow-overflow' will be deleted?

Z3rio:
> @KaelWD @Z3rio  Sorry, my issue was about v-combobox. So in api-documentation for v-combobox 'allow-overflow' will be deleted?

What do you mean?

Z3rio:
> f16c2c506
> 
> VMenu already has allowOverflow throu the menuable mixin, and it can be passed through menuProps from the select components. It was supposed to be removed as a separate VAutocomplete prop in v2.0.0

Ah okay, thanks for the explanation

Z3rio:
@KaelWD Upon trying to put allow-overflow into "menu-props" it seems to do nothing at all?

PR #16038, opened 2022-11-06
fix(VListItem): infer link if click event is bound

### LINKED ISSUE #16014: [Bug Report][3.0.0] No `v-list-item--link` if there is only `@click` on `VListItem`
### Environment
**Vuetify Version:** 3.0.0
**Last working version:** 3.0.0-beta.15
**Vue Version:** 3.2.41
**OS:** Linux x86_64

### Steps to reproduce
Add `@click` on `VListItem`

### Expected Behavior
There is `v-list-item--link` class to make cursor be `pointer`

### Actual Behavior
No `v-list-item--link` on `VListItem`

### Reproduction Link
[https://codepen.io/0140454/pen/mdxePOw?editors=1010](https://codepen.io/0140454/pen/mdxePOw?editors=1010)

NoraH1to:
## Description

fixes #16014 

## Markup:

<details>

```vue
<template>
  <v-container>
    <v-menu>
      <template #activator="{props}">
        <v-btn v-bind="props">
          Dropdown Menu
        </v-btn>
      </template>
      <v-list>
        <v-list-item title="cursor = pointer" link="foo" />
        <v-list-item title="cursor != pointer" @click="foo" />
      </v-list>
    </v-menu>
  </v-container>
</template>

<script>
export default {
  methods: {
    foo() {
      console.log("foo");
    }
  }
};
</script>

```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -30,7 +30,7 @@ import { computed, watch } from 'vue'
 import { genericComponent, useRender } from '@/util'
 
 // Types
-import type { SlotsToProps } from '@/util'
+import type { EventProp, SlotsToProps } from '@/util'
 import type { PropType } from 'vue'
 
 type ListItemSlot = {
@@ -83,6 +83,10 @@ export const VListItem = genericComponent<new () => {
     title: [String, Number, Boolean],
     value: null,
 
+    // declare event as prop to check it has exist
+    onClick: Function as EventProp,
+    onClickOnce: Function as EventProp,
+
     ...makeBorderProps(),
     ...makeDensityProps(),
     ...makeDimensionProps(),
@@ -92,6 +96,10 @@ export const VListItem = genericComponent<new () => {
     ...makeTagProps(),
     ...makeThemeProps(),
     ...makeVariantProps({ variant: 'text' } as const),
+  },
+
+  emits: {
+    click: (e: Event) => true,
   },
 
   setup (props, { attrs, slots, emit }) {
@@ -143,6 +151,8 @@ export const VListItem = genericComponent<new () => {
     }))
 
     function onClick (e: MouseEvent) {
+      emit('click', e)
+
       if (isGroupActivator || !isClickable.value) return
 
       link.navigate?.(e)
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/components/VList/VListItem.tsx
```diff
@@ -147,8 +143,6 @@ export const VListItem = genericComponent<new () => {
     }))
 
     function onClick (e: MouseEvent) {
-      emit('click', e)
```
KaelWD, line 150: It needs to emit click on space or enter keydown

KaelWD (changes requested):

packages/vuetify/src/components/VList/VListItem.tsx
NoraH1to: if add this, should add `emits: ['click']` too, or the event will trigger twice

and if add `emits: ['click']`, `useLink`  will not work well because **Events declared in emits will not exist in attrs**:

```typescript
export function useLink (props: LinkProps, attrs: SetupContext['attrs']): UseLink {
  ...
  const isClickable = computed(() => {
    // Events declared in emits will not exist in attrs
    return isLink?.value || !!(attrs.onClick || attrs.onClickOnce)
  })
  ...
}
```

so if we want to check a bound listener exists,  it needs to be declared in the component's props

```typescript
...
props: ['onClick'],
emits: ['onClick'],
...
```

and need to modify `isClickable` in `useLink` like

```typescript
isLink?.value || !!(attrs.onClick || attrs.onClickOnce || props.onClick || props.onClickOnce)
```

KaelWD: Yeah that's probably what I'd go with, we've done that in a couple of other places already. There's a `callEvent` function and `EventProp` interface you can use in `@/util`. I don't know if `@click.once` is passed as a separate prop, you'd have to check that.

NoraH1to: I notice that EventProp could be an array, dose it mean the listener could be like this?

```vue
<VListItem @click="[foo, bar]" />
```

If it is ,I dont think thats a good way

It will make `props.onClick` be a func like `($event) => [foo, bar]` instead of an array, if the user also passes a similar func, it's hard to tell if the func in the returned array should be executed

If not like this, i had already fixed the issue and could commit right now

KaelWD: Yes, that's how events work in vue: 

```js
mergeProps({ onClick: () => {} }, { onClick: () => {} }) === { onClick: [() => {}, () => {}] }
```

If the prop type is just `Function` you'll get runtime errors.

PR #15904, opened 2022-10-16
fix: properly passthrough non-bubbling events to input elements

KaelWD:
Bubbling events should bind to the root, and non-bubbling events to both the root and input element. Currently all events are passed to the input element which makes it impossible to intercept clicks on other parts of the component. 

fixes #16812

```vue
<template>
  <v-app>
    <v-container>
      <v-btn>
        Menu

        <v-menu activator="parent">
          <v-list>
            <v-list-item link>
              <v-list-item-title>Dialog</v-list-item-title>

              <v-dialog activator="parent" width="auto">
                <v-sheet>
                  Dialog
                </v-sheet>
              </v-dialog>
            </v-list-item>
          </v-list>
        </v-menu>
      </v-btn>

      <v-switch
        v-model="sw"
        label="Nightly"
        @click="onClick"
        @mousedown="onMousedown"
        @focus="onFocus"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { computed, ref, watch } from 'vue'

  const sw = ref(false)

  function onClick (e) {
    console.log('click', e.target)
  }

  function onMousedown (e) {
    console.log('mousedown', e)
  }

  function onFocus (e) {
    console.log('focus', e)
  }
</script>
```

https://discord.com/channels/340160225338195969/1068832136863432724:
```vue
<template>
  <v-app>
    <v-main>
      <v-container class="bg-grey">
        <v-card color="" width="400" @click="onClick">
          <v-card-text>
            <v-select v-ripple.stop label="select" :items="['foo']" @click.stop />
            <v-autocomplete v-ripple.stop label="autocomplete" :items="['foo']" @click.stop />
          </v-card-text>
        </v-card>
      </v-container>
    </v-main>
  </v-app>
</template>

<script setup>
  const onClick = () => {
    window.alert('hi')
  }
</script>
```

https://discord.com/channels/340160225338195969/1045346180802039888:
```vue
<template>
  <div>
    <v-form id="inputForm" @submit.prevent="submitGuess()">
      Animation playing: {{ animateInputShake }}
      <v-responsive
        class="mx-auto"
        max-width="200"
      >
        <v-row>
          <v-col>
            <v-text-field
              id="textField"
              ref="guessInputField"
              variant="outlined"
              type="text"
              max-width="1"
              :class="{'guess-input-shake': animateInputShake}"
              @animationend="animateInputShake = false"
            />
          </v-col>
        </v-row>
      </v-responsive>
    </v-form>
[...]

PR #16079, opened 2022-11-12
fix(VSlider): add missing prepend and append icon click events

### LINKED ISSUE #16071: [Bug Report][3.0.1] `v-slider` append and prepend event is not working
### Environment
**Vuetify Version:** 3.0.1
**Vue Version:** 3.2.44
**OS:** Mac OS 10.15.7

### Steps to reproduce
Check codepen

### Expected Behavior
Append and prepend events should be emitted when clicking on `prepend-icon` and `append-icon`

### Actual Behavior
Not Working

### Reproduction Link
[https://codepen.io/Alokdev/pen/ZERePMB?editors=101](https://codepen.io/Alokdev/pen/ZERePMB?editors=101)

vexleet:
## Motivation and Context
There were props for adding click events for the append and prepend icons in the v slider but they weren't used in the component which led to not being able to have events for the icons, which would fix #16071

## How Has This Been Tested?
I created a unit test using cypress which adds a slider with 2 icons and putting events on them, after that I used cypress spies to check if the button events worked when both icons were clicked.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


nekosaur:
@KaelWD we're filtering out event props here https://github.com/vuetifyjs/vuetify/blob/f627eb1726f1e195b11a984bedc79772a682b2e8/packages/vuetify/src/components/VInput/VInput.tsx#L191 which is the root cause of this issue. What's the reason we're doing this?

KaelWD:
I think because VTextField does some other shit first then waits a tick before re-emitting clear:
https://github.com/vuetifyjs/vuetify/blob/f6ee9a33af5bc0b312158738c97dd226837cd45d/packages/vuetify/src/components/VTextField/VTextField.tsx#L126-L136
We could probably do something similar to line 145 though and just 
`const [{ 'onClick:clear': _, ...fieldProps }] = filterFieldProps(props)`

johnleider:
Is that the advised direction @KaelWD? To add exceptions for those events?

johnleider:
@KaelWD @nekosaur any updates?

johnleider:
Possibly related: #15904, #16945, 133b4172550ec7d5ddaae19085761f31c5ce3548

yuwu9145:
> I think because VTextField does some other stuff first then waits a tick before re-emitting clear:
> 
> https://github.com/vuetifyjs/vuetify/blob/f6ee9a33af5bc0b312158738c97dd226837cd45d/packages/vuetify/src/components/VTextField/VTextField.tsx#L126-L136
> 
> 
> If clear was part of fieldProps it'd happen twice
> We could probably do something similar to line 145 though and just
> `const [{ 'onClick:clear': _, ...fieldProps }] = filterFieldProps(props)`

VField is emitting "prependInner" rather than "prepend". We should focus on `filterInputProps`. 

Removing `.filter(k => !isOn(k))` makes sense, however, would this be a breaking change?

packages/vuetify/src/components/VInput/VInput.tsx
```diff
@@ -190,6 +190,6 @@ export const VInput = genericComponent<VInputSlots>()({
 export type VInput = InstanceType<typeof VInput>
 
 export function filterInputProps (props: Record<string, unknown>) {
```
KaelWD, line 192: This function can be removed and replaced with `VInput.filterProps` if it isn't handling events differently anymore.

yuwu9145: nice ! 🎉

johnleider (approved):

Issue #14773, opened 2022-03-01
[Documentation] Information on use of `!important` in flex display helper `d-flex`


Adminext:
### Problem to solve
So far there's no info in Vuetify documentation on the risk `d-flex` affecting behaviours of `v-show` and others. CSS class `d-flex` introduces flex layout by setting style of according element to `display: flex !important;` in which case overrides any other settings on `display`. Therefore, once `d-flex` is set anything based on setting CSS `display` fails.
Although this problem seem to be minor once it's realized that `!important` rule is used, lack of any info on it damanges developing experience. Thanks for consideration.

### Proposed solution
Adding a warning to Vuetify documentation, preferrably on page [flex](https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/pages/en/styles/flex.md), of something as follows:

> Please note that `d-flex` works by setting CSS of according element to `display: flex !important` thus overrides any other settings.

PR #16080, opened 2022-11-12
docs(display.md + flex.md): add caveats section

Z3rio:
## Description

I have added information about the usage of !important in the display classes. Such as how this could be an issue and what would happen with an example noted.

#14773

## Motivation and Context

This is a needed addition to the documentation because some people might overlook that `!important` is added in the display classes.

#14773

## How Has This Been Tested?
Since this is just documentation, it has been tested locally on my computer, both with a dev server and after building.

## Markup:
N/A

## Types of changes

## Checklist:

### DIFF
--- a/packages/docs/src/pages/en/styles/display.md
+++ b/packages/docs/src/pages/en/styles/display.md
@@ -79,12 +79,6 @@ Additionally, **media types** can be targeted using the `only` condition. Both `
 
 It is important to note that using any of the display classes above will result in any display style previously added being overwritten. This is because of the classes using `!important` in their display styling.
 
-#### Example
-
-If you have an element which already has `display: block` in it's style.
-
-If you then were to add the class `d-flex` to it, this would be overwritten and it would now use the style from `d-flex` instead of the original one (`display: flex!important`)
-
 </alert>
 
 ## Display in print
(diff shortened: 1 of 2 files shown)


packages/docs/src/pages/en/styles/display.md
```diff
@@ -73,6 +73,18 @@ The _condition_ applies the class base on:
 
 Additionally, **media types** can be targeted using the `only` condition. Both `hidden-screen-only` and `hidden-print-only` are currently supported.
 
+### Drawbacks
```
MajesticPotatoe, line 76: This can be placed in an alert under a section ##Caveats which can be placed right above Line 20.
See<https://github.com/vuetifyjs/vuetify/blob/4f49f08e5a86cf9f2f3c07a84cdef5aa53b39545/packages/docs/src/pages/en/components/grids.md?plain=1#L55> for example

MajesticPotatoe (changes requested):

packages/docs/src/pages/en/styles/display.md
Z3rio: If you take a look at the updated code now, is that what you meant?

packages/docs/src/pages/en/styles/display.md
```diff
@@ -73,6 +73,20 @@ The _condition_ applies the class base on:
 
 Additionally, **media types** can be targeted using the `only` condition. Both `hidden-screen-only` and `hidden-print-only` are currently supported.
 
+### Caveats
+
+<alert type="info">
+
+It is important to note that using any of the display classes above will result in any display style previously added being overwritten. This is because of the classes using `!important` in their display styling.
+
+#### Example
```
MajesticPotatoe, line 82: The example section can be removes, the previous portion already implies this.

Z3rio: This should also be resolved now.

PR #16130, opened 2022-11-19
chore: fix importMap paths on windows

jacekkarczmarczyk:
Building Vuetify on windows was producing incorrect paths in importMap ~~(+ few eslint fixes because autofix on save)~~

Also fixed generating .css files in wrong folder when building Vuetify

### steps
- build vuetify
- build api
- check api-generator/src/attributes.json (should not be empty)


MajesticPotatoe (approved):
👍

johnleider (approved):

Issue #13480, opened 2021-04-22
[Task] v-date-picker


chris-sab:
# Overview

- Related Tasks: n/a
- Related Issues: n/a

A conversion should not start until *all* assigned issues with the milestone v2.2.x are resolved and in the `next` branch. Reference our [Coding Guidelines](https://github.com/vuetifyjs/vuetify/wiki/coding-guidelines) for information regarding the team's practices or reach out to us on [Discord](https://community.vuetifyjs.com).

lorand-horvath:
May I ask whether it would be possible to prioritize the v-date-picker port to v3.1 sooner than Q2 2023 ?
I would very much like to see it in Vuetify Labs... I need it in a few projects I'm currently porting from Vuetify 2 to 3.

Thank you all for your hard work releasing Vuetify 3.1 + Labs - that's actually pretty awesome and exciting!

lorand-horvath:
@johnleider @KaelWD May I ask whether it would be possible to prioritize the v-date-picker port to v3 sooner than Q2 2023 ?
I would very much like to see it in Vuetify Labs... it's quite needed in a few projects I'm currently porting from Vuetify 2 to 3.

Thank you all for your hard work releasing Vuetify 3.1 + Labs - that's actually pretty awesome and exciting!

johnleider:
This task is already in process by @nekosaur https://github.com/vuetifyjs/vuetify/tree/feat/v3-date-picker

spmsupun-certscanner:
@johnleider when will this be released?

chuannguyen1208:
Anything new Sir. ? I'm watching this page every day

mrserzhan:
> Anything new Sir. ? I'm watching this page every day

Same here, the only thing that stops me from migrating to vuetify 3

johnleider:
Here are the P.R.s to keep an eye on:

https://github.com/vuetifyjs/vuetify/pull/16691
https://github.com/vuetifyjs/vuetify/pull/16234

Working on getting #16691 closed out this week because it's what prevents us from duplicating everything we are writing for pickers in calendars. Once that's in the pieces should fall into place considering #16234 is very far along in development.

Stefanosapk:
Cannot find the v-time-picker in Vuetify 3. Do we have the plans to include it? @KaelWD @johnleider

sulton-max:
Still can't find time picker, is this task finished ?

johnleider:
[...]

PR #16234, opened 2022-12-07
feat: v-date-picker

nekosaur:
## Motivation and Context
Port v-date-picker to v3

## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4">
      <h3>Display date {{ displayDate }}</h3>

      <!-- <v-date-picker
        v-model="date"
        :locale="locale"
      /> -->

      <v-date-field label="Date" />

      <!-- <v-date-range-card
        v-model="multiple"
        :locale="locale"
      /> -->

      <!-- <v-date-range-field from-label="From" to-label="To" v-model="multiple" /> -->
    </div>
  </v-app>
</template>

<script>
  import { changeMonth } from '../src/labs/VDatePicker/utils'
  import enUS from 'date-fns/locale/en-US'
  import sv from 'date-fns/locale/sv'

  export default {
    data: () => ({
      date: '2022-01-01',
      displayDate: '2022-06',
      locale: sv,
      mode: 'year',
      multiple: [],
      foo: null,
    }),
    methods: {
      prev () {
        const [year, month] = changeMonth(this.year, this.month, -1)
        this.year = year
        this.month = month
      },
      next () {
        const [year, month] = changeMonth(this.year, this.month, 1)
        this.year = year
        this.month = month
      },
    },
  }
</script>

<style scoped>
  .foo {
    width: 500px;
    height: 500px;
  }
</style>

```
</details>


lorand-horvath:
So this completely depends on https://github.com/vuetifyjs/vuetify/pull/16691 being implemented, according to https://github.com/vuetifyjs/vuetify/issues/13480#issuecomment-1476626896

johnleider:
> So this completely depends on #16691 being implemented, according to [#13480 (comment)](https://github.com/vuetifyjs/vuetify/issues/13480#issuecomment-1476626896)

#16691 was specifically made to accommodate this P.R., no worries.

jannis6023:
Hi, as I see this PR is going further really fast! Thanks!!! What will be an ETA? Is it already working? Thanks to everyone doing this - missing the Date Picker so much for our next project!

thanks in advance

johnleider:
This will be a heavy focus after 2.7-lts which is will be within the next few months. I'd like to get it into labs by the v3.4 release.

jannis6023:
Oh, thanks - that‘s great!!! Good job 👍

packages/vuetify/dev/vuetify.js
```diff
@@ -5,18 +5,30 @@ import { aliases, mdi } from 'vuetify/src/iconsets/mdi'
...
+import DateIoDateFnsAdapter from '@date-io/date-fns'
 
 export default createVuetify({
   directives,
   ssr: !!process.env.VITE_SSR,
   locale: {
+    locale: 'en-US',
     messages: {
-      en,
+      'en-US': en,
```
KaelWD, line 19: Should this be part of the locale object instead somehow?

johnleider:
Updated playground:

```html
<template>
  <v-app>
    <div class="ma-4 pa-4" style="max-width: 350px;">
      <v-date-input v-model="model" variant="outlined" label="Date" />

    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref()
</script>
```

packages/vuetify/dev/vuetify.js
KaelWD: Maybe match `navigator.languages` with the key as a prefix? Mine is `['en-AU', 'en-GB', 'en-US', 'en']` so with `locale: 'en'` my formatting would be `en-AU`

packages/vuetify/src/labs/VDatePicker/VDatePickerMonth.tsx
```diff
@@ -0,0 +1,361 @@
...
+                  'v-date-picker-month__day--start': item.isStart,
+                  'v-date-picker-month__day--end': item.isEnd,
+                  'v-date-picker-month__day--adjacent': item.isAdjacent,
+                  'v-date-picker-month__day--hide-adjacent': item.isHidden,
+                  'v-date-picker-month__day--week-start': item.isWeekStart,
+                  'v-date-picker-month__day--week-end': item.isWeekEnd,
+                  'v-date-picker-month__day--hovered': item.isHovered,
+                },
+              ]}
+              data-v-date={ !item.isHidden ? item.formatted : undefined }
```
KaelWD, line 322: This should be an ISO string, it only works with en-US currently.

packages/vuetify/src/labs/VDateInput/VDateInput.tsx
```diff
@@ -0,0 +1,128 @@
...
+  setup (props, { slots }) {
+    const { mobile } = useDisplay()
+    const { adapter, model, inputMode, viewMode, displayDate, parseKeyboardDate } = createDateInput(props, false)
+
+    const dialog = ref(false)
+    const inputModel = ref(model.value.length ? adapter.format(model.value[0], 'keyboardDate') : '')
+
+    function onBlur () {
+      const { isEqual } = adapter
+      const date = parseKeyboardDate(inputModel.value)
```
KaelWD, line 53: This assumes US format, so with en-AU the month and day get swapped when you click on a date. There doesn't seem to be a way to specify the input format for `Date.parse()`, it only really works with US or ISO.

KaelWD: Temporal doesn't support parsing either: https://tc39.es/proposal-temporal/docs/strings.html#localized-parsing

packages/vuetify/dev/vuetify.js
KaelWD: This guy says that's a bad idea though https://blog.sffc.xyz/post/190943794505/why-you-should-not-parse-localized-strings#:~:text=User%20Language%20!%3D%3D%20Device%20Language

jacob-carlborg-apoex: As a non-English native, I totally agree. I have all my devices set to English for the language but I'm using a 24 hour clock and the formatting of time, dates, currency and so on are all set to my native language.

KaelWD: What's `navigator.languages` for you?

jacob-carlborg-apoex: For Safari (which I usually use) it's `["en-US"]`. For Chrome it's `['en-US', 'en', 'sv']`.

KaelWD: And `new Date(1999, 1, 1).toLocaleDateString()`?

jacob-carlborg-apoex: `"2/1/1999"`.

MartinX3: Here my `navigator.languages` are `de-DE,de,en-US,en`
`new Date(1999, 1, 1).toLocaleDateString()` gives me `1.2.1999`

KaelWD: Yeah I have en-AU with en-SE for dates, but `<input type="date">` and `toLocaleDateString` just use en-AU.
I think we have three options here:
1. display and accept only ISO8601 like in v2
2. parse the result of `new Date(1999, 1, 1).toLocaleDateString()` internally and use it to infer a format. this could get pretty complicated though
3. either [1] or [2] but with `formatIn`/`formatOut` props to override it with something like moment.js that does support localisation

jacob-carlborg-apoex: For `<input type="time">` Chrome seems to follow my date time settings, but not Safari. Safari follow the language setting.

packages/vuetify/src/iconsets/mdi.ts
```diff
@@ -43,6 +43,7 @@ const aliases: IconAliases = {
   file: 'mdi-paperclip',
   plus: 'mdi-plus',
   minus: 'mdi-minus',
+  calendar: 'mdi-calendar',
```
KaelWD, line 46: Update in other sets

packages/vuetify/src/labs/VDateInput/VDateInput.tsx
johnleider: We are adding a default for now and will further address as part of our regular patch cycle.

johnleider (approved):

PR #16244, opened 2022-12-08
fix(VListGroup): improve accessibility

paul-thebaud:
## Description
This PR proposes to improve VListGroup accessibility by adding appropriate ARIA attributes (`aria-expanded` and `aria-owns`).

## Motivation and Context
Currently, the VListGroup is not threaten by screen readers as an openable menu. This might confuse users when placed in a list of VListItem and VListGroup.

## How Has This Been Tested?
This have been tested using the Vuetify playground. See the test markup bellow.

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-list v-model:opened="opened">
        <v-list-group value="group">
            <v-list-item
              title="My group"
              v-bind="props"
            />
          </template>
          <v-list-item title="My child 1" />
          <v-list-item title="My child 2" />
        </v-list-group>
      </v-list>
    </v-container>
  </v-app>
</template>

<script setup>
import { ref } from 'vue';

const opened = ref();
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VList/VListGroup.tsx
+++ b/packages/vuetify/src/components/VList/VListGroup.tsx
@@ -73,6 +73,7 @@ export const VListGroup = genericComponent<new <T extends InternalListItem>() =>
   setup (props, { slots }) {
     const { isOpen, open, id: _id } = useNestedItem(toRef(props, 'value'), true)
     const id = computed(() => `v-list-group--id-${String(_id.value)}`)
+    const contentId = computed(() => `${id.value}-content`)
     const list = useList()
 
     function onClick (e: Event) {
@@ -83,6 +84,8 @@ export const VListGroup = genericComponent<new <T extends InternalListItem>() =>
       onClick,
       class: 'v-list-group__header',
       id: id.value,
+      'aria-expanded': String(isOpen.value),
+      'aria-owns': contentId.value,
     }))
 
     const toggleIcon = computed(() => isOpen.value ? props.collapseIcon : props.expandIcon)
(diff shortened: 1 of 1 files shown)


KaelWD:
Before
Screenshot_20221211_011347
NVDA only reads `list    My group`
After
Screenshot_20221211_011416
NVDA reads
```
list    My group My group My child 1 My child 2
list    My group
list    grouping    clickable  My child 1
list    clickable  My child 2
```
if I add values to the items it's a bit better, but now repeats them twice
```
list    My group
list    grouping    clickable  My child 1
list    My child 1
list    clickable  My child 2
list    My child 2
```
It also still doesn't say there's sub items or anything, you just get `list    clickable  My group` if it isn't expanded yet

paul-thebaud:
Hello @KaelWD,
It was working for me on Chrome + NVDA.
Do you think we can add a role or something else onto the activator to fix this behavior?

KaelWD:
I can't really get a good result. Replacing `aria-owns` with `aria-controls` and adding `role="listitem"` to the items fixes the activator title but stops the children from being read.

KaelWD:
`role="tree"` with `aria-owns` + `role="treeitem"` seems to do what we want, but probably isn't suitable for non-nested lists. Will have to figure out some way to set that in the root based on how its children are structured.

paul-thebaud:
Hi @KaelWD, thanks for all of your testing. It looks like this is bigger than I thought.
If I can do anything, please let me know.

PR #16255, opened 2022-12-09
fix(VAutocomplete/VCombox/VSelect): add check for readonly form

### LINKED ISSUE #16161: [Bug Report][3.0.2] (VForm) readonly states does not apply to VSelect
### Environment
**Vuetify Version:** 3.0.2
**Vue Version:** 3.2.45
**OS:** Linux x86_64

### Steps to reproduce
Create a VSelect with items (multiple or not), wrap it inside a VForm where you set the readonly props.

### Expected Behavior
The readonly prop of VForm should apply to VSelect like if we were defining the prop directly on VSelect component.

### Actual Behavior
VSelect is still writable when wrapped in a readonly VForm.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqFUstOwzAQ/BXLl4DUxIjHJSoIOPABcKx7cJsNBOKH7E0RqvLvrJ20Sh+CXKLZ2Z0Zr3ax5U/OFZsOeMnnYe0bhywAdu5BmkY765FtmYea9az2VrOMWjNppFlbE5BtVNsBu48dF1ltbXa5YxoEHUZmIQ0jGWywhZJlL9Q3G0YJxSnWzw5bnpWftKwIpZYlyc/FEJMCEiAX1yoEQozNN7lyMTmjj4BWjRlRwpQMqQR+X0zl2npNOVVlTfuzpyRKJDJAC2t6aK5tBe295CmU5KxVq4Rfx8FhP1GLyDI9n9j0p4KYWorB8zDFv0ZvqEylWmtgH/ac046bWEbD47fH2n4/EQybm4vJRvmMDzdAna74DNbQlWzjANklggxLliqxRrcRseQfiC6UQnTGfb0Xa6vFI3HCdwYbDXll9eNNcV3c3omqCTitFxB0vvL2O4AnR8nTYYzi2NT05PMGI026V8X1IDuWouSJlCD9Dfjcg6nAg/8z91HvQfYj7iR/9Oyl6Xk/48McbTWF5Mtfl0El3A==)

### Other comments
Looks like it is working with VTextField, so this might not work with other inputs too.
VSelect readonly prop is working fine.

vexleet:
Used useForm composable in order to get if select is wrapped inside a form. After that I used isReadonly (if a form is found) and added it in the functions that open the vmenu on click. They were already checking for props readonly but were missing a check for the form.

fix #16161

## Description

I also thought of having a check in `VTextField` in the `onClick:control` event with something like `onClick:control={ isReadonly ? undefined : onControlClick }` but also thought that having `useForm` in the selects components themselves it will be easier to do some other checks for other possible cases.

## How Has This Been Tested?

I created unit tests checking if a menu becomes active when you press it or you press down key when you focus the input.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:


packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -215,7 +217,7 @@ export const VSelect = genericComponent<new <
             },
           ]}
           appendInnerIcon={ props.menuIcon }
-          readonly
+          readonly={ props.readonly }
```
KaelWD, line 220: Should never be able to type in here

KaelWD, line 220: Should never be able to type in here

KaelWD (changes requested):

packages/vuetify/src/components/VSelect/VSelect.tsx
vexleet: Ok now it make sense why it was always `true`, didnt quite thought about that. Thought it was just missed.

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -138,13 +140,13 @@ export const VAutocomplete = genericComponent<new <
     function onClickControl () {
       if (
         (props.hideNoData && !items.value.length) ||
-        props.readonly
+        props.readonly || form?.isReadonly
```
KaelWD, line 143: This condition is always true if form is defined

```suggestion
        props.readonly || form?.isReadonly.value
```

KaelWD (changes requested):

packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
```diff
@@ -1,5 +1,6 @@
-/// <reference types="../../../../types/cypress" />
+/// <reference types="..\..\..\..\types\cypress" />
```
KaelWD, line 1:
```suggestion
/// <reference types="../../../../types/cypress" />
```

vexleet:
@KaelWD resolved conflicts and comments

johnleider (approved):

PR #16272, opened 2022-12-13
fix(VAutocomplete): enable jest click to open menu

AntonioDell:
VAutocomplete now opens it's menu in testing environments using jest `.trigger("click")`.

## Description
The `onClick` method of VAutocomplete didn't trigger the menu to open in a testing environment, because the `onFocus` method of VInput, which was called before this change, didn't actually focus the VAutocomplete. This was due to detachement from root in a jest envrionment.
Since VSelect does this differently, I thought I could simply use the same behavior as VSelect here.

## Motivation and Context
When testing components which use VAutocomplete, until now it wasn't possible to open the VAutocomplete menu with jest methods, such as `.trigger('click')` on the `v-input__slot`. This made testing anything containing VAutocomplete very hard.
Now VAutocomplete acts more like VSelect when clicked and thus can be tested more easily, f.e. like this:
```
  const autocomplete = wrapper.find("#autocomplete-field");
  const slot = autocomplete.find(`.v-input__slot`);
  const listId = slot.attributes("aria-owns");

  await slot.trigger("click")
  
  const optionList = createWrapper(document.body).find(`#{listId}`);   // <-- This previously couldn't find the option list, because the click wouldn't open the menu
  const options = optionList.findAll("[role=option]");
```

This is related to https://github.com/nuxt-community/vuetify-module/issues/466.

## How Has This Been Tested?
I created a test before the changes checking if the `document.body` contains the open menu after clicking the `v-input__slot` and saw, that `document.body` was empty. Then I added the changes and saw that it wasn't empty anymore.

Result of `console.log(createWrapper(document.body).html())` after `slot.trigger("click")` before the changes of the commit:
```
<body>
  <div data-app="true">
  </div>
</body>

``` 
Result of `console.log(createWrapper(document.body).html())` after `slot.trigger("click")` after the changes of the commit:
```
<body>
  <div data-app="true">
    <div
      class="v-menu__content theme--light menuable__content__active v-autocomplete__content"
      style="
        max-height: 304px;
        min-width: 0px;
        max-width: auto;
        top: 12px;
        left: 0px;
        transform-origin: top left;
        z-index: 8;
        display: none;
      "
    >
      <div
        role="listbox"
        tabindex="-1"
        class="v-list v-select-list v-sheet theme--light theme--light"
        id="list-1"
      >
        <div
          tabindex="0"
          aria-selected="false"
          id="list-item-7-0"
          role="option"
          class="v-list-item v-list-item--link theme--light"
        >
          <div class="v-list-item__content">
            <div class="v-list-item__title">foo</div>
          </div>
        </div>
        <div
          tabindex="0"
          aria-selected="false"
          id="list-item-7-1"
          role="option"
          class="v-list-item v-list-item--link theme--light"
        >
          <div class="v-list-item__content">
            <div class="v-list-item__title">bar</div>
          </div>
        </div>
      </div>
    </div>
  </div>
</body>
``` 

I wasn't sure if I should have this test in the commit too, so I removed it. But if you like, I can add it again.

Also I needed to change a couple of failing tests because of my changes, but these should not be critical to the functionality.

## Markup:
The changes only affect how VAutocomplete behaves in testing environment, so no markup needed (I suppose).

## Types of changes

## Checklist:


Dimas-e63s:
One question regarding styles of the `v-menu` component mentioned in the result section: why does it have `display: none` styles applied when it has class `menuable__content__active`?

`display: none` property disallows all interactions with the expanded content of `v-menu` (if it has interactive elements such as buttons)

Can that be fixed? (I mean removing `display: none` style from `v-menu` element when it has class `menuable__content__active`)

P.s

one snapshot shows that `v-menu` doesn't contain `display: none` style applied when active. But other 
https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VMenu/__tests__/__snapshots__/VMenu.spec.ts.snap#L33

P.p.s
Also notices that tests, for that component are skiped :D
https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.ts#L17

AntonioDell:
> One question regarding styles of the `v-menu` component mentioned in the result section: why does it have `display: none` styles applied when it has class `menuable__content__active`?
> 
> `display: none` property disallows all interactions with the expanded content of `v-menu` (if it has interactive elements such as buttons)
> 
> Can that be fixed? (I mean removing `display: none` style from `v-menu` element when it has class `menuable__content__active`)
> 
> P.s
> 
> one snapshot shows that `v-menu` doesn't contain `display: none` style applied when active. But other https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VMenu/__tests__/__snapshots__/VMenu.spec.ts.snap#L33
> 
> P.p.s Also notices that tests, for that component are skiped :D https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.ts#L17

Since I tried to change as little as possible, I didn't touch any `display: none` styling on v-menu, so I cannot say why it is there, sorry. But if that is a problem, maybe open a issue for it, explaining it in more detail?
Thank you for your interest in this PR, hope it get's merged :)

johnleider (approved):

arnegns:
When can I expect this fix?

PR #16278, opened 2022-12-14
feat(V-Menu): allow functions for closeOnClick in v-menu.

jaydiazzz:
## Description
allow for closeOnClick property in the v-menu component to use a function provided by the developer, instead of the standard close procedure.

## Motivation and Context
This is required for the case that you want to handle the value of v-menu manually while not having direct access. For example, if this is nested within v-select we need to be able to have access to the close event. This helps with customizability for the component and better control over it.

## How Has This Been Tested?
By using this in my local instance and it having the intended affects while also allowing my other v-menu components to function normally.

## Markup:

<details>

```vue
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VMenu/VMenu.ts
+++ b/packages/vuetify/src/components/VMenu/VMenu.ts
@@ -327,7 +327,7 @@ export default baseMixins.extend({
         directives.push({
           name: 'click-outside',
           value: {
-            handler: () => { typeof this.closeOnClick === 'function' ? this.closeOnClick() : ( () => { this.isActive = false } )() },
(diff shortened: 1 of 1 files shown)


jaydiazzz:
That’s not very conclusive of an answer. Will you be able to further explain what actions are necessary from this point?

jaydiazzz:
> This is a feature, not a fix

Will you be able to further explain what actions are necessary from this point?

KaelWD:
It means rebase to v2-dev. How do you intend to use this anyway, you can already prevent menus from closing with `:closeOnClick="false"

PR #16300, opened 2022-12-20
Added missing menu prop for combobox

Shdorsh:
The `menu` prop from Vuetify still works, and is the only way how one can toggle the selection properly.

## Description
Added the menu prop to the documentation, as it still works. A simple boolean that either makes the selection menu of a combobox visible or not

## Motivation and Context
People will use a bad framework with good documentation, but none will use a good framework with a bad documentation.

## How Has This Been Tested?
Had to use it to create a combobox prop that takes a store and AJAX fetch function to dynamically load the data in it on typing. Problem is on changing items, the selection menu wouldn't trigger as I wanted so I had to change this.
</details>

## Types of changes

- [X] Documentation

## Checklist:


Shdorsh:
Imported some stuff from the next branch unwittingly, going to make a new one with only the right file changed.

PR #16343, opened 2022-12-29
feat(VList): add lazy mode to VListGroup

tinou98:
## Description
Add a `lazy` props to skip the rendering of children if the group has never been opened

## Motivation and Context
This allows to render instantly a list with ~1k elements (see example in docs)

## How Has This Been Tested?
Manually (patched in my project, used to crash Chrome and Firefox, now work instantly).
Also added 2 Cypress test.

## Markup:
Available in the documentation

<details>

```vue
<template>
  <v-card>
    <v-card-title>List with {{ size }} elements</v-card-title>

    <v-card-text v-if="data">
      <v-list>
        <recursive-list :entry="data"></recursive-list>
      </v-list>
    </v-card-text>
    <v-card-actions v-else class="justify-center">
      <v-btn @click="data = MakeEntries()">Prepare</v-btn>
    </v-card-actions>
  </v-card>
</template>

<script lang="ts" setup>
  import { computed, h, ref, VNode } from 'vue'
  import { VListGroup, VListItem } from 'vuetify/components'
  export type Entry = {
    name: string,
    children?: Entry[],
  }
  /*
  <VListGroup v-if="entry.children" lazy>
      <VListItem v-bind="props">
        {{ p.val.name }}
      </VListItem>
    </template>
    <RecursiveList :val="e" v-for="e in p.val.children" />
  </VListGroup>
  <VListItem v-else>
    {{ p.val.name }}
  </VListItem>
  */
  const RecursiveList = ({ entry }: { entry: Entry }): VNode =>
    entry.children
      ? h(VListGroup, { lazy: true }, {
        default: () => entry.children?.map(entry => RecursiveList({ entry })),
        activator: ({ props }: any) => h(VListItem, props, { default: () => entry.name }),
      })
      : h(VListItem, null, { default: () => entry.name })
  function MakeEntries (prefix = 'Prop', depth = 5): Entry {
    let children
    if (depth > 0) {
      children = []
      for (let i = 0; i < 5; ++i) {
        children.push(MakeEntries(prefix + '.' + i, depth - 1))
      }
    }
    return { name: prefix, children }
  }
  function GetSize (v: Entry): number {
    return v.children?.map(GetSize).reduce((acc, v) => acc + v, 1) ?? 1
  }
  const data = ref<Entry>()
  const size = computed(() => data.value ? GetSize(data.value) : 0)
</script>
```
</details>

## Types of changes

## Checklist:

### DIFF
--- a/packages/api-generator/src/locale/en/VListGroup.json
+++ b/packages/api-generator/src/locale/en/VListGroup.json
@@ -11,6 +11,7 @@
     "expandIcon": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-group.json))",
     "fluid": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-group.json))",
     "subgroup": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-group.json))",
-    "items": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-group.json))"
+    "items": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-group.json))",
+    "lazy": "Lazilly render children"
   }
 }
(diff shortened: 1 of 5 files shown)


nekosaur:
There is a lazy composable that you might be able to use

tinou98:
It does indeed simplify the code, but has the drawback of making lazy the default behaviour (might break/slow some application).
I've updated (but not the test/docs). If it's good, I'll update the rest.

johnleider:
Please update your fork and rebase this onto dev.

tinou98:
Rebased

PR #16459, opened 2023-01-18
feat(defaults): allow setting class and style

### LINKED ISSUE #16896: [Feature Request] Support style prop for global defaults
### Problem to solve
More flexibility in overriding specific styling in certain situations e.g. I want to allow scrolling and limit the max height for a v-container inside a v-window-item, but currently I am unable to do that without writing a custom component for this case, which is clunky and not great for bundling logic together in one component since it's just for styling.

### Proposed solution
Allow style as a global prop? This would have higher priority than all other defaults but be overwritten by the style prop in sfc

Example:

image

### LINKED ISSUE #17188: [Feature Request] Enable "class" property on global configuration
### Problem to solve
I have tried to define components with custom styles, this using "aliases" (eg: aliases: {myButton: VBtn}). This allows me to modify the reference component's props without modifying the parent.

But there is a limitation. What could I do if I wanted all "myButton" buttons to have "text-transform: lowercase". For this, it only occurred to me to go from button to button by adding a class to it.

### Proposed solution
I think that it would be ideal to be able to assign a reference class from the global configurations. Maybe creating a custom composable.

johnleider:
* [x] Implement in all components #16854
* [x] Create unit tests #16854

## Motivation and Context
Makes it easier to handle aliased component styling through global defaults.

Closes #16896
Closes #17188

## Markup:
<details>

```vue
<template>
  <div class="ma-4 pa-4">
    <v-btn
      color="primary"
      class="text-blue"
      style="border-radius: 0;"
    >
      Inline
    </v-btn>

    <br>
    <br>

    <v-btn
      color="primary"
      :class="['text-blue']"
      :style="['border-radius: 0;']"
    >
      Array
    </v-btn>

    <br>
    <br>

    <v-btn
      color="primary"
      :class="{ 'text-blue': true }"
      :style="{ borderRadius: 0 }"
    >
      Object
    </v-btn>
  </div>
</template>

<script setup>
  import { provideDefaults } from '@/composables/defaults'

  provideDefaults({
    global: {
      class: 'v-global-class',
      style: {
        opacity: 1,
      },
    },
    VBtn: {
      class: 'v-btn--alt',
      style: {
        opacity: 0.5,
      },
    },
  })
</script>

<style lang="sass">
  .v-global-class
    width: 300px

  .v-btn--alt
    width: 200px
</style>

```
</details>


blalan05:
🔥 LGTM... for what that's worth.

yuwu9145:
This pattern of using props breaks `inheritAttrs: false` which has been extensively used in Vuetify.

Here is a simplified "Vue only" [reproduction link](https://sfc.vuejs.org/#eNqVUsFOwzAM/ZUoQtplbbRrKUgTJ85cc+natM1ok8hxO9DUf8dJujFAIHGpHNvv+fm5Z753Lp8nxQte+hq0Q+YVTu5RGj06C8ieej00rAU7sk0u4isANtKUIiGot0Q1uqFCRTFjZcLUQ+X9g+SgmuzQSS5Co7h28i1PI7KxcvnRW0MizgEv14KXvGAxE3I0NLwl7xGdL4TwbR2UHH1uoRMU5TAZ1KPKlR+zA9iTV0DEkm9vOAQlZwUZKNMoUPAX57fWH7yBdpFmoVWuxnw6Seuqt+hho9pqGjDtok2vQOMeEXzB2mrwKupzYEnCZd9oXsFeELTpYn2hLw373fVGzzGgsN+xOTto05D9d1WYJPlak/jMTpVBhvYihaULpZnxAFiKfhcBFBJZcbnlXVSZp85bysaaDf6HuBRJr8QvfwU9PL4PaaV8xa+eHKr6tQM7maYIzPer/eRIRPDlA0Dy9LY=)

In Vuetify, it has been causing problems to nested components, VCombobox one of the examples:

VCombox has root element:
```js
<VTextField class={['v-combox'...]}
```

VTextField `{ inheritAttrs: false }` has template:
```vue
<VInput class={[''v-text-field''...]}  { ...rootAttrs }> <!--manually apply fall through class aims to apply "v-combox" class here-->
  <VField /> <!--"v-combox" class is NOT supposed to assigned here-->
</VInput>
```
**Expected result**: "v-combox" class should not be assigned to `VField`

**Actual result**: "v-combox" class is assigned to both `VInput` and `VField`


**Proposed solution** : Instead of holding class & style in props, modify global defaults logic to keep them in `attrs`. This shall prevent arbitrary classes bypass `{ inheritAttrs: false }` and "fall through" via props

KaelWD:
We can't control fallthrough from a composable so the alternative would be another proxy with `inheritAttrs: false` and `{ ...attrs }` on everything, which also makes it hard to expose this for public use (#16461). Do you think adding exceptions for class and style to `filterVTextFieldProps` etc. would be sufficient?

yuwu9145:
I have gone through all nested components and applied cy tests against them. `filterVTextFieldProps` is the only place causing this issue and only three components `VSelect` `VCombobox` `VAutocomplete` are impacted.

For existing code base, add exceptions for class and style for  filterVTextFieldProps could be labeled as "sufficient".

However, this new feature means props will start to hold class/style which are usually targeting a certain parent component, in future if someone does this,

```vue
<template>
  <parent>
    <child { props } />
  </parent>
</template>
````

[props.class would contaminate childs classes](https://sfc.vuejs.org/#eNqVUk1PwzAM/StRhLTL2mjXUpAmTpy55tKtXpvRJpHjbqCp/x03accAgcQl8ufz83Mucut9fhpAFrIMezSeRAAa/KO2pvcOSTy1pqvFAV0vVrmK3tSw0rZUqYNrS4LedxUB20KUqWffVSE8aNkggM12jZZqKlXXWrmWaUjWVz4/BmeZxmVC0HMiaFmIGJliPHbytWyJfCiUCof9xOUYcoeNYivHwZLpIYfQZzt05wDIwFqubzAUB0+AGYKtAQH/wvxW+gN3gh21HXmVqzSfWvK68BZVrOFQDR2lXYxtAQ1tiTAU4lB1ASI/j44pLPtG+QrxQmhsE/Mjvzzsd91rc4oGm+1GnLKdsTUf4K6aJmk55zQ9i3NlSZBbqIjlRmlqPAGVqt3EFjYZrljueRd55qnyFrR2dkX/gy5V4qzpy89gJ9B7l9bKrwizMrtq/9qgG2xdJPT7+QysTOyS4weH5vjk).

Is this something need to worry? Is it worth to spend some time on doing some R&D for the approach of setting up attrs in proxy?

**Comment by yuwu9145** [comment]
LGTM

File: packages/docs/src/pages/en/features/global-configuration.md [hunk 1]
```diff
@@ -57,6 +58,73 @@ This is used internally by some components already:
...
+import { createVuetify } from 'vuetify'
+import { VSheet } from 'vuetify/components/VSheet'
+
+export default createVuetify({
+  aliases: {
+    GSheetPrimary: VSheet,
+  },
+  defaults: {
+    GSheetPrimary: {
+      class: 'border',
```
yuwu9145, line 112: Should `border` be `g-sheet-primary`?

praburangki:
Any plan on defining custom class for other elements other the root of the component?
For example in the button I want to add a custom class for the `v-btn__content`

johnleider:
> Any plan on defining custom class for other elements besides the root of the component? For example in the button I want to add a custom class for the `v-btn__content`

No, only components.

PR #16465, opened 2023-01-18
fix(VBtn): accept value prop for underlying DOM element 

### LINKED ISSUE #16188: [Bug Report][3.0.2] v-btn's value props does not works
### Environment
**Vuetify Version:** 3.0.2
**Last working version:** 3.0.2
**Vue Version:** 3.2.45
**OS:** Windows 10

### Steps to reproduce
1. Put a value prop in v-btn.
2. Assign a method to the `click` event to display its own value (`e.target.value`) attribute as an `alert`.
3. When clicked, it becomes undefined and does not contain the value defined in `value` props.

### Expected Behavior
The value of the value prop is displayed

https://next.vuetifyjs.com/en/api/v-btn/#props-value

### Actual Behavior
The value of the `value` prop is ignored. 
When I checked the DOM in the console log, the value attribute was not included.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNp9UstOwzAQ/JXFl7ZSYyMel9JWhRN/wAFzSJNtCfVL9qaAqv47m6SU0ApyiDQ745nxys87cR+C3NYoJmKailgFgoRUh7l2lQ0+Euwg4gr2sIrewoClA+20K7xLBDatYdbww8EjGuPhyUdTXgxGPxJC/s1giCOYzWGnHUBuMNIQJeVxjSS3ualxdKfdXrup6kpwPANCG0xOyAhgus3y0PQC/hjYvHIH1GLCD8pWFZoSmPQlmpkWXFALUH3dkhy0kUw35ZhfFKYqNkc8f6iJPHdpxd+JjI6RDejKTFWvpBiLbmmsDPItecdrba+sD0TSYtItoZnxMhusxStRSBOlahc2a1l4qxbMqVg7qixmpbeLa3klb25VWSXqzyUmmy2jf08YOVGLcc+cqtXnXwEHmn0v5VVnexg1lmdWiv23GLOIrsSI8d/eJ9pf3U+4s/5NJj+EvdiPRXeOt9qWFC9ftdXkxQ==)

floroz:
## Description

- Implements the logic to allow consumers to provide a `value` prop that will be attached as Node attribute to the underlying DOM element.

## Motivation and Context

fix #16188

## How Has This Been Tested?

Added Cypress Component testing

## Types of changes

## Checklist:

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.tsx
+++ b/packages/vuetify/src/components/VBtn/VBtn.tsx
@@ -112,6 +112,12 @@ export const VBtn = defineComponent({
     const isElevated = computed(() => {
       return props.variant === 'elevated' && !(props.disabled || props.flat || props.border)
     })
+    const valueAttr = computed(() => {
+      if (props.value === undefined) return undefined
+
+      return Object(props.value) === props.value
+        ? JSON.stringify(props.value, null, 0) : props.value
+    })
 
     useSelectLink(link, group?.select)
 
@@ -168,6 +174,7 @@ export const VBtn = defineComponent({
             link.navigate?.(e)
             group?.toggle()
           } }
+          value={ valueAttr.value }
         >
           { genOverlays(true, 'v-btn') }
 
--- a/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
+++ b/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
@@ -219,18 +219,40 @@ describe('VBtn', () => {
     })
   })
 
-  describe.skip('value', () => {
-    // none of the "value" props are implemented yet
-    it('should stringify non string|number values', () => {
-      const objectValue = { value: { hello: 'world' } }
-      const numberValue = { value: 2 }
+  describe('value', () => {
+    it('should pass string values', () => {
+      const stringValue = 'Foobar'
 
-      cy.mount(<VBtn value={ objectValue }></VBtn>)
+      cy.mount(<VBtn value={stringValue}></VBtn>)
         .get('button')
-        .should('contain.text', JSON.stringify(objectValue, null, 0))
-        .mount(<VBtn value={ numberValue } />)
+        .should('have.value', stringValue)
+    })
+
+    it('should stringify object', () => {
+      const objectValue = { value: {} }
+      cy.mount(<VBtn value={objectValue}></VBtn>)
         .get('button')
-        .should('contain.text', numberValue.value)
+        .should('have.value', JSON.stringify(objectValue, null, 0))
+    })
+
+    it('should stringify number', () => {
+      const numberValue = 15
+      cy.mount(<VBtn value={numberValue}></VBtn>)
+        .get('button')
+        .should('have.value', JSON.stringify(numberValue, null, 0))
+    })
+
+    it('should stringify array', () => {
+      const arrayValue = ['foo', 'bar']
+      cy.mount(<VBtn value={arrayValue}></VBtn>)
+        .get('button')
+        .should('have.value', JSON.stringify(arrayValue, null, 0))
+    })
+
+    it('should not generate a fallback value when not provided', () => {
+      cy.mount(<VBtn></VBtn>)
+        .get('button')
+        .should('not.have.value')
     })
   })
 


packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -113,6 +113,17 @@ export const VBtn = defineComponent({
       return props.variant === 'elevated' && !(props.disabled || props.flat || props.border)
     })
 
+    const buttonId = String(Math.random()).replace('.', '')
+
```
floroz, line 117: this could be extracted into a separate function `generateRandomId` but I wanted to check first where would be the correct place for this helper.

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -155,23 +166,24 @@ export const VBtn = defineComponent({
             locationStyles.value,
             sizeStyles.value,
           ]}
-          disabled={ isDisabled.value || undefined }
-          href={ link.href.value }
+          disabled={isDisabled.value || undefined}
+          href={link.href.value}
```
floroz, line 170: I've reset my VSCode settings to make sure I wasn't picking up any other configuration, and run separately the linter.

These fixes seem to be legit so the file was not formatted correctly previously.

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -212,24 +220,26 @@ export const VBtn = defineComponent({
...
+              <span class="v-btn__append">{slots.append?.() ?? <VIcon />}</span>
             </VDefaultsProvider>
-          ) }
+          )}
 
-          { !!props.loading && (
+          {!!props.loading && (
             <span key="loader" class="v-btn__loader">
-              { slots.loader?.() ?? (
+              {slots.loader?.() ?? (
```
floroz, line 229: wasn't clear from the issue open whether the `value` prop should actually be displayed in place of the `slot`.

I don't think that would be the correct behaviour as the `value` on button seems to only be useful when passed via a Group?

floroz:
If these changes are fine, I'd be happy to also tackle the `group:selected` event which seems to have an incorrect signature, since it set to only emit `true` but never invoked upon click.

I assume the correct behaviour should be to emit the `props.value` ad added in the `onClick`

packages/vuetify/src/components/VBtn/VBtn.tsx
nekosaur: No it is correctly formatted. There should be spaces.

nekosaur:
> If these changes are fine, I'd be happy to also tackle the `group:selected` event which seems to have an incorrect signature, since it set to only emit `true` but never invoked upon click.
> 
> I assume the correct behaviour should be to emit the `props.value` ad added in the `onClick`

That emit declaration is correct iirc. It is emitted from the useGroupItem composable

packages/vuetify/src/components/VBtn/VBtn.tsx
KaelWD: You have prettier or something running

packages/vuetify/src/components/VBtn/VBtn.tsx
KaelWD: What is the purpose of a random ID?

floroz: > What is the purpose of a random ID?

The docs state: 

>The value used when the component is selected in a group. If not provided, a unique ID will be used.

KaelWD: It already uses the index of the item in the group.

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -113,6 +113,14 @@ export const VBtn = defineComponent({
...
     })
 
+    const valueAttr = computed((): string | undefined => {
+      if (props.value === undefined) {
+        return undefined
+      }
+
+      return typeof props.value === 'string' ? props.value : JSON.stringify(props.value, null, 0)
+    })
+
```
floroz, line 123: I can inline this computed property as

```ts
const valueAttr = computed((): string | undefined => typeof props.value === 'undefined'
      ? undefined : typeof props.value === 'string'
        ? props.value : JSON.stringify(props.value, null, 0))
```

not a fan of shortening a few chars and losing readability, but happy to change if better reflects the style

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -113,6 +113,14 @@ export const VBtn = defineComponent({
       return props.variant === 'elevated' && !(props.disabled || props.flat || props.border)
     })
 
+    const valueAttr = computed((): string | undefined => {
+      if (props.value === undefined) {
+        return undefined
+      }
+
+      return typeof props.value === 'string' ? props.value : JSON.stringify(props.value, null, 0)
```
johnleider, line 121: What is the use-case for object value?

johnleider (changes requested):

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -113,6 +113,14 @@ export const VBtn = defineComponent({
       return props.variant === 'elevated' && !(props.disabled || props.flat || props.border)
     })
 
+    const valueAttr = computed((): string | undefined => {
```
johnleider, line 116: you shouldn't need to define the return type

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -113,6 +113,14 @@ export const VBtn = defineComponent({
       return props.variant === 'elevated' && !(props.disabled || props.flat || props.border)
     })
```
johnleider, line 115:
```suggestion
```

packages/vuetify/src/components/VBtn/VBtn.tsx
johnleider: if you want you could do the following to remove 2 loc

`if (props.value === undefined) return undefined`

johnleider (changes requested):

floroz:
Thanks for the detailed feedback @johnleider. 

I've polished the PR and implemented your suggestions.

packages/vuetify/src/components/VBtn/VBtn.tsx
```diff
@@ -167,7 +173,8 @@ export const VBtn = defineComponent({
 
             link.navigate?.(e)
             group?.toggle()
-          } }
+          }}
```
johnleider, line 176:
```suggestion
          } }
```

johnleider (changes requested):

packages/vuetify/src/components/VBtn/VBtn.tsx
floroz: is there a way for eslint (via prettier?) to automatically apply this formatting rule?

johnleider: I wish lol. For now it's just an internal team guideline.

johnleider (approved):

PR #16525, opened 2023-01-25
fix(vswitch): thumb inherits `currentColor` as default when no `props.color` is provided

### LINKED ISSUE #16486: [Bug Report][3.1.2] VSwitch default color should be 'primary'
### Environment
**Vuetify Version:** 3.1.2
**Vue Version:** 3.2.45
**Browsers:** Chrome 109.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Create a v-switch using default settings

### Expected Behavior
There should be more of an indication beside the switch in left or right position.

### Actual Behavior
The color, outline, border does not change when the switch is on.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNUMtOwzAQ/BXLl1waWxS4RAkq4gPgApemBzfZtqZ+yXZSVVX+HTuJREgF9ObZGc/OzvqCn40hbQM4w7kHaQTz8FQqhPI2Zcb0zx5IxtWIeuxO3FcHlEldg/hgooEi2THhIEH0P523ze2ySgtti8RYLpk9J0iwLYgieVXo3XG1R28DgV6i7ts1p5PMEQzX5HRyJV5gLo22PigN+XRahRou8UM5Eq7EGeoncRZ6irjEB++NyyhtlDnuSaUlXQWO2kZ5LiGttVzdkyV5eKQ1d346J+BkurX65MCGjSVeTMw9351/WzDSwfeOLAfbcRQtr6xo8G/BphZUDRbsn7ln2h/ZZ9xV/rizK1UXuhRcHd2sxsr1Fa43UdIt8GAdxP0dePMFzF/S8g==)

### Other comments
If there were 2 labels, on each side of the switch, then I would agree the color does not need to change.

The workaround is to set the primary color in the Vuetify options when creating the plugin.

floroz:
## Description

This PR implements a default behaviour for the VSwitch thumb, when `props.color` is not defined, to inherit the parent's color via `currentColor` CSS property.

It also adds a layer of component test that verifies that all `props.color` passed to this component are converted to RGB color format.

fix #16486

## Markup:

```vue
<div style="color: rgb(50, 50, 50); --v-theme-surface: rgb(255,255, 255)">
    <v-switch />
    <v-switch color="#10563a" />
    <v-switch color="hsla(303, 100%, 50%, 1)" />
    <v-switch color="rgba(20,57,224, 0.2)" />
  </div>
```


packages/vuetify/src/components/VSwitch/VSwitch.sass
```diff
@@ -33,7 +37,7 @@
 .v-switch__thumb
   align-items: center
   border-radius: $switch-thumb-radius
-  color: rgb(var(--v-theme-surface))
```
floroz, line 36: this SASS conversion was redundant, and also was breaking Cypress component testing.

Not experienced enough with `.sass` files, I'd be curious since `rgb` is a compile-time function, and `var(--v-theme-surface)` is a runtime value, how the conversion is handled?

Nevertheless, the conversion to RGB happens elsewhere and the unit test is going to lock in that behaviour to avoid any breaking change.

packages/vuetify/src/components/VSwitch/__tests__/VSwitch.spec.cy.tsx
```diff
@@ -0,0 +1,72 @@
...
+    it('supports HSL custom colors provided as prop', () => {
+      const hsl = 'hsla(303, 100%, 50%)'
+      const rgb = 'rgb(255, 0, 242)'
+
+      expectColorPropConversionToRGBColor(hsl, rgb, styleContext)
+    })
+  })
+})
+
+function expectColorPropConversionToRGBColor (
```
floroz, line 51: I would have expected to be able to use `it.each` for this run, but it seems that the Cypress test runner doesn't support that? TypeScript tells me should be using Mocha, which supports it.

So that's strange, I tried googling around but considering the Cy Component Testing is something quite recent, I could not find any useful answer.

If there is a way to run a suite iteratively over a sample array, that'd be better for code maintenance and semantics.

johnleider:
> Is this a fix? It sounds more like a feature to me. Do any of our other components work this way?

`v-switch` should be able to indicate it's active status without a color like the other selection controls.

packages/vuetify/src/components/VSwitch/VSwitch.sass
nekosaur: Not sure how it was breaking cypress, but `rgb(var(--v-theme-surface))` is correct.

nekosaur:
> > Is this a fix? It sounds more like a feature to me. Do any of our other components work this way?
> 
> `v-switch` should be able to indicate it's active status without a color like the other selection controls.

sure, I'd agree with that

packages/vuetify/src/components/VSwitch/VSwitch.sass
floroz: @nekosaur Yes, the issue not related to Cypress. If you add the below code to the Playground and run `yarn dev`

```vue
<template>

  <div id="parent" :style="{ color: 'rgb(5,45,5)', '--v-theme-surface': 'rgb(200, 200, 200)' }">
    <v-switch />
  </div>

</template>
```

You will notice that the `color` property of the thumb fails to correctly parse the `--v-theme-surface` when wrapped in the `rgb` function. If you remove `rgb` the color is computed correctly.

Any idea?

floroz: Again, to me make sense that it doesn't work correctly because the value of `--v-theme-surface` will only be known at browser-render time, while the `rgb` will be invoked during the compilation of `node-sass`, so the value cannot be known/parsed ahead of time.

Is there something I am missing?

packages/vuetify/src/components/VSwitch/__tests__/VSwitch.spec.cy.tsx
floroz: - looked at it, doesn't seem supported yet by cypress test runner -

johnleider (changes requested):
`v-switch` has **v-selection-control--dirty**. This should be a small one liner I think:

```sass
.v-switch__thumb
  .v-selection-control--dirty &
    color: currentColor
```

johnleider (commented):
Looks good but I'd like to get feedback from the @vuetifyjs/core-team on the need for tests; I don't think they are necessary here.

floroz:
> Looks good but I'd like to get feedback from the @vuetifyjs/core-team on the need for tests; I don't think they are necessary here.

Yes, the first one is the one required because verifies the correct behaviour introduced by the new CSS rule (thumb color changes when no `props.color` is provided. 

The others are verifying the behaviour of the color switch when props is provided, but they were also created by my initial misunderstanding of the computed value of `--v-theme-surface`, so happy to remove them if they are redundant.

johnleider (changes requested):
Can you simplify the tests to do a screen test for the switch on and off with and without color. Your current tests are essentially retesting our color composable.

y-takebe:
I have one request. There seems to be a potential problem with the VSwitch in Vuetify 3 where setting the color slows down the toggle transition in the Safari browser.

You can check the [official page](https://vuetifyjs.com/en/components/switches/). We have tried on an iPhone 14 pro and iPhone 12 pro.

For this reason, my project avoids this problem by not setting a color on the VSwitch.

I would prefer to have the initial color, but would be happy if it does not slow down the toggle transition.

If that is not feasible, you must be able to disable the color settings.

Thank you for everything!

floroz:
> Can you simplify the tests to do a screen test for the switch on and off with and without color. Your current tests are essentially retesting our color composable.


@johnleider I've updated the tests and added Percy snapshots.



> I have one request. There seems to be a potential problem with the VSwitch in Vuetify 3 where setting the color slows down the toggle transition in the Safari browser.
> 
> You can check the [official page](https://vuetifyjs.com/en/components/switches/). We have tried on an iPhone 14 pro and iPhone 12 pro.
> 
> For this reason, my project avoids this problem by not setting a color on the VSwitch.
> 
> I would prefer to have the initial color, but would be happy if it does not slow down the toggle transition.
> 
> If that is not feasible, you must be able to disable the color settings.
> 
> Thank you for everything!

Thanks for reporting this problem.

I am sure @johnleider can advise whether we might need a separate issue to track this bug (I have not had time to investigate your report, so can't tell whether they are related regressions or not)

KaelWD:
We prefer to do a single snapshot per component, see the "Showcase" test in other files for examples.

floroz:
> We prefer to do a single snapshot per component, see the "Showcase" test in other files for examples.

considering that the test has to include a change in state (switch from off to on), how would you include that in a single snapshot?

KaelWD (approved):

floroz:
Thank you @KaelWD for completing the work. 

I am sorry I couldn't find the time lately to follow up on the requested changes.

PR #16526, opened 2023-01-25
docs(VColorPalette): implement setup and Typescript

s-rigaud:
## Description

I tried to implement setup style and Typescript for ColorPalette docs as many other docs are using them.
I'm not sure about the **colors** import from vuetify, can you verify it's okay please ?

### DIFF
--- a/packages/docs/src/components/features/ColorPalette.vue
+++ b/packages/docs/src/components/features/ColorPalette.vue
@@ -66,54 +66,39 @@
   </section>
 </template>
 
-<script>
+<script setup lang="ts">
   // Utilities
   import kebabCase from 'lodash/kebabCase'
-  import colors from 'vuetify/lib/util/colors'
+  import { colors as vuetifyColors } from 'vuetify/lib/util/colors'
+  import { computed } from 'vue'
 
-  export default {
-    name: 'ColorPalette',
+  type Colors = Record<string, Record<string, string>>
 
-    data: () => ({
-      colors,
-      search: '',
-    }),
+  const colors: Colors = vuetifyColors
+  const search = ""
 
-    computed: {
-      computedColors () {
-        const colors = {}
-        const search = this.search.toLowerCase()
+  const computedColors: Colors = computed(() => {
+    const matchingColors: Colors = {}
+    const colorSearch = search.toLowerCase()
 
-        Object.keys(this.colors).forEach(key => {
-          const kebabKey = kebabCase(key).toLowerCase()
+    Object.keys(colors).forEach(key => {
+      const kebabKey = kebabCase(key).toLowerCase()
 
-          if (kebabKey.indexOf(search) > -1) {
-            colors[kebabKey] = this.colors[key]
-          }
-        })
+      if (kebabKey.indexOf(colorSearch) > -1) {
+        matchingColors[kebabKey] = colors[key]
+      }
+    })
 
-        return colors
-      },
-    },
+    return colors
+  })
 
-    methods: {
-      convertToClass (base, variant) {
-        if (variant === 'base') return base
+  function convertToClass (base: string, variant: string): string {
+    if (variant === 'base') return base
 
-        const lastChar = variant.at(-1)
+    const lastChar = variant.at(-1)
 
-        if (isNaN(Number(lastChar))) return variant
+    if (isNaN(Number(lastChar))) return variant
 
-        return `${base}-${variant.slice(0, -1)}-${lastChar}`
-      },
-      getColorClass (key) {
-        if (['white', 'transparent'].includes(key) ||
-          key.indexOf('light') > -1 ||
-          key.indexOf('accent') > -1
-        ) return 'black--text'
-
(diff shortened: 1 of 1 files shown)


s-rigaud:
Seems like I failed typehinting properly according to CI pipeline, might need some help 😓

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
-      search: '',
-    }),
+  const colors: Colors = vuetifyColors
+  const search = ""
 
-    computed: {
-      computedColors () {
-        const colors = {}
-        const search = this.search.toLowerCase()
+  const computedColors: Colors = computed(() => {
```
floroz, line 80:
```suggestion
  const colors = computed((): Colors  => {
```

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
 
-  export default {
-    name: 'ColorPalette',
+  type Colors = Record<string, Record<string, string>>
 
-    data: () => ({
-      colors,
-      search: '',
-    }),
+  const colors: Colors = vuetifyColors
```
floroz, line 77:
```suggestion
```

This assignment is redundant

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
+<script setup lang="ts">
   // Utilities
   import kebabCase from 'lodash/kebabCase'
-  import colors from 'vuetify/lib/util/colors'
+  import { colors as vuetifyColors } from 'vuetify/lib/util/colors'
+  import { computed } from 'vue'
 
-  export default {
-    name: 'ColorPalette',
+  type Colors = Record<string, Record<string, string>>
```
floroz, line 75: My feeling is that `Colors` is not a type that is exclusive to the `ColorPalette` component, therefore perhaps it should live elsewhere?

Either in `'vuetify/lib/util/colors'` or `colors` should already be typed, what do you think?

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
 
-        const lastChar = variant.at(-1)
+    const lastChar = variant.at(-1)
 
-        if (isNaN(Number(lastChar))) return variant
+    if (isNaN(Number(lastChar))) return variant
 
-        return `${base}-${variant.slice(0, -1)}-${lastChar}`
-      },
-      getColorClass (key) {
```
floroz, line 109: was this method unused? I don't see it being ported to the `script setup`

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
-      computedColors () {
-        const colors = {}
-        const search = this.search.toLowerCase()
+  const computedColors: Colors = computed(() => {
+    const matchingColors: Colors = {}
+    const colorSearch = search.toLowerCase()
 
-        Object.keys(this.colors).forEach(key => {
-          const kebabKey = kebabCase(key).toLowerCase()
+    Object.keys(colors).forEach(key => {
```
floroz, line 84:
```suggestion
    Object.keys(vuetifyColors).forEach(key => {
```

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
-  export default {
-    name: 'ColorPalette',
+  type Colors = Record<string, Record<string, string>>
 
-    data: () => ({
-      colors,
-      search: '',
-    }),
+  const colors: Colors = vuetifyColors
+  const search = ""
```
floroz, line 78: Search now is a static assignment without any reactivity (also, since its a string primitive declared as const, is also an immutable value).

```suggestion
  const search = ref("")
```

I would suggest some further reading about [Vue Reactivity Fundamentals](https://vuejs.org/guide/essentials/reactivity-fundamentals.html)

packages/docs/src/components/features/ColorPalette.vue
```diff
@@ -66,54 +66,39 @@
...
 
-    computed: {
-      computedColors () {
-        const colors = {}
-        const search = this.search.toLowerCase()
+  const computedColors: Colors = computed(() => {
+    const matchingColors: Colors = {}
+    const colorSearch = search.toLowerCase()
```
floroz, line 82:
```suggestion
    const colorSearch = search.value.toLowerCase()
```

floroz (commented):
Not a team member, just leaving my thoughts here :)

There's a bug introduced as the previously reactive state `search` was declared as `const` assignment, therefore losing reactivity and becoming immutable.

`search` is used in `v-model`, therefore you might want to look at the Reactivity Fundamentals I've linked in the PR and refactor that part.

I also think there are some considerations around type declarations in this SFC, but a team member could be best place to advise on that 👍

PR #16546, opened 2023-01-29
fix(VSelect): return-object with plain items

### LINKED ISSUE #16512: [Bug Report][3.1.2] VSelect with 'multiple' and 'return-object' has inverted selection unless 'items' are wrapped in ref
### Environment
**Vuetify Version:** 3.1.2
**Vue Version:** 3.2.45
**Browsers:** Chrome 109.0.0.0
**OS:** Windows 10

### Steps to reproduce
Create a v-select with attributes: "multiple" and "return-object", binding a plain array of items with a v-model containing a matching array of items.

### Expected Behavior
The checkbox in the multi-select dropdown should be selected

### Actual Behavior
The 'checked' state of the checkbox in the multi-select dropdown is an inversion of the actual selected state.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVVMGO2jAQ/ZWRL6ESJOq2vSC22h572VZV1R7wHkI8LF4S27Id2Arl3zu2CQ2gAnuJMjOe956fx57v2KZFL5d/8hfHpkw2RlsPO6gslh5/pRp0sLS6gWy/NuOKK3yNSyutnId9Ae6PG0c7rgD8ChucEqjAZdnW/meKM1HadQbdmKvuHVdszL4YkxMS6Zh5bExNQJ8DwGwzKY2JvzFoSqn2EcBvbddSPU/hK/U42FpaigKkAotL2Eq/gqwhVmlqzKBUAjKLvrVqohcvWHnaTAIiYIc1ZYAYtMD6nrOUQMEZTGXAp9ymrFt0lOlRiWiAB8VB2qP2/+R9r0k1RJBbVMHIIUK1wmq90K9hO8JqI/RWkVfXBUe2c9V9+qr0W2TfrONR/4g03yLLdVW9jFkxOOsQpCmYFYPpoNBVVhoPjjhC+TDDdP5j+pSVlxscznBUzlUa3V4pzS41jOY7L30dxtOj89kYokQKLYqseyLzj9uSP8e9b2k99qYHii2ho+9JRl3WeMozMJcaLzcFV5OP5CBdxWQiuW/oYdCKrmS8ynxfoPGnC51OiU6xxRBztvLeuGlRtMqsn/NKN8UD1QrbKi8bnAjdPHzI7/KPnwohnR/mc3TNZGH11qElRs7oVTiAh6fkfwT7MuG+z+8SbP+iEeQZVEH4G7QTi0qgRXtR98naI+0ntTP9gbOjl428rKVauxMbKxctnJPxHevGLEHT4rgP9vQXWxTsBg==)

### Other comments
Probably related to https://github.com/vuetifyjs/vuetify/issues/16388.

nekosaur:
closes #16512

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-select v-model="selected" :items="values" multiple return-object />

      <v-select v-model="selectedPlain" :items="valuesPlain" multiple return-object />

      <v-combobox
        v-model="foo"
        v-model:search="search"
        :items="items"
      />

      <pre>{{ foo }}</pre>
      <pre>{{ search }}</pre>
    </v-main>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selected = ref([{title: 'test', value: 'red'}])
  const selectedPlain = ref([{title:'test', value: 'red'}])
  const values = ref([{title: 'test', value:'red'}])
  const valuesPlain = [{title: 'test', value:'red'}]
  const items = [
    { title: 'Item 1', value: 'item1' },
    { title: 'Item 2', value: 'item2' },
    { title: 'Item 3', value: 'item3' },
    { title: 'Item 4', value: 'item4' },
  ]
  const foo = ref()
  const search = ref()
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -166,7 +166,7 @@ export const VCombobox = genericComponent<new <
 
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => props.valueComparator(item.title, v.title)) || v
       })
     })
 
--- a/packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
+++ b/packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
@@ -263,4 +263,77 @@ describe('VSelect', () => {
       cy.get('.v-overlay__content .v-list-item .v-list-item-title').eq(1).should('have.text', 'Item 4')
     })
   })
+
+  describe('return-object', () => {
+    it('should return whole object', () => {
+      const items = ref([
+        {
+          title: 'Item 1',
+          value: 1,
+        },
+        {
+          title: 'Item 2',
+          value: 2,
+        },
+        {
+          title: 'Item 3',
+          value: 3,
+        },
+      ])
+
+      const selected = ref([])
+
+      cy.mount(() => (
+        <div class="ma-10">
+          <VSelect label="Select" v-model={selected.value} items={items.value} multiple returnObject />
+        </div>
+      ))
+        .get('.v-select')
+        .click()
+        .get('.v-list-item')
+        .eq(1)
+        .click()
+        .then(() => {
+          expect(selected.value).deep.equal([{ title: 'Item 2', value: 2 }])
+        })
+        .find('.v-checkbox-btn input')
+        .should('be.checked')
+    })
+
+    // https://github.com/vuetifyjs/vuetify/issues/16512
+    it('should work with plain items array', () => {
+      const items = [
+        {
+          title: 'Item 1',
+          value: 1,
+        },
+        {
+          title: 'Item 2',
+          value: 2,
+        },
+        {
+          title: 'Item 3',
+          value: 3,
+        },
+      ]
+
+      const selected = ref([])
+
+      cy.mount(() => (
+        <div class="ma-10">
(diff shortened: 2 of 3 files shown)


packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -166,7 +166,7 @@ export const VCombobox = genericComponent<new <
 
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => props.valueComparator(item.title, v.title)) || v
```
KaelWD, line 169: How is this related?

KaelWD:
What you have here is probably better anyway, but the underlying cause seems to be proxiedModel returning a reactive object on line 49, while nested@230 is calling map.get() with the raw value.

packages/vuetify/src/components/VCombobox/VCombobox.tsx
nekosaur: Not strictly related, could make separate commit.

To see why, go back to `.value` then type `item1` in the combobox in playground, then focus somewhere else. See that it changes `item1` to `Item 1`

KaelWD: Hmm yeah I see what you mean, and it only does that because of the transformItem change. I don't know if I really like this change though, it's inconsistent with other components and means if you type `Item 1` then the list item shows as selected even though the model is actually a string instead of the item object (this might be ok though?? idk how people are actually using it)

johnleider (approved):

PR #16548, opened 2023-01-30
feat: Add click events for VListItem prepend/append icon/avatars

blalan05:
markup
```
<template>
  <v-app>
    <div class="ma-4">
      <v-list>
        <v-list-item prependIcon="mdi-vuetify" @click:prepend="clickIcon(1)" @click="clickItem(1)" appendAvatar="https://cdn.vuetifyjs.com/images/lists/1.jpg" @click:append="clickAvatar(1)">List Item 1</v-list-item>
        <v-list-item prependIcon="mdi-vuetify" @click:prepend="clickIcon(2)" @click="clickItem(2)" appendAvatar="https://cdn.vuetifyjs.com/images/lists/2.jpg" @click:append="clickAvatar(2)">List Item 2</v-list-item>
        <v-list-item prependIcon="mdi-vuetify" @click:prepend="clickIcon(3)" @click="clickItem(3)" appendAvatar="https://cdn.vuetifyjs.com/images/lists/3.jpg" @click:append="clickAvatar(3)">List Item 3</v-list-item>
      </v-list>
    </div>
  </v-app>
</template>

<script>
  export default {
    data: () => ({}),
    methods: {
      clickAvatar: (msg) => { console.log(`Avatar Clicked: ${msg}`)},
      clickIcon: (msg) => { console.log(`Icon Clicked: ${msg}`)},
      clickItem: (msg) => { console.log(`Item Clicked: ${msg}`)},
    },
  }
</script>
```


blalan05:
It does match VTextField, which as these handlers.

johnleider:
I think this should probably add these events as props similar to our input components:

https://github.com/vuetifyjs/vuetify/blob/97b22d2113dd2e1ae2ad162674aede4e79e92891/packages/vuetify/src/components/VField/VField.tsx#L79
https://github.com/vuetifyjs/vuetify/blob/97b22d2113dd2e1ae2ad162674aede4e79e92891/packages/vuetify/src/components/VInput/VInput.tsx#L50

blalan05:
Updated to use props, and match VInputs use of InputIcon

johnleider:
Ideally we should abstract this functionality into something reusable to avoid duplication.

J-Sek:
I think I'd keep the only path to bind events via slot. Anything interactive landing in list item rendered as `<a>` is incorrect, while items with `role="option"` could trigger a11y tools on ARIA violation. `.stop` / `.prevent` (or lack thereof) are also decisions with consequences.

Since this PR stalled for 3.5y anyway, the call to keep slot-click events on the form inputs only is a bit easier. Given the trajectiory, v5 would probably keep them, but ideally only `click:clear` would survive (opinion).

Issue #15515, opened 2022-07-26
[Bug Report][3.0.0] VBtn size options does not change icon size


davidramosweb:
### Environment
**Vuetify Version:** 3.0.0-beta.5
**Vue Version:** 3.2.37
**Browsers:** Safari 15.5
**OS:** Mac OS 10.15.7

### Steps to reproduce
<v-btn
    color="secondary"
    icon="mdi-television"
    size="x-small"
></v-btn>

### Expected Behavior
The class `v-icon--size-x-small` should be added to the icon

### Actual Behavior
The icon has `v-icon--size-default`class

### Reproduction Link
<a href="https://codepen.io/davidramosweb/pen/jOzLRXw?editors=101" target="_blank">https://codepen.io/davidramosweb/pen/jOzLRXw?editors=101</a>

KaelWD:
#15516 was reverted

@johnleider I think it did just have a `font-size: 1.5rem` (or something) override at one point.

johnleider:
fixed by 569b75d4

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

Issue #15746, opened 2022-09-07
[Bug Report][3.0.0-beta.10] size not working for v-rating


LeoTuet:
### Environment
**Vuetify Version:** 3.0.0-beta.10
**Vue Version:** 3.2.38
**Browsers:** Chrome 104.0.0.0
**OS:** Linux x86_64

### Steps to reproduce
Add the `size` prop to the `Rating` component and set it to any numeric string value.

### Expected Behavior
The stars of the `Rating` component adapt to the set size.

### Actual Behavior
The stars format to a fixed size with no spacing whatsoever.

### Reproduction Link
<a href="https://codepen.io/leotuet/pen/bGMVYqg?editors=101" target="_blank">https://codepen.io/leotuet/pen/bGMVYqg?editors=101</a>

KaelWD:
Buttons and chips have some pretty complex css to generate the correct sizes: https://github.com/vuetifyjs/vuetify/blob/7fe2e2671f44002d43384d74d3aaaf38610eb72c/packages/vuetify/src/components/VChip/_mixins.scss#L8-L85

I've made buttons work with just a basic width/height for now, but that needs to be refactored to be based off a single value to be able to set font size too.

KaelWD:
Follow #15515 for that.

PR #15516, opened 2022-07-26
fix(VBtn): add size prop to internal v-icon

davidramosweb:
Fixes #15515

## Description

fixes #15515 
## Motivation and Context

fixes #15515 
## How Has This Been Tested?

visually
## Markup:

<details>

```vue
// Paste your FULL Playground.vue here
```
</details>

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [x] My code follows the code style of this project.
- [ ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #16550, opened 2023-01-30
feat(VBtn): add iconSize prop

### LINKED ISSUE #16288: [Bug Report][3.0.5] VBtn size as number does not affect font size
### Environment
**Vuetify Version:** 3.0.5
**Vue Version:** 3.2.45
**OS:** Mac OS 10.15.7

### Steps to reproduce
Set the size property of the `v-rating` component to a numeric value. For example `<v-rating :size="100" />`.

### Expected Behavior
The size of the icons increases to match the specified size.

### Actual Behavior
The padding around the icon increases but the icon does not.

### Reproduction Link
https://play.vuetifyjs.com/#eNqNkMtqAyEUhl/l9GymhWac0m4aTKC7vkEXNQuZmEHwhpohJOTdozgZkpCbLvTjP+dTzv8Og2/Jj3N1vxY4RRpaL12EIOLazZmR2lkfYQderGAPK281VKm0YoaZ1poQQYcOZjl/rX6FUhb+rFfLl+qNGUqKLokSRKGd4lEkAqD9hLv8AqSVQHNpBkq8lH3uiWWn2PMoTTenZLyWSoCxh5Sm+wYIcitmDBX3nWD4jHAwHGEIL4WbyU3lgx9eCqeD8aNpzm3XPDkeR5ehDJWSk2Hj/h0/66+6wXJ+4+IApYmbBw==

### Other comments
https://github.com/vuetifyjs/vuetify/issues/15746 seems to be the same issue but it doesn't seem like it was resolved by the mentioned VBtn fix https://github.com/vuetifyjs/vuetify/issues/15515.

floroz:
## Description

Starting point to discuss the potential implementation of `iconSize` prop and its behaviour

# The problem

Currently, `<VIcon />` rendered from the `<VBtn />` do not receive any `size` prop, and therefore they don't scale accordingly to the users' needs.

One potential approach would be to automatically scale with the same sizes provided to the `<VBtn />`, but it would limit the ability for the user to customize the look for built-in icons.

The alternative approach is to add a new `iconSize` prop to the `<VBtn />` (and potentially replicate this approach to other components that leverage icons?).

Open to thoughts and considerations.

# Related issues

closes #16288

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.tsx
+++ b/packages/vuetify/src/components/VBtn/VBtn.tsx
@@ -191,7 +191,7 @@ export const VBtn = defineComponent({
               }}
             >
               <span class="v-btn__prepend">
-                { slots.prepend?.() ?? (<VIcon />) }
+                {slots.prepend?.() ?? (<VIcon size={ props.iconSize } />) }
               </span>
             </VDefaultsProvider>
           ) }
@@ -207,7 +207,7 @@ export const VBtn = defineComponent({
             >
               { slots.default?.() ?? (
                 hasIcon && (
-                  <VIcon key="icon" />
+                  <VIcon key="icon" size={ props.iconSize } />
                 )
               ) }
             </VDefaultsProvider>
@@ -223,7 +223,7 @@ export const VBtn = defineComponent({
               }}
             >
               <span class="v-btn__append">
-                { slots.append?.() ?? (<VIcon />) }
+                { slots.append?.() ?? (<VIcon size={ props.iconSize } />) }
               </span>
             </VDefaultsProvider>
           ) }
--- a/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
+++ b/packages/vuetify/src/components/VBtn/__tests__/VBtn.spec.cy.tsx
@@ -3,6 +3,7 @@
 import { VBtn } from '../VBtn'
 import { generate, gridOn } from '@/../cypress/templates'
 import { createRouter, createWebHistory } from 'vue-router'
+import { VIcon } from '@/components/VIcon'
 
 const anchor = {
   href: '#my-anchor',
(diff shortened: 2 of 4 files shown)


floroz:
Currently the implementation breaks the scaling of the button when the icon overflows:

Screenshot 2023-01-30 at 10 45 21

Before solving this issue, I prefer to validate that this approach could be valid/correct, since the Btn implementation has many dependencies and there might be quite a few things I am missing.

I would also like to hear thoughts about potentially using the same `props.size` provided to the `VBtn` and also to the `VIcon` and how it should behave.

floroz:
However, even using the `props.size` instead of `props.iconSize` causes issue in correctly generating the layouts

Screenshot 2023-01-30 at 10 52 56
Screenshot 2023-01-30 at 10 53 16

KaelWD:
That's why #15516 was reverted. 

Related: #15746, #15515

floroz:
> 


> That's why #15516 was reverted.
> 
> Related: #15746, #15515
> 
> Basically the button-sizes sass mixin has to be rewritten to accept arbitrary values from a prop instead of using $size-scales

Thanks, @KaelWD for the clarification! Is this something still available to work? I am not a SASS expert but can give it a try.

Do we have any capability for Cypress Stories to generate some Visual testing? 

This feature seems a good candidate for that.

floroz:
Oh, also to clarify further, @KaelWD this means that `props.size` should be sufficient to implement the feature and we don't need a `props.iconSize` for further customization?

KaelWD:
It's pretty complex with size+density so I don't have high hopes that you'll be able to solve it, you're always welcome to try though. 

> Do we have any capability for Cypress Stories to generate some Visual testing?

Yes, the last test ("Showcase") is a percy snapshot. I don't think it's publicly accessible though.

> we don't need a props.iconSize

Correct. Size presets (`size="xs"`) already affect font size too, it's just arbitrary pixel values that don't. The way you'd done it here isn't great anyway because multiple components use the size composable, and not all of them have icons.

floroz:
> It's pretty complex with size+density so I don't have high hopes that you'll be able to solve it, you're always welcome to try though.

Makes sense. It might become a bit of a rabbit hole to rewrite that mixin, I'll close this PR and pick something else 👍 
 
> Yes, the last test ("Showcase") is a percy snapshot. I don't think it's publicly accessible though.

Good to know!

PR #16582, opened 2023-02-03
docs(VTextField): update api descriptions

aleg001:
## Description

docs:  added missing descriptions in definitions of v-text-field. 
Used Vuetify 2 as main inspiration for the descriptions.


johnleider (approved):
Some of these may end up moving to `generic.json` but otherwise looks good.

PR #16607, opened 2023-02-07
feat: add support for Google Material Design font icon variants

june07:
## Description

[Google Material Icon](https://fonts.google.com/icons) have several variants available including: outlined, filled, rounded, sharp, and two-tone, however prior to this change Vuetify only supported the default.

This change makes the others available by setting a variant prop on the VIcon component. So...

```
<v-icon icon="fiber_new" variant="outlined"></v-icon>
```
 
image
vs the default
image

## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/composables/icons.tsx
+++ b/packages/vuetify/src/composables/icons.tsx
@@ -53,6 +53,7 @@ export interface IconProps {
   tag: string
   icon?: IconValue
   disabled?: Boolean
+  variant?: string
 }
 
 type IconComponent = JSXComponent<IconProps>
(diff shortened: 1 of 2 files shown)


johnleider:

june07:
Not sure why this was closed, or what the problem was with the code? As far as being a "proprietary property"?! I have no idea of any other icons sets that are setup similar to Google's Material Icons, but the fact remains that said icons set is broken without this addition. Tons of the Material icons simply and plainly do not work.

Again, I'm running into the issue where a bunch of the icons in the set DO NOT WORK with vuetify and thus one is required to fall back to plain html syntax:

<span class="material-symbols-outlined"> reset_image </span>

If there is some other magical way to use all of the icons in the Material Icon set... I would love to know what that is...

june07:
Here's what I get now:

`<v-icon icon="reset_image" />`
image

vs what it should be:

`<span class="material-symbols-outlined"> reset_image </span>`
image

And simply adding the variant would make this possible:

`<v-icon icon="reset_image" variant="outlined"></v-icon>`
image

PR #16619, opened 2023-02-08
feat(VAutocomplete): add auto-select-first prop

### LINKED ISSUE #16508: [Feature Request] [v3] restore v-autocomplete's auto-select-first property
### Problem to solve
`v-autocomplete`'s property `auto-select-first` was lost during upgrade from 2.x to 3.x. This feature allowed us to e.g. confirm item selection on enter key press, now we need to extra press down key to focus (select) the first item.

### Proposed solution
Restore said property. Thank you!

krausvo1:
## Description

Resolves #16508.

Mainly allows user to select first item from menu with `Enter`.

Autoselected item is not respected when user presses `ArrowDown` / `ArrowUp` in a way users might expect, starting position is still the input field. The same happens when some items is selected, this just follows current behavior of autcomplete.

The prop's effect is disabled when `multiple` prop is used since it incorrectly marks first item as selected (checkbox is ticked) and I don't know how to workaround that.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <h1>Value: {{ value }}</h1>
      <v-autocomplete
        v-model="value"
        :items="[
          { title: 'foo', value: 1 },
          { title: 'bar', value: 2 },
          { title: 'baz', value: 3 }
        ]"
        auto-select-first
      />
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const value = ref()
      return {
        value,
      }
    },
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -65,6 +65,7 @@ export const VAutocomplete = genericComponent<new <
   V extends Value<T, ReturnObject, Multiple> = Value<T, ReturnObject, Multiple>
 >() => {
   $props: {
+    autoSelectFirst: Boolean
     items?: readonly T[]
     returnObject?: ReturnObject
     multiple?: Multiple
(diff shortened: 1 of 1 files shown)


krausvo1:
Please let me know if this approach is acceptable, I can try to fix the failing check(s) afterwards.

KaelWD:
Duplicate of #17257

PR #16660, opened 2023-02-14
docs(location): update property description

marius-thogersen:
## Markup:

```vue

```


KaelWD:
Some components also allow "center"

johnleider:
Do we need to define the values it can be? It's already covered in the type definition.

marius-thogersen:
I was missing it in the documentation. Maybe, maybe not :)

PR #16667, opened 2023-02-15
fix(VSlider): respect decimals from min

### LINKED ISSUE #16634: [Bug Report][3.1.4] v-slider step wrong rounding/decimals
### Environment
**Vuetify Version:** 3.1.4
**Vue Version:** 3.2.41
**Browsers:** Firefox 109.0
**OS:** Windows 10

### Steps to reproduce
Problem when using v-slider with initial value=1.001, step=1, max=10.001, min=1.001
Move slider to see unexpected values.
Use plus and minus buttons to see expected values.

### Expected Behavior
Expected jumps: 2.001, 3.001, 4.001...

### Actual Behavior
Actual jumps: 2, 3, 4...

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqtVttu20YQ/ZUBgSZSLVFS4idVdt1bgDwYfUhRoAjzsCKH4trkLsFdSjIM/XtnL7xKTm2gNmxydmfOnDmzF359Dn4py3BfY7AONhqLMmcabyMBsNnPWVnaV2sUjAtvWVvlPMEKaEImmN9EQVxXFQr9N8trjAJYF1zQKP23Bjsagx2NoTSWZJlHFLSYhNoQgDhnSpFLyeZLijDZpF6XFZYokn7Efr7VorMB7uKcx48G3fL7HeMKC6IVBZ3X7XyzsIFdPYtB7UMyPjup8cbkn8Wl5FevSU4uDqMbanphfqzIa3h+7ot+OnXhnbOPi7T53Ww7QIB/ZA0xE6AQIcMKoazkNscCDhkKqBUXO/BtPnCdARdcc5bD3mS7WYXL5WoGtpn0NB1eLd2Yab2ddvlc7kj/cSwx1pjAQ12UCg6yzhPYInxwUR/d49o8wjAccv211rYRnk2eQyVrQeFMEaCkIogHJBjzguUKlASdYZPIuJs8lMPgW+ghfEttixnbc1kBJxBe1KYtFv8KGKWbw7bWWgrVw2jK+wyxLDl5P8m6ovcEISWgmOWxgTFqWuWURSoYzZPgpVQkacqP1tmG2sqMe1qLWHMpLiT7i6pr2sViXbM8fzKcM3kwIMConf3Moi62pJxM4X0j0vsLsPdPoOrdDpVJa/C0hDjD+LGTlv6k5Uodd5VwYZ9xJqVCq/uWWwyfNTxfl2aBtyeKMdxZM9gKZKq44qWmd16UstLwTDRSLvA3SbaghQ8nSCtZQBTQGRYFP0UCj9aT/Fid22fff/JsEgpW0O6JzNkXBTMzkjDNJlOwswAV6roSjQXQ32VrcAu/mTPrjsZam9QYu5BQNOT2RjPWyLmGpRs6EXf6bx0KWgC07DpCOuMqbDtw4+ymvUiHnJ2YTPsYqDOZEL6HuODdK5BWtAZd8aLA5IvZST6HqS7U8gtNid1kGhoXn+Ys7p7WgQ8jEV4dRauoiWLH70QNow2xphIT31EPuYjzOkE1iYIwCqbdJv954Jaj2NGhNh/FJnj8M/WhNLnqwqlVoxKozHMOJMMrKBivMQMX+SYC7HiBAMn4AoEuPfmcpbdxL6XvJ2+efp/cM52Z5k36TbG3QM/omDYtdeuUSslZUU7s0WiDqJKlDaCXVW+RjrOR68xbXExoZObO1+k4gT1P7eZ1Wc4WvmVAB3ezsSyhTndHrbObJX42RJ83fmS82GWaKtS97QE/dBts5CvwYMnCTYdvy7R1TCYN27lHncKiw5rCj51Bl5ZzoQwjFUtWKfyUS6Z7hbZiNhRmbWG0HeUnfqRDaXAWtStrrPno86t/2PAUHEj/WIXbTpp37/r1tHFe5UGUF7TX4XMfWuAtWiu1/1Ya0m0/2P6b7qZV5v+me9VDu8iWjJMpZLNor8dgFrgbki7VMnxQUtD3vGUS+QkVBe1V4C9LcwNmWpdqvVjUonzchbEsFnc0t6jo/uEFzhNZ3H0MP4TXq0XCle6Ph6iK+baSB0UX/APBeyUtuObp00sJ/DThrsJrB+uHDOQZ1ILw91jNSSDqEFbf5T3yHXAfzZ3xb6QlLXMuHtVIxlhZCb821N5UmLkPLYCJ/maSnGaBI0fpbEDw7V//LFPr)

### Other comments
I copied your code for calculating values and made proposal fix for your rounding function which you can see in the methods.
The problem actually is how you are calculating number of 'decimals'.
My suggestion is to check decimals also for max, min and initial value and choose the biggest number of decimals.

yuwu9145:
fixes #16634

## Description

fixes #16634

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-slider v-model="currentValue" :min="min" :max="max" :step="step">
        <template class="pa-0" v-slot:prepend>
        <v-btn
          @click="sliderDecrement"
        >-</v-btn>
      </template>
      <template v-slot:append>
        <v-btn
          @click="sliderIncrement"
        >+</v-btn>
      </template>
      </v-slider>
      <p>
        Value: {{currentValue}}
      </p>
      <p>
  			<br>
        You can see here problem when using slider with initial value=1.001, step=1, max=10.001, min=1.001<br>
				Expected jumps would be 2.001, 3.001, 4.001...<br>
        But v-slider will round based on step decimals so the jumps will be 2, 3, 4..<br><br>
        Expected behavior is simulated on + and - buttons.<br><br>
				I copied your code for calculating values and made proposal fix for your rounding function.<br><br>
				The problem actually is how you are calculating number of 'decimals'.<br><br>
				My suggestion is to check decimals also for max and min and choose the biggest number.
      </p>
    </v-main>
  </v-app>
</template>

<script>
import { defineComponent } from "vue";
  name: "App",
  data() {
    return {
      currentValue: 1.001,
      step: 1,
      min: 1.001,
      max: 10.001,
      decimals: 0
    };
  },
  mounted() {
    this.decimals = this.calculateDecimals();
  },
  methods: {
    calculateDecimals() {
      const trimmedStep = this.step.toString().trim();
      const trimmedMin = this.min.toString().trim();
      const trimmedMax = this.max.toString().trim();
      
      const stepDecimals = trimmedStep.includes(".")
        ? trimmedStep.length - trimmedStep.indexOf(".") - 1
        : 0;
      const minDecimals = trimmedMin.includes(".")
        ? trimmedMin.length - trimmedMin.indexOf(".") - 1
        : 0;
      const maxDecimals = trimmedMax.includes(".")
      ? trimmedMax.length - trimmedMax.indexOf(".") - 1
      : 0;
      
      return Math.max(stepDecimals, minDecimals, maxDecimals);
    },
    clamp(value, min = 0, max = 1) {
      return Math.max(min, Math.min(max, value));
    },
    roundValue(value) {
      const clamped = this.clamp(
        value,
        this.min,
        this.max
      );
      const offset = this.min % this.step;
      const newValue =
        Math.round((clamped - offset) / this.step) * this.step + offset;

      return parseFloat(
        Math.min(newValue, this.max).toFixed(this.decimals)
      );
    },
    sliderDecrement() {
      if (this.currentValue > this.min && this.step) {
        this.currentValue = this.roundValue(this.currentValue - this.step);
      }
    },
    sliderIncrement() {
      if (this.currentValue < this.max && this.step) {
        this.currentValue = this.roundValue(this.currentValue + this.step)
      }
    },
  }
});
</script>
```


yuwu9145:
in js, `3+1.001===4.0009999999999994`.

Should we defined a default max decimals for this case? E.g. if `maxDecimals=3`, it will equal to 4.001

johnleider:
> in js, `3+1.001===4.0009999999999994`.
> 
> Should we defined a default max decimals for this case? E.g. if `maxDecimals=3`, it will equal to 4.001

Could @KaelWD or @nekosaur chime in on this?

johnleider:
This is still having issues with rounding.

yuwu9145:
Solution has been refactored:

In rounding logic, decimals was only calculated based on step, it should also consider the decimals from min

packages/vuetify/src/components/VSlider/slider.ts
```diff
@@ -158,10 +158,10 @@ export const useSlider = ({
   const max = computed(() => parseFloat(props.max))
   const step = computed(() => props.step > 0 ? parseFloat(props.step) : 0)
```
johnleider, line 159: I'd change this to:

```ts
const step = computed(() => parseFloat(props.step > 0 ? props.step : 0))
```

So that it always returns a string.

packages/vuetify/src/components/VSlider/slider.ts
```diff
@@ -158,10 +158,10 @@ export const useSlider = ({
...
   const decimals = computed(() => {
-    const trimmedStep = step.value.toString().trim()
-    return trimmedStep.includes('.')
-      ? (trimmedStep.length - trimmedStep.indexOf('.') - 1)
-      : 0
+    const stepDecimals = getDecimals(step.value.toString())
```
johnleider, line 161: With the above comment you don't have to call **.toString()**

packages/vuetify/src/components/VSlider/slider.ts
```diff
@@ -158,10 +158,10 @@ export const useSlider = ({
...
+    const minDecimals = getDecimals(min.value.toString())
```
johnleider, line 162: min is always a string

packages/vuetify/src/components/VSlider/slider.ts
```diff
@@ -158,10 +158,10 @@ export const useSlider = ({
...
+
+    return Math.max(stepDecimals, minDecimals)
```
johnleider, line 164: this could probably be:

```ts
return Math.max(
  getDecimals(step.value),
  getDecimals(min.value)
)
```

johnleider (changes requested):

packages/vuetify/src/components/VSlider/slider.ts
yuwu9145: It's probably better to keep it as `number` type, because there are other places using it to do arithmetic calculation.

Instead, I put `toString` as part of `getDecimals` function in [new commit](43e1a317a35f2a5807469beae2dc5ad99bd5a0b4)

image

packages/vuetify/src/components/VSlider/slider.ts
yuwu9145: Resolved, thanks

johnleider (approved):

packages/vuetify/src/components/VSlider/__tests__/VSlider.spec.cy.tsx
```diff
@@ -193,4 +194,42 @@ describe('VSlider', () => {
       </VApp>
     ))
   })
+
+  // #16634
```
nekosaur, line 198: Use link instead of just issue number

packages/vuetify/src/components/VSlider/__tests__/VSlider.spec.cy.tsx
```diff
@@ -193,4 +194,42 @@ describe('VSlider', () => {
...
+  describe('rouding & decimals', () => {
```
nekosaur, line 199: No need to put it in separate describe

packages/vuetify/src/components/VSlider/__tests__/VSlider.spec.cy.tsx
```diff
@@ -193,4 +194,42 @@ describe('VSlider', () => {
...
+    it('should respect the decimals from both step and min', () => {
+      const currentValue = ref(1.001)
+
+      cy.mount(() => (
+        <VApp>
```
nekosaur, line 204: Use `<Application>` from cypress/templates

packages/vuetify/src/components/VSlider/__tests__/VSlider.spec.cy.tsx
```diff
@@ -193,4 +194,42 @@ describe('VSlider', () => {
...
+        .trigger('mousemove', 190, 0, { force: true }) // move to fifth step
+        .should(_ => {
+          expect(currentValue.value).equal(6.0051)
+        })
+        .trigger('mousemove', 360, 0, { force: true }) // move to the final step
+        .should(_ => {
+          expect(currentValue.value).equal(10)
+        })
+        .trigger('mouseup')
+    })
```
nekosaur, line 233: This is just a suggestion, but you could possibly rewrite this without the ref and v-model. Instead you can check `.emitted('.v-slider', 'update:modelValue').should('deep.equal', ...)` at the end.

nekosaur (changes requested):

packages/vuetify/src/components/VSlider/__tests__/VSlider.spec.cy.tsx
yuwu9145: Thanks for the advice. It's much cleaner now.

KaelWD (approved):

PR #16669, opened 2023-02-15
refactor(transitions): move styling to component implementation

johnleider:
## Motivation and Context
transition and animation styles are always included in the final bundle

image


KaelWD (changes requested):
This potentially prevents transition classes from being loaded if no transition components are used

PR #16691, opened 2023-02-16
refactor(VPicker): implement date calculation locally

johnleider:
## Motivation and Context

I'd like to explore an interface where a single component does day / date / month / year calculation and provides it through slots. This would provide a solid baseline for `v-calendar`, `v-date-picker`, and other implementations that involve date related data; `v-date-time-picker` maybe?

This would be a step towards homogenizing usage of the date composable.

## Markup:
<details>

```vue
<template>
  <div class="ma-4 pa-4">
    <v-picker-alt>
      <template #day="props">
        <v-sheet class="pa-4 ma-2" border>
          <pre>{{ JSON.stringify(props, null, 2) }}</pre>
        </v-sheet>
      </template>

      <template #week="props">
        <v-sheet class="pa-4 ma-2" border>
          <pre>{{ JSON.stringify(props, null, 2) }}</pre>
        </v-sheet>
      </template>
    </v-picker-alt>
  </div>
</template>

<script setup>
  import { ref } from 'vue'
</script>

```
</details>


nekosaur:
yep something like this is what I had planned. the built in adapter is not very locale friendly but that can be a later problem

PR #16721, opened 2023-02-20
fix(theme): support vueuse/head v1.0

KaelWD:
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-btn color="primary" @click="toggle">Toggle theme</v-btn>
    </v-container>
  </v-app>
</template>

<script setup>

  const theme = useTheme()

  function toggle () {
    theme.themes.value.light.colors.primary = theme.themes.value.light.colors.primary === '#1976d2' ? '#ff0000' : '#1976d2'
  }
</script>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,5 +82,8 @@
     "vue-analytics": "^5.16.1",
     "vue-router": "^4.1.6",
     "yargs": "^17.6.2"
+  },
+  "resolutions": {
+    "hookable": "npm:strict-hookable@5.4.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "@typescript-eslint/eslint-plugin": "^5.51.0",
     "@typescript-eslint/parser": "^5.51.0",
     "@vue/compiler-sfc": "^3.2.47",
+    "@vueuse/head": "^1.0.26",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^28.1.3",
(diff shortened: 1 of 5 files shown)


harlan-zw:
Not sure why the computed isn't triggered with watchEffects but you can make use of the API returned by `push`.

```ts
const getHead = () => ({
  style: [
    {
      children: styles.value,
      id: 'vuetify-theme-stylesheet',
      nonce: parsedOptions.cspNonce || false as never,
    },
  ],
})
// vueuse/head >= v1
if (head.push) {
  const entry = head.push(getHead)
  watch(styles, () => { entry.patch(getHead) })
} else {
  // vueuse/head < v1
  if (IN_BROWSER) {
    head.addHeadObjs(computed(getHead))
    watchEffect(() => head.updateDOM())
  } else {
    head.addHeadObjs(getHead())
  }
}
```

That seems to work but I only tested it with the playground code you posted

harlan-zw:
Oh wait I know, because all of the reactivity is handled by the `useHead` function itself, these lower-level APIs are language agnostic, so they don't handle any of the reactivity for you.

See https://github.com/unjs/unhead/blob/main/packages/vue/src/runtime/composables/useHead/clientUseHead.ts

KaelWD:
Makes sense, thanks. The unhead docs are a bit light on internals.

KaelWD:
This was a breaking change: https://github.com/harlan-zw/zhead/commit/20884dc

PR #16828, opened 2023-03-03
fix(display): add hydrate checks for getPlatform

### LINKED ISSUE #16825: [Bug Report][3.1.4] useDisplay v-if Hydration node mismatch
### Environment
**Vuetify Version:** 3.1.4
**Vue Version:** 3.2.47
**Browsers:** Chrome 110.0.0.0
**OS:** Linux x86_64

### Steps to reproduce
1.  `import { useDisplay } from "vuetify";`
2. `const { mobile } = useDisplay();`
3. `<div v-if="mobile">Whatever markup</div>`
4. Load page with SSR with a reduced screen (mobile) so that on the server `mobile == false` while on the client `mobile == true`

### Expected Behavior
While doing SSR, `mobile` should be `false`, but when hydrating, the value change should be taken in account and generate a re-rendering.

### Actual Behavior
I get a 

```
chunk-3NMN3MUW.js?v=ca98b82d:1381 [Vue warn]: Hydration node mismatch:
- Client vnode: Symbol(Comment) 
- Server rendered DOM: <!--[--> (start of fragment) 
  at <VDefaultsProvider defaults= {VTabs: {…}} > 
  at <VToolbar ref=Ref< undefined > class="v-app-bar top-screen" style= {top: '0px', zIndex: 1004, transform: 'translateY(0%)', position: 'fixed', height: undefined, …}  ... > 
  at <VAppBar color="white" class="top-screen" flat=true  ... > 
  at <Header top-transparent=true > 
  at <PageWrapper no-container="" > 
  at <Index onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< undefined > > 
  at <RouteProvider key="/" routeProps= {Component: {…}, route: {…}} pageKey="/"  ... > 
  at <FragmentWrapper > 
  at <RouterView name=undefined route=undefined > 
  at <NuxtPage> 
  at <VApp> 
  at <Default > 
  at <LayoutLoader key="default" name="default" hasTransition=false > 
  at <FragmentWrapper > 
  at <NuxtLayout> 
  at <App key=2 > 
  at <NuxtRoot>
```
warning and then the app stops being reactive.
My guess is that the error comes from the `v-if="mobile"` which produces different markup in the frontend and the backend.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVUFtLwzAU/ishTwprgxcQZtUJPvgy/AHLHrL2VOOaC0laGaP/3ZNeRu1EsH1oz+W7nc2RPlubNjXQJc0CKFuJAI9cE5I1ibC2+x2LZCfc0Ji2Ei2aROZGE/yUD5wqs5MVcErYCGdzPKKVkBorHuKbjUrxeT0UTgSJhOCccU8nSXbyg4QD/sSO/xmbRMDS507aQDyEOo6lssYFciS1hxfpcfFAWlI6owineIMgywOn9xGKaTxu9kla8jCBXFxONqRfdytr0PWbBY2bpag8dCsZ6w2gNF3QXh1t2/TTG40HP0b3fBh4Tpek68Qeuok1px8hWL9krNZ2/57mRrEVzpirdZAKksKo1U16nd7esUL6MO2n4FWyc+bLg0NFThcT8j7q7wLDGHmv0oF2aCWV2PnIe8bHUKQBlzjQBThwf5qf7f4IMJudhYiaLdctHrSSeu9nt8x9d8fNaO3/6TqGCN9GlXZBe3eo1yHo9htpFhFo)

### Other comments
The error doesn't appear on the reproduction link because I don't know how to make SSR with it, but it is the same code :(

johnleider:
## Motivation and Context

fixes #16825

## Markup

Not sure the best way to test this but from @KaelWD's [comment](https://github.com/vuetifyjs/vuetify/issues/16825#issuecomment-1453810253), I assume this is the fix

### DIFF
--- a/packages/vuetify/src/composables/display.ts
+++ b/packages/vuetify/src/composables/display.ts
@@ -1,5 +1,5 @@
 // Utilities
-import { inject, reactive, ref, toRefs, watchEffect } from 'vue'
+import { inject, reactive, ref, shallowRef, toRefs, watchEffect } from 'vue'
 import { mergeDeep } from '@/util'
 
 // Globals
@@ -124,7 +124,6 @@ function getPlatform (isHydrate?: boolean): DisplayPlatform {
   const win = match(/win/i)
   const mac = match(/mac/i)
   const linux = match(/linux/i)
-  const ssr = match(/ssr/i)
 
   return {
     android,
@@ -139,7 +138,7 @@ function getPlatform (isHydrate?: boolean): DisplayPlatform {
     mac,
     linux,
     touch: SUPPORTS_TOUCH,
-    ssr: ssr || userAgent === 'ssr',
+    ssr: userAgent === 'ssr',
   }
 }
 
@@ -147,13 +146,17 @@ export function createDisplay (options?: DisplayOptions, ssr?: boolean): Display
   const { thresholds, mobileBreakpoint } = parseDisplayOptions(options)
 
   const height = ref(getClientHeight(ssr))
-  const platform = getPlatform(ssr)
+  const platform = shallowRef(getPlatform(ssr))
   const state = reactive({} as DisplayInstance)
   const width = ref(getClientWidth(ssr))
 
-  function update () {
+  function updateSize () {
     height.value = getClientHeight()
     width.value = getClientWidth()
+  }
+  function update () {
+    updateSize()
+    platform.value = getPlatform()
   }
 
   // eslint-disable-next-line max-statements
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/composables/display.ts
```diff
@@ -170,9 +175,9 @@ export function createDisplay (options?: DisplayOptions, ssr?: boolean): Display
       : xl ? 'xl'
       : 'xxl'
     const breakpointValue = typeof mobileBreakpoint === 'number' ? mobileBreakpoint : thresholds[mobileBreakpoint]
-    const mobile = !platform.ssr
+    const mobile = !platform.value.ssr
       ? width.value < breakpointValue
-      : platform.android || platform.ios || platform.opera
+      : platform.value.android || platform.value.ios || platform.value.opera
```
KaelWD, line 180: Unrelated but this seems kinda off. If ssr the other three will always be false, so really it's just checking width < mobileBreakpoint except that would be true for ssr

PR #16831, opened 2023-03-04
fix(VOverlay): overlay location in RTL mode

kingyyy:
Overlay location in RTL is showing outside the screen because locationStrategies.ts is not calling updateLocation when direction is changed.

rollback the locationStrategies.ts vuetifyjs/vuetify@a11138e to vuetifyjs/vuetify@33ab244


nekosaur:
1. Please include markup to reproduce the issue
2. The solution here is not a rollback

PR #16841, opened 2023-03-05
fix(VImg): pass alt prop to img element

### LINKED ISSUE #16203: [Bug Report][3.0.0] v-img alt text prop support is missing
### Environment
**Vuetify Version:** 3.0.0
**Vue Version:** 3.2.45
**OS:** Mac OS 10.15.7

### Steps to reproduce
1. Put a alt prop in v-img with "test alt test"
2. Hover over v-img once rendered
3. Inspect <img tag in dev tools

### Expected Behavior
If hovering, should show the "test alt test" alt text.
If inspecting, should show alt="test alt test"

### Actual Behavior
No alt text appears when hovering
The alt attribute appears when inspecting, but it is empty.

### Reproduction Link
[https://codepen.io/TwinHits/pen/VwdErmv](https://codepen.io/TwinHits/pen/VwdErmv)

### Other comments
This is important for 508 compliance. This appears to be an issue with line https://github.com/vuetifyjs/vuetify/blob/next/packages/vuetify/src/components/VImg/VImg.tsx#L195

ffxsam:
## Description

`<v-img>` `alt` prop was not being passed down to the underlying `` tag. It was only being set on a higher level as an `aria-label`, which is not sufficient for a11y requirements. Fixes #16203.


KaelWD:
Both NVDA and TalkBack read this the same as before (just fine) but ok.

PR #16843, opened 2023-03-06
chore(dev): check component types in playground.vue

yuwu9145:
## Description

A script command that can perform type check against `playground.vue`, steps:

- Make sure SFC `playground.vue` has `lang="ts"` in script tag
- Run `yarn dev` 
- Load localhost in browser (this will trigger "load on-demand" by [unplugin-vue-components](https://github.com/antfu/unplugin-vue-components)) and generates playground components type definitions into `components.d.ts`
- Cd to packages/vuetify
- Run `yarn dev:typecheck`
- It will perform check based on playground file via [vue-tsc](https://github.com/vuejs/language-tools/tree/master/packages/vue-tsc)

## Markup:

The command will produce error from Markup:
```bash
$ vue-tsc --noEmit --skipLibCheck --project ./tsconfig.dev.json
dev/Playground.vue:5:9 - error TS2322: Type '{ randomProp: string; }' is not assignable to type 'Partial<{ symbol: any; flat: boolean; block: boolean; stacked: boolean; active: boolean; ripple: boolean; rounded: string | number | boolean; density: Density; disabled: boolean; replace: boolean; exact: boolean; size: string | number; tag: string; variant: NonNullable<...>; }> & Omit<...>'.
  Object literal may only specify known properties, and 'randomProp' does not exist in type 'Partial<{ symbol: any; flat: boolean; block: boolean; stacked: boolean; active: boolean; ripple: boolean; rounded: string | number | boolean; density: Density; disabled: boolean; replace: boolean; exact: boolean; size: string | number; tag: string; variant: NonNullable<...>; }> & Omit<...>'.

5         randomProp="fail"
          ~~~~~~~~~~~~~~~~~



Found 1 error in dev/Playground.vue:5
```

```vue
<template>
  <v-app>
    <v-main>
      <v-btn
        randomProp="fail"
      ></v-btn> 
    </v-main>
  </v-app>
</template>

<script lang="ts"></script>

```


KaelWD:
You can probably enable dts in unplugin-vue-components instead of adding them all to GlobalComponents manually

packages/vuetify/vite.config.mjs
```diff
@@ -20,7 +20,7 @@ const components = files.filter(file => file.startsWith('src/labs') || !block.so
 const map = new Map(components.flatMap(file => {
   const src = readFileSync(file, { encoding: 'utf8' })
   const matches = src.matchAll(/export const (V\w+)|export { (V\w+) }/gm)
-  return Array.from(matches, m => [m[1] || m[2], file.replace('src/', '@/')])
+  return Array.from(matches, m => [m[1] || m[2], file.replace('src/', '@/').replace('.ts', '')])
```
yuwu9145, line 23: To assist `vue-tsc` work properly, have to remove .ts suffix for import paths, so unplugin-vue-components can yield

`VBtn: typeof import('@/components/VBtn/index')['VBtn']` 👍 
as opposed to 
`VBtn: typeof import('@/components/VBtn/index.ts')['VBtn']` 👎 

TS 4.x does not support .ts suffix in import path, and it looks like to be one of tasks in 5.x roadmap

Issue: https://github.com/microsoft/TypeScript/issues/37582
PR: https://github.com/microsoft/TypeScript/pull/51669

packages/vuetify/dev/components.d.ts
```diff
@@ -0,0 +1,78 @@
...
+    VThemeProvider: typeof import('@/components/VThemeProvider/index')['VThemeProvider']
+    VTimeline: typeof import('@/components/VTimeline/index')['VTimeline']
+    VTimelineItem: typeof import('@/components/VTimeline/index')['VTimelineItem']
+    VToolbar: typeof import('@/components/VToolbar/index')['VToolbar']
+    VToolbarTitle: typeof import('@/components/VToolbar/index')['VToolbarTitle']
+    VTooltip: typeof import('@/components/VTooltip/index')['VTooltip']
+  }
+}
+
+export {}
```
KaelWD, line 78: Just add this to gitignore, otherwise it'll change every time you modify the playground

KaelWD: also eslintignore

yuwu9145: this file is ignored by git so it doesn't bother lint command in CI. Assuming this is for IDE experience, then a `.eslintignore` file must be placed in [root directory](https://eslint.org/docs/latest/use/configure/ignore#the-eslintignore-file)

KaelWD (approved):

PR #16859, opened 2023-03-07
fix(VAutocomplete/VSelect/VCombobox): compare items using item value if passed

### LINKED ISSUE #16279: [Bug Report][3.1.4] V-Autocomplete pre selected option not working when using return-object prop
### Environment
**Vuetify Version:** 3.1.2
**Vue Version:** 3.2.45
**Browsers:** Firefox 108.0
**OS:** Ubuntu 22

### Steps to reproduce
Click v-autocomplete field and see there is no option pre selected while vmodel isnt empty and item value is good for the comparison.
Remove firstname prop from data object and see it working.

### Expected Behavior
Make pre selected option work properly when using filled vmodel and return object prop

### Actual Behavior
Pre selected option doest not work when using filled vmodel and return object prop when the vmodel has additional prop

### Reproduction Link
[Play vuetify](https://play.vuetifyjs.com/#eNqlU9uK2zAQ/RWhl20httntQsHNliz0qdAvWOVBsSaNsrohySkl5N87IyeunWwLpX4wnjlHZ2bOyC9H/hxCfeiBt3yZuqhDZglyHz4Lp23wMbMji7BlJ7aN3rI7pN4JJxxjnXcpMyWzZE9EefdC2SO9GAMrtWmZ4Hu/cyvloe68FXwxoFq17P78vdUxZSctEPsrsgUn4FTgazUro4bVXrpbvYe39L4R/7fg+v0n6n3oPIGBLoP68n8TjLqMCbdsBg/RPQwy2GBkBowYWx4qGchWOo4ByrtzNCRknz2WCAYyjHl8kOoVmCfBpx0PQ12eVmOthBRaxxzqdjqkaSLieqOr/GaPYlPAyE0p85yS7zT2rVifIM7lqFCVdTaAzOLRG/hBmp5wreag7U3WOOA497KZz32xB9OjP4VTnFs2E0f5gg8XFJmh3ifv8AqX3WHdAiTB28s2BceLS7Hgu5xDapumd+H1Oy21WSHWxN5lbaFS3q4+1A/148dG6ZSn+RqSrTbR/0BXsOJ4GYp41tuffypwhlH3vn4cZM+pCk1PpHuj12CRA8QqglMQaQ1/af6KOxvgCrsZolxi4U5oqNHuNV152aXiY/k1KP736YoCHV9TldOCD91hvXKCr38BLj5vwA==)

### Other comments
This is a side issue from https://github.com/vuetifyjs/vuetify/issues/16046 which is closed and fixed. I left a new comment but though it was better to create a new issue. 
This time it is maybe an expected behavior, but i think it could be a good functionality also depending on your goal.

vexleet:
fixes #16279

not sure if thats the best solution but I think it makes sense to what @ThomasCarbon said in the issue `if you set "item-value" props i expect it to rely on it in priority instead of anything else`. Thats why I made an if check to see whether itemValue is not default and to use the passed value as a priority when going through the items.

If this is ok as a solution I will add unit tests and fix formatting.

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -21,7 +21,7 @@ import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utility
 import { computed, mergeProps, nextTick, ref, watch } from 'vue'
-import { genericComponent, omit, useRender, wrapInArray } from '@/util'
+import {genericComponent, getPropertyFromItem, omit, useRender, wrapInArray} from '@/util';
 import { filterVTextFieldProps, makeVTextFieldProps } from '../VTextField/VTextField'
 
 // Types
@@ -132,7 +132,10 @@ export const VAutocomplete = genericComponent<new <
     const { filteredItems, getMatches } = useFilter(props, items, computed(() => isPristine.value ? undefined : search.value))
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => {
+          if(props.itemValue !== 'value') return props.valueComparator(getPropertyFromItem(item.value, props.itemValue), getPropertyFromItem(v.value, props.itemValue))
+          return props.valueComparator(item.value, v.value)
+        }) || v
       })
     })
 
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -22,7 +22,7 @@ import { useTextColor } from '@/composables/color'
 
 // Utility
 import { computed, mergeProps, nextTick, ref, watch } from 'vue'
-import { genericComponent, omit, useRender, wrapInArray } from '@/util'
+import {genericComponent, getPropertyFromItem, omit, useRender, wrapInArray} from '@/util';
 import { filterVTextFieldProps, makeVTextFieldProps } from '../VTextField/VTextField'
 
 // Types
@@ -174,7 +174,10 @@ export const VCombobox = genericComponent<new <
 
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => {
+          if(props.itemValue !== 'value') return props.valueComparator(getPropertyFromItem(item.value, props.itemValue), getPropertyFromItem(v.value, props.itemValue))
+          return props.valueComparator(item.value, v.value)
+        }) || v
       })
     })
 
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -22,7 +22,7 @@ import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utility
 import { computed, mergeProps, ref } from 'vue'
-import { deepEqual, genericComponent, omit, propsFactory, useRender, wrapInArray } from '@/util'
+import {deepEqual, genericComponent, getPropertyFromItem, omit, propsFactory, useRender, wrapInArray} from '@/util';
 
 // Types
 import type { VInputSlots } from '@/components/VInput/VInput'
@@ -134,7 +134,10 @@ export const VSelect = genericComponent<new <
     const form = useForm()
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => {
+          if(props.itemValue !== 'value') return props.valueComparator(getPropertyFromItem(item.value, props.itemValue), getPropertyFromItem(v.value, props.itemValue))
+          return props.valueComparator(item.value, v.value)
+        }) || v
       })
     })
     const selected = computed(() => selections.value.map(selection => selection.props.value))


ThomasCarbon:
Any update on the futur of this ? :'(

yuwu9145:
I wonder probably we should start use an primitive-only "internalValue" that holds `item-value` when using `return-object`. This would make value comparison easier.

packages/vuetify/src/components/VCombobox/__tests__/VCombobox.spec.cy.tsx
```diff
@@ -326,6 +327,65 @@ describe('VCombobox', () => {
...
+          },
+        ]
+      )
+
+      cy.mount(() => (
+        <VCombobox
+          v-model={selectedItems.value}
+          items={items.value}
+          multiple
+          chips
```
johnleider, line 365: is the usage of chips necessary for the test?

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -132,7 +132,15 @@ export const VAutocomplete = genericComponent<new <
     const { filteredItems, getMatches } = useFilter(props, items, computed(() => isPristine.value ? undefined : search.value))
     const selections = computed(() => {
       return model.value.map(v => {
-        return items.value.find(item => props.valueComparator(item.value, v.value)) || v
+        return items.value.find(item => {
+          if (props.itemValue !== 'value' && props.returnObject) {
```
KaelWD, line 136: It's only affected by return-object
```suggestion
          if (props.returnObject) {
```

packages/vuetify/src/components/VAutocomplete/__tests__/VAutocomplete.spec.cy.tsx
```diff
@@ -1,3 +1,4 @@
+/// <reference types="..\..\..\..\types\cypress" />
```
KaelWD, line 1: Windows is illegal

quatrochan:
This looks like a very important feature but PR is stale. Any updates on the future of this?

johnleider:
> This looks like a very important feature but PR is stale. Any updates on the future of this?

We are waiting on the author to address feedback.

packages/vuetify/src/components/VAutocomplete/__tests__/VAutocomplete.spec.cy.tsx
```diff
@@ -98,6 +99,63 @@ describe('VAutocomplete', () => {
...
+        id: 'item2',
+      },
+    ])
+
+    cy.mount(() => (
+      <VAutocomplete
+        v-model={selectedItems.value}
+        items={items.value}
+        multiple
+        chips
```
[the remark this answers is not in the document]
vexleet: nope, not really. i changed the tests to not use the chips prop.

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
vexleet: Something funny happens when we use `return-object` and the object doesnt have `value`

image

Code taken from first unit test in `VSelect`

```vue
<script setup>
import { ref } from 'vue'

const items = [
  { title: 'a' },
  { title: 'b' },
  { title: 'c' },
]
let model = ref([{ title: 'b' }])

</script>

<template>
  <v-app>
    <v-main>
      <VSelect
        :multiple="true"
        :return-object="true"
        :items="items"
        v-model="model"
        >
        <template #selection="{item, index}">
          {{item.raw.title.toUpperCase()}}
        </template>

      </VSelect>
    </v-main>
  </v-app>
</template>
```

vexleet: Saw that when using `return-object` and the object doesnt have `value` but only `title` it breaks.

I saw from Vuetify 2 the way the values are taken and compared to and used it here too

Would that make sense?

```diff
+ const getItemValue = (item: object) => getPropertyFromItem(item, props.itemValue, getItemTitle(item))
+ const getItemTitle = (item: object) => getPropertyFromItem(item, props.itemTitle, item.raw)
const selections = computed(() => {
  return model.value.map(v => {
    return items.value.find(item => {
      if (props.returnObject) {
        return props.valueComparator(
+        getItemValue(item.raw),
+         getItemValue(v.raw)
        )
      }
      return props.valueComparator(item.value, v.value)
    }) || v
  })
})
```

johnleider (approved):

vexleet:
I saw that `transformItem` already does what I have suggested [here](https://github.com/vuetifyjs/vuetify/pull/16859#discussion_r1261687732) and is the `title` property inside `item` so I guess it makes sense to use just that and all tests pass like that.

vexleet:
@johnleider the test failing now is that since `returnObject` is looking for the value inside the objects it finds `Item 3` and makes it the selected one rather than creating a new one. Should the test be updated to something like this:

```diff
cy.get('input').type('item3')
cy.should(() => {
  expect(model.value).to.equal('item3')
  expect(search.value).to.equal('item3')
})
cy.get('input')
  .should('have.value', 'item3')
  .blur()
cy.get('.v-combobox__selection')
-  .should('contain', 'item3')
+  .should('contain', 'Item 3')

// same test as what was before but writing item5 rather than item3
```

johnleider:
What is the reason for adding the fallback? That's what the `|| v` is for.

johnleider:
I think that the fallback should be removed.

johnleider (approved):

johnleider (approved):

oruman:
Changes break work with modelValue, what not Object. Before changes code below work fine, but now vModel MUST BE Object.
**Example:**
`<script setup>
import { ref } from 'vue'

const items = [
  { id: '1', title: 'a' },
  { id, '2', title: 'b' },
  { id: '3', title: 'c' },
]
let model = ref(2)

const comporator = (a, b) => {
  if (a === b) return true
  const c = (a ?? "").toString()
  const d = (b ?? "").toString()
  return c === d
}
</script>

<template>
  <v-app>
    <v-main>
      <VSelect
        v-model="model"
        :items="items"
        item-value="id"
        :value-comparator="comporator"
      />
    </v-main>
  </v-app>
</template>`

johnleider:
> Changes break work with modelValue, when modelValue is not Object. Before changes code below work fin....

Create a new issue please.

oruman:
> > Changes break work with modelValue, when modelValue is not Object. Before changes code below work fine...
> 
> Create a new issue please.

Done: https://github.com/vuetifyjs/vuetify/issues/17997

PR #16952, opened 2023-03-19
docs(VBottomNavigation): update example text

xuejianma:
Change deprecated "v-model" for v-bottom-navigation in doc to ":active".


johnleider (approved):

johnleider:
You had them backwards. **input-value** is for **v-model** which corresponds to the currently selected value. **active** controls whether or not the `v-bottom-navigation` is shown.

PR #16955, opened 2023-03-19
feat(validation composable): add validate immediately option

guilhermesdev:
## Description
Add a new option to `useValidation.props` called  `validateImmediately` that is optional and receives a boolean (defaults to `false`).

It has effect only if the `validateOn` option is `'blur'` or `'input`' since this behavior is not necessary when validation runs only on submit

## Motivation and Context

Without this feature there was no easy and intuitive way to achieve this behavior so we needed to write some workarounds like this:

```vue
<template>
  <v-app theme="dark">
    <v-container class="my-auto">
      <v-textarea
        v-model="text"
      	variant="outlined"
        label="Validate on input and immediately"
        :rules="validationRules"
        validate-on="input"
       @vue:mounted="forceVTextareaImmediateValidation"
      />
    </v-container>
  </v-app>
</template>

<script>
export default {
  name: 'Playground'
};
</script>

<script setup>
import { ref, nextTick } from 'vue';

const validationRules = [text => (text.length > 10 ? 'error' : true)];

const text = ref('a'.repeat(100));
  
async function forceVTextareaImmediateValidation() {
  text.value += ' ';
  await nextTick();
  text.value += text.value.trim();
}
</script>
```

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container class="my-auto">
      <v-row>
        <v-col>
          <v-textarea
            variant="outlined"
            label="Validate on input"
            v-model="text"
            :rules="validationRules"
            validate-on="input"
          />
        </v-col>

        <v-col>
          <v-textarea
            variant="outlined"
            label="Validate on input and immediately"
            v-model="text"
            :rules="validationRules"
            validate-on="input"
            validate-immediately
          />
        </v-col>
      </v-row>

      <v-row>
        <v-col>
          <v-textarea
            variant="outlined"
            label="Validate on blur"
            v-model="text"
            :rules="validationRules"
            validate-on="blur"
          />
        </v-col>
        <v-col>
          <v-textarea
            variant="outlined"
            label="Validate on blur and immediately"
            v-model="text"
            :rules="validationRules"
            validate-on="blur"
            validate-immediately
          />
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script>
export default {
  name: 'Playground'
};
</script>

<script setup>
import { ref } from 'vue';

const validationRules = [text => (text.length > 10 ? 'error' : true)];

const text = ref('a'.repeat(100));
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VRating/VRating.tsx
+++ b/packages/vuetify/src/components/VRating/VRating.tsx
@@ -186,7 +186,7 @@ export const VRating = genericComponent<VRatingSlots>()({
             type="radio"
             value={ value }
             checked={ normalizedValue.value === value }
-            tabindex={-1}
+            tabindex={ -1 }
             readonly={ props.readonly }
             disabled={ props.disabled }
           />
--- a/packages/vuetify/src/components/VRating/__tests__/VRating.spec.cy.tsx
+++ b/packages/vuetify/src/components/VRating/__tests__/VRating.spec.cy.tsx
@@ -100,7 +100,7 @@ describe('VRating', () => {
         <VRating>
           {{
             item: ({ value, rating }) => (
-              <VBtn variant="tonal" class="mx-1" color={rating === value ? 'primary' : undefined}>{ value }</VBtn>
+              <VBtn variant="tonal" class="mx-1" color={ rating === value ? 'primary' : undefined }>{ value }</VBtn>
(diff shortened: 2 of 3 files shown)


KaelWD:
Duplicate of #17111

PR #16956, opened 2023-03-19
chore: restructure directives folder

johnleider:


KaelWD:
By... duplicating all the files?

johnleider:
> By... duplicating all the files?

WDYM? It moves the directives out of folders with the exception of ripple and groups up their tests.

KaelWD:
I think you want ctrl+x not ctrl+c

PR #16974, opened 2023-03-22
docs(aliasing): update import code

santiagoaloi:
## Description

A minor tweak in the importing of components in `dev`resulting in slight performance benefit.
This only affects `dev` mode as in `prod` threeshaking will automatically optimize these kind of imports.

from:
```js
import { VBtn } from 'vuetify/components'
```

to:
```js
import { VBtn } from 'vuetify/components/VBtn'
```

### DIFF
--- a/packages/docs/src/pages/en/features/aliasing.md
+++ b/packages/docs/src/pages/en/features/aliasing.md
@@ -21,7 +21,7 @@ Aliasing allows you to use built-in Vuetify components as a baseline for your cu
 
 ```js { resource="src/plugins/vuetify.js"}
 import { createVuetify } from 'vuetify'
-import { VBtn } from 'vuetify/components'
+import { VBtn } from 'vuetify/components/VBtn'
 
 export default createVuetify({
   aliases: {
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/en/features/aliasing.md
```diff
@@ -31,13 +31,19 @@ export default createVuetify({
 })
 ```
 
+<alert type="info">
+
+Although treeshaking is automatically applied during production builds, it is advantageous to import components by specifying their full path in development mode. For instance, using 'vuetify/components/VBtn' instead of 'vuetify/components' ensures that the compiler loads fewer components, thus optimizing performance.
```
johnleider, line 36:
```suggestion
Although treeshaking is automatically applied during production builds, it is advantageous to import components by specifying their full path in development mode. For instance, using `vuetify/components/VBtn` instead of `vuetify/components` ensures that the compiler loads fewer components, thus optimizing performance.
```

johnleider (changes requested):
small tweak, otherwise lgtm

PR #16984, opened 2023-03-22
fix: check if CSS.supports exists

fireharp:
## Description
`CSS.support` checkup now fails with some environments. I suggest adding an extra check on its existence similar to the way it's done in many other projects.

Specifically, I was checking it with `nuxt-vitest` (it uses `happy-dom`). 

The issue it produces is similar to that one https://github.com/wojtekmaj/react-fit/issues/2

image

This fix made earlier seems to be not enough
https://github.com/vuetifyjs/vuetify/commit/59370b0ebd7ad9029c9ec14ad0c580435af6b967

## Markup:

```vue

```


KaelWD:
https://github.com/capricorn86/happy-dom/issues/352

johnleider (approved):

PR #16993, opened 2023-03-24
feat(VCardItem): add prepend and append click handlers

blalan05:
### Markup
``` vue
<template>
  <v-app>
    <v-main>
      <v-card width="400px">
        <v-card-item @click:prepend="doOne()" @click:append="doTwo()">
          <template #prepend>
            <v-sheet :color="one ? 'primary': null">Prepend</v-sheet>
          </template>
          <template #default>
            Default
          </template>
          <template #append>
            <v-sheet :color="two ? 'primary': null">Append</v-sheet>
          </template>
        </v-card-item>
      </v-card>
    </v-main>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const one = ref(false)
  const doOne = () => one.value = !one.value

  const two = ref(false)
  const doTwo = () => two.value = !two.value

</script>
```


packages/vuetify/src/components/VCard/VCardItem.tsx
```diff
@@ -46,9 +66,9 @@ export const VCardItem = genericComponent<VCardItemSlots>()({
       const hasSubtitle = !!(props.subtitle || slots.subtitle)
 
       return (
-        <div class="v-card-item">
+        <div class="v-card-item" onClick={ onClick } >
```
johnleider, line 69:
```suggestion
        <div class="v-card-item" onClick={ onClick }>
```

johnleider (changes requested):

J-Sek:
Hard to tell if it was meant as a convenience for appended close icons or anything else.

I am cutting it off, leaving slot-click emits on form fields only. While it looks like a small amount of code, after adding a11y attributes and ensuring hover/focus... too much for a convenience. Would also end up leaving buttons, chips, breadcrumbs, and alerts inconsistent. We can always revisit it once v5.0 lands. I imagine re-worked internals could unlock "on{event}:{slot}" binding as a framework-wide feature.. although probably leaving a11y to the end user.

PR #16997, opened 2023-03-24
fix(btn): use line height from typography settings

jd-solanki:
## Description
Fixes issue where line height is not updated if we update button line height in typography

## Markup:
N/A


johnleider:
@KaelWD would this be considered breaking?

PR #17005, opened 2023-03-25
feat(VStepper): migrating stepper to vue 3

MFJamil:
## Description
Migrating the Stepper component to Vue 3, the change was based on initial changes made by Albasaur, which includes the stepper to populate the steps via properties, the migration was extended to include populating the elements defined via tags/slots, like the **v-stepper-header**, **v-stepper-step**, **v-stepper-items**, **v-stepper-content**. The implementation took into consideration the possibility to define the steps in different order, for example, they might not being all properly be defined under the main v-stepper-header. The same applies to the v-stepper-content. Also, the migration allows the mixing between the two, so it is possible to define the steps via the properties as well as the tags, and it resolves the conflicting steps defined between the two via a flag which specifies the priority. All the cases mentioned are defined in playgrounds with a description for each case. There are 10 playgrounds defined, the first 4 are about different combinations of steppers created via Properties and Slots. The other 6 playgrounds are derived simply from the same Markup mentioned in the documentation of Steppers on Vuetify2.
Therefore it is hard to mention only one markup here on this page, better to run "yarn dev" to see all the playgrounds all together. 

## Markup:
 There are 10 Playgrounds with different markups, below is only one sample of them 

```
<template>
  <v-app>
    <v-locale-provider :rtl="rtl">
      <v-btn @click="rtl = !rtl">rtl</v-btn>
      <p style="padding: 20px;">Below all the steps were provided via the tags, like stepper-header/stepper-step, as well as the contents via stepper-items/stepper-content </p>
     
      <!-- *******************************************                Vertical Component               ********************************************-->
      <v-stepper v-model="e1">
        <v-stepper-header>
          <v-stepper-step
            title="Name of step 1"
            :complete="e1 > 1"
            step="1"
          />

          <v-divider />

          <v-stepper-step
            title="Name of step 2"
            :complete="e1 > 2"
            step="2"
          />

          <v-divider />

          <v-stepper-step
            title="Name of step 3"
            step="3"
          />
        </v-stepper-header>

        <v-stepper-items>
          <v-stepper-content step="1">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 1 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e1 = '2'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>

          <v-stepper-content step="2">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 2 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e1 = '3'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>

          <v-stepper-content step="3">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 3 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e1 = '1'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>
        </v-stepper-items>
      </v-stepper>

      <!-- *******************************************                Horizontal Component               ********************************************-->

      <v-stepper v-model="e2" direction="vertical">
        <v-stepper-header>
          <v-stepper-step
            title="Name of step 1"
            :complete="e2 > '1'"
            step="1"
          />

          <v-divider />

          <v-stepper-step
            title="Name of step 2"
            :complete="e2 > '2'"
            step="2"
          />

          <v-divider />

          <v-stepper-step
            title="Name of step 3"
            step="3"
          />
        </v-stepper-header>

        <v-stepper-items>
          <v-stepper-content step="1">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 1 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e2 = '2'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>

          <v-stepper-content step="2">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 2 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e2 = '3'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>

          <v-stepper-content step="3">
            <v-card
              class="mb-12"
              color="grey lighten-1"
              height="200px"
            >
              Step 3 Contents
            </v-card>

            <v-btn
              color="primary"
              @click="e2 = '1'"
            >
              Continue
            </v-btn>

            <v-btn text>
              Cancel
            </v-btn>
          </v-stepper-content>
        </v-stepper-items>
      </v-stepper>

    </v-locale-provider>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    data: () => ({
      e1: 0,
      e2: 0,
      rtl: false,
      open: null,
      active: null,
      items: [
        {
          title: 'Foo',
          subtitle: 'Subtitle',
          value: 'foo',
        },
        {
          title: 'Bar',
          value: 'bar',
        },
      ],
      loadedItems: [
        {
          value: 'root',
          text: 'root',
          loaded: false,
          children: [],
        },
      ],
      opened: [],
      isCtrl: false,
      count: 1,
    }),
  }
</script>

```


packages/vuetify/src/components/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,222 @@
...
+    'update:modelValue': (v: any) => true,
+  },
+
+  setup (props, { slots, emit }) {
+    // console.log('Stepper -Setup') // Can be removed after approval
+    const rootEl = ref<HTMLElement>()
+
+    const model = useProxiedModel(props, 'modelValue')
+    const group = useGroup(props, VStepperGroupProvideSymbol)
+    const pitems = computed(() => props.items)
```
johnleider, line 78: we have an items composable for this

packages/vuetify/src/util/helpers.ts
```diff
@@ -322,6 +322,28 @@ type Writable<T> = {
   -readonly [P in keyof T]: T[P]
 }
 
+/*
+ Collects the VNodes insdie the provided slots that holds the name specified
+Note: for some reason, recursively returning array and adding the content via concat did not work, therefore passing the array as parameter!
+ */
+export function addNodesInSlots (name: string, slots: Slots, nodes?: VNode[]) {
```
johnleider, line 329: I'd like @KaelWD's opinion on this. Early on we decided against VNode parsing in v3 which is why `v-tabs` is the way it is now.

packages/vuetify/src/components/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,222 @@
...
+        'v-stepper',
+        `v-stepper--${props.direction}`,
+      ]
+      if (props.altLabels) classList.push('v-stepper--alt-labels')
+      if (props.flat) classList.push('v-stepper--flat')
+      if (props.nonLinear) classList.push('v-stepper--non-linear')
+      return classList
+    })
+    return () => {
+      if (props.direction === 'vertical') {
```
johnleider, line 156: This was one of the reasons for the hesitancy behind stepper. It's really difficult to do vertical and horizontal in the same component, especially without VNode parsing (which we are trying to avoid).

packages/vuetify/src/components/VStepper/VStepperContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+// Utilities
+import { computed, inject } from 'vue'
+import { defineComponent } from '@/util'
+import { VStepperGroupProvideSymbol, VStepperProvideSymbol } from './VStepper'
+
+export const VStepperContent = defineComponent({
+  name: 'VStepperContent',
+
+  inheritAttrs: false,
```
johnleider, line 19: You're not explicitly binding attrs anywhere.

packages/vuetify/src/components/VStepper/VStepperContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+              >
+                { hasContent && content }
+              </div>
+            </VExpandTransition>
+          </div>
+        )
+      }
+
+      return (
+        <VWindowItem value={ groupItemId.value }>
```
johnleider, line 80: You will need to parse out props and add a makeProps function for VWindowItem

packages/vuetify/src/components/VStepper/VStepperStep.tsx
```diff
@@ -0,0 +1,151 @@
...
+export const VStepperStep = genericComponent()({
+
+  name: 'VStepperStep',
+
+  // inheritAttrs: false,
+
+  props: {
+    color: {
+      type: String,
+      default: 'primary',
```
johnleider, line 28: No default colors

packages/vuetify/src/components/VStepper/VStepperStep.tsx
```diff
@@ -0,0 +1,151 @@
...
+    },
+    complete: Boolean,
+    completeIcon: {
+      type: String,
+      default: '$complete',
+    },
+    editable: {
+      type: Boolean,
+      default: false,
```
johnleider, line 37: No need to set defaults for primitives. We use the intrinsic default values from Vue

packages/vuetify/src/components/VStepper/VStepperStep.tsx
```diff
@@ -0,0 +1,151 @@
...
+    },
+    /*
+    rules: {
+      type: Array as PropType<any[]>,
+      default: () => [],
+    },
+    */
+    step: {
+      type: [Number, String],
+      required: true,
```
johnleider, line 59: Any component should not require a default value

packages/vuetify/src/components/VStepper/VStepperStep.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const icon = computed(() => {
+      if (props.errors && props.errorIcon) return props.errorIcon
+
+      if (completed.value && props.completeIcon) return props.completeIcon
+      if (props.editable && props.editIcon) return props.editIcon
+
+      return undefined
+    })
+
+    const color = computed(() => {
```
johnleider, line 91: We might have to tweak this to allow the user to change these values.

johnleider (changes requested):
This is just a general overview. I'm mostly interested in the conversation regarding how we want to handle VNode parsing.

Another point, the user should be able to recreate the stepper using individual elements if they choose not to opt into items.

packages/vuetify/src/components/VStepper/VStepper.tsx
nekosaur: I am not a fan of the parsing. But like you say v-stepper is a hard one because of the large difference in markup between the directions. Maybe we just make them different components with different markup? It would be harder for users to switch between the two directions at runtime, but I have to think that scenario is very uncommon.

PR #17033, opened 2023-03-30
feat(style): add cursor class utilities 

santiagoaloi:
TODO: Documentation

Added all cursor classes to be used in any element:

auto default pointer text progress wait not-allowed grab grabbing help

```html
<!-- example --> 
<div class="cursor-pointer" />
<div class="cursor-grab" />
```

### DIFF
--- a/packages/docs/src/data/nav.json
+++ b/packages/docs/src/data/nav.json
@@ -61,6 +61,7 @@
       },
       "border-radius",
       "content",
+      "cursor",
       "display",
       "elevation",
       "flex",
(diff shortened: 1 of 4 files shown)


nekosaur:
If we're adding this, then I think just including the most commonly used ones is enough.

santiagoaloi:
> If we're adding this, then I think just including the most commonly used ones is enough.

John and you have the same opinoin, I think the most common ones wins then.

Suggestion:

- pointer
- not-allowed
- grab
- grabbing
- help

johnleider:
There are a lot of unnecessary changes to the utilities file. This should be based off of dev.

PR #17036, opened 2023-03-30
fix(VSelect/VAutocomplete): transform raw object if item-title is a function

### LINKED ISSUE #16210: [Bug Report][3.0.3] VSelect item-title function gets inconsistent arguments
### Environment
**Vuetify Version:** 3.0.3
**Last working version:** 2.6.12
**Vue Version:** 3.2.45
**OS:** Windows 10

### Steps to reproduce
```vue
<template>
  <v-select
    :items="values"
    :item-title="(item) => 'Item: ' + JSON.stringify(item)"
    item-value="id"
  />
</template>
<script setup lang="ts">
const values = [
  { id: 1, name: 'a' },
  { id: 2, name: 'b' },
];
</script>
```
Try selecting a value.
image

### Expected Behavior
The item-title function should always get the whole selected object as an argument.

### Actual Behavior
The item-title function gets only the value of the key specified by item-value as an argument, but only for the selected value.

### Reproduction Link
[https://stackblitz.com/edit/node-nykzap?file=pages/index.vue](https://stackblitz.com/edit/node-nykzap?file=pages/index.vue)

yuwu9145:
fix #16210

## Description

Transform Result

```js
{
   "children":null,
   "title":"Item: {\"id\":1,\"name\":\"a\"}",
   "value":1,
   "props":{
      "title":"Item: {\"id\":1,\"name\":\"a\"}",
      "value":1
   },
   "raw":{
      "id":1,
      "name":"a"
   }
}
```

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <h1>{{ selected }}</h1>
      <v-select
        v-model="selected"
        :items="values"
        :item-title="(item) => 'Item: ' + JSON.stringify(item)"
        item-value="id"
      />
      <v-autocomplete
        v-model="selected"
        :items="values"
        :item-title="(item) => 'Item: ' + JSON.stringify(item)"
        item-value="id"
      />
      <!-- <v-combobox
        v-model="selected"
        :items="values"
        :item-title="(item) => 'Item: ' + JSON.stringify(item)"
        item-value="id"
        :return-object="false"
      /> -->
    </v-main>
    
  </v-app> 
</template>
<script setup lang="ts">import { ref } from 'vue';

const values = [
  { id: 1, name: 'a' },
  { id: 2, name: 'b' },
];
const selected = ref(null)
</script>


```


johnleider:
Does this fix need to occur in the other select components?

* https://github.com/vuetifyjs/vuetify/blob/36f727bd81084e4efa239d9729be3f0ecbb96e62/packages/vuetify/src/components/VCombobox/VCombobox.tsx#L130
* https://github.com/vuetifyjs/vuetify/blob/36f727bd81084e4efa239d9729be3f0ecbb96e62/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx#L125

yuwu9145:
Thanks, yes, 

- need to apply to `VAutocomplete`
- but no need for  `VCombobox` because it has `return-object`. Unless there was any instance that user would explicitly override `:return-object="false"` on `VCombobox`

* https://github.com/vuetifyjs/vuetify/blob/36f727bd81084e4efa239d9729be3f0ecbb96e62/packages/vuetify/src/components/VCombobox/VCombobox.tsx#L94

johnleider (approved):

nekosaur:
I wonder if something like this would be enough to fix it

```ts
function transformIn (value: any[]): InternalItem[] {
  if (!props.returnObject) {
    return items.value.filter(item => value.includes(item.value))
  }
  return value.map(item => transformItem(props, item))
}
```

yuwu9145:
Thanks @nekosaur . Inspired by your idea, I refactored it a little bit to make it more readable. Few key points:

- Cause is `transformItem` is unable to figure out which concrete item object attached to model value, so it cannot be called directly during "transform In"
- However, `items` has correct transform data because `transformItem` was called based on item object itself
- During "transform In" stage, try to get `InternalItem` from existing `items`
- If no `InternalItem` was found based on modelValue, then it must be a combobox custom input, so call `transformItem` to create a new `InternalItem` based on input

johnleider (approved):

PR #17044, opened 2023-03-31
docs(section): add search functionality

santiagoaloi:


packages/docs/src/components/api/Section.vue
```diff
@@ -1,46 +1,42 @@
...
-        @input="filter = $event"
+        prepend-inner-icon="mdi-magnify"
+        single-line
+        hide-details
       />
-    </div> -->
+
+    </div>
+
+    <!-- TODO: investigate what this component does, if it even exists -->
```
KaelWD, line 17: Enable inline API in settings then go to a component page.

packages/docs/src/components/api/Section.vue
```diff
@@ -51,15 +47,48 @@
...
+    const component = sectionMap[props.section as keyof typeof sectionMap]
+    return component || null
+  })
+
+  // 👉 Seection Search logic
+  const search = ref()
+  const searchableFields = ['name', 'description']
+
+  const filteredItems = computed(() => {
+    return items.value.filter(item =>
```
KaelWD, line 78: `(item: any) =>` or define a proper interface for `items`

KaelWD:
The v2 inline API had a single search for all item types, it was only separate on API pages.
Screenshot_20230331_224711

s-rigaud:
@KaelWD I think this PR might be closed as this feature as recently been implemented 😊

PR #17077, opened 2023-04-04
fix(VFileInput): expose InputHTMLAttributes type

### LINKED ISSUE #17069: [Feature Request] [VFileInput] Types: Add `accept` attribute (Volar gives error now)
### Problem to solve
Volar would not give the error below (which basically says that `accept` is not defined on the `VFileInput` type):
```
Argument of type '{ $children: any; modelValue: File[]; accept: string; label: string; prependIcon: string; 'prepend-icon': string; rules: ValidationRule[] | undefined; }' is not assignable to parameter of type 'Partial<{ reverse: boolean; error: boolean; active: boolean; direction: "horizontal" | "vertical"; disabled: boolean; multiple: boolean; readonly: boolean; counter: boolean; messages: string | string[]; ... 17 more ...; showSize: boolean | ... 1 more ... | 1024; }> & Omit<...>'.
  Object literal may only specify known properties, and 'accept' does not exist in type 'Partial<{ reverse: boolean; error: boolean; active: boolean; direction: "horizontal" | "vertical"; disabled: boolean; multiple: boolean; readonly: boolean; counter: boolean; messages: string | string[]; ... 17 more ...; showSize: boolean | ... 1 more ... | 1024; }> & Omit<...>'.ts(2345)
```

### Proposed solution
Add the `accept` prop in `VFileInput`.

P.S.: Kind request: Briefly comment on your experience regarding the different Volar changes over time that affected Vuetify, please!

yuwu9145:
fix #17069 

It is supported in [v2](https://v2.vuetifyjs.com/en/components/file-inputs/#accept)

`InputHTMLAttributes` including `accept` should be type supported in VFileInput.

input attributes like `accept` will be passed to input via attributes

Allows users to safely specify any legit input(file) attributes. e.g. autofocus, accept ...

## Description

fix #17069
## Markup:

```vue
<template>
  <v-app>
    <v-main>
      
      <v-file-input id="a" autofocus accept="image/png, image/jpeg"></v-file-input>
    </v-main>
  </v-app>
</template>

<script lang="ts">

</script>


```

### DIFF
--- a/packages/vuetify/src/components/VFileInput/VFileInput.tsx
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.tsx
@@ -40,6 +40,7 @@ export const VFileInput = genericComponent<VFileInputSlots>()({
   inheritAttrs: false,
 
   props: {
+    accept: String,
     chips: Boolean,
     counter: Boolean,
     counterSizeString: {
@@ -207,6 +208,7 @@ export const VFileInput = genericComponent<VFileInputSlots>()({
                       <input
                         ref={ inputRef }
                         type="file"
+                        accept={ props.accept }
                         readonly={ isReadonly.value }
                         disabled={ isDisabled.value }
                         multiple={ props.multiple }
--- a/packages/vuetify/src/components/VFileInput/VFileInput.tsx
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.tsx
@@ -35,9 +35,7 @@ export type VFileInputSlots = VInputSlots & VFieldSlots & MakeSlots<{
 }>
 
 export const VFileInput = genericComponent<new () => {
-  $props: InputHTMLAttributes & SlotsToProps<{
-    default: [VFileInputSlots]
-  }>
+  $props: InputHTMLAttributes & SlotsToProps<VFileInputSlots>
 }>()({
   name: 'VFileInput',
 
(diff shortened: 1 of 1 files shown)


KaelWD:
This shouldn't be necessary, we don't want to list every possible html attribute in the docs:

Screenshot_20230404_171642

Arbitrary attributes should be supported:

https://discord.com/channels/340160225338195969/660898563139567625/1091801131128205423
Screenshot_20230404_171743

KaelWD:
```ts
declare module '@vue/runtime-core' {
  export interface AllowedComponentProps {
    [key: string]: any;
  }
}
```
I don't think we should include this ourselves though

yuwu9145:
> This shouldn't be necessary, we don't want to list every possible html attribute in the docs:

Yes, `InputHTMLAttributes` should not be in our docs. However, the current runtime behaviour is these attributes are being passed to input tag via `{ ...inputAttrs }` (if you inspect `VInput`, you will see accept & autofocus are rendered into the dom), the problem is this isn't reflected in type check. For VFIleInput, allowing `InputHTMLAttributes` types should be the same as allowing "class" & "style" by default

> Arbitrary attributes should be supported on every component:

I agree with you, this seems should be a default Vue type feature. Having said this, allowing arbitrary attributes does not mean allowing **arbitrary values** , we will still  need `InputHTMLAttributes` to guard input value is in correct type:

image

yuwu9145:
> Arbitrary attributes

https://github.com/vuejs/language-tools/issues/1077#issuecomment-1145361472

websitevirtuoso:
I have no idea why we need this but I can set any attr to any elements. I also use vite +TS and everything works. it doesn't make sense  define each attr for TS. It can be added in docs or somewhere else I think

KaelWD:
Closing in favour of #17082

PR #17102, opened 2023-04-07
fix: unreliable dependencies of `useProxiedModel()`

johnsoncodehk:
`model` is relying on `isControlled` to propagate the dirty event. After https://github.com/vuejs/core/pull/5912, the dirty event will no longer be propagated if the new value of `isControlled` hasn't changed. This has caused some tests to fail.

### DIFF
--- a/packages/vuetify/src/composables/proxiedModel.ts
+++ b/packages/vuetify/src/composables/proxiedModel.ts
@@ -46,8 +46,8 @@ export function useProxiedModel<
 
   const model = computed({
     get (): any {
-      void props[prop]
-      return transformIn(isControlled.value ? props[prop] : internal.value)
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/composables/proxiedModel.ts
```diff
@@ -46,6 +46,7 @@ export function useProxiedModel<
 
   const model = computed({
     get (): any {
+      void props[prop]
       return transformIn(isControlled.value ? props[prop] : internal.value)
```
KaelWD, line 50: Save a proxy access?
```suggestion
      const externalValue = props[prop]
      return transformIn(isControlled.value ? externalValue : internal.value)
```

PR #17111, opened 2023-04-09
fix(validation): validate initial input value

### LINKED ISSUE #15568: [Bug Report][3.0.0-beta.6] VForm validation: model-value is null instead of true when form is valid
### Environment
**Vuetify Version:** 3.0.0-beta.6
**Vue Version:** 3.2.37
**Browsers:** Chrome 103.0.0.0
**OS:** Windows 10

### Steps to reproduce
1. Enter something into the form until a rule is broken (e.g. more than 10 characters in the name input)
2. See console to check the current model value.
3. Change input until the validation rules are observed again (e.g. 8 characters in the name input)
3. See console to check the current model value.

### Expected Behavior
The model-value (validation result) of the `v-form` should be `true` if all rules are observed.

### Actual Behavior
The model-value (validation result) of the v-form is `null` if all rules are observed.   
It works well, if >=1 rule isn't observed (model value is `false`)

### Reproduction Link
<a href="https://codepen.io/jriedel-osp/pen/abYYvRr" target="_blank">https://codepen.io/jriedel-osp/pen/abYYvRr</a>

nekosaur:
## Description

closes #15568

this is a change to default behaviour, but I _think_ the old behaviour could be classified as a bug.

## Markup:

```vue
<template>
  <v-form v-model="valid">
    <v-text-field autofocus label="name" v-model="name" required :rules="nameRules"></v-text-field>
    <v-text-field label="type" v-model="type" required :rules="typeRules"></v-text-field>
  </v-form>
  Valid: {{ valid === null ? 'null' : valid }}
  <v-btn :disabled="!valid">Sa</v-btn>
</template>

<script>

  export default {
    data () {
      return {
        valid: false,
        name: '',
        type: 'asdasdasdasd',
        nameRules: [(v) => (!!v && v.length > 5) || "EDITOR.stories.storyNameRequired"],
        typeRules: [(v) => !!v || "EDITOR.stories.typeIsRequired"],
      }
    },
  }
</script>
```


johnleider (dismissed):

nekosaur:
TODO: this has the same issue as optional PR in that empty value is different for different components so just checking `if(model.value)` is probably not enough

hafizjavaid:
> ## Description
> closes #15568
> 
> this is a change to default behaviour, but I _think_ the old behaviour could be classified as a bug.
> 
> ## Markup:
> ```
> <template>
>   <v-form v-model="valid">
>     <v-text-field autofocus label="name" v-model="name" required :rules="nameRules"></v-text-field>
>     <v-text-field label="type" v-model="type" required :rules="typeRules"></v-text-field>
>   </v-form>
>   Valid: {{ valid === null ? 'null' : valid }}
>   <v-btn :disabled="!valid">Sa</v-btn>
> </template>
> 
> <script>
> 
>   export default {
>     data () {
>       return {
>         valid: false,
>         name: '',
>         type: 'asdasdasdasd',
>         nameRules: [(v) => (!!v && v.length > 5) || "EDITOR.stories.storyNameRequired"],
>         typeRules: [(v) => !!v || "EDITOR.stories.typeIsRequired"],
>       }
>     },
>   }
> </script>
> ```

Thanks

KaelWD:
This should be more similar to v2. 
- Inputs should always be validated initially no matter what the value is
- There should be some way to run validation "silently" without displaying errors to the user
- There should be some way to disable the initial validation, for example if async rules are used

johnleider (approved):

johnleider (dismissed):

PR #17112, opened 2023-04-09
fix(VSelect): replicate html select hotkeys

### LINKED ISSUE #7260: [Bug Report] Keyboard enter key searches v-select items beginning with 'Enter'
### Versions and Environment
**Vuetify:** 1.5.7
**Vue:** 2.6.10
**Browsers:** Chrome 74.0.3729.131
**OS:** Windows 10

### Steps to reproduce
Ensure v-select includes an option who's text begins with "Enter" (e.g.: Entertainment).
Click or focus on the v-select control.
Press the enter key on the keyboard.
The v-select selects the option which begins with "Enter".

### Expected Behavior
The v-select control should not search for items beginning with "Enter" when the user presses their keyboard's enter key.

### Actual Behavior
The v-select control searches for (and selects if found) items beginning with "Enter" when the user presses their keyboard's enter key.

### Reproduction Link
<a href="https://codepen.io/anon/pen/OYOQdX" target="_blank">https://codepen.io/anon/pen/OYOQdX</a>

### LINKED ISSUE #16055: [Bug Report][3.0.0] Pressing letters on the keyboard does not select items in VSelect
### Environment
**Vuetify Version:** 3.0.0
**Last working version:** 2.6.12
**Vue Version:** 3.2.42
**OS:** Linux x86_64

### Steps to reproduce
Create a simple select menu with VSelect:

```
<v-select
  label="Select"
></v-select>
```

### Expected Behavior
Pressing a letter on the keyboard should select the first item beginning with that letter, as is the deafult behaviour of browsers in html select menus

### Actual Behavior
Pressing a letter does nothing

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpVUN1qwjAUfpVDbrJBl4ruSupgCNsDbGwX1ousnkpG/kiiKKXv7mlTUXP1/Z2TL9l07HjApNqz+I9syZTxLiTooAkoE/5kD3pogzPApyyvbW3xNEYbZ2OCyYDV4+BT1z/XlhXs3XtBGbqhik1QPkHEdPBvta3KLBAkktB4TfPEAKrji/RDBugQMVLZiY08osYmXQUALf9Qr2r2lXV2c5aKFkeyNnwttWpdsEryAvjaaRfkzg34g6DajfInurDPiW88yTiA37Mzyu759ra4vHYr78oNJNeuyrvnsL5gCzEXr3P6joWYiRnbXgA3CHo7)

### Other comments
Compare 
https://next.vuetifyjs.com/en/components/selects/
to
https://vuetifyjs.com/en/components/selects/

See also: https://github.com/vuetifyjs/vuetify/pull/15632

yuwu9145:
fixes #16055
fixes #7260

- Port #4853 to V3

## Markup:

```vue
<script setup>
import { ref } from 'vue' 
      
const items = ref(['Foo', 'Bar', 'Fizz', 'Buzz'])
  
</script>

<template>
  <v-select
    :items="items"
    label="Standard"
  ></v-select>
</template>

```

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -188,7 +188,7 @@ export const VSelect = genericComponent<new <
       } else if (e.key === 'End') {
         listRef.value?.focus('last')
       }
-    
+
       // html select hotkeys
       ((e: KeyboardEvent) => {
         const KEYBOARD_LOOKUP_THRESHOLD = 1000 // milliseconds
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -189,22 +189,29 @@ export const VSelect = genericComponent<new <
       } else if (e.key === 'End') {
         listRef.value?.focus('last')
       }
-    }
-    // html select hotkeys
-    function onKeypress (e: KeyboardEvent) {
-      if (props.multiple) return
+    
+      // html select hotkeys
+      ((e: KeyboardEvent) => {
+        function checkPrintable (e: KeyboardEvent) {
+          const isPrintableChar = e.key.length === 1 && e.key !== ' '
+          const noModifier = !e.ctrlKey && !e.metaKey && !e.altKey
+          return isPrintableChar && noModifier
+        }
 
-      const now = performance.now()
-      if (now - keyboardLookupLastTime > KEYBOARD_LOOKUP_THRESHOLD) {
-        keyboardLookupPrefix = ''
-      }
-      keyboardLookupPrefix += e.key.toLowerCase()
-      keyboardLookupLastTime = now
+        if (props.multiple || !checkPrintable(e)) return
 
-      const item = items.value.find(item => item.title.toLowerCase().startsWith(keyboardLookupPrefix))
-      if (item !== undefined) {
-        model.value = [item]
-      }
+        const now = performance.now()
+        if (now - keyboardLookupLastTime > KEYBOARD_LOOKUP_THRESHOLD) {
+          keyboardLookupPrefix = ''
+        }
+        keyboardLookupPrefix += e.key.toLowerCase()
+        keyboardLookupLastTime = now
+
+        const item = items.value.find(item => item.title.toLowerCase().startsWith(keyboardLookupPrefix))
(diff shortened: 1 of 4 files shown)


yuwu9145:
https://github.com/vuetifyjs/vuetify/blob/d96378ae13627fcdc3833317e69957f7e3f4d94d/packages/vuetify/src/components/VSelect/VSelect.tsx#L259

What was the reason for `readonly` here? If it could be removed, we can use [beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event) to capture printable characters, it is supported in firefox now. Otherwise, we will have to stick with [keypress](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) which is deprecated

KaelWD:
Stop it from being typed into. We especially don't want mobile keyboards to appear when focused.

yuwu9145:
Thanks for the comprehensive suggestion. I refactored it to use keydown and wrapped it inside iife for encapsulation purpose, because it's not going to be used anywhere else

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -185,6 +188,31 @@ export const VSelect = genericComponent<new <
...
+      ((e: KeyboardEvent) => {
+        const KEYBOARD_LOOKUP_THRESHOLD = 1000 // milliseconds
+
+        function checkPrintable (e: KeyboardEvent) {
+          const isPrintableChar = e.key.length === 1 && e.key !== ' '
+          const noModifier = !e.ctrlKey && !e.metaKey && !e.altKey
+          return isPrintableChar && noModifier
+        }
+
+        if (props.multiple || !checkPrintable(e)) return
```
KaelWD, line 202: - We want to keep spaces
- Not sure why you'd use an iife here
```suggestion
      const KEYBOARD_LOOKUP_THRESHOLD = 1000 // milliseconds
      const isPrintable = e.key.length === 1 && !e.ctrlKey && !e.metaKey && !e.altKey
      if (isPrintable && !props.multiple) {
```

yuwu9145: Not particular good reason, I was just paranoid at that moment wanting to group this logic in a dedicated place

KaelWD (approved):

PR #17178, opened 2023-04-19
docs: add composition api to examples

MajesticPotatoe:
## Description
Adds the ability to provide `<script setup>` to examples and parse them accordingly:
- Displays options + composition on examples
- Sends composition/options to playground based on user preferences and availability 
- Removes "send to codepen"

## Markup:
Example provided in `examples/alert/props-closable.vue`

```vue
<template>...</template>
<script setup>
  import { ref } from 'vue'

  const alert = ref(true)
</script>

<script>
  export default {
    data: () => ({
      alert: true,
    }),
  }
</script>
```


johnleider:
is it possible to do this for code fences as well?

MajesticPotatoe:
> is it possible to do this for code fences as well?

that would be a separate PR, but given that its pretty much handled in the composable, yeah it would be possible.

KaelWD:
A vue bug was hiding the warning but both script blocks are running:

Screenshot_20230529_180233

MajesticPotatoe:
> A vue bug was hiding the warning but both script blocks are running:
> 
gotta love hidden warnings xD fixed.

KaelWD (dismissed):
If we have to do everything in comment blocks I'm out

PR #17212, opened 2023-04-23
chore(dev): upgrade vue-tsc to 1.4.4 and move to root package.json

### LINKED ISSUE #17211: [Bug Report][3.1.15] VSelect prop return-object type error
### Environment
**Vuetify Version:** 3.1.15
**Vue Version:** 3.2.47
**Browsers:** Edge 112.0.1722.48
**OS:** Windows 10

### Steps to reproduce
Add a `<v-select>` with prop `return-object` in template. Type error:
```
Type 'true' is not assignable to type 'false'.
```

image

### Expected Behavior
`return-object` can be either `true` or `false`

### Actual Behavior
`return-object` can only accept `false`

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVU9tq3DAQ/RUhCtvCyiJtoGDckBDavJS89CnEC5HXs4mysmQk2WlZ/O8dWbbXuyG3J2vOzJwzN9/u6EVdJ20DNKWZh6pWwsNZrgnJWuZAwdoHg5CWVaYE9SOnA0ojnj5I7RG9+7SLjsR5pOiWZAJEUdjubkqQKOMwo/+OaDCYl15BUAgMB55WqCZ4AtXoUKLo6/lzUE8N1knnQXsWKougBd9YzUzxOPXjpL5XwJTUEICzjI/9YvcZn40CTbe2sg4OQuBvbawnJWxEozzZRbZSeEE+fxnNUXFvo2BPnpId6dtLyeKXMlaWYrEkoa0A/F6QbrlP6SeUkts9Qj6SPg++AmPv5Sz46uLl4GsorHDbWfT1z5ejL4WSG2P1nP3yVfYncmPsdsZ+cxi9mt5dfEQnGhkfVkGXFFe3dcmjMxpvtx90TtcOL2qaWE4fvK9dyvm61BiJ9ytbm2jwXNcVx6v3cvPv/Ftykpyc8hLPZsQYHpdLerZAtcp1h4qyCrtnlaiPZKOjlx42nlNkCvYbNZxjGLeN9rICVpoKi/manH6PxczwBFzFCmueHFhkyekwol4nVPwOrTf7RY1n3BwFW7DMgi7BAv597+vpKO2gryPfs96GZXe0W9KYh8OPVdPVfw/6fSw=)

yuwu9145:
- fixes #17211
- Upgrade vue-tsc version to 1.4.4 and move it root package.json file

## Description

This issue starts to appear since [volar@1.3.9](https://github.com/vuejs/language-tools/commit/e7a81dae5b0a2c9a939a7145c93bf664607eadbd), but I couldn't figure out the reason yet. 

volar@1.3.8 works.

However, this is an issue supposed to be complained: [prove link](https://www.typescriptlang.org/play?ts=4.8.4#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBVDtUAAWTwACF4PAWpx9gAeAAikGSDCiqAAagxoOwUGsFDxhMDGR4vABeHECwSQAB8oFlHK5PP5gqUigQnEENAU+vy1LpDKZ+xVaDIQuCxQpRBN9MlZlZau50D5AqF4NWou44rNUstXOgCuVqs57s9WuCKl1+rQ+lQxppTvNlNlqGtSjjjUEkoC0HgzAAFIbUABKWNgHDNDxkLyIeB9BjA9gGu6MTjMdgMZiQfiwaDqRKt+B6EReABEAFFVgxEAt9FOktip+AHKgAOTCABykAkU6CKgA8l8yO8Ffx+0dUI4FK5pHhQE4mzphlh9CgEIhocJLl8kjNtEIgGBw2ILEwrAcDwQQlEAA)

this seems to be an issue which didn't show up before, also seems like there was an enhancement from volar@1.3.9 that makes this issue show up. So it doesn't bother we make a fix in vuetify side

Since we can use [`vue-tsc` against dev playground](https://github.com/vuetifyjs/vuetify/pull/16843#issue-1611312999), it would make sense to move `vue-tsc` to root package.json

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-select
        return-object
        multiple
      />
      <v-autocomplete
        return-object
        multiple
      />
      <v-combobox
        return-object
        multiple
      />
    </v-main>
  </v-app>
</template>

<script lang="ts">
</script>

```


johnleider:
2 questions:

- What other areas could this crop up in?
- What is the reason that we can't infer for these specific props?

KaelWD:
1.5.2 is out now
I didn't try a separate project yet but I couldn't reproduce the original issue in dev with either 1.2 or 1.4

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -90,7 +90,7 @@ export const VAutocomplete = genericComponent<new <
     search: String,
 
     ...makeFilterProps({ filterKeys: ['title'] }),
-    ...makeSelectProps(),
+    ...makeSelectProps({ multiple: false, returnObject: false }),
```
KaelWD, line 93: Why'd you add this? Boolean props already default to false.

yuwu9145: Thanks! That was something I didn't consider thoroughly. Removed.

yuwu9145:
> 2 questions:
> 
> * What other areas could this crop up in?

This is the issue regarding the usage of `extends boolean = true/false`, when I do a global search, only `multiple` & `returnObject` doing this.

> * What is the reason that we can't infer for these specific props?

It has been an issue which should have shown up from the right beginning. I believe there was something in `genericComponent` that hides this issue before volar 1.3.8. 1.3.9 was released three weeks ago and it shows up. I am still digging what was it. However, regardless of the reason, this issue needs to be fixed

yuwu9145:
> 1.5.2 is out now

This version is too new, 1.4.4 is more stable. 

image

> Need to make sure this is still working in tsx too, volar should be following tsx

will do

KaelWD:
This is why the generics have `= false`:
```js
const items = [
  { title: 'a' },
  { title: 'b' },
  { title: 'c' },
]
const Comp = (
  <VSelect items={ items } onUpdate:modelValue={ v => v } returnObject />
)
```
`v` should be `{title: string}`, but if typescript doesn't know what `multiple` is then you get `{title: string} | readonly {title: string}[]` instead. Volar doesn't seem to be resolving the generic correctly as it's always `any` in templates.

KaelWD:
Closing in favour of #17231

PR #17282, opened 2023-05-03
feat: support vue 3.3

KaelWD:
- [ ] https://github.com/vuejs/language-tools/issues/3122
- [ ] https://github.com/vuejs/language-tools/issues/3171


jd-solanki:
@KaelWD shoud we upgrade to vue 3.3 as there are two mentioned issues are pending?

KaelWD:
Should be fine they're just a couple of obscure issues I found in our docs. I think even vuetify 3.2 should work with vue 3.3, this PR is more of an internal refactor making use of the new slot types. Try 3.2 and nightly@dev and let us know how it goes.

PR #17284, opened 2023-05-03
fix(VExpansionPanel): remove duplicate aria-expanded

### LINKED ISSUE #15514: [Bug Report][2.6.7] Expansion Panel uses `aria-expanded` attribute on a div
### Environment
**Vuetify Version:** 2.6.7
**Vue Version:** 2.7.7
**Browsers:** Chrome 103.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
1. Install a tool similar to [Axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US). Alternatively, add a Vuetify expansion panel into a [Storybook](https://storybook.js.org/) story and install the [Accessibility addon](https://storybook.js.org/addons/@storybook/addon-a11y/).
2. Screen a page that includes an Expansion panel.
 

### Expected Behavior
a11y screening shouldn't fail on the following [criteria](https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?application=axeAPI).
The button should be the element that has `aria-expanded` attribute.

### Actual Behavior
Div element wrapping an expansion panel (div with the `v-expansion-panel` class on it) has `aria-expanded` attribute attached to it.

### Reproduction Link
<a href="https://codepen.io/marjev/pen/MWVvRmm" target="_blank">https://codepen.io/marjev/pen/MWVvRmm</a>

### Other comments
See [this Vanilla JS example](https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion.html) which demonstrates HTML which doesn't fail this screening.

marjev:
fixes #15514

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-expansion-panels>
        <v-expansion-panel
          v-for="i in 3"
          :key="i"
          title="Item"
          text="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
        />
      </v-expansion-panels>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VExpansionPanel/VExpansionPanel.tsx
+++ b/packages/vuetify/src/components/VExpansionPanel/VExpansionPanel.tsx
@@ -91,7 +91,6 @@ export const VExpansionPanel = genericComponent<VExpansionPanelSlots>()({
             backgroundColorStyles.value,
             props.style,
           ]}
-          aria-expanded={ groupItem.isSelected.value }
         >
           <div
             class={[


KaelWD (dismissed):
Don't just remove the attribute to make your validation tool pass, make it work in a screenreader

marjev:
@KaelWD I've tested with VO on Safari and it works because there is a child button that actually contains the `aria-expanded`. With which screen reader did you test this?

marjev:
@KaelWD thank you for following up. I can check with my QA team tomorrow to verify that it also works on the following:
- NVDA + Firefox
- JAWS + Chrome

Let me get back to you on those. I want to ensure that I don't accidentally introduce a new issue.

KaelWD:
Looks alright in NVDA + Chrome

Issue #15885, opened 2022-10-14
[Feature Request] Add support for moduleResolution: NodeNext in v3


viell-dev:
### Problem to solve
When using `moduleResolution: NodeNext` in `tsconfig.json` the `exports` field is used for imports and types. So `package.json` must have `types` fields in the `exports` field. And the type definition files need to have .d.mts extension when using .mjs files.

### Proposed solution
Change the build to produce `.d.mts` definition files or change to `.js` files instead of `.mjs`. And change `package.json` `exports` to something like the following:
```json
  "exports": {
    ".": {
      "import": {
        "types": "./lib/index.d.mts",
        "default": "./lib/framework.mjs"
      },
      "require": {
        "types": "./dist/vuetify.d.ts",
        "default": "./dist/vuetify.js"
      }
    },
    "./styles": "./lib/styles/main.css",
    "./styles/*": "./lib/styles/*",
    "./framework": {
      "types": "./lib/index.d.mts",
      "default": "./lib/framework.mjs"
    },
    "./blueprints": {
      "types": "./lib/blueprints/index.d.mts",
      "default": "./lib/blueprints/index.mjs"
    },
    "./blueprints/*": {
      "types": "./lib/blueprints/*.d.mts",
      "default": "./lib/blueprints/*.mjs"
    },
    "./components": {
      "types": "./lib/components/index.d.mts",
      "default": "./lib/components/index.mjs"
    },
    "./components/*": {
      "types": "./lib/components/*/index.d.mts",
      "default": "./lib/components/*/index.mjs"
    },
    "./directives": {
      "types": "./lib/directives/index.d.mts",
      "default": "./lib/directives/index.mjs"
    },
    "./directives/*": "./lib/directives/*/*.mjs",
    "./locale": {
      "types": "./lib/locale/index.d.mts",
      "default": "./lib/locale/index.mjs"
    },
    "./locale/adapters/*": {
      "types": "./lib/locale/adapters/*.d.mts",
      "default": "./lib/locale/adapters/*.mjs"
    },
    "./iconsets/*": {
      "types": "./lib/iconsets/*.d.mts",
      "default": "./lib/iconsets/*.mjs"
    },
    "./*": "./*"
  },
```

viell-dev:
[...]

PR #17305, opened 2023-05-06
fix: add types field to esm exports map

alexzhang1030:
## Description

When set `moduleResolution` to `nodenext` or `bundler`, import `vuetify` will shows an error:

image

Looks like we need to add types field to esm exports map to resolve this case.

related:

- https://github.com/microsoft/TypeScript/issues/50223
- https://github.com/chakra-ui/chakra-ui/pull/6945

## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/package.json
+++ b/packages/vuetify/package.json
@@ -54,7 +54,8 @@
     ".": {
       "sass": "./lib/styles/main.sass",
       "style": "./lib/styles/main.css",
-      "default": "./lib/framework.mjs"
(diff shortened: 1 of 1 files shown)


ml1nk:
I had the same problem a while ago https://github.com/vuetifyjs/vuetify/issues/15885#issuecomment-1485068275. Your pull only fix the default import, but not for example /labs as .mts tries to import from .d.mts not .d.ts by default. I added my patch as #17319.

PR #17316, opened 2023-05-07
refactor(framework): use shallowRef instead of ref for primitives

johnleider:
Just a micro optimization.


nekosaur:
Is this actually an optimization? I would have assumed that ref and shallowRef would be equal for primitives

johnleider:
It's my understanding that shallowRef is still faster.

PR #17340, opened 2023-05-10
fix(VSlider, VTextarea, VTextField): add aria name to semantic divs

### LINKED ISSUE #16397: [Bug Report][3.0.7] ARIA input fields must have an accessible name
### Environment
**Vuetify Version:** 3.0.7
**Vue Version:** 3.2.45
**OS:** Mac OS 10.15.7

### Steps to reproduce
Use Axe or a similar static page analyzer to perceive the issue.

1. Set a label for one of the following form input elements: Slider, Text area, or Text fields.
2. Axe should report that the div element with the role (slider or textbox) misses an accessible name - depending on the input element used in step 1.

### Expected Behavior
The label (provided string or the label element) should be used as a source of aria-label (string) or aria-labelledby (element). If the label element is used for this purpose, it would be good to ensure that it has a unique ID to ensure that this fix works in case more than one input has been added to the same page.

### Actual Behavior
`ARIA input fields must have an accessible name` issue is reported and the div with role attribute isn't labeled. 

### Reproduction Link
[https://codepen.io/marjev/pen/eYjNrKo](https://codepen.io/marjev/pen/eYjNrKo)

marjev:
## Description
fixes #16397

`VSlider`, `VTextarea` & `VTextField` have wrapped native input elements into div elements with specific roles. Those roles should not be left without an ARIA name.

Since Vuetify already adds the ability to provide a label, I am proposing to use that label as the accessible name. However, I also noticed that there is a bug in the codebase which causes both the label and the input to have the same ID. In order to be able to reference the label correctly in `aria-labelledby` attribute, I had to resolve that issue as well so that the assistive technologies could resolve which element holds the label text.

## Markup:

```vue
<v-slider
  id="1"
  label="slider"
  model-value="30"
/>
<v-responsive
  class="mx-auto"
  max-width="1000"
>
  <v-text-field
    label="First name"
    hide-details="auto"
  />
</v-responsive>

<v-textarea label="Label" />
```

### DIFF
--- a/packages/vuetify/src/labs/VDatePicker/VDatePicker.sass
+++ b/packages/vuetify/src/labs/VDatePicker/VDatePicker.sass
@@ -7,17 +7,15 @@
 
 .v-date-picker
   overflow: hidden
-  width: 358px
 
   .v-date-picker-month
-    padding-bottom: 8px
     padding-inline-start: 12px
     padding-inline-end: 12px
 
     .v-date-picker-month__day
-      height: 40px
-      width: 40px
+      height: 48px
+      width: 48px
 
       .v-btn
-        --v-btn-height: 24px
+        --v-btn-height: 28px
         --v-btn-size: 0.85rem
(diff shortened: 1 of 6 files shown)


marjev:
This PR is only a partial improvement. Our a11y team is investigating the a11y UX issues in more depth and there will probably be a follow-up PR for further improving the a11y of the form controls that Vuetify is providing.

Gandhi11:
@marjev Is there any reason that this PR was closed? the `aria-labelledby` is still missing.

PR #17342, opened 2023-05-10
feat(VWindow): set class when window is transitioning

fahmifitu:
## Description

This PR adds the css class 'v-window__container--is-active' when the window is in transitioning. this existed in vuetify 2 but didn't make to vuetify 3. 

My use case: I have a window that contains elements with shadows, I needed to make the shadows overflow the container in normal state by setting `overflow: visible`. but this breaks the ui when navigating the windows. and there's no class to apply styling to without this PR.
I'm sure there are more advanced use cases that would need style when the windos are transitioning

## Markup:

Test this markup in before and after the PR to notice the difference

```vue
<template>
  <v-app>
    <v-container style="max-width: 500px">
      <v-window show-arrows>
        <v-window-item>
          <v-row>
            <v-col cols="12" xl="8">
              <v-card class="mb-10 pa-10" elevation="10" />
              <v-card elevation="10" class="mb-10 pa-10" />
            </v-col>
            <v-col cols="12" xl="4">
              <v-card class="mb-10 pa-10" elevation="10" />
              <v-card elevation="10" class="mb-10 pa-10" />
            </v-col>
          </v-row>
        </v-window-item>
        <v-window-item>
          <v-row>
            <v-col cols="12" xl="8">
              <v-card class="mb-10 pa-10" elevation="10" />
              <v-card elevation="10" class="mb-10 pa-10" />
            </v-col>
            <v-col cols="12" xl="4">
              <v-card class="mb-10 pa-10" elevation="10" />
              <v-card elevation="10" class="mb-10 pa-10" />
            </v-col>
          </v-row>
        </v-window-item>
      </v-window>
    </v-container>
  </v-app>
</template>

<style>
.v-window {
  overflow: visible;
}
</style>

```

### DIFF
--- a/packages/vuetify/src/components/VWindow/VWindow.sass
+++ b/packages/vuetify/src/components/VWindow/VWindow.sass
@@ -10,6 +10,9 @@
     position: relative
     transition: $window-transition
 
+    &--is-active
+      overflow: hidden
+
   &__controls
     position: absolute
     left: 0
--- a/packages/vuetify/src/components/VWindow/VWindow.tsx
+++ b/packages/vuetify/src/components/VWindow/VWindow.tsx
@@ -230,7 +230,12 @@ export const VWindow = genericComponent<VWindowSlots>()({
         v-touch={ touchOptions.value }
       >
         <div
-          class="v-window__container"
+          class={[
+            'v-window__container',
+            {
+              'v-window__container--is-active': transitionCount.value > 0,
+            },
+          ]}
           style={{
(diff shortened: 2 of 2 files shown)


J-Sek:
Possible with modern CSS:

```css
.v-window:not(:has([class*='-transition-'])) {
  overflow: visible;
}
```

I don't think overriding `overflow` is a good idea. The recommended solution would be to add some padding on each item (.e.g. `pa-12`).

PR #17425, opened 2023-05-23
chore: Add percy tests for select components

yuwu9145:
## Description

Percy tests to capture snapshots for select components in [v3.3.0](https://github.com/vuetifyjs/vuetify/releases/tag/v3.3.0)


johnleider:
Should we add VField and VInput to this?

KaelWD (approved):

PR #17431, opened 2023-05-23
chore: add sort-imports rule

KaelWD:
@johnleider now we can stop going back and forth between "lines sorted" and "sources sorted"

It also puts imports in the proper group


packages/vuetify/src/components/VSlider/VSlider.tsx
```diff
@@ -2,16 +2,16 @@
 import './VSlider.sass'
 
 // Components
-import { makeVInputProps, VInput } from '@/components/VInput/VInput'
-import { VLabel } from '@/components/VLabel'
 import { VSliderThumb } from './VSliderThumb'
 import { VSliderTrack } from './VSliderTrack'
+import { makeVInputProps, VInput } from '@/components/VInput/VInput'
+import { VLabel } from '@/components/VLabel'
```
johnleider, line 8: Due to how these are imported they look like they should be swapped in order. Should we just change all imports to go directly to the component and not that component's index file?

```tsx
import { VLabel } from '@/components/VLabel/VLabel'
```

packages/vuetify/src/labs/VDataTable/VDataTableRows.tsx
```diff
@@ -1,23 +1,25 @@
+
```
johnleider, line 1:
```suggestion
```

packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.cy.tsx
```diff
@@ -1,8 +1,9 @@
 /// <reference types="../../../../types/cypress" />
 
+// Components
 import { VBadge } from '..'
-import { VBtn } from '@/components/VBtn'
 import { generate, gridOn } from '@/../cypress/templates'
+import { VBtn } from '@/components/VBtn'
```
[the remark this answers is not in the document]
KaelWD: `@/..` < `@/co`

johnleider (approved):

PR #17432, opened 2023-05-23
fix: Keyframe property value “” is invalid warning

### LINKED ISSUE #16258: [Bug Report][3.0.4] VTab generates invalid keyframe background-color value
### Environment
**Vuetify Version:** 3.0.4
**Vue Version:** 3.2.45
**Browsers:** Mozilla Firefox
**OS:** Linux x86_64

### Steps to reproduce
- open codepen in firefox (used 107.0 on linux)
- open browser console (not just codepen console)
- flip between tabs and watch warning in console:
```
Keyframe property value “” is invalid according to the syntax for “background-color”. animation.ts:46:23
Keyframe property value “” is invalid according to the syntax for “transform”.
```
- using a breakpoint, this is caused by https://github.com/vuetifyjs/vuetify/blob/next/packages/vuetify/src/components/VTabs/VTab.tsx#L98

### Expected Behavior
No warnings

### Actual Behavior
```
Keyframe property value “” is invalid according to the syntax for “background-color”. animation.ts:46:23
Keyframe property value “” is invalid according to the syntax for “transform”.
```

### Reproduction Link
[https://codepen.io/codesurfr/pen/MWXRvWv](https://codepen.io/codesurfr/pen/MWXRvWv)

### Other comments
I'm flagging this because it is polluting the browser log with warnings and I can't see other actually useful warnings. Dunno what your policy on such warnings is. The warnings are not shown in an up-to-date vivaldi browser, for example.

alexandreec:
Fixes #16258

### DIFF
--- a/packages/vuetify/src/components/VTabs/VTab.tsx
+++ b/packages/vuetify/src/components/VTabs/VTab.tsx
@@ -86,11 +86,11 @@ export const VTab = genericComponent()({
 
         const sigma = 1.5
         animate(nextEl, {
-          backgroundColor: [color, ''],
+          backgroundColor: [color, 'transparent'],
           transform: [
             `translate${XY}(${delta}px) scale${XY}(${initialScale})`,
             `translate${XY}(${delta / sigma}px) scale${XY}(${(scale - 1) / sigma + 1})`,
-            '',
+            'none',
           ],
           transformOrigin: Array(3).fill(origin),
         }, {


KaelWD:
I already tried this, it causes a flicker as the transition ends.

alexandreec:
@KaelWD yes, you are right, in Firefox not happens, we can just remove the empty string from the array, in both browsers, I can't see the difference in the animation :

`backgroundColor: [color],`

KaelWD:
Now set a different slider-color on each tab.

Issue #7295, opened 2019-05-26
[Feature Request] Optimize controls for test automation


Phantom75:
### Problem to solve
Test automation is key for DevOps. Whereas most (if not all) vuetify-components **are** automatable with selenium, ranorex, tosca & co, some do make live of a test automator unnecessary hard.
The wish is: make each component in every configuration as easy automatable as possible.

### Proposed solution
E.g. a select box:
```
v-select  id="contest_persons_groups" :items="registrationtypes" v-model=editedItem.type
                              label="Einzelperson oder Gruppe" item-text="value" item-value="ID"></v-select>
```
results in the following html code:
```
<div role="combobox" class="v-input v-text-field v-select v-input--is-label-active v-input--is-dirty v-input--is-focused theme--light primary--text">
    <div class="v-input__control">
        <div class="v-input__slot">
            <div class="v-select__slot">
                <label for="contest_persons_groups" class="v-label v-label--active theme--light primary--text" style="left: 0px; right: auto; position: absolute;">Einzelperson oder Gruppe</label>
                <div class="v-select__selections">
                    <div class="v-select__selection v-select__selection--comma">Einzelperson</div>
LOOK HERE -->         <input id="contest_persons_groups" readonly="readonly" type="text" autocomplete="on" aria-readonly="false">**
                </div>
                <div class="v-input__append-inner">
                    <div class="v-input__icon v-input__icon--append"><i aria-hidden="true" class="v-icon fas fa-caret-down theme--light primary--text"></i></div>
                </div>
            </div>
            <div class="v-menu"></div>
        </div>
        <div class="v-text-field__details">
            <div class="v-messages theme--light primary--text">
                <div class="v-messages__wrapper"></div>
            </div>
        </div>
    </div>
</div>
```
The id went to the input which is not clickable, as it is readonly. You had to solve this with nasty xpath-give-me-your-parent-parent-magic:
``` $('//*[@id="country"]/..').click() ```
if the id was set to a clickable div, the automation code would be so much nicer and more readable:
```$('#country"').click()```
Now I am aware, that this may be tricky. If the dropdown was typable, of course, having the id on the input were totally correct. So there is decision magic necessary.

nicoloboschi:
Perhaps a solution could be to add a "semantic" tag to each component, for example to click a select:

`$('[data-vuetify-ref=v-select-open]').click()`

in this way you avoid using the id (components in read and write normally use different elements).
[...]

PR #17435, opened 2023-05-23
feat(VList, VSelect): add support for test IDs

jacob-carlborg-apoex:
Test IDs are added to provide stable points in the rendered HTML which tests can rely on when creating selectors. The test will be less dependent on the exact structure of the HTML and not having to rely on CSS classes, which are primarily used for styling. I primarily thinking of tests for sites which uses Vuetify components and not necessarily testing of Vuetify itself (but it can be used for that as well).

I would normally create separate pull requests but the internal types for `VList` and `VSelect` are so tightly coupled that it would be difficult.

Ideally I would like to create a new section in the API documentation which would list a component's test IDs and a description, but I did not managed to do that. I don't understand enough about how the documentation is generated.

Related: https://github.com/vuetifyjs/vuetify/issues/7295.

### DIFF
--- a/packages/api-generator/src/locale/en/list-items.json
+++ b/packages/api-generator/src/locale/en/list-items.json
@@ -2,6 +2,7 @@
   "props": {
     "itemTitle": "Property on supplied `items` that contains its title",
     "itemValue": "Property on supplied `items` that contains its value",
+    "itemTestId": "Property on supplied `items` that contains its test ID",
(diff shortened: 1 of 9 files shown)


jacob-carlborg-apoex:
Ping. Is there anyone that can take a look at this PR?

jacob-carlborg-apoex:
Related: https://github.com/vuetifyjs/vuetify/issues/7295.

PR #17447, opened 2023-05-24
fix(VDataTable): expose sub components

MathisTLD:
## Description

When using slots like `data-table-group` with the `VDataTable` component, we could need to use components like `VDataTableColumn` (see example in `Playground.vue`). Then we need to expose these.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <VDataTable
        :headers="headers"
        :items="desserts"
        :group-by="[{key:'fat_group', order:'asc'}]"
        item-value="name"
        class="elevation-1"
      >
        <template #data-table-group="{ item, props }">
          <VDataTableColumn class="v-data-table-group-header-row__column">
            <v-btn
              v-bind="props"
              size="small"
              variant="text"
            />
            <span :style="{fontWeight:300+item.value*100}">{{ getGroupName(item.value) }}</span>
            <span>({{ item.items.length }})</span>
          </VDataTableColumn>
        </template>
      </VDataTable>
    </v-container>
  </v-app>
</template>

<script>
  import { VDataTable, VDataTableColumn } from 'vuetify/src/labs/VDataTable'
  export default {
    name: 'Playground',
    components: {
      VDataTable,
      VDataTableColumn,
    },
    setup () {
      const headers = [
        {
          title: 'Dessert (100g serving)',
          align: 'start',
          sortable: false,
          key: 'name',
        },
        { title: 'Calories', key: 'calories' },
        { title: 'Fat (g)', key: 'fat' },
        { title: 'Carbs (g)', key: 'carbs' },
        { title: 'Protein (g)', key: 'protein' },
        { title: 'Iron (%)', key: 'iron' },
      ]
      const desserts = [
        {
          name: 'Frozen Yogurt',
          calories: 159,
          fat: 6.0,
          carbs: 24,
          protein: 4.0,
          iron: 1,
        },
        {
          name: 'Ice cream sandwich',
          calories: 237,
          fat: 9.0,
          carbs: 37,
          protein: 4.3,
          iron: 1,
        },
        {
          name: 'Eclair',
          calories: 262,
          fat: 16.0,
          carbs: 23,
          protein: 6.0,
          iron: 7,
        },
        {
          name: 'Cupcake',
          calories: 305,
          fat: 3.7,
          carbs: 67,
          protein: 4.3,
          iron: 8,
        },
        {
          name: 'Gingerbread',
          calories: 356,
          fat: 16.0,
          carbs: 49,
          protein: 3.9,
          iron: 16,
        },
        {
          name: 'Jelly bean',
          calories: 375,
          fat: 0.0,
          carbs: 94,
          protein: 0.0,
          iron: 0,
        },
        {
          name: 'Lollipop',
          calories: 392,
          fat: 0.2,
          carbs: 98,
          protein: 0,
          iron: 2,
        },
        {
          name: 'Honeycomb',
          calories: 408,
          fat: 3.2,
          carbs: 87,
          protein: 6.5,
          iron: 45,
        },
        {
          name: 'Donut',
          calories: 452,
          fat: 25.0,
          carbs: 51,
          protein: 4.9,
          iron: 22,
        },
        {
          name: 'KitKat',
          calories: 518,
          fat: 26.0,
          carbs: 65,
          protein: 7,
          iron: 6,
        },
      ].map(x => ({ ...x, fat_group: Math.round(x.fat / 10) }))
      return {
        headers,
        desserts,
        getGroupName (value) {
          if (value > 1) {
            return Array(value - 1).fill('very').join(' ') + ' fat'
          } else {
            return value === 1 ? 'fat' : 'not fat'
          }
        },
      }
    },
  }
</script>
```


nekosaur:
I am not sure we want to expose these components. There is value in being able to make large changes to internal structure and architecture without having to make them breaking changes requiring a major version release.

nekosaur:
> we already expose the row why not exposing the col ? Plus now this is a labs component so breaking changes might occur anyway

I'm not sure row component _should_ be exposed. And using labs as a reason doesn't really work because at some point data-table will move out of labs.

MathisTLD:
> > we already expose the row why not exposing the col ? Plus now this is a labs component so breaking changes might occur anyway
> 
> I'm not sure row component _should_ be exposed. And using labs as a reason doesn't really work because at some point data-table will move out of labs.

Then what solutions do you propose for people trying to use slots such as `data-table-group`. They should re-implement VDataTableColumn themselves using CSS classes that might be changed by "structural changes" you mentioned earlier ? Then they won't be able to detect breaking changes (as miss-spelled CSS classes won't raise errors when Typescript does). I mean not exposing anything to be able to make changes without authoring new version would mean we should remove v-list-item-<anything> and many other similar component which would make developing with vuetify a nightmare.

I agree that exposing everything isn't a good practice but in this case I find it quite fair considering how we dealt with other components

johnleider:

PR #17483, opened 2023-05-28
fix(VDataTableVirtual): render expanded and grouped rows correctly

nekosaur:
## Description
Change how expanded rows work so that they also can use virtual scroll item.

## TODO

This can probably be cleaned up a bit. Not super happy about the `hasKey` thing in virtual composable.

## Markup:

```vue
<v-data-table-virtual
    :headers="headers"
    :items="virtualDesserts"
    class="elevation-1"
    height="500"
    show-expand
    :group-by="[{ key: 'name', order: 'asc'}]"
  >
    <template #expanded-row="{ props }">
      <tr v-bind="props">
        <td colspan="7">
          <div>-> expanded</div>
        </td>
      </tr>
    </template>
    <template #item="{ item, props }">
      <tr v-bind="props">
        <td>FOOO</td>
      </tr>
    </template>
  </v-data-table-virtual>
```

### DIFF
--- a/packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
+++ b/packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
@@ -10,7 +10,7 @@ import { makeDimensionProps, useDimension } from '@/composables/dimensions'
 import { makeVirtualProps, useVirtual } from '@/composables/virtual'
 
 // Utilities
-import { ref, toRef } from 'vue'
+import { toRef } from 'vue'
 import {
   convertToUnit,
   genericComponent,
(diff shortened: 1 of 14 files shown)


packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
```diff
@@ -82,13 +81,13 @@ export const VVirtualScroll = genericComponent<new <T>(
             paddingBottom: convertToUnit(paddingBottom.value),
           }}
         >
-          { computedItems.value.map(item => (
+          { virtualItems.value.map((item, index) => (
             <VVirtualScrollItem
-              key={ item.index }
+              key={ index }
```
KaelWD, line 86: Items need to be keyed to the original index so the element is reused.

KaelWD (changes requested):
This will also conflict with #17265 again

J-Sek:
partially covered already, closing in favor of #23014

PR #17495, opened 2023-05-30
fix(VTab): remove firefox invalid keyframe property

### LINKED ISSUE #16258: [Bug Report][3.0.4] VTab generates invalid keyframe background-color value

alexandreec:
Fixes #16258


KaelWD:
This has the same problem as your previous PR.

https://github.com/vuetifyjs/vuetify/assets/16421948/31f86713-e14b-4924-a482-3668efdd2f87

alexandreec:
@KaelWD I see now, sorry.

I just removed the `backgroundColor` option from the animate function, now it seems that it works properly, 
* Tested it in Firefox, Opera, Edge and Chrome (linux versions)

https://github.com/vuetifyjs/vuetify/assets/15526814/abe3426b-4ed7-406e-a772-1973cf65f92e

alexandreec:
@KaelWD I prefer this rather than having an annoying warning in Firefox.

lacksfish:
This fix would be needed for Firefox, currently without it warnings are being thrown. 

@alexandreec Possibly another similar fix is needed for the "transform" warning ?

image

PR #17569, opened 2023-06-09
feat: utility classes in color props work with or without prefixes

cgodo:
## Description

Some Vuetify components accept color properties using utility classes without prefixes, such as `success` instead of `text-success`. This works seamlessly with Vuetify as all color classes are bundled in the final build. But, when a different utility class library like Tailwind is used, which purges classes based on usage, a class like `red-500` will be purged by Tailwind, due to the absence of the `text-` or `bg-` prefix.

The proposed modification allows the color composable to accept utility classes either with or without the prefix. Therefore, both `text-success` and `success` will function identically, which will enable Tailwind users to specify the complete class to prevent its removal during the purge process.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <h3>Using utility colors without prefixes:</h3>
      <v-text-field variant="outlined" color="blue" base-color="pink" bg-color="cyan-lighten-5"></v-text-field>
      <v-chip color="primary"> Primary </v-chip>
      <v-switch color="primary" loading="secondary"> Primary </v-switch>

      <v-divider class="my-10"></v-divider>

      <h3>Using utility colors with prefixes:</h3>
      <v-text-field variant="outlined" color="text-blue" base-color="text-pink" bg-color="bg-cyan-lighten-5"></v-text-field>
      <v-chip color="text-primary"> Primary </v-chip>
      <v-switch color="text-primary" loading="text-secondary"> Primary </v-switch>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
      }
    },
  }
</script>

```


cgodo:
Not sure I can do much about the failing Cypress. It seems to be a known issue with Cypress unrelated to this PR. More info here: https://github.com/cypress-io/cypress/issues/25913.

```
  Running:  components/VAlert/__tests__/VAlert.spec.cy.tsx                                 (1 of 50)


  1) An uncaught error was detected outside of a test

  0 passing (680ms)
  1 failing

  1) An uncaught error was detected outside of a test:
     TypeError: The following error originated from your test code, not from Cypress.

  > Failed to fetch dynamically imported module: http://localhost:5173/__cypress/src/cypress/support/index.ts
```

PR #17568, opened 2023-06-09
fix(VOverlay): allow click outside even if content is not on top

yuwu9145:
fixes #17488

tested #15475, #15276

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-row>
        <v-col>
          <v-select
            class="item"
            placeholder="Select1"
            :menu-props="{ }"
            :items="[
              { key: 1, title: 'title1' },
              { key: 2, title: 'title2' },
            ]"
          />
        </v-col>
        <v-col>
          <v-autocomplete
            class="item"
            placeholder="Select2"
            :items="[
              { key: 3, title: 'title3' },
              { key: 4, title: 'title4' },
            ]"
          />
        </v-col>
      </v-row>

      <div class="d-flex justify-space-around">
        <v-btn>
          click me and press tab
          <v-tooltip activator="parent" location="start">Tooltip</v-tooltip>
        </v-btn>

        <v-btn>
          default
          <v-tooltip activator="parent" location="end">Tooltip</v-tooltip>
        </v-btn>

        <v-btn>
          open on focus
          <v-tooltip
            activator="parent"
            location="top"
            open-on-focus
          >Tooltip</v-tooltip>
        </v-btn>

        <v-btn>
          default
          <v-tooltip activator="parent" location="bottom">Tooltip</v-tooltip>
        </v-btn>
      </div>
      <v-tooltip text="I'm a cool tip">
        <template #activator="{ props }">
          <v-btn color="primary" @click="snackbar = true" v-bind="props">Click Me</v-btn>
        </template>
      </v-tooltip>
    </v-container>

    <v-snackbar v-model="snackbar" :timeout="-1">
      Snackbar content
      <template #actions>
        <v-btn @click="snackbar = false">Close</v-btn>
      </template>
    </v-snackbar>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const snackbar = ref(false)
</script>


```

PR #17594, opened 2023-06-13
chore: improve stack logic to better serve a convincible cy test

yuwu9145:
This PR is opened because found a hidden issue in stack when constructing a convincible cy test case for #17448

The test case should fail on master branch, but it's successful, this is because `globalTop` inside setTimeout behaves differently in cypress from normal user behaviour

I'm not entirely sure the initial purpose of using `setTimeout` here but replacing it with `nextTick` performs more safely

The purpose of opening this PR is to start conversation around and get feedback

### DIFF
--- a/packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
+++ b/packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
@@ -9,6 +9,7 @@ import { VListItem } from '@/components/VList'
 import { cloneVNode, ref } from 'vue'
 import { generate } from '../../../../cypress/templates'
 import { keyValues } from '@/util'
+import { VAutocomplete, VCol, VCombobox, VRow } from '@/components'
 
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VSelect/__tests__/VSelect.spec.cy.tsx
```diff
@@ -436,6 +437,46 @@ describe('VSelect', () => {
...
+
+      cy.get('.v-autocomplete')
+        .click()
+        .trigger('keydown', { key: keyValues.down, waitForAnimations: false })
+        .trigger('keydown', { key: keyValues.down, waitForAnimations: false })
+        .get('.v-overlay__content.v-select__content')
+        .should('not.exist')
+        .get('.v-overlay__content.v-autocomplete__content')
+        .should('exist')
+
```
johnleider, line 477:
```suggestion
```

yuwu9145:
This PR is not supposed to be merged.

It's created to raise this issue and see if it is right direction. The proper implementation is in #17568

PR #17618, opened 2023-06-15
docs: add v-img error slot

Path-A:
## Description

Added information about the v-img error slot and included a common use-case for a fallback image if the source image is cannot be loaded.


MajesticPotatoe (approved):
👍 great example

PR #17653, opened 2023-06-19
fix(VMenu): don't close on tab if focusable content

avarixe:
## Description

fixes #17645 

## Markup:

```vue
<template>
  <v-app>
    <v-app-bar>
      <v-toolbar-title>
        Hello World
      </v-toolbar-title>
      <v-spacer />
      <v-menu>
        <template #activator="{ props }">
          <v-btn color="blue" v-bind="props">
            <div>Joe User</div>
          </v-btn>
        </template>
        <v-list>
          <v-list-item @click=";">
            <v-list-item-title style="font-size: 14px">
              Logout
            </v-list-item-title>
          </v-list-item>
        </v-list>
      </v-menu>
    </v-app-bar>
    <v-main>
      <v-container>
        <v-btn>
          Test 1
          <v-menu activator="parent" :close-on-content-click="false">
            <v-card>
              <v-card-text>
                <v-text-field />
                <v-textarea />
                <v-select />
                <v-radio />
                <v-slider />
                <v-switch />
                <v-text-field />
              </v-card-text>
            </v-card>
          </v-menu>
        </v-btn>
        <v-select label="Select 1" :items="['Item 1', 'Item 2']" />
        <v-select label="Select 2" :items="['Item 1', 'Item 2']" />
        <v-btn>Test 2</v-btn>
      </v-container>
    </v-main>
  </v-app>
</template>
```

PR #17674, opened 2023-06-22
fix(VMenu): don't close on tab if content is tabbable

### LINKED ISSUE #17645: [Bug Report][3.3.4] VTextField does not navigate correctly on tab press in VMenu
### Environment
**Vuetify Version:** 3.3.4
**Last working version:** 3.3.2
**Vue Version:** 3.3.4
**Browsers:** Chrome 114.0.0.0
**OS:** Windows 10

### Steps to reproduce
1. add VMenu with at least 2 VInputs inside (type does not seem to matter)
2. focus on first input field
3. press TAB key

### Expected Behavior
Browser focuses on second input field

### Actual Behavior
VMenu closes and browser focuses on next element at activator level

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlUk1v00AQ/SvDXgJSbFdqT1GKyg0OiAsSh7qHjT0mS/dLu2OnKMp/Z3btNo4rpApO9rw3+9583R/FJ+/LoUexEVtC47Uk/FhbgO1QSO/zbw4aZ0kqi2GCMrgj+xICfPNo4Sva/gxxjrkAADaNdhELZ7MkWioarZrH21p0UkesxTxZNqQGSS4w7WXg7Ev+oFraM3dzdeWf5tSsrLF8GdoL7AUtCJ9oQWUy4UWnULfAXbgWNRuZ+LMWoOUuR9/lDrrgDOwxcOUge3Kda/oI1T8qfoGDtATkYNSZhJdy2+pvxT8zMzBBaQvnzTFwXl1+cbHdhIzL31azo+AwNkF5gojUj7ehjHeB4AgBOziNw1jxPa1SOgDrRgJuEW5TxvvVZ9TawQ8XdPtu9SEZjJIsJtZiVCuM9OWv6Cwf5TGp1BMRa7GBjCSMXVJciz2Rj5uqalrLz3iqagilRaqsN9Udp1Wht6QMFq0zd9fldXlTtSrSHC4xmmIX3CFiYJFarGc2pLrfb7CaMucOE1TwemOyeCVdsd+AoeDTbnnV4a0dLZ7NPRfUq86S+6m2J563VvYxLkbdxDzm++ci/6vlLJaUHpLhaS1yIjuP34c/mQBqWA==)

avarixe:
## Description

~~this PR adds a `closeOnTab` prop to VMenu, default value to true for backwards compatibility. Just like closeOnContentClick, a developer would set this prop to false to disable closing menu on tab press, so that one can navigate the menu contents if necessary.~~

fixes #17645

My original PR (#17653) tried to use closeOnContentClick instead, but was rejected as breaking Test 2 below (being able to use tab to close out of the Test 2 Menu).

Checking for focusable content inside the menu would also break Test 2 since the VListItems all have tabindex="-2" (and the fact that they are focusable is used for the keyboard navigation).

edit: revised to check for tabbable content as advised

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-btn>
        Test 1
        <v-menu activator="parent" :close-on-content-click="false">
          <v-card>
            <v-card-text>
              <v-text-field />
              <v-textarea />
              <v-select />
              <v-radio />
              <v-slider />
              <v-switch />
              <v-text-field />
            </v-card-text>
          </v-card>
        </v-menu>
      </v-btn>
      <v-select label="Select 1" :items="['Item 1', 'Item 2']" />
      <v-select label="Select 2" :items="['Item 1', 'Item 2']" />
      <v-btn>
        Test 2
        <v-menu activator="parent" :close-on-content-click="false">
          <v-list>
            <v-list-item v-for="x in ['a','b','c']" :key="x" link>
              {{ x }}
            </v-list-item>
          </v-list>
        </v-menu>
      </v-btn>
    </v-container>
  </v-app>
</template>
```


yuwu9145 (commented):
```ts
export function tabbableChildren (el: Element) {
  const focusable = focusableChildren(el)
  return focusable.filter(el => el.tabIndex >= 0)
}
```

```ts
function onKeydown (e: KeyboardEvent) {
    const tabbable = tabbableChildren(overlay.value?.contentEl)
    if (props.disabled) return

    if (e.key === 'Tab' && !tabbable.length) {
        isActive.value = false
        overlay.value?.activatorEl?.focus()
   }
}
```

yuwu9145 (approved):

johnleider:
Using the example provided in the original issue is still broken.

```vue
<template>
  <v-app>
    <v-container>
      <v-btn>
        Open Menu
        <v-menu
          activator="parent"
          width="400px"
        >
          <v-card>
            <v-card-text>
              <v-text-field v-model="msg" label="Tab from here" autofocus />
              <v-text-field v-model="msg" label="I want to focus here" />
            </v-card-text>
          </v-card>
        </v-menu>
      </v-btn>

      <v-text-field label="And then focus here" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

</script>
```

Also, the menu stays open even after tabbing out of it.

avarixe:
adjusted so that it checks for the next (previous is Shift+Tab) tabbable element instead

johnleider:
Using your example the menu inputs are put at the end of the tabindex.

focus

KaelWD:
This breaks down a bit when you mix tabbable and non-tabbable elements:
```html
<v-menu activator="parent" :close-on-content-click="false">
  <v-sheet>
    <v-text-field />
    <v-list>
      <v-list-item v-for="x in ['a','b','c']" :key="x" link>
        {{ x }}
      </v-list-item>
    </v-list>
  </v-sheet>
</v-menu>
```

If the last list item is focused the menu doesn't close on tab. Adding `|| !~targetIndex` to the condition prevents the following text field from being focused:

```html
<v-menu activator="parent" :close-on-content-click="false">
  <v-sheet>
    <v-list>
      <v-list-item v-for="x in ['a','b','c']" :key="x" link>
        {{ x }}
      </v-list-item>
    </v-list>
    <v-text-field />
  </v-sheet>
</v-menu>
```

@johnleider unrelated, it does that on master too.

avarixe:
@KaelWD to address the 2nd example, I "floated" the focus to a tabbable element before checking for the next element. So the user can then press tab to go from the VList to the VTextField.

I needed to remove the tabindex="-1" when VList is focused to make this work. Is that ok? I did some checks to see whether it affected keyboard navigation and it seemed to work the same.

@johnleider this "close on tab" behavior happens even if the `persistent` prop is set to true. Is that intended? If not, should I make a separate issue and PR for it, or just handle it in this PR?

johnleider:
> @johnleider this "close on tab" behavior happens even if the persistent prop is set to true. Is that intended? If not, should I make a separate issue and PR for it, or just handle it in this PR?

I imagine it would benefit a11y to let it close on tab. @KaelWD what are your thoughts?

KaelWD:
persistent generally means it can only be closed by changing v-model

packages/vuetify/src/components/VList/VList.tsx
```diff
@@ -228,7 +228,7 @@ export const VList = genericComponent<new <T>(
             dimensionStyles.value,
             props.style,
           ]}
-          tabindex={ (props.disabled || isFocused.value) ? -1 : 0 }
```
KaelWD, line 231: This was so the isn't an "invisible" tab stop on shift+tab

```
<v-btn>Button</v-btn>
<v-list>
  <v-list-item link>List</v-list-item>
</v-list>
```

avarixe: understood, I changed the approach using a different helper getTabbableChild that uses a modified version of focusableChildren and focusChild so that the VList tabindex doesn't need to change

packages/vuetify/src/util/helpers.ts
```diff
@@ -596,6 +596,22 @@ export function focusChild (el: Element, location?: 'next' | 'prev' | 'first' |
   }
 }
 
+export function getTabbableChild (el: Element, location?: 'next' | 'prev') {
+  const targets = ['button', '[href]', 'input:not([type="hidden"])', 'select', 'textarea', '[tabindex]']
+    .map(s => `${s}:not([disabled])`)
+    .join(', ')
+  const children = [...el.querySelectorAll(targets)] as HTMLElement[]
```
johnleider, line 603: This is already exists in a function: https://github.com/vuetifyjs/vuetify/pull/17674/files#diff-6daf9735657f26888ecc2f70a2fa7b67dfef7600ab7275e0ae89b111dd461c6dR567

packages/vuetify/src/util/helpers.ts
```diff
@@ -596,6 +596,22 @@ export function focusChild (el: Element, location?: 'next' | 'prev' | 'first' |
...
+
+  let _el
+  let idx = children.indexOf(document.activeElement as HTMLElement)
+  const inc = location === 'next' ? 1 : -1
+  do {
```
johnleider, line 608: This might be worth abstracting into a shared function since it's basically this: https://github.com/vuetifyjs/vuetify/pull/17674/files#diff-6daf9735657f26888ecc2f70a2fa7b67dfef7600ab7275e0ae89b111dd461c6dR590

avarixe: abstracted into getNextElement()

packages/vuetify/src/util/helpers.ts
avarixe: I had to add an optional arg to focusableChildren to use it for what I needed it for (without the `:not(tabindex="-1")` condition).

johnleider:
I'm still seeing this:

menu1

If I press **DownArrow** instead of **Tab**, it properly highlights to the first element. From this point, tabbing works as expected.

menu2

avarixe:
@johnleider I'm seeing this behavior in v2 as well:

Recording 2023-06-29 142034

I copied the same logic that focuses into the content of VDialog after it is opened to do the same for VMenu. If you have a different idea of how to approach this, I'm all for trying it if you want me to.

I initially thought the first part of the listener could be abstracted into a helper function in @/components/VOverlay/VOverlay.tsx, but ran into a typescript argument error when I tried to pass overlay.value into this function:

```ts
export function onVOverlayFocusIn(e: FocusEvent, overlay: VOverlay, callback: () => void) {
  const before = e.relatedTarget as HTMLElement | null
  const after = e.target as HTMLElement | null

  if (
    before !== after &&
    overlay?.contentEl &&
    // We're the topmost overlay
    overlay?.globalTop &&
    // It isn't the document or the overlay body
    ![document, overlay.contentEl].includes(after!) &&
    // It isn't inside the overlay body
    !overlay.contentEl.contains(after)
  ) {
    callback()
  }
}
```

Is there something I'm missing to make this work?

johnleider:
> @johnleider I'm seeing this behavior in v2 as well:

Fair enough. I'd still like a final okay from @KaelWD though.

avarixe:
@KaelWD is there anything else I need to correct? 

perfectly understand if you guys have been busy with other PR's and issues. I love using vuetify and am excited to see it continue to reach parity with v2

johnleider (approved):

PR #17685, opened 2023-06-23
feat(VStepper): port to v3

### LINKED ISSUE #2452: [Feature Request] Stepper custom step label
### New Functionality
Allow you to override the label currently set by the 'step' property (limited to first 2 characters).

### Improvements
there currently is no workaround that i know of. 

this would allow you to dynamically customize the label of the step without sacrificing the order
the step could still be numerical (1,2,3) but the labels will not need to be so.

### LINKED ISSUE #3546: [Feature Request] Stepper with large number of steps-  Only show current and nearby steps.
### New Functionality
It will allow you to use large number of steps in stepper with better look.
Right now, if you make a stepper with say 20 steps in my case, it looks cumbersome and scrolling doesn't look good. 

I was wondering if there was some way for only the steps around the current step to be visible
(Say i was on step 11 so i could see steps 10 through 14 in my header and the rest would be dotted)

### Improvements
I'm fairly new to this but I believe this will make it look a lot better and user-friendly.

I'm not sure... I'm sorry...

### LINKED ISSUE #3963: [Feature Request] Additional stepper icon support
## Issue

I'd like to create a stepper that allows you to edit previously completed steps, but not allow jumping ahead until a step is completed. I'd like all completed steps to have a check mark, the current step to have an editing icon, and the default styling for future steps (as shown in the photo and code pen below).

I'd like the ability to further customize the icons for the stepper.

Here's what I'm trying to accomplish: https://codepen.io/XanderLuciano/pen/ELZwQL

image
So in order to accomplish this I am setting step properties like so:

```
<v-stepper-step
:complete-icon="'edit'"
:edit-icon="'check'"
:complete="(index + 1 ) <= stepper" 
:editable="(index + 1) < stepper" ...>
```

i.e. All previous steps are editable, and all steps are complete *including* the current one, and all future steps are not editable and not completed. Then in order to get the icons to display as I want, I have to swap the edit and completed icons.

This is because the "edit" icon overrides the completed icon. So by reversing the roles, I can have a step that appears correctly, but is coded "backwards"

## Proposed solution:

I think the idea of keeping it simple should be used and instead of making some type of configuration to set icon priority, I'd like the ability to just directly set the icon or icon state for each step. Something like: `:icon-state="editable"` / `:icon-state="completed"` or :icon="edit"` / `:icon-state="check"`

## Why is this an actual issue?

The only reason why this is an "issue" is because I don't like the idea of having to set an incomplete step as "completed" prematurely. My options are either:  
1) Change the edit icon to a check mark - issue: The current step won't have the edit icon
2) Don't make previous editable, instead use custom styling and click event handler to manually create editable functionality - issue: too overkill
3)Don't make previous editable, add a back button - Poor UX compared to editable behavior
4) Swap editable and completed icons - cons: current step now registers as complete, while it's incomplete

It's not a big issue, but seems like it should be fairly straightforward to override the current icon with a custom property. If you're open to the idea, but feel it is more effort than it's worth, I wouldn't mind creating a PR with the proposed solution. Just wanted to start a discussion before going through the trouble of learning the inner workings of veutify.

### LINKED ISSUE #5064: [Feature Request] v-stepper should be more interactive like vue-form-wizard

it will be having universal action control for next, previous and finish step. which current stepper doesn't have as of now.
It should have basic functionality like before step change function, dynamic rendering for steps.

v-stepper shoud give basic functionalities like vue-form-wizard

### LINKED ISSUE #6745: [Bug Report] Property edit-icon of v-stepper-step doesn't work unless step is both editable and complete
### Versions and Environment
**Vuetify:** 1.5.6
**Vue:** 2.6.6
**Browsers:** Chrome 72.0.3626.121
**OS:** Mac OS 10.14.3

### Steps to reproduce
Create v-stepper-step with edit-icon and editable properties. Doesn't show the custom icon. Add complete property and it works.

### Expected Behavior
Show the custom icon passed in the edit-icon property as long as step is editable, even though it is not complete.

### Actual Behavior
Custom icon only shows when step is also complete.

### Reproduction Link
<a href="https://codepen.io/anduhd/pen/ZEpOBdM?editors=1010" target="_blank">https://codepen.io/anduhd/pen/ZEpOBdM?editors=1010</a>

**edit: fixed link**

### LINKED ISSUE #7049: [Feature Request] Add a slot to be able to change the icon template in v-stepper-step
### Problem to solve
This would allow for more costumization of the stepper icon, currently the icon can only be passed as editableIcon, errorIcon, completedIcon

This would allow to change the logic of the displayed icons and to give more flexibility for their display (For example to add an image, or to change the default number or text to an icon, and to change the size and color of the stepper icon)

Related #3963

### Proposed solution
To add a scoped slot named "icon" that will pass the props of the parent (hasError, completed, editable, isActive, isInactive and the icons props) so that we can declare our components like so
```HTML
<v-stepper-step>
    <template #icon="step">
        <v-icon v-if="step.isActive">ActiveIcon</v-icon>
        <v-icon v-else-if="step.editable">{{step.editableIcon}}</v-icon>
        //Etc ...
    </template>
    Step Label
</v-stepper-step>
```

### LINKED ISSUE #8705: [Feature Request] Disabled property for v-stepper-step
### Problem to solve
I have a situation where 2 of the 6 steps of the v-stepper depends on the selection the user makes on the first step. They don't make any sense, if the user chooses option A, and makes sense when chooses option B.

### Proposed solution
v-stepper-step shoud have a disabled property. If the step is disabled, then it should be gray, not selectable, and skipped by navigation.

### LINKED ISSUE #9047: [Feature Request] Stepper - Small stepper label font-weight too light
### Problem to solve
Currently, the optional second label you can use in a stepper step is font-weight: 300. I strongly believe that a light font-weight should be not used for text this small. It's very hard to read and we've had multiple complaints about it. 
image

### Proposed solution
I would say just make it regular weight (400).

### LINKED ISSUE #9613: [Feature Request] Stepper tile prop, to remove the small border-radius.
### Problem to solve
Steppers being inconsistent with other card-like components in having props for borders/elevation/etc.

### Proposed solution
Add the `tile` prop and any other relevant props.

### LINKED ISSUE #9975: [Bug Report] v-stepper-step[editable] not accessible with keyboard
### Environment
**Vuetify Version:** 2.1.14
**Vue Version:** 2.6.10
**Browsers:** Chrome 79.0.3945.79
**OS:** Mac OS 10.15.1

### Steps to reproduce
1. Open the codepan
2. Use tab navigation to move through the page. Notice that Step 1 and Step 2 are never focused.
3. Click on Step 2, then on Step 1 using the mouse pointer. Notice these two steps are interactive

### Expected Behavior
When given the "editable" property, v-stepper-focus should be focusable, have a role="button" attribute, and and respond to pressing enter or space.

Also noticed that even when tabindex is added to v-stepper-step, @click events can still not be fired using keyboard.

### Actual Behavior
The v-stepper-step with an editable attribute is not focusable, has no role, and when these are added, @click can still not be fired from it.

### Reproduction Link
<a href="https://codepen.io/wilcofiers/full/zYxKXpg" target="_blank">https://codepen.io/wilcofiers/full/zYxKXpg</a>

### LINKED ISSUE #10007: [Feature Request] V-Stepper footer component
### Problem to solve
At the moment, there is no way to define a common footer, which appears on every step, without having to repeatedly define it in each step.

The anticipated use case would be where you have a button bar at the bottom which has buttons like Back, Continue, Cancel for example. Although I could add a separate bar which does this immediately below the v-stepper component, it doesn't look as though it's part of the stepper if I do that, whereas I want it to look like an integrated part of the stepper.

Given that we already have a v-stepper-header component, it seems reasonable to have a footer component too.

### Proposed solution
Implement a v-stepper-footer component. Any content in it's default slot would be shown in the step footer.

For horizontal steppers, it would appear in the same place on every step, at the bottom of the v-stepper.

For vertical steppers, it would appear at the bottom of the current step, rather than at the very bottom of the stepper itself.

### LINKED ISSUE #10088: [Feature Request] Add Default Icon for V-Stepper-Step
### Problem to solve
Currently you are only able to show a Number or text in the default mode of the Step.  I would like to show an Icon instead.  I have tried using `<i class="v-icon notranslate mdi mdi-account theme--dark"></i>` but it doesn't render html.

### Proposed solution
Along with error-icon, complete-icon, edit-icon, I would like to have just `icon`
If not set then it uses the default number, Otherwise I would like to show an Icon.

### LINKED ISSUE #10579: [Feature Request] Add ripple prop to v-stepper
### Problem to solve
We deactivate the ripple effect on all Vuetify components that we use, because it doesn't fit our design. Right now, the ripple effect on steppers cannot be controlled or turned off. 

### Proposed solution
Add the ripple prop that is present on other components like v-btn to v-stepper or one of its sub components.

### LINKED ISSUE #10586: [Bug Report] v-stepper doesn't render first step content if added dynamically
### Environment
**Vuetify Version:** 2.2.12
**Vue Version:** 2.6.11
**Browsers:** Chrome 80.0.3987.100
**OS:** Windows 10

### Steps to reproduce
Create a stepper with no steps, add steps in mounted()

### Expected Behavior
Should display first step

### Actual Behavior
Doesn't display first step

### Reproduction Link
<a href="https://codepen.io/jkarczm/pen/QWbyeMd" target="_blank">https://codepen.io/jkarczm/pen/QWbyeMd</a>

### LINKED ISSUE #13509: [Task] v-stepper
# Overview

- Related Tasks: n/a
- Related Issues: n/a

A conversion should not start until *all* assigned issues with the milestone v2.2.x are resolved and in the `next` branch. Reference our [Coding Guidelines](https://github.com/vuetifyjs/vuetify/wiki/coding-guidelines) for information regarding the team's practices or reach out to us on [Discord](https://community.vuetifyjs.com).

### LINKED ISSUE #13934: [Feature Request] v-stepper create next steps content only when accessed
### Problem to solve
When adding a v-stepper all the steps contents are rendered, the ideal would be to render only the content of the current step

### Proposed solution
The v-expansion-panels component works that way, maybe it would be a solution to copying itself

johnleider:
## Motivation and Context
closes #2452
closes #3963
closes #5064
closes #7049
closes #8705
closes #9613
closes #10007
closes #10088
closes #10579
closes #13934
fixes #3546
fixes #6745
fixes #9047
fixes #9975
fixes #10586

## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-btn class="d-none">leave here</v-btn>

      <v-switch v-model="toggle" label="Toggle" />

      <v-stepper editable>
        <template #default="{ prev, next }">
          <v-stepper-header>
            <v-stepper-item
              title="Item 1"
              value="1"
              :rules="[() => false]"
            />

            <v-divider />

            <v-stepper-item title="Item 2" value="2" />

            <v-divider />

            <v-stepper-item title="Item 3" value="3" />
          </v-stepper-header>

          <v-stepper-window>
            <v-stepper-window-item value="1">
              <v-card title="one" height="200" color="grey-lighten-1" />

              <v-stepper-actions
                v-if="toggle"
                @click:prev="prev"
                @click:next="next"
              />
            </v-stepper-window-item>

            <v-stepper-window-item value="2">
              <v-card title="two" height="300" color="grey-lighten-2" />

              <v-stepper-actions
                v-if="toggle"
                @click:prev="prev"
                @click:next="next"
              />
            </v-stepper-window-item>

            <v-stepper-window-item value="3">
              <v-card title="three" height="250" color="grey-lighten-3" />

              <v-stepper-actions
                v-if="toggle"
                @click:prev="prev"
                @click:next="next"
              />
            </v-stepper-window-item>
          </v-stepper-window>

          <v-stepper-actions
            v-if="!toggle"
            @click:prev="prev"
            @click:next="next"
          />
        </template>
      </v-stepper>
    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref()
  const toggle = ref(false)
</script>
```
</details>


nekosaur:
Will look at this

packages/docs/src/data/page-to-api.json
```diff
@@ -123,10 +123,10 @@
   "components/sparklines": ["VSparkline"],
   "components/steppers": [
     "VStepper",
-    "VStepperContent",
     "VStepperHeader",
-    "VStepperItems",
-    "VStepperStep"
+    "VStepperItem",
+    "VStepperWindow",
+    "VStepperWindowItems"
```
nekosaur, line 129: singular?

packages/vuetify/src/labs/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,185 @@
...
+            </VStepperHeader>
+          )}
+
+          { hasWindow && (
+            <VStepperWindow key="stepper-window">
+              { items.value.map((item, index) => {
+                return (
+                  <VStepperWindowItem
+                    value={ item.value }
+                    v-slots={{ default: slots[`window-item.${item.value}`] }}
```
nekosaur, line 157: `window-item.x` is very wordy. I think it would be better to switch the name with the header slot. So that content is `item.x` and header is `header-item.x` or something similar. My assumption is that it will be a lot more common to have content slot than header slot.

packages/vuetify/src/labs/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export type VStepperSlots = {
+  default: VStepperSlot
+  actions: VStepperSlot
+  header: never
+  icon: never
+  item: never
+  title: never
+  subtitle: never
```
nekosaur, line 41: These need to have proper types

packages/vuetify/src/labs/VStepper/VStepperItem.tsx
```diff
@@ -0,0 +1,182 @@
...
+            { hasSubtitle && (
+              <div
+                key="subtitle"
+                class="v-stepper-item__subtitle"
+              >
+                { slots.subtitle?.(slotProps.value) ?? props.subtitle }
+              </div>
+            )}
+
+            { slots.default?.(slotProps.value) }
```
nekosaur, line 175: I think it might be unexpected that `item` slot does not override the content but simply appends at the end.

packages/vuetify/src/labs/VStepper/VStepperWindow.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const model = computed({
+      get () {
+        if (_model.value != null || !group) return _model.value
+
+        return group.items.value.find(item => group.selected.value.includes(item.id))?.value
+      },
+      set (val) {
+        _model.value = val
+      },
+    })
```
nekosaur, line 45: Could you add a comment why this is done?

packages/vuetify/src/labs/VStepper/VStepperWindowItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { genericComponent, propsFactory, useRender } from '@/util'
+
+// Types
+import type { VImgSlots } from '@/components/VImg/VImg'
+
+export const makeVStepperWindowItemProps = propsFactory({
+  ...makeVWindowItemProps(),
+}, 'VStepperWindowItem')
+
+export const VStepperWindowItem = genericComponent<VImgSlots>()({
```
nekosaur, line 14: What relation does VImgSlots have with VWindowItem?

packages/docs/src/examples/v-stepper/misc-horizontal.vue
```diff
@@ -1,21 +1,120 @@
 <template>
-  <v-stepper value="1">
-    <v-stepper-header>
-      <v-stepper-step step="1">
-        Select campaign settings
-      </v-stepper-step>
+  <v-stepper
+    v-model="step"
+    :items="items"
+    show-actions
```
nekosaur, line 5: I think there needs to be some easy way of

1. Making actions not circular (removing previous on first, next on last)
2. Adding some custom button/content in place of previous on first, next on last

In this example there needs to be a "place order" button on last screen, and possibly a "back" or "cancel" button on first one.

nekosaur (commented):
What's happened with vertical stepper? Are we no longer doing it, or is it just not in this PR?

packages/vuetify/src/labs/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,185 @@
...
+      const value = getPropertyFromItem(item, props.itemValue, index + 1)
+
+      return {
+        title,
+        value,
+        raw: item,
+      }
+    }))
+
+    const slotProps = computed(() => ({ next, prev }))
```
nekosaur, line 96: Not sure this needs to be a computed

packages/vuetify/src/labs/VStepper/VStepper.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+    provideDefaults({
+      VStepperItem: {
+        editable: computed(() => props.editable),
+        prevText: computed(() => props.prevText),
+        nextText: computed(() => props.nextText),
```
nekosaur, line 102: `toRef`?

nekosaur:
You need to update docs with the slot name change.

johnleider:
> What's happened with vertical stepper? Are we no longer doing it, or is it just not in this PR?

It's going into a separate component using expansion panels.

PR #17760, opened 2023-07-09
Refactor input control height logic

yuwu9145:
## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/components/VField/VField.sass
+++ b/packages/vuetify/src/components/VField/VField.sass
@@ -69,9 +69,9 @@
     $root: &
 
     @at-root
-      @include input-density('v-input', $input-density, $input-chips-margin-bottom) using ($modifier, $chips-modifier)
+      @include tools.density('v-input', $input-density) using ($modifier)
         @at-root #{selector.nest(&, $root)}
-          --v-input-control-height: #{$field-control-height + $modifier - math.div($chips-modifier, 2) - $input-chips-margin-top}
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/components/VTextarea/__tests__/VTextarea.spec.cy.tsx
```diff
@@ -27,8 +27,8 @@ describe('VTextarea', () => {
       .should(el => expect(el.outerHeight()).to.equal(56))
       .click()
       .type('Lorem ipsum dolor sit amet consectetur adipisicing elit. ')
-      .should(el => expect(el.outerHeight()).to.equal(80))
+      .should(el => expect(el.outerHeight()).to.equal(78))
```
yuwu9145, line 30: When it comes two lines, it should be

height(78px) = line-height (24px) * 2 + padding top/bottom (30px)

johnleider:
Looking good. Only thing I've found so far is the label shifts up when focusing the input on your branch:
label-twitch

johnleider (approved):

KaelWD (approved):

yurks:
@KaelWD the same bug still remains for `density="compact"`at least - default height is 40px, but rendered height is bigger

johnleider:
> @KaelWD the same bug still remains for `density="compact"`at least - default height is 40px, but rendered height is bigger

Create a new issue please.

Issue #17583, opened 2023-06-12
[Bug Report][3.3.3] VInfiniteScroll does not set root element correctly unless margin prop is explicitly supplied


eldjh3:
### Environment
**Vuetify Version:** 3.3.3
**Vue Version:** 3.3.4
**Browsers:** Chrome 114.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Use a `v-infinite-scroll` element without the `margin` prop.  The `root` is not passed to the `IntersectionOberserver` which therefore uses the viewport as the root for intersection calculations.

#### More details
In the linked playground run as-is.  It will use the viewport for the root element.  It will only load one extra set of data (3 more items) [and get stuck - as per issue 17358].  Resize the viewport such that the intersection `div` leaves and re-enters the viewport.  The infinite scroll will be triggered to load more data.
Comment/uncomment the two `v-infinite-scroll` tags in the playground.  Behaviour changes such that the wrapper `div` is now correctly the root element.  But, no amount of scrolling will now get more data to load.

### Expected Behavior
Whether the `margin` prop is specified or not, the `v-infinite-scroll` wrapper `div` is used for the root element.

### Actual Behavior
With no explicit margin prop the `IntersectionObserver` uses the viewport as the root for intersection calculations.

### Reproduction Link
[...]

PR #17813, opened 2023-07-15
fix(VVirtualScroll): correctly calculate first/last index

### LINKED ISSUE #13199: [Bug Report] Virtual Scroller Renders all items when no height property is specified
### Environment
**Vuetify Version:** 2.4.5
**Vue Version:** 2.6.11
**Browsers:** Firefox 84.0
**OS:** Windows 10

### Steps to reproduce
1. Remove height property from any example
2. Wait for it to render
3. Note the unbounded height

### Expected Behavior
It limits the size of the scroller to say the flex-container it's within.

### Actual Behavior
It grows endlessly

### Reproduction Link
<a href="https://codepen.io/douglasg14b/pen/poNKgbB?editors=101" target="_blank">https://codepen.io/douglasg14b/pen/poNKgbB?editors=101</a>

### Other comments
The docs specifically say:

> The virtual scroller displays just enough records **to fill the viewport** and uses the existing component, rehydrating it with new data.

I also was unable to find any mention of the `height` or `max-height` attribute being required for the scroller to behave.

NOTE: The example link will take some time to render, it's the docs example, with the height prop removed off.

**Why:** A scrollable area that is affected by the size of the window (ie. something that is in the main body of the layout, like a table, or list ), that in most cases isn't set to a specific height in pixels.

### LINKED ISSUE #17801: [Bug Report][3.3.8] v-virtual-scroll broken
### Environment
**Vuetify Version:** 3.3.8
**Last working version:** 3.3.6
**Vue Version:** 3.3.4
**Browsers:** Google Chrome
**OS:** Ubuntu undefined

### Steps to reproduce
Use v-virtual-scroll with any length simple items, but more so with more, e.g. 150, 250.

### Expected Behavior
All items show when the user scrolls.

### Actual Behavior
When you scroll to top or bottom, some items are not shown.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlUd1uwiAUfhXClSYCW9ySpdFFn8N6gfaYMYESoF1M03ffgdpF643Jrlo+vp9zPnYd3TrH2wZoQVcRjNMywmdpCVm1TDqXf/PhWNsolQV/hTLYKh8bqVk4+lprUii0COuS5m9J/6hIHs1Jy4KuY1HBSTY6IrkjiU565BPbmAN40o1YT1bibqzsJabJ45h4cT9nQoY17nzwiELlIgkQm2FLFIaYYwNZkx3nfOu9vMxe31/m/AyXMJvvk80gRAldUGVc7SMz0vHvUFsssUtWWEC+wAoKkpGEYcvpXNKvGF0ohDhWFmUVaNV6biEK64zYIE34xkZlgFW12Sz5kr+JSoV4C3MIhh18/RPAo0lJFzcxUZ0uT0RdmTnhY0i4QkzLQ0gRD9YC81rwzIOtwIN/dqOJ7HarydXDZim9L22PfWtlz2FS9THkmnfjkP9aOZslJ3zqnvYLmufE5Cyg+1/aMBJx)

### Other comments
It is totally reproducible for us.  It just takes scrolling up and down, maybe a bit faster, maybe just anyway, and it happens.  In my playground example I made it show numbers 0 to 149, and in one trial it stopped at 131, in the other at 136.  When going back to Vuetify 3.3.6 it works as expected.

### LINKED ISSUE #17996: [Bug Report][3.3.11] Virtual table removes rows when row height is changed
### Environment
**Vuetify Version:** 3.3.11
**Vue Version:** 3.3.4
**Browsers:** Chrome 115.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
When scrolling down and up again, the first entry disappears. This strange behavior only appears when the header is hidden.

### Expected Behavior
No rows should disappear. No extra white space at top and bottom of table should appear. 

### Actual Behavior
Some rows disappear, inconsistent white spacing at top and bottom of table.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqllltv2zYUx78KobVIAkS05FsSLQlapLt03cNeh9hAaYmxuVCUQFJuPcPfvYcW6Uiy2MZoXkLzXPg7f5LiedwG78sSrysaJMGtpnnJiab3M4HQ7TrMiCahJgtOwzWTuiLcGBBKVpRkVKq7WWBHs8BaGOQw89b/A1WKSn2wp5woY6acrolmhQhjlG/qNZzPirLlSoPTOIrcnEkbrgmvKMwLklvnPSigOnC0DhUvtOO7vx20Sxr01QSmlh/8VHrD6xDs6NC2XisM1zZcFl/CmjVBw6j8+qtx2JlkLtxkSiUr9T4V/VoWUqOMPpGKa5fP8CTo/ALd3aNzO2c02BeQoEc3g1xA/aeZ5jRBZ1ZgdB5H0RLBcM3E8uLssulLOFsK8FWaSN02KSAyxSToiXBFW7ZnuoEgI3YzZtcYbw8YD4QXklF1dnlYjYoMftVJUmf2hP9OoAKD3Rv9RLQv8IHIhfpOaGrsvuB/ZKEpE98JL2sPX4KPsoDot75oBuZm6Pwwyuy18O+vkd0II4v/qUD/Fsuqu3VO0wTFk5uWBfRK0BRHHXdQAg7quDVrC0zQuONu2GH92Lv1PbAfU4pSSUmOFBHZF5aufMTD0dUx8U0/cce1QTz6aeLf4HvEpJdyOjymjH3CtmkOmF13i3l1CuZDVabkuXUNm5yjaHLMOcJt3Szm9DQ1r0/B/AM+PVQu4ATAHfCgTqavl3TcPtYH1hFuG9zOT0+B/YtyvkELSoSX9apH1qgf9cZzrbruFjU6hfTvgnNWFqWX86bnmEa4Pek4rz2cfZTDUyj/LATdpEW+8GGOo/ba9pT2Yl57TukUt7fEgo4np5B+KETl/ZaOJz1iDif9uz6JfZep94AOTxL0E9Of4OHzcE7iHjWHnos0bat24GyrbCk9t8g9XrsLO4C9LitNs+SFvtP0QVPTKExSXUmBHjHG76UkG9OxRNEFhrcSPOc4J+U5M11QS4y0EMr0THWXcwePL8TrFVPYvaGPDL1FcTQ3rddLnLViIyaEfX6zbc7s0C9vtmz32bRoLzGW0Do2RLhwYydH/b/u9VyDF1wGnIlnhf9ThYBuel/GLEgVNL+Hdx4aZq1LlQwGaSbAM6OcrSUWVA9EmQ+gD9fsafNuhEc4jgcZU9rNhZwsFN5nM6nmM7GDFVluesoQtOssWxv2S1s9oSevoG1OfsTwDtwGshKa5TTMinwPM65ZGtOYqjxcQA8MWkGSWWCl2S9jgF+x1A/LhTWOcg9Mk0tlKKHVopLK15bUCWuW1TEdlWY3exfsLoN9GChfMwfzb2ronaU=)

### Other comments
I only had to resort to changing the css variable as the compact setting for density was not as dense as in vuetify2. If there is a more proper way to make table rows more compact, I am eager to know!

### LINKED ISSUE #18198: [Bug Report][3.3.15] v-virtual-scroll fails with items larger than viewport
### Environment
**Vuetify Version:** 3.3.15
**Vue Version:** 3.3.4
**OS:** Windows 10

### Steps to reproduce
Use a ```<v-virtual-scroll>``` component with **items larger than the viewport** of the component (like the provided playground example.

1. Scroll down slowly. After passing the preloaded items, a new item is loaded only after it is already visible for about 1/3 of the viewport. In the provided example this can be seen by the white space (items have red or green background)
scroll_down

2. When scrolling down completely, the last item (in this example 'Section 5') is not loaded at all.

3. Slowly scroll up again. The previous item at the bottom of the viewport will get unloaded before it leaves the screen. Also in some configurations the complete list then glitches. 
scroll_up

Of course this happens with any scroll speed.

### Expected Behavior
Item size beeing larger or smaller than the viewport does not matter and items get loaded / unloaded seemlessly while scrolling without any glitches and white spaces.

### Actual Behavior
**Works** fine for items **smaller** than viewport. 
Does **not work** with items **larger** than the viewport.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNq1VW1v2jAQ/iunaBOgNS9AO2kRrG/qh03TOrXbJ4LUkBhwceLIdlIQ4r/v7CQ0QLtWmsincvfcPc899rmjtSVF5F5mmVPkxPKtgYwEzdTXIAUgy4wLBTGZhjlTsNYxgDhUoQ/tDgy/QruKAbguKK5CBnwKXQ9s6HkLoEk4IxJoCt0FSBIpylNZV5jkn7sf8mp1X6b8mkJ/rSoI3ZYPI8dxLoUIV+2+13EWZCXbnbGThFn7uQA7xkst6mGuVCZ9181oJPPEyeZccen2PO/ccA4/rBG5eXgu7Zy8wNs7Di98gr73Jnn/aOSf3yY/PRr5l7fJz45G3u29wr6p/tqUoU2QDtztGuAPRZKMhYqYpRgUdkGFykNmI4YzVtb6FFFyGFi3k0ecpBR9eMU7gVXi54TO5grxZ55Xx55orOYY6p7WMcOInLUCKGzJuPKrlUTsGjTxCe5YTJYo3apKsCimBfhSrRjRuEkYLWaC52l8zRkXflXxEXowHIIH59ASJG6BD62ZICRtNXthN5mFKUQslHpIRZbKnvcQsS4FwGaDniGkWVLYgj81AiYU8cqx+ivsKRfYc+tWNUxHvxuHDo403bi2rP589Bt7mML9HDJqzai2Gd/RZZTRZAYE6QS6JiLdrSIPLHB3x3DNHM1ZMdIcduCi+/XpubsXCKG7NwjDOxjrBKn102vjlXceJU/xYTZPo9akEzKwto9lYOHLrX8HVr0JUZxiWUwYLYSTEuWmWeJeIMwVeapoQuyYJxd9p++cok6pmmGHyMSe4CySCGwSWNVyGBpFp6t3UFVIw9A9KymqmM3CidQcB71dJCyIsAXBQxREvHekvbLmWHupg9Gqbd+g4YymC7nndSSNz6Na5P/NbLrpVuOSUcmIp1M62yflSUYZEbdZ+d+yedAhY/zpu4kpkZOte9GcRIsX4o8Sl0G7+EsQ40XDcRWKGVFl+ub+Jy50I5nwOGfVpXoleUckZ7nWWMKu8GVB2Q2cUfvN3Feazn7Lm6UiqayH0kKN/wZvbsD1P0Z/losGb89tc2KZ40Y7S9+t8V8GFrJn)

### Other comments
I would consider this a **severe bug** as viewport size may be changed by the user resizing his browser window and thus this case **can occur in basically every application** using ```<v-virtual-scroll>```.

yuwu9145:
fixes #17801
fixes #17996
fixes #13199
fixes #18198

- [x] fix scroll is broken in big height - [demo](https://play.vuetifyjs.com/#eNqlUUtuwjAQvYrlVZGwTUVbVRFUcA7CIiRDcfFPtpMKRbl7xwYkCBukriK/eZ95k01P187xrgVa0EUE7VQV4as0hCw61kkf20qxUHurFCkkEsKypPlbUnIA+X2ICHzMZiXNKtRdXUjHgrKxaGBftSrRepKUZEAuMa3egSf9FRvIQtzni/ECCN9x8IkD6SIJEFuXVbU1IWbLQJZkwzlfe1+dXl7fZxN+hFN4mWyTzVmIEjqlUjvrI9OV4z/BGrxEn6ywZx5g04JkJGF4qvQu6SFGFwoh6sagrAElO88NRGGcFiukCd+aKDWwxurVnM/5m2hkiLcwh6DZztvfAB5NSjq9iYlyf3oi6sLMCZ/nhAvEVLULKeLBWmBeB555MA148M82GsluW41GD81S+lCaAe+tpDmG0anrkM+8uS75r8rZLDnhrx7oMKV5T0zOArr9A1rNAdc=)
- [ ] Allow dynamic buffer ratio - unnecessary at this stage? 1/3 works pretty good
- [x] Handle multiple scrolling elements - scroll based on the closest scrollable parent

image

### 1. Combine dynamic height with item height beyond container
This senario requires user to set up possible smallest `item-height`, in order to allow the component always prepare enough buffer - This should be an edge case and I'm unsure if this was the right direction, feel free to comment
```vue
<template>
  <v-virtual-scroll :items="items" height="400" :item-height="15">
    <template v-slot:default="{ item, index }">
      <div
        :class="[
          index % 2 === 0 ? 'bg-grey-lighten-2' : index % 5 === 0 ? 'bg-grey-darken-2' : '',
          'px-2'
        ]"
        :style="{height: `${item.height}px`}"
      >
        Dynamic item {{ item.index }} - height {{item.height}}px
      </div>
    </template>
  </v-virtual-scroll>
</template>

<script>
  export default {
    computed: {
      items() {
        return Array.from({ length: 1000 }, (k, v) => ({
          index: v,
          height: this.randomIntFromInterval(),
        }))
      },
    },
    methods: {
      randomIntFromInterval() {
        // min and max included
        return Math.floor(Math.random() * (1200 - 15 + 1) + 15)
      },
    },
  }
</script>
```

### 2. Dynamic height
```vue
<template>
  <v-virtual-scroll :items="items" height="400">
    <template v-slot:default="{ item, index }">
      <div
        :class="[
          index % 2 === 0 ? 'py-2' : index % 5 == 0 ? 'py-8' : 'py-4',
          index % 2 === 0 ? 'bg-grey-lighten-2' : index % 5 === 0 ? 'bg-grey-darken-2' : '',
          'px-2'
        ]"
      >
        Dynamic item {{ item }}
      </div>
    </template>
  </v-virtual-scroll>
</template>

<script>
  export default {
    computed: {
      items() {
        return Array.from({ length: 1000 }, (k, v) => v + 1)
      },
    },
  }
</script>


```

### 3. Rendless and no height, html document is the scrolling container
```vue
<template>
  <v-virtual-scroll renderless :items="items" :item-height="40">
    <template v-slot:default="{ item }">
      <v-list-item>number {{ item }}</v-list-item>
    </template>
  </v-virtual-scroll>
</template>

<script setup>
  const items = [...Array(350).keys()]
</script>
<style>
html {
  height: 100%;
}

</style>

```

### 4. Item height is beyond container
```vue
<script>
  export default {
    data: () => ({
      // total of 10 - 20k images in 1k sections
      imageURLsBySection: {
        'Section 0': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx}`
        ),
        'Section 1': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 30}`
        ),
        'Section 2': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 60}`
        ),
        'Section 3': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 90}`
        ),
        'Section 4': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 120}`
        ),
      },
    }),
  }
</script>

<template>
  <v-virtual-scroll
    :items="Object.keys(imageURLsBySection)"
    height="500"
    item-height="1000"
  >
    <template v-slot:default="{ item, index }">
      <div
        :style="{height: '1000px',backgroundColor: index % 2 == 0 ? 'red' : 'green'}"
      >
        <span class="text-h2">{{ item }}</span>
      </div>
    </template>
  </v-virtual-scroll>
</template>


```


packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
```diff
@@ -59,7 +59,10 @@ export const VVirtualScroll = genericComponent<new <T, Renderless extends boolea
 
   setup (props, { slots }) {
     const vm = getCurrentInstance('VVirtualScroll')
-    const { dimensionStyles } = useDimension(props)
+    const { dimensionStyles } = useDimension({
+      ...props,
+      height: props.height ?? window.innerHeight
+    })
```
KaelWD, line 65: 1. this isn't reactive
2. it can have any height, shouldn't be limited to the window

KaelWD (changes requested):

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
yuwu9145: Makes sense.

I also think It's user's responsibility to ensure VVirtualScroll is either placed in a fixed height container or has height prop defined

packages/vuetify/src/composables/virtual.ts
```diff
@@ -51,7 +51,7 @@ export function useVirtual <T> (props: VirtualProps, items: Ref<readonly T[]>, o
         ? display.height.value
         : contentRect.value.height
     ) - (offset?.value ?? 0)
-    return Math.ceil((height / itemHeight.value) * 1.7 + 1)
```
KaelWD, line 54: There's an extra 1/3 buffer on each side of the visible area

packages/vuetify/src/composables/virtual.ts
```diff
@@ -105,7 +104,7 @@ export function useVirtual <T> (props: VirtualProps, items: Ref<readonly T[]>, o
     containerRef.value.scrollTop = offset
   }
 
-  const last = computed(() => Math.min(items.value.length, first.value + visibleItems.value))
+  const last = computed(() => Math.min(items.value.length, Math.ceil(first.value + visibleItems.value)))
```
KaelWD, line 107: These should both already be integers

packages/vuetify/src/composables/virtual.ts
```diff
@@ -81,20 +81,19 @@ export function useVirtual <T> (props: VirtualProps, items: Ref<readonly T[]>, o
...
     const midPointIndex = calculateMidPointIndex(scrollTop + height / 2)
     const buffer = Math.round(visibleItems.value / 3)
-    const firstIndex = midPointIndex - buffer
-    const lastIndex = first.value + (buffer * 2) - 1
+    const firstIndex = midPointIndex - buffer * 3 / 2 
+    const lastIndex = first.value + (buffer * 3) - 1
     if (direction === UP && midPointIndex <= lastIndex) {
       first.value = clamp(firstIndex, 0, items.value.length)
-    } else if (direction === DOWN && midPointIndex >= lastIndex) {
+    } else if (direction === DOWN && ((midPointIndex >= lastIndex) || (midPointIndex + visibleItems.value / 2 >= items.value.length))) {
```
KaelWD, line 94: It updates in pages instead of every scroll event

KaelWD:
> It's user's responsibility to ensure VVirtualScroll is either placed in a fixed height container or has height prop defined

It should ideally work anywhere like VInfiniteScroll, some related discussion about this in #17583. Currently the scroll parent is only used with the renderless prop as there's then no root element.

KaelWD:
There's a couple of unfinished changes I have related to this:

[refactor/configure-virtual-buffer](https://github.com/vuetifyjs/vuetify/tree/refactor/configure-virtual-buffer) - moving the buffer amount to a constant and maybe eventually a prop. Doesn't really work with anything but 1/3 atm
[fix/virtual-scroll-any-element](https://github.com/vuetifyjs/vuetify/tree/fix/virtual-scroll-any-element) - the whole scroll listener needs rewriting to handle multiple scrolling elements. This was much easier for VInfiniteScroll because IntersectionObserver already does it for us.

yuwu9145:
> > It's user's responsibility to ensure VVirtualScroll is either placed in a fixed height container or has height prop defined
> 
> It should ideally work anywhere like VInfiniteScroll, some related discussion about this in #17583. Currently the scroll parent is only used with the renderless prop as there's then no root element.

Without a height restriction, scroll/overflow container could be stretched to content height. (scroll bar cannot be guaranteed) [demo](https://play.vuetifyjs.com/#eNrtls9uwjAMxl8lCtqNNiBAmroOsXfYjXJokwAZaRIloQUh3n3uP6mCC9KOyy3+OfZnfydvb/jLmLg6c5zg1PPSyNzzdaYQSpmokGCfGdYVt3up62/ufIbbJKTN2kOcEhNAAAEEEEAAAQTwn0BK4EqCV0pGtxOEzl9ld0ZNxtcTunVdipyeDlafFUvQZEnz/Wr20WWoltomqD4Kz3tkcsaEOiRovjKXntWC+WOCVrO3HgwqCXLUail7XGjLOPSbmwtyWgqGJpTSNnlvph7mxFMsSqOtj8rcxD9OK7gH22GzPuEynAzjZxgOxibO8NF74xJCKFNQxrgUlY0V90SZkmzgG7Fn5UXJI6bLzSJexEuwzPkxjrkro8Lq2nELTTI8Hcl4sb++INX/bBXeO4UeRTIvXCPx1JqAHtgWWa7AJG5f3eihbLzVQ+pps974O/gthTq5B6upa23eDkP+aeW2WdNp1wjep7idE5TbArz7BYW/rGg=)

At least there should be a logic to default height to `window.innerHeight` or it should be totally a user responsibility

yuwu9145:
> the whole scroll listener needs rewriting to handle multiple scrolling elements. This was much easier for VInfiniteScroll because IntersectionObserver already does it for us.

@KaelWD I understand what you are trying to say for VInfiniteScroll in terms of intersecting viewpoint as root can be automatically triggered.

However, what's the use case of handle multiple scrolling elements for VVirtualScroll?

I'm playing around this [example](https://play.vuetifyjs.com/#eNrtWsFu4zYQ/ZWBTruAJadIWqBKdrFB0UPvvcV7oMWxxZoiBZKS4wb+9z6SctfJHrpAT4slcrBMzryZeTNPAQZ+eqkex7GZJ67a6iHwMGoR+OPGED2M6YPoN2u8kuxI0JG3NIo9U+hFoMmzJ2V2yqjA5DtntVZm39Afy52gmY20rh6dnQEhM6BWWyfciYKlQZgMxyTkzC4ozwOb4AmJdCxpZKesVJ3QGg69s9O+t1OIHhksprOiXiCaUQOSl7R3YuxV56lnB1wjo7WDVXxcEusmH+zwJRXUk/GkE0dPxga1Q9SgrKGtfY4+GQcuB27od9H1ZHfxxHMKr5C0PZqMokxg57lL/pLD8oT0gzLskYrW5CZjwBfhIgIPQsUHx0I29GdkZAq9dRkwh4oN8JHsQZxijsQgmOCh1d+RRYU0ULYYR47IK/LKdIneU3I5GHvMeOD6hFJC0IizXRhF2gZ9Plp3gDfKyUHD0S5EKfCQwASInQCOZj/6jBgtQa5bBiEZLt3JJP0HJwlzpxxAM2CHuQvCBDReTvE4IWZ01GhkHDZYSURw7CcNmD2ioFPEz3FwOBd/aa1mMecCcqI7hyFwaWSOKvTpYuvsEXcp5wvbeXBwoJZmdHYYJxTTJJ2ss1CKXopeil6KXopeil6KXopeil6KXopeil6+G71INZMPJ80fNpVUHgN8ammn+fkeg6j2fWjpl5ub8XlTLcJ65RENW/oJf2i7vSeLDu20PbbL99EiFfDVolbsGNTM/+IAaa5n5cIkdL3w4yA8dpq9pxYlDB4h0uemygd1zgnHdzdXSMC6bDForr22oZW8EyAXli8Ych7o/Mo+RdfKhzpefjTTsAW3Lxfb88P6+voqzvrVuiQfva3jwtQaVGWWvzyV91N5P5X3U/l/XvRS9FL0UvRS9FL0UvRS9FL0UvRS9FL0UvRS9PLD6uXVdg1fkaYaA3kOU9ZRqibt6Dx9oKemaR6dE6d3tz/fvG8OfPLv3n+OMNlxAYm7yuQ90ksOfVQy9C3dxcXmfTw6J6fFsFpVahitC/UgxuYvbw1+JpU8sY1MF9hHthesTYXfUcXvm6oPYfTtet1JAzfJWs2uMRzWZhzWn2C2xrAGNXAt7fDptrlt7rAY9OH6uGE/1At/ANlUq6swkNPpG0ItlinCrznCclRrsfUxxFfQa8TDONd5BcvuWyt643Zd1ZurrypbiD+Dbwzhwb+huvOJ5qdLkv+r5AQWkTAe5+q8qlKeiJwcqs//AB4x37M=). But I think its working makes sense by having just one scroll parent listener? What is missed?

yuwu9145:
Maybe I get it, you want scrolling behaves correctly when no height is specified and scrolling element is the window

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
```diff
@@ -72,11 +72,19 @@ export const VVirtualScroll = genericComponent<new <T, Renderless extends boolea
 
     useToggleScope(() => props.renderless, () => {
       onMounted(() => {
-        containerRef.value = getScrollParent(vm.vnode.el as HTMLElement, true)
-        containerRef.value?.addEventListener('scroll', handleScroll)
+        containerRef.value = getScrollParent(vm.vnode.el as HTMLElement, false)
+        if (containerRef.value === document.documentElement) {
```
yuwu9145, line 76: I tested in multiple live html pages:

- `document.documentElement.addEventListener("scroll", (event) => {console.log("htmltag scrolling");});` doesn't work

- However, `document.addEventListener("scroll", (event) => {console.log("document scrolling");});` works

KaelWD: Yeah idk some archaic netscape navigator shit probably. document.documentElement.scrollTop is the one that changes.

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
```diff
@@ -72,11 +72,19 @@ export const VVirtualScroll = genericComponent<new <T, Renderless extends boolea
 
     useToggleScope(() => props.renderless, () => {
       onMounted(() => {
-        containerRef.value = getScrollParent(vm.vnode.el as HTMLElement, true)
-        containerRef.value?.addEventListener('scroll', handleScroll)
+        containerRef.value = getScrollParent(vm.vnode.el as HTMLElement, false)
```
KaelWD, line 75: includeHidden needs to be true for places like VAutocomplete

packages/vuetify/src/composables/virtual.ts
```diff
@@ -81,20 +83,18 @@ export function useVirtual <T> (props: VirtualProps, items: Ref<readonly T[]>, o
...
-    const buffer = Math.round(visibleItems.value / 3)
-    const firstIndex = midPointIndex - buffer
-    const lastIndex = first.value + (buffer * 2) - 1
-    if (direction === UP && midPointIndex <= lastIndex) {
+    const buffer = Math.ceil(visibleItems.value * BUFFER_RATIO)
+    const firstIndex = Math.ceil(midPointIndex - buffer * BUFFER_RATIO_RECEPROCAL / 2)
+    if (direction === UP && midPointIndex <= first.value + buffer - 1) {
       first.value = clamp(firstIndex, 0, items.value.length)
-    } else if (direction === DOWN && midPointIndex >= lastIndex) {
+    } else if (direction === DOWN && (midPointIndex >= first.value + buffer + 1)) {
```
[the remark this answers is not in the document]
yuwu9145: Achieved the outcome of only updating first.value when buffer is reached, but the new logic is bit different (it calculates the threshold midIndex based on buffer & first/last values)

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
yuwu9145: Rolled back. Tested it is unnecessary, must be a debugging code. Thanks for picking up

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
johnleider: What's the status on this feedback?

johnleider (dismissed):

packages/vuetify/src/composables/virtual.ts
```diff
@@ -37,6 +38,13 @@ export function useVirtual <T> (props: VirtualProps, items: Ref<readonly T[]>, o
     },
   })
   const containerRef = ref<HTMLElement>()
+
+  const isItemBeyondContainer = computed(() => containerRef.value && containerRef.value.clientHeight <= itemHeight.value)
```
KaelWD, line 42: DOM APIs aren't reactive, there's `contentRect` for this

KaelWD:
Just messing around but something like this seems like it might work better for dynamic item height
```js
const start = calculateMidPointIndex(scrollTop)
let end = start
let endOffset = sizes[start]
while ((endOffset += sizes[end]) < sizes[start] + contentRect.value.height) {
  end++
}

// ... work out buffer/paging somehow ...

first.value = clamp(start - buffer, 0, items.value.length)
visibleItems.value = clamp(end - start + buffer * 2, 0, items.value.length)
```

That way visibleItems and the buffer size are always based on what's actually displayed instead of the largest possible item.

packages/vuetify/src/composables/virtual.ts
yuwu9145: changed and dynamic size items are working now

yuwu9145:
> Just messing around but something like this seems like it might work better for dynamic item height

implemented

KaelWD:
I was thinking something more like #18392

yuwu9145:
close in favour of #18392

Issue #17816, opened 2023-07-15
[Feature Request] Custom sort with other properties of the object


Employee87:
### Problem to solve
When I use :custom-key-sort="customKeySort" on v-data-table, the function gives me 2 values taking the key directly from the object. If the key was "a", and the 2 objects it is sorting is {b:1} and {b:2}, I would get undefined as the 2 parameters to the sort function (since it is giving me object[a] directly). 

### Proposed solution
Get back the 2 original objects so I can do my own custom sort logic. This would be very useful as I create my own headers, and the values I show in the table are a calculation of some nested properties in the original objects. It would look like this

```
const customSortFunction = {
    a : (o1, o2) => {
        if (o1.b > o2.b) return 1;
       if (o1.b < o2.b) return -1;
       return 0;
  };
};
```

KaelWD:
Duplicate of #11226

PR #17832, opened 2023-07-18
fix(filter): always pass InternalItem to custom filter function

### LINKED ISSUE #17487: [Bug Report][3.3.1] VDataTable customFilter cannot access item.raw
### Environment
**Vuetify Version:** 3.3.1
**Last working version:** 3.2.5
**Vue Version:** 3.3.4
**Browsers:** Google Chrome
**OS:** Windows 10

### Steps to reproduce
Use the reproduction link with Vuetify 3.3.1 and enter `vanilla` in the "Flavor search" input field. Notice there is nothing found. Then, set Vuetify to 3.2.5 and enter `vanilla`. Notice there are 2 items found. The flavor search uses a custom filter that searches a hidden "flavors" item property, which is accessed using `item.raw` within the filter function.

### Expected Behavior
Custom filter should be able to access `item.raw`.

### Actual Behavior
Custom filter throws an error in the console: "cannot read properties of undefined" (because `item.raw` is undefined).

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVVm1v0zAQ/ismErSVmoRuA0S0TiBg0j7BV7Tug5tcVzPHDvYl25j63zk7SZu0y17ypc69Ps+dc9fLh+BrUURVCUESnCLkheQIZwvF2GkVZhx5iHwpwQkYS9LSos7DlZAIZr4I6sMiaNRr4BkYS4rmtNUICu3kGVgLBncKC9yka9LUh1aeSm6dPUioOAqtwlmrcqHCissSSK94DrXCYybULQdWhVZqTFAXjcpTQrhDwg8ya4WMLHOdgTwA4R7Jl15zLnmlDTs0aJEWPDzZyc9O426uFlzcrzDZ7EpMop7+1KZGFOgt4a7QBlkGK15KZA91OOebsPGEzc/YuJGxBmLCRqNpK0q51EaA7QnbViTscsdmG8U9rrjkcm70P1Dst74uDe78+4FnHz73NCuOCfsYvd8zN0uyPTrpSQujEYRK2MmeuTCapKPZ237SlW+FI2PR8NslGHM/Zelap9qVrmu86Zwfo3aRAksN8JxZrrJbka6H+B0dfzrk9/lxfnumHX7Hr+NXcSWk5K9h9IPuozCDLD4eHbKYDbWpj3ZLY9+8ofHpORrP9uiqOW0mzSHVeVEiZMmOajNY6NJ32BvA0qjuPabiMBQoXUm+1xd9NGVcimsH1SL37zdwT2+ucqNeYbve35riddxBZVvntrbDAc45svH1ZMCfOvBUbmrEE86+UcPuv+qOPRGg6elwiAvqLRu/HfJ3re87X7XHVtj+5oBrndF12PatXh5jP8qn7G8J7jN20/2x3o67+LwLezNnqpSSvXvX1fk4Azq8L0CvGvf5fO4niFDXoz07ByKi0RI1dzgSKoO7n6uxDz6h6HNGC2nrMXmc9MZN9O0UD6aBFOrGRn+sVrRtPcdFkFpahtsZTJsTsbBJHKeZIktaS6IykQKMVZHHtKdRrO6/HEfH0SzOhMVWFNKispEP5iJRGzaUUORub4Q5L/ay1gqfuan1IqBI7v0ZCF/ILDalQpFDmOncYzmpsXTEEdg8XBp9S58eBVkETWV8Ggf4BameY0spDkLHlK8CExq6pmDcf5OXMdpz67LaUx0wa1q9CTbTwLtR4T3k4Oo/XEq4oA==)

### Other comments
The line below was changed from 3.2.5 to 3.3.0. As of 3.3.0, only `item.columns` are accessible within custom filters in VDataTable. In v3.2.5 and earlier, `item.raw` was also accessible, allowing us to implement custom filters that access hidden item properties.

https://github.com/vuetifyjs/vuetify/blob/9d4b9d1b76b095db8637761c2f2595d9472ccb29/packages/vuetify/src/labs/VDataTable/VDataTable.tsx#L120

Can "item.raw" accessibility be restored in custom filters in VDataTable? Thank you for considering this issue.

johnleider:
## Motivation and Context
fixes #17487

## Markup:
<details>

```vue
<template>
  <v-data-table
    :custom-filter="filter"
    :headers="headers"
    :items="desserts"
    :search="search"
    class="elevation-1"
    item-value="name"
  >
    <template #top>
      <v-text-field
        v-model="search"
        label="Flavor search"
        class="pa-4"
      />
    </template>
  </v-data-table>
</template>
<script>
  export default {
    data: () => ({
      search: 'st',
      calories: '',
      desserts: [
        {
          name: 'Frozen Yogurt',
          calories: 159,
          fat: 6.0,
          carbs: 24,
          protein: 4.0,
          iron: '1%',
          flavors: 'strawberry, chocolate',
        },
        {
          name: 'Ice cream sandwich',
          calories: 237,
          fat: 9.0,
          carbs: 37,
          protein: 4.3,
          iron: '1%',
          flavors: 'vanilla',
        },
        {
          name: 'Eclair',
          calories: 262,
          fat: 16.0,
          carbs: 23,
          protein: 6.0,
          iron: '7%',
          flavors: 'vanilla, chocolate',
        },
      ],
    }),
    computed: {
      headers () {
        return [
          { title: 'Dessert', align: 'start', key: 'name' },
          { title: 'Calories', align: 'end', key: 'calories' },
          { title: 'Fat (g)', align: 'end', key: 'fat' },
          { title: 'Carbs (g)', align: 'end', key: 'carbs' },
          { title: 'Protein (g)', align: 'end', key: 'protein' },
          { title: 'Iron (%)', align: 'end', key: 'iron' },
        ]
      },
    },
    methods: {
      filter (value, query, item) {
        if (
          value == null ||
          query == null ||
          typeof value !== 'string'
        ) return false

        const flavors = item?.item?.raw?.flavors?.indexOf(query) !== -1
        const search = value.toString().toLocaleLowerCase().indexOf(query.toString().toLocaleLowerCase())

        return (flavors || search)
      },
    },
  }
</script>

```
</details>


Employee87:
Can this be done for custom sort functions too, so it can get the original row object?

nekosaur:
> Can this be done for custom sort functions too, so it can get the original row object?

That's a separate feature request.

packages/vuetify/src/composables/filter.ts
```diff
@@ -135,7 +135,12 @@ export function useFilter <T extends { value: unknown }> (
 ) {
   const filteredItems: Ref<T[]> = ref([])
   const filteredMatches: Ref<Map<unknown, Record<string, FilterMatch>>> = ref(new Map())
-  const transformedItems = computed(() => options?.transform ? unref(items).map(options?.transform) : unref(items))
+  const transformedItems = computed(() => options?.transform
+    ? unref(items).map(item => ({
+      item,
+      ...(options?.transform?.(item) ?? {}),
```
nekosaur, line 141: Not a huge fan of this. When implementing a custom filter function it would be `item.item.x` (without a spread). There is also a potential key conflict if table has a column called `item`.

Possible alternative solution.

```js
items.map(item => [item, options.transform(item)])
```

And then in filter function on L71

```
const [item, transformed] = items[i]
```

That way `transformed` could be used for `getPropertyFromItem` and `item` could be passed as is to `keyFilter/filter`

nekosaur:
I think we need to add some tests here as well. We've had too many regression issues in the past with filtering.

Employee87:
> > Can this be done for custom sort functions too, so it can get the original row object?
> 
> That's a separate feature request.

https://github.com/vuetifyjs/vuetify/issues/17816

packages/vuetify/src/composables/filter.ts
johnleider: Changes made. I'm still figuring out how I want to approach the tests.

Issue #17628, opened 2023-06-15
[Bug Report][3.3.4] Elements that overflow are covered by those under them, even ignoring z-index


scscgit:
### Environment
**Vuetify Version:** 3.3.4
**Last working version:** 2.5.7
**Vue Version:** 3.3.4
**Browsers:** Opera, Google Chrome, Mozilla Firefox
**OS:** Windows 10

### Steps to reproduce
Create `<v-card style="overflow: initial">`, place an element inside and make it overflow, for example by `style="transform: scale(7)`.

### Expected Behavior
Element covers anything outside the container, especially when using higher `z-index`.

### Actual Behavior
![image](https://github.com/vuetifyjs/vuetify/assets/3118390/d68f7179-5400-4d5a-a3d1-a2b57c3611e9)

`z-index` is also ignored - in general, it works way worse compared to Vuetify 2, and the rules behind how some components sometimes have increased `z-index` seem too arbitrary.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlU8tu3DAM/BVBpxZYWW42QFF3WyRA7/2AOAfFpmO1ekGinU2D/fdSdhZ2HKAJUN84JGc4FH3zxK9DKMYBeMUPCDYYhfC9dowdRtGo2LKEjwa+1dyPEDvjHyqmnUatTM1Zn0HWGJUSVVglLmo+Nf93+0Sg7f05YktZKy5rvuBnAYzKpc5HW7HUKAMfPn/8yv4I7Vo4VuwLfS+6YkM9PWJIlZRBN2mwReg9+iQvylLuy3JdbtVR9KDve6SuT2UZjktWnh3L2fKyAIQjik6DaTcmmVF3YCj8AZ0aDBIys7zg+CfDqKJWLo/jBzTaQbui/blAz7zPr7HZ9bXz2OcnoNxK+iBXl8B3nLh+p+JX8o6u5Cnz1bxJqeYVu5nNLqtsWkeVLRg9xsIBShespPtC3T1e7Yt9cSlbnfAMCRo5FRNZZrqt3YkEtQ0+orAqbFTnxKQ8IRkjphy/McIVlck4ONQWROvtNEs5z7KCC0hW3EX/kCASSc13K5k88Duk3nJLEq+oJenRzyAi0L1GiO91tGlbu9qkXjnL6qe879OOT220+GlkfvsXLQ9VxQ==)

### Other comments
[...]

PR #17839, opened 2023-07-19
docs(VCard): workaround to allow content to overflow, fixes #17593, #17628

### LINKED ISSUE #17593: [Bug Report][3.3.3] VOverlay contained is cropped if content is larger than container
### Environment
**Vuetify Version:** 3.3.3
**Last working version:** 2.6.15
**Vue Version:** 3.3.4
**Browsers:** Mozilla Firefox, Opera
**OS:** Windows 10

### Steps to reproduce
Use `<v-overlay contained>` with content that exceeds the boundaries of outer container.

### Expected Behavior
Both behaviors should be supported depending on our use-case: cropping the content or exceeding the outer container.

Vuetify 2 behavior with `absolute` prop:

image

### Actual Behavior
Content is cropped.

image

### Reproduction Link
* Vuetify 3 with `contained`: [https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNUUtuwjAQvYrldRxLbVdRiugJegDMwiTT4tY/2U4oQty9YwdQBIuQbOb7PuPNiX54X48D0Ia2CYzXMsFKWELakUnvS1iSToaedFrG+C6okexF0EuTkPoa4JwbIWh5JI1xPWg2Sj0AbqQwgKCkczZJZaG/7S6C5+/zAnpQaU8wPlbtLhC+yuEUaWe/SYK/NMPlE/ANKRcu8q6+5iM5mSy3fHYKWlGt7G+sf6KzeKZTnhW0i1HQhmwmIEH3KfnYcN71FifRuhpDbSFx6w3HAyf1dVy/1vjzXsV0LTEtd7EuYBlpK+wZCZXxLiRmpL9jnRqFuVRyDZFyviBhjWM8DDYpA6x3pmh5m7TMyjVEw3bBHSIEBBG0mtFkwU9QLblFigdojnz4NCyA7SFAeNbR3drc1V3rwVlmP+d7nyta1vDwRTLd/gOCzwJK)
* Vuetify 2 with `absolute`: https://codepen.io/3e30K5dl6JkG/pen/yLQYMgx

### Other comments
* In Vuetify 2, `<v-overlay absolute>` had the behavior of exceeding the outer container. Documentation doesn't provide migration alternative.
* As a **feature request**, I'd really like an option to enforce that after exceeding the outer container, the element **never goes outside the screen**; i.e. its top position will always be above 0px and, if needed, page will be scrollable.

scscgit:
## Description
1. Add an **upgrade step** for `v-card` to allow a content to overflow the card by adding `style="overflow: initial"` as per
    * #17593

    Also document a fix for `z-index` stacking context behavior by adding `style="z-index: initial"` as per
    * #17628

2. Create a **Content overflow** section under `v-card` documentation with an explanation of the default behavior<del>, and an interactive example of how to disable it using both of these styles</del>.


johnleider:
I don't think that we need an example for this. Just a mention in the guide with links to the corresponding issues.

scscgit:
@johnleider Is the text in the guide sufficient as it is, only the example should be removed? Also I don't think these two issues are perfectly representative of this happy-path scenario (one is related to `v-overflow` and other covers additional cases), it will feel more like a complex workaround that has to be studied instead of being by design. Should I still link both of them?

Personally I feel like the text
> `v-card` doesn't let its content overflow outside the card. It also creates a `z-index` stacking context, making its content unable to display on top of elements outside `v-card` even if it defines a higher `z-index` value

may get misunderstood by beginners, as it mentions two new concepts that require a specific setup to replicate. For example, there is a parallel `<v-card>` explicitly illustrating the `z-index` issue, and the `scale` (or an absolute position) illustrates the overflow use-case.

More importantly, I also intended this example to be a manual test coverage of the current behavior, so that whenever the scenario or workarounds change (due to any breaking change, a bug, or when new edge cases get discovered -- I'm not confident if this is an exhaustive list of all possible overflow issues, and given the different implementation compared to Vuetify 2, there can be other side-effects), it will be updated as a source of truth for any overflow scenarios. But I can't find any better place for this than a guide. I'd prefer at least linking a `play.vuetifyjs.com` example there instead of a live demo, what do you think?

scscgit:
Updated PR with a less intrusive `play.vuetifyjs.com` link example and linked the issues. Please provide a feedback if you see any reason to completely remove the example.

johnleider:
I'm not sure how the other @vuetifyjs/core-team feel about this but I don't think this belongs in the docs.

MajesticPotatoe:
would agree with @johnleider. Like he had said initially, just the mention in the upgrade guide with corresponding issue links should suffice.

scscgit:
Moved both issue links to the upgrade guide and removed the example.

packages/docs/src/pages/en/components/cards.md
```diff
@@ -75,6 +75,10 @@ In some cases it is possible to combine the different options, like the example
 
 <example file="v-card/basics-combine" />
 
+### Content overflow
+
+By default, `v-card` doesn't let its content overflow outside the card. It also creates a `z-index` stacking context, making its content unable to display on top of elements outside `v-card` even if it defines a higher `z-index` value. If you want to opt-out from this behavior, you can use it like `<v-card style="overflow: initial; z-index: initial">`.
```
johnleider, line 80:
```suggestion
`v-card` does not allow its content to overflow outside the card by default. It also establishes a z-index stacking context, which prevents its content from displaying on top of elements outside the `v-card`, even when it sets a higher z-index value. To override this default behavior, apply the following usage: `<v-card style="overflow: initial; z-index: initial">`.
```

packages/docs/src/pages/en/getting-started/upgrade-guide.md
```diff
@@ -226,3 +226,7 @@ app.use(vuetify)
 - `v-expansion-panel-header` has been renamed to `v-expansion-panel-title`.
 - `v-expansion-panel-content` has been renamed to `v-expansion-panel-text`.
 - `v-expansion-panel` now has `text` and `title` props that can be used instead of subcomponents.
+
+### v-card
+
+- Content is no longer allowed to overflow card or use higher `z-index` values to display on top of elements outside the card. To disable this behavior, use `<v-card style="overflow: initial; z-index: initial">` ([#17593](https://github.com/vuetifyjs/vuetify/issues/17593), [#17628](https://github.com/vuetifyjs/vuetify/issues/17628))
```
johnleider, line 232:
```suggestion
- `v-card` does not allow content to overflow or use higher `z-index` values to display on top of elements outside the it. To disable this behavior, use `<v-card style="overflow: initial; z-index: initial">` ([#17593](https://github.com/vuetifyjs/vuetify/issues/17593), [#17628](https://github.com/vuetifyjs/vuetify/issues/17628))
```

johnleider (changes requested):

packages/docs/src/pages/en/getting-started/upgrade-guide.md
scscgit: "on top of elements outside the it", I assume it'll be "outside it" right?

johnleider (approved):

Issue #17874, opened 2023-07-26
[Bug Report][3.3.9] v-date-picker wrong date returned after Daylight Saving Time


PGLongo:
### Environment
**Vuetify Version:** 3.3.9
**Vue Version:** 3.3.4
**Browsers:** Chrome 114.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
* Change month to December
* Select 25

### Expected Behavior
Return Mon, Dec 25

### Actual Behavior
Return Sun, Dec 24

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNkD1uwzAMha9icLakIV1quEV6jiiDY7GoEusHkuyiCHz3UnIKuM6QbNIj+b5HHq7w4T2fRoQG2oTGD13Cd2mrqp1Y72zqtMVQhCIF912dx5j058+bhOi7HlkX3GiVhFtX6VNkw7zuLzTciv//mxmp5Law6L2GtWIVBWoYtL1Efo7OUsxrnpDQxyihqQ6Lm4SvlHxshOiVpU6Fg54Ct5iE9UbQgjnyfsd3/FUoHdOfxIbuFHkxy05HaWcCauNdSMx0fkNdCoVclKyRU/4/iLCnNhFGm7RBppwpWV6WLCuZYzTsRIeJGMhEQr3C5MBPoB5tS4g7a0G8CQMLaBUGDM9utBlbb7Up3W2W6XO+91xDGaPDl8hw/AXo5N/n)

### Other comments
It works correctly until Sun, Oct 29, probably due to a change from Central European Summer Time to  Central European Standard Time.

29: Date: Sun Oct 29 2023 00:00:00 GMT+0200 (Central European Summer Time)
30: Date: Sun Oct 29 2023 23:00:00 GMT+0100 (Central European Standard Time)

PR #17876, opened 2023-07-26
fix(date): parse string dates in local time zone

nekosaur:
## Description

I was still having issues with date picker selecting incorrect date after https://github.com/vuetifyjs/vuetify/pull/17721. This _should_ work, but Date is a pain so who knows. @abea please check if I haven't broken anything.

I feel like we're going to regret using Date instead of plain strings :) 

Should fix https://github.com/vuetifyjs/vuetify/issues/17874

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker v-model="selectedDate" hide-actions />
      {{ selectedDate }}
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selectedDate = ref(new Date(2023, 0, 1))
</script>
```

### DIFF
--- a/packages/vuetify/package.json
+++ b/packages/vuetify/package.json
@@ -164,6 +164,7 @@
     "rollup-plugin-sass": "^1.2.19",
     "rollup-plugin-sourcemaps": "^0.6.3",
     "rollup-plugin-terser": "^7.0.2",
+    "timezone-mock": "^1.3.6",
     "ts-json-schema-generator": "^1.2.0",
     "vite": "^4.3.5",
     "vite-ssr": "^0.17.1",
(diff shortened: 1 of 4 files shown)


abea:
I tested a few varying time zones and this is looking good. Much more elegant, as well. Do you know what kind of dates or time zones were still causing problems?

nekosaur:
I'm on GMT+2 (GMT+1 plus DST). When clicking on 2nd of January in the provided markup it was selecting January 1st. And so on.

nkamuo:
I'm still having this issue with Vuetify3 VDatePicker component on GMT+1

qxygene1:
i get 21:00:00 GMT anyone has same issue?

xannz:
> I'm on GMT+2 (GMT+1 plus DST). When clicking on 2nd of January in the provided markup it was selecting January 1st. And so on.

I'm still having the same issue and I'm in the same timezone. Selecting a date such as July 9th gives July 8th as value.

PR #17901, opened 2023-07-28
feat(VInput): Add hideSpinButtons prop for VInput

blalan05:
``` vue
<template>
  <v-text-field label="Control" type="number" />
  <v-text-field label="Test" type="number" hide-spin-buttons />
</template>
<script setup>

</script>
```


johnleider:
Since this was never ported I believe it should be considered a feature and go to the dev branch.

Issue #17908, opened 2023-07-29
[Bug Report][3.3.10] VListItem missing anchor title


bogdanciuca:
### Problem to solve
Similar to `VBtn` and `VTab`, `VListItem` can be used as an anchor. Due to the fact that `title` is already used as a prop for the component, it can't be used as an anchor title, thus limiting accessibility and SEO.

### Proposed solution
Define another prop for `VListItem`, e.g. `linkTitle` or `anchorTitle` and assign its value to the tag's `title` property (when the component is rendered as a link).

### Accepted solution
Use item's `title` prop instead, see [discussion](https://github.com/vuetifyjs/vuetify/pull/17909) for further details.

KaelWD:
I brought this up when we added the title prop, if you want title as an attribute you have to use `:title.attr="'Title'"` instead.

johnleider:
> I brought this up when we added the title prop, if you want title as an attribute you have to use `:title.attr="'Title'"` instead.

Add it to the docs

bogdanciuca:
@KaelWD thanks for the suggestion. However, it seems that `title.attr` doesn't work with SSG / SSR, only with CSR. I've opened a [follow up issue](https://github.com/vuetifyjs/vuetify/issues/18457).

PR #17909, opened 2023-07-29
feat(VListItem): support link title when rendered as anchor

bogdanciuca:
## Description

Fixes: #17908

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-card>
        <v-list>
          <v-list-item
            :key="index"
            :title="item.title"
            :link-title="item.linkTitle"
            :href="item.href"
          />
        </v-list>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const items = ref([
    {
      title: 'Vuetify',
      linkTitle: 'Vuetify Website',
      href: 'https://vuetifyjs.com/en/',
    },
    {
      title: 'Vue',
      linkTitle: 'Vue Website',
      href: 'https://vuejs.org/',
    },
  ])
</script>

```

### DIFF
--- a/packages/api-generator/src/locale/en/VListItem.json
+++ b/packages/api-generator/src/locale/en/VListItem.json
@@ -6,7 +6,8 @@
     "title": "Generates a `v-list-item-title` component with the supplied value",
     "value": "The value used for selection.",
     "nav": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-item.json))",
-    "lines": "MISSING DESCRIPTION ([edit in github](https://github.com/vuetifyjs/vuetify/tree/master/packages/api-generator/src/locale/en/v-list-item.json))"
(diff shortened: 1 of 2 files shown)


bogdanciuca:
@johnleider thanks for the feedback. Please advise on the concerns below.

> I agree that this is an issue but I'd rather apply the item's **title** to the **title** attribute explicitly instead of coming up with a new property. 

This is how I initially thought of implementing it. However, there were some concerns that came up:

1. List item links could potentially contain only icons (or maybe only subtitles / other content provided via templates), so the `title` prop may not be provided. In this situations `a[title]` will not be set.
2. Inability of setting a `title` which differs from the link text.

> I also imagine there are other places this could be useful.

Do you have any other components in mind?

> And finally, if this is a feature, it needs to go to the **dev** branch.

Gotcha, will do in the case this remains a feature and not a fix.

johnleider:
> This is how I initially thought of implementing it. However, there were some concerns that...

I feel like those use-cases are very niche and I would suggest the user wrap the `v-list-item` component in a `router-link`.

bogdanciuca:
Closed in favor of https://github.com/vuetifyjs/vuetify/pull/17940.

Issue #17872, opened 2023-07-25
[Bug Report][3.3.9] v-date-picker emits update:modelValue twice when selecting a date


mikestu:
### Environment
**Vuetify Version:** 3.3.9
**Vue Version:** 3.3.4
**Browsers:** Chrome 114.0.0.0
**OS:** Windows 10

### Steps to reproduce
Set v-model on a v-date-picker instance, select a date, then in the Vue devtools, you can see two update:modelValue events being emitted.  In the reproduction link, open the console, select a date, and see two lines output.

### Expected Behavior
A single update:modelValue event should be emitted.

### Actual Behavior
Two update:modelValue events are emitted.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkctqwzAQRX9lEIU6EFuLdtPghBb6Dd3EgTj2JFFjPZAUhxL87x3JNs1jE+jKnsedM3e0PLMPY7L2iGzGco/SNKXHRaEA8jYtjYm/Mai08qVQaIdUTNbUnRpRHdBCm0pdYzMvmMMGK4/1J1ULtsj5VeM4krLXM0OmR+b8YhUKXWWF8eDQH/uNhDTaejiDxe0UTqWv9tDB1moJz2TmOagAaLwLqr9tYB4UicIThDCZTPrOOCHpF0smMF9cqbK2bI447cuk/QphaDqPpwgk3WDW6F2yvgJW+1LtsAavZ/B0HsXdmsBB2IUP/ee890ju2JT19lJZmuzbaUVvE0nFUHAFm43sgpHfEBds771xM86rWpGMnkK0NlPouTKSv1Mbt0flhcS01vL9JXvJXnktnL9MZ+hkurH65NDSkIINriPGi+3PA6ihMxLeesKQSpty4wLibjQnXos2tahqtGgfdXQju3R1U7pzFuh0/o7u3Qh1cDenrlw883Jc8l+W47AwaRWA3ZTFPYkcBWz1C0iXL5s=)

Issue #17867, opened 2023-07-24
[Bug Report][3.3.9] v-date-picker with hide-actions causes 'Maximum recursive updates exceeded'


mikestu:
### Environment
**Vuetify Version:** 3.3.9
**Vue Version:** 3.3.4
**Browsers:** Chrome 114.0.0.0
**OS:** Windows 10

### Steps to reproduce
Have a v-model set on v-date-picker, as well as "hide-actions".  Click any date in the picker, then you'll see this in the console:

[Vue warn]: Maximum recursive updates exceeded in component <VContainer>. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.

### Expected Behavior
It just emits modelValue, without a recursive issue.

### Actual Behavior
Selecting a date will emit for modelValue, but it takes longer, and has some recursive issue.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkE1uwjAQha9ieQNIOF7QTVFAVOotMIsQD8Il/pFtgiqUu3fsBDXABomd5/nN+2Zme6VfzhXtGeiSlhG0a6oIa2EIKVtWOZefuaitiZUy4AcpixLdzKn6BJ60TFsJzUrQAA3UEeQ3/gpKjkoCq+qorAnrkt913fJRvQckpeeXfDQXlqH2ykUSIJ778ZR21kdyJR4OpCMHbzWZ4EqTZCcEc0Oy/89EVsk6NXAhqZzOZonS52IindM+kunKFT/BGjzONUWJ4SMIuiRZSRqiUi3oMUYXlpzX0mAb3kK1vjAQuXGab9DG/dlEpYFJqzeLYlF8cKlCHMsFBM323l4CeAwRdD7CRHX4fQE1ODPhsycMEmuqfUiIp2iOvBY882AkePCvbvTQNt7q4etps0TvhOnw3o0yp/Bw6jrkM29vQ761cg5LSbsE7OY0z4nk3EB3f2nzBzk=)

PR #17914, opened 2023-07-29
fix(VDatePicker): recursive update when using hide-actions

nekosaur:
## Description

closes https://github.com/vuetifyjs/vuetify/issues/17867, closes https://github.com/vuetifyjs/vuetify/issues/17872

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker v-model="selectedDate" />
      <v-date-picker v-model="selectedDate" hide-actions />
    </v-container>
  </v-app>
</template>

<script setup>

  const selectedDate = ref(new Date())

  watch(
    () => selectedDate.value,
    newValue => {
      console.log(`selectedDate changed to: ${newValue}`)
    }
  )
</script>

```

### DIFF
--- a/packages/vuetify/src/labs/VDateInput/composables.ts
+++ b/packages/vuetify/src/labs/VDateInput/composables.ts
@@ -78,8 +78,12 @@ export function createDateInput (props: DateInputProps, isRange: boolean) {
 
   function isEqual (model: readonly any[], comparing: readonly any[]) {
     if (model.length !== comparing.length) return false
-    if (model[0] && comparing[0] && !adapter.isEqual(model[0], comparing[0])) return false
-    if (model[1] && comparing[1] && !adapter.isEqual(model[1], comparing[1])) return false
+
+    for (let i = 0; i < model.length; i++) {
+      if (comparing[i] && !adapter.isEqual(model[i], comparing[i])) {
+        return false
+      }
+    }
 
     return true
   }
--- a/packages/vuetify/src/labs/VDateInput/composables.ts
+++ b/packages/vuetify/src/labs/VDateInput/composables.ts
@@ -76,6 +76,14 @@ export function createDateInput (props: DateInputProps, isRange: boolean) {
     return adapter.isValid(date) ? date : fallback
   }
 
+  function isEqual (model: readonly any[], comparing: readonly any[]) {
+    if (model.length !== comparing.length) return false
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/labs/VDatePicker/VDatePicker.tsx
```diff
@@ -175,9 +200,10 @@ export const VDatePicker = genericComponent<VDatePickerSlots>()({
             ) : (
               <div class="v-date-picker__input">
                 <VTextField
-                  v-model={ inputModel.value }
+                  modelValue={ inputModel.value[0] }
+                  onUpdate:modelValue={ v => updateFromInput(v, 0) }
```
johnleider, line 204: updateFromInput is only used here. Does it need to accept an index arg?

packages/vuetify/src/labs/VDateInput/composables.ts
```diff
@@ -76,12 +76,21 @@ export function createDateInput (props: DateInputProps, isRange: boolean) {
     return adapter.isValid(date) ? date : fallback
   }
 
+  function isEqual (model: readonly any[], comparing: readonly any[]) {
+    if (model.length !== comparing.length) return false
+    if (model[0] && comparing[0] && !adapter.isEqual(model[0], comparing[0])) return false
+    if (model[1] && comparing[1] && !adapter.isEqual(model[1], comparing[1])) return false
+
+    return true
+  }
```
johnleider, line 85:
```suggestion
  function isEqual (model: readonly any[], comparing: readonly any[]) {
    if (model.length !== comparing.length) return false

    for (let i = 0; i < model.length; i++) {
      if (comparing[i] && !adapter.isEqual(model[i], comparing[i])) {
        return false
      }
    }

    return true
  }
```

packages/vuetify/src/labs/VDatePicker/VDatePicker.tsx
nekosaur: It will be needed for range picker as well, at which point it'll be moved somewhere else.

packages/vuetify/src/labs/VDateInput/composables.ts
nekosaur: I mean, sure. But it'll only ever have a maximum of two dates.

johnleider (approved):

johnleider:
I verified all tests locally.

Issue #17848, opened 2023-07-20
[Feature Request] Add prop to set crossorigin attr on v-img


martinEtflais:
### Problem to solve
Well the title speak for itself. At the moment there is no way except of trying to target every img in the document by query to set the crossorigin attribute so images from different origin doesnt get loaded

### Proposed solution
Add some prop to v-img?

websitevirtuoso:
Hope PR will be accepted soon

PR #17927, opened 2023-07-31
feat(VImg): Add prop to set crossorigin attr on v-img

websitevirtuoso:
## Description

https://github.com/vuetifyjs/vuetify/issues/17848

## Markup:

```vue
  <v-app>
      <v-img src="https://t3.ftcdn.net/jpg/00/92/53/56/360_F_92535664_IvFsQeHjBzfE6sD4VHdO8u5OHUSc6yHF.jpg" />
</v-app>
  <v-app>
      <v-img  crossorigin="use-credentials" src="https://t3.ftcdn.net/jpg/00/92/53/56/360_F_92535664_IvFsQeHjBzfE6sD4VHdO8u5OHUSc6yHF.jpg" />
</v-app>
```


KaelWD:
> If the crossorigin attribute is specified, then a CORS request is sent (with the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) request header); but if the server does not opt into allowing cross-origin access to the image data by the origin site (by not sending any [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) response header, or by not including the site's origin in any [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) response header it does send), then the browser blocks the image from loading, and logs a CORS error to the devtools console.

Sounds like the default should be undefined.

websitevirtuoso:
Yes we can set undefined or even empty string but this is the same

image

websitevirtuoso:
About undefined very clear. But not clear about boolean shorthand? How it can be shorthand if it can have 2 options?
'anonymous' | 'use-credentials'

websitevirtuoso:
@KaelWD Could you assist with this PR please? I would create more PR but want to close this one first

Issue #17629, opened 2023-06-15
[Bug Report][3.2.5] A11y issues with VDataTable and VAutocomplete


chriskulwik:
### Environment
**Vuetify Version:** 3.2.5
**Vue Version:** 3.2.45
**Browsers:** Chrome 110.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Using the Vue Axe Popup tool, I can see there are accessibility issues with VDataTable and VAutoComplete

VAutoComplete: 

	On div.v-field element
	- ARIA attribute is not allowed: aria-expanded="false"
	On input element 	
    - ARIA role listbox is not allowed for given element

VDataTable: 

	On div.v-field element (dropdown for items for page) 
	- ARIA attribute is not allowed: aria-expanded="false"
	- Form element has explicit <label> that is hidden
	- aria-label attribute does not exist or is empty

	On input element (dropdown for items for page)
	- form elements should have a visible label

### Expected Behavior
VAutoComplete and VDataTable should pass the Vue Axe Popup test

### Actual Behavior
They do not

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkMtOwzAQRX/FeBOQmljisakCKjv+oenCTabUED9kT4JQlX9n7AQpTTeVuvNc37lnZrYn/u5c0XfA17xE0K6VCG+VYazsc+lceqaitgalMuAnaXR0aGtLXUBdpVgIM2MjUeYo9+1om5UTgMRzQkpLA5RiNhiVofbKIQuA3Tif0s56ZCfm4cAGdvBWs4x2yqKdMcoNyHT4ZK/RcZ99QNtahkfwcJc9RMAYSWF8xce0XEtXfAVr6DCnmFJNH6Hia5aUqBEl1hU/IrqwFqJuDLU10KreFwZQGKfFhmzCdwaVhryxevNUPBbPL6JRAed6AUHne29/AnhKqfhqxkF1+L2CNTkTYiJMUt7KfYiIi2hBvB587sE0dBR/7UqLtrO1Fn8Xq0X8UJmBLt4q8x0Wx65DOvT2f8qbdk5hMWkXgcOKj4MSOnXw3R/owAUf)

### Other comments
Vue Axe Popup repo: https://github.com/vue-a11y/vue-axe-next

I added Vue Axe Popup to my project like this:

import('vue-axe').then(({ default: VueAxe, VueAxePopup }) => {
[...]

PR #17938, opened 2023-08-01
fix(VAutocomplete): fix accessibility issues

tmasrat:
## Description

1. v-autocomplete__menu-icon (the drowdown icon) should be hidden from screen readers so I am setting aria-hidden to true

    - This change is fixing the following AXE issue: _Ensures every ARIA button, link and menuitem has an accessible name_

2.  role has to be comobox for autocompletes but that's being set in VTextField so I am passing in roleType to VTextField. When roleType is defined, the role will be combox if not it will be textbox so it shouldn't affect other components that are using VTextField (including VTextField)

    - This change is fixing: Ensures ARIA attributes are allowed for an element's role

resolves part of #17629

## Markup:

```
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        placeholder="Test Autocomplete"
        persistent-placeholder
        :items="itemsForSelect"
        aria-label="Testing Aria Label"
        item-value="Test Value"
        item-title="Test Value"
        density="compact"
        variant="outlined"
        clearable
      />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',

    setup () {
      const itemsForSelect = ['one', 'two', 'three']
      return {
        itemsForSelect,
      }
    },
  }
</script>

```


bogdanciuca:
I believe this is a duplicate of https://github.com/vuetifyjs/vuetify/pull/17786

tmasrat:
> I believe this is a duplicate of #17786

Yeah you're right. I checked out the branch and it solves the a11y problems in v-texts and autocompletes which is what I needed

Issue #17933, opened 2023-08-01
[Bug Report][3.3.10] contain class on v-img not working


stanig2106:
### Environment
**Vuetify Version:** 3.3.10
**Vue Version:** 3.3.4
**Browsers:** Chrome 115.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
use content-class props on v-img

### Expected Behavior
the class go to the \<img\> html element

### Actual Behavior
the class doesn't apply

### Reproduction Link
[https://codepen.io/stanig2106/pen/LYXMJzB](https://codepen.io/stanig2106/pen/LYXMJzB)

### Other comments
https://github.com/vuetifyjs/vuetify/blob/4ec7f13e39f4e6da4dc70910eeab1852f47363d2/packages/vuetify/src/components/VImg/VImg.tsx#L203C5-L207C1
https://github.com/vuetifyjs/vuetify/blob/4ec7f13e39f4e6da4dc70910eeab1852f47363d2/packages/vuetify/src/components/VImg/VImg.tsx#L212C9-L213C55
```typescript
// source code of the bug : 
const containClasses = computed(() => ({
      'v-img__img--cover': props.cover,
      'v-img__img--contain': !props.cover,
    }))

// ...
<img class={['v-img__img', containClasses.value]}
```

the classes is never applied in the source code

bogdanciuca:
If you want the css to be applied to the image container, use `class` instead of `content-class`.

stanig2106:
`class` apply class on the div around the image

API doc : 
https://vuetifyjs.com/en/api/v-img/#props-content-class

bogdanciuca:
Gotcha, I raised a [PR](https://github.com/vuetifyjs/vuetify/pull/17939).

stanig2106:
@KaelWD 
The documentation does state that content-class is to 'Apply a custom class to the internal content element.' 
However, the debate isn't about that, given that content-class isn't applied anywhere. It just isn't applied at all...

KaelWD:
The content element is only rendered if the default slot is used.

joshuadabejero:
Do we have fix on this? Got the same issue:

```
.hero-image > .v-image__image {
  filter: brightness(10%);
  -webkit-filter: brightness(10%);
}
```

Tyler-widd:
Same

PR #17939, opened 2023-08-01
fix(VImg): add content-class

bogdanciuca:
## Description

`content-class` value is not set as a class on ``. Fixes https://github.com/vuetifyjs/vuetify/issues/17933
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-img
        width="100"
        src="https://play.vuetifyjs.com/assets/vuetify-play.41631d9c.svg"
        content-class="my-class"
      />
    </v-container>
  </v-app>
</template>

<script setup></script>
```


johnleider:
Features should be based off of the `dev` branch.

PR #17940, opened 2023-08-01
fix(VListItem): add anchor title

bogdanciuca:
## Description

`VListItem` anchor is missing the `title` prop. Fixes: https://github.com/vuetifyjs/vuetify/issues/17908

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-card>
        <v-list>
          <v-list-item
            :key="index"
            :title="item.title"
            :href="item.href"
          />
        </v-list>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const items = ref([
    {
      title: 'Vuetify',
      href: 'https://vuetifyjs.com/en/',
    },
    {
      title: 'Vue',
      href: 'https://vuejs.org/',
    },
  ])
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -220,6 +220,7 @@ export const VListItem = genericComponent<VListItemSlots>()({
           ]}
           href={ link.href.value }
           tabindex={ isClickable.value ? (list ? -2 : 0) : undefined }
+          title={ slots.title?.({ title: props.title }) ?? props.title }
           onClick={ onClick }
           onKeydown={ isClickable.value && !isLink.value && onKeyDown }
           v-ripple={ isClickable.value && props.ripple }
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VList/VListItem.tsx
```diff
@@ -220,6 +220,7 @@ export const VListItem = genericComponent<VListItemSlots>()({
           ]}
           href={ link.href.value }
           tabindex={ isClickable.value ? (list ? -2 : 0) : undefined }
+          title={ slots.title?.({ title: props.title }) ?? props.title }
```
johnleider, line 223: I've never seen a slot used like this before. It also means you'd have to update the available slots for this component. I think it's safe for this to simply equal props.title.

johnleider (changes requested):

johnleider (approved):

KaelWD:
I'm reverting this, it adds `title="false"` if the title prop is not set and even if it is set you don't always want a tooltip.

johnleider:
> I'm reverting this, it adds `title="false"` if the title prop is not set and even if it is set you don't always want a tooltip.

Then why not adjust it to be undefined if not set instead of reverting it?

KaelWD:
Because the second part of that sentence.

chunlampang:
Because of `title="false`, when hover the list item, it showing "false" by the default html title attribute.

PR #17952, opened 2023-08-03
feat(VColorPicker): add eyedropper button when supported

### LINKED ISSUE #14790: [Feature Request] Support eye dropper for color picker
### Problem to solve
Chromium based browsers now support native eye dropper. This would be useful for users in selecting a color from e.g. an image they have open on their computer.

https://caniuse.com/mdn-api_eyedropper

### Proposed solution
It could be a small icon button within the color picker that is only displayed when eye dropper has browser support.

prashantsinghb:
resolves #14790

## Description

In EyeDropper API compatible browser, an icon will be displayed in VColorPicker to sample colors from their screens, including outside of the browser window.

**Queries** 
- Position of EyeDropper Icon in color picker?
- Should EyeDropper icon be prop?

image

## Markup:

```vue
<template>
  <div class="d-flex justify-space-around">
    <v-color-picker
      v-model="picker"
    ></v-color-picker>
  </div>
</template>
<script>
  export default {
    data () {
      return {
        picker: null,
      }
    },
  }
</script>
```


packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
```diff
@@ -3,18 +3,25 @@ import './VColorPickerPreview.sass'
...
 
 // Types
 import type { PropType } from 'vue'
 import type { HSV } from '@/util'
 
+declare global {
+  interface Window {
+      EyeDropper?: any;
+  }
+}
```
KaelWD, line 23: Move to globals.d.ts and maybe add a proper interface instead of any

packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
```diff
@@ -35,6 +42,18 @@ export const VColorPickerPreview = defineComponent({
   },
 
   setup (props, { emit }) {
+    const supportEyePicker = 'EyeDropper' in window
```
KaelWD, line 45: check `IN_BROWSER` first

packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
```diff
@@ -35,6 +42,18 @@ export const VColorPickerPreview = defineComponent({
...
+
+    async function openEyePicker(){
+      const eyeDropper = new EyeDropper()
+      try{
+        const result = await eyeDropper.open();
```
KaelWD, line 50: Should probably be passed AbortSignal when the component unmounts

packages/vuetify/src/iconsets/mdi.ts
```diff
@@ -43,6 +43,7 @@ const aliases: IconAliases = {
   file: 'mdi-paperclip',
   plus: 'mdi-plus',
   minus: 'mdi-minus',
+  eyeDropper: 'mdi-eyedropper'
```
KaelWD, line 46: Add for the other icon sets too.

KaelWD (changes requested):

KaelWD:
> Position of EyeDropper Icon in color picker?

Chrome has it on the left before the circle preview

> Should EyeDropper icon be prop?

We don't have one for the expand icon either, it's probably fine.

packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
prashantsinghb: Moved to global.d.ts

packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
prashantsinghb: added in IN_BROWSER check

packages/vuetify/src/components/VColorPicker/VColorPickerPreview.tsx
prashantsinghb: Done.

packages/vuetify/src/iconsets/mdi.ts
prashantsinghb: Done.

prashantsinghb:
Moved position of eyedropper to left before circle preview

image

PR #17953, opened 2023-08-03
docs(VOtpInput): fix closing tag in documentation

GaetanWi:
## Description

Fixing the ending tag from the Vuetify documentation of the new OtpInput https://vuetifyjs.com/en/components/otp-input/ (ending tag is the stepper's one)

## Markup:

```vue

```


GaetanWi:
Oh well, seems like unit tests are broken  :'(

Issue #4843, opened 2018-08-12
[Feature Request] Add speech bubble prop to v-dialog / v-card?


axdyng:
### What problem does this feature solve?

Vuetify users don't have to create their own "speech bubble" ui since it's visually close.

### What is your proposed solution?

include a prop where "speech-bubble" = true/false

CSS-side:
Add an :after pseudo-element to style the element

A good example is this 
https://leaverou.github.io/bubbly/

johnleider:
@abhaywawale is currently working on this

johnleider:
Unfortunately the champion for this is mia. I'm going to tag as help wanted but will be removing it from the v2.0 milestone.

kamikazechaser:
This would be a good feature.

KareemDa:
I've tried to implement this feature but I had a problem with `elevation` & `v-ripple` directive

reuwi:
any progress？

johnleider:
> any progress？

There was https://github.com/vuetifyjs/vuetify/pull/18079 but it was a little heavier handed than I had hoped.

J-Sek:
If anyone would have time to give it a shot, here is how I would do it in user-land: [playground](https://play.vuetifyjs.com/playgrounds/1rYAPg)

J-Sek:
- CSS `anchor(...)` is getting us closer ([article](https://frontendmasters.com/blog/perfectly-pointed-tooltips-a-foundation/)), but we are still missing `border-shape` [[status](https://chromestatus.com/feature/5459864205393920)]

J-Sek:
Even with border-shape, it is going to require too much work for a simple decoration. Closing to clear the backlog as there are no plans nor resources to pull off anything better than those demos linked above.

PR #17958, opened 2023-08-04
feat(VTooltip): added speechBubble prop

### LINKED ISSUE #12939: [Feature Request] v-tooltip with arrow to emphasize direction
### Problem to solve
This feature would improve user experience (UX) and help with giving the right direction to the tip being show. Currently, there is no direction, and after thousands of user tests we've run across our platforms, we've noticed that user's believe the tooltips are the buttons instead of telling the user what the button actually does. This is because the tooltip themselves look as if they are independent from their original object.  

### Proposed solution
Like many tooltips across many online packages, they usually always have an arrow pointing towards its object. Most notably, the [Bootstrap tooltip](https://getbootstrap.com/docs/4.0/components/tooltips/). 

**Example with tooltip and arrow**

Screen Shot 2021-01-10 at 9 11 20 PM

**Compared to what it currently does:** 

This is just one example of many, for the user's its hard for them to understand the tooltip is a helpful guide rather than the button they should click. Especially if you activate tooltips without hover, its even more confusing towards users, 78% of the sample of users tested clicked the tooltip message instead of the buttons themselves. 

Screen Shot 2021-01-10 at 9 17 15 PM

prashantsinghb:
## Description

Resolves #12939 (Closed as duplicate of #4843) 
This PR added speechBubble prop to VTooltip

**NOTE:**  

- For #4843 have raised separate PR #18079

**Queries:**
- Name of "speechBubble" composable and its props?
- Show arrow position be decided by user (By prop pointerPosition) or we should compute it for the user(as attempted in this PR)?

## Markup:

```vue
<template>
  <v-container
    fluid
    class="text-center"
  >
    <v-row
      class="flex"
      justify="space-between"
    >
      <v-col cols="12">
        <v-btn @click="show = !show">
          toggle
        </v-btn>
      </v-col>

      <v-col
        cols="12"
        class="mt-12"
      >
        <v-tooltip
          v-model="show"
          speech-bubble
        >
            <v-btn
              icon
              v-bind="props"
            >
              <v-icon color="grey-lighten-1">
                mdi-cart
              </v-icon>
            </v-btn>
          </template>
          <span>Programmatic tooltip</span>
        </v-tooltip>
      </v-col>
    </v-row>
  </v-container>
</template>
<script>
  export default {
    data () {
      return {
        show: false,
      }
    },
  }
</script>

```


J-Sek:
Slightly too much overhead for something outside of Material spec anyway. And it is just a PoC. We would need more code to actually support rounded corners and make the outline look correct.

PR #17979, opened 2023-08-07
fix(VTab): changed classes from v-slide-group to v-tabs

### LINKED ISSUE #17975: [Bug Report][3.3.11] VTabs SASS Variables customization not working because v-slide-group--horizontal class is not applied
### Environment
**Vuetify Version:** 3.3.11
**Vue Version:** 3.3.4
**Browsers:** Firefox 116.0
**OS:** Linux x86_64

### Steps to reproduce
It's difficult give a good reproduction link so I will explain what's happening in more detail here:

In the file https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VTabs/VTab.sass the customization for `$tab-max-width` is only inside `.v-slide-group--horizontal &`, which means that `.v-slide-group--horizontal` must be included in the parent element for the customization to work. Unfortunately the direction only gets directly specified when prop `direction="vertical"` is passed, which means that this variable is never used for horizontal tabs.

### Expected Behavior
`$tab-max-width` sets the maximum tab width

### Actual Behavior
Nothing gets changed

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlUEFugzAQ/AryGdui5IRQlP6htzgHAlvVLTaWvaGqIv5e25iKkiqK1JtnxruzM8creTaGjRcgFakRlOkbhL3QWVaPtDEmPiNoB42N1GATFUlszu4HL8z+BRxmRc1n9Lf8dF8u78u7rZzwckyAvw8OzJyn5qucJCdSmcEiVY1h727QvohrGBBJcIJUWWQC55sKWJA3ROMqzttO+7EOejlapgG5Noof/DduLxqlAtoN6lCyku14Jx2uaQZO0bMdPh1Yv0SQfGWD8vXrAav0MzoUxWyRONr7QoLHzW7uDUew1ILuwIJ9NNJmbB1rI91EC+6T0JMvvJf6w226bl3s+bgc+b/McVtYdQqOU07iod56niCnb6Hm9aU=)

### Other comments
The fix to this issue would be to add the `.v-slide-group--horizontal` class to the parent component by default.

RCRalph:
fixes #17975

## Description
This PR changes SCSS customization of `VTab`, which previously used classes such as `.v-slide-group--horizontal` and `.v-slide-group--vertical`. Now it accomplishes the same by using classes such as `.v-tabs--horizontal` and `.v-tabs--vertical`. These classes are already being added by Vuetify, but previously, as far as I know, there weren't any uses for them.

## Markup:
Instead of setting up [treeshaking](https://vuetifyjs.com/en/features/treeshaking/) and [component specific variables](https://vuetifyjs.com/en/features/sass-variables/#component-specific-variables), I made a playground which edits these styles using component scoped styles. The first one works only for direction="horizontal" and, since the max-width style could not specified before this fix, doesn't require !important to work. The other three styles work only for direction="vertical" and were already working before this change, hence !important. I added these to the PR because I think it would help with code consistency.

```vue
<template>
  <v-app>
    <v-container>
      <v-tabs direction="horizontal">
        <v-tab>Iteeeeeeeem 1</v-tab>
        <v-tab>Item 2</v-tab>
        <v-tab>Item 3</v-tab>
      </v-tabs>
    </v-container>
  </v-app>
</template>

<style>
  /* Requires setting direction='horizontal' */
  .v-tabs--horizontal .v-tab {
    max-width: 100px;
  }

  /* Requires setting direction='vertical' */
  .v-tabs--vertical .v-tab {
    justify-content: end !important;
  }

  .v-tabs--vertical .v-tab__slider {
    width: 10px !important;
  }

  .v-tabs.v-tabs--vertical {
    --v-tabs-height: 100px !important;
  }
</style>
```

### DIFF
--- a/packages/docs/src/components/app/Toc.vue
+++ b/packages/docs/src/components/app/Toc.vue
@@ -89,7 +89,7 @@
             </v-card>
           </v-col>
 
-          <v-col cols="12">
+          <!-- <v-col cols="12">
             <a
               href="https://themeselection.com/item/category/vuejs-admin-templates/?utm_source=vuetify&utm_medium=banner&utm_campaign=category_page&utm_id=12"
               target="_blank"
@@ -98,7 +98,7 @@
             >
               <v-img src="https://cdn.vuetifyjs.com/docs/images/promotions/theme-selection-dashboard-2023/vuetify-ad-banner.png" />
             </a>
-          </v-col>
+          </v-col> -->
         </v-row>
       </v-container>
     </template>
@@ -112,7 +112,7 @@
   // Composables
   import { RouteLocation, Router, useRoute, useRouter } from 'vue-router'
   import { useAppStore } from '@/store/app'
-  import { useGtag } from 'vue-gtag-next'
+  // import { useGtag } from 'vue-gtag-next'
   import { useSponsorsStore } from '@/store/sponsors'
   import { useTheme } from 'vuetify'
 
@@ -219,7 +219,7 @@
   const route = useRoute()
   const router = useRouter()
   const theme = useTheme()
-  const { event } = useGtag()
+  // const { event } = useGtag()
 
   const { scrolling } = useUpdateHashOnScroll(route, router)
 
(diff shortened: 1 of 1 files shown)


KaelWD:
Why not add v-slide-group--horizontal to v-slide-group?

johnleider:
I'm okay with this change.

PR #17988, opened 2023-08-08
fix(locale): update no translations

Bjornar97:
datePicker, pagination, stepper, rating, loading and infiniteScroll

## Description
Add some missing translation in locale/no

### DIFF
--- a/packages/vuetify/src/locale/no.ts
+++ b/packages/vuetify/src/locale/no.ts
@@ -31,15 +31,15 @@ export default {
   },
   datePicker: {
     ok: 'OK',
-    cancel: 'Cancel',
+    cancel: 'Avbryt',
     range: {
-      title: 'Select dates',
-      header: 'Enter dates',
+      title: 'Velg datoer',
+      header: 'Velg datoer',
     },
-    title: 'Select date',
-    header: 'Enter date',
+    title: 'Velg dato',
+    header: 'Velg dato',
     input: {
-      placeholder: 'Enter date',
+      placeholder: 'Velg dato',
     },
   },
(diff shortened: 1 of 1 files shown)


KaelWD:
@rosvik @marduv Review please

packages/vuetify/src/locale/no.ts
```diff
@@ -31,15 +31,15 @@ export default {
...
+      title: 'Velg datoer',
+      header: 'Velg datoer',
     },
-    title: 'Select date',
-    header: 'Enter date',
+    title: 'Velg dato',
+    header: 'Velg dato',
     input: {
-      placeholder: 'Enter date',
+      placeholder: 'Velg dato',
```
rosvik, line 42: Maybe the context where this is used suggests otherwise, but could "Fyll inn dato" be a better translation for "Enter date"? That would differentiate between "enter" and "select", and might make sense for a placeholder at least.

rosvik (commented):
Looks good to me 👍 Have one tiny suggestion on the date picker translations.

packages/vuetify/src/locale/no.ts
Bjornar97: image

This is where the title and header is used. I feel like it "Fyll inn dato" doesnt work here. 

Here is with input-mode="keyboard", and here I agree, it makes more sense with "Fyll inn dato":
image

rosvik (commented):
@Bjornar97 I don't have permissions to approve PRs, but would if I could. Looks great ✅

marduv:
Just a minor thing I can see is: `to` should be `til`
```
dateRangeInput: {
    divider: 'to',
},
```

PR #18010, opened 2023-08-09
feat(display): add local mobile-breakpoint support

### LINKED ISSUE #15381: [Feature Request] Add missing `mobile-breakpoint` prop in navigation drawer like v2
### Problem to solve
Same as v2

### Proposed solution
Same as v2

johnleider:
## Motivation and Context
resolves #15381

## Markup:
<details>

```vue
<template>
  <v-app>
    <v-navigation-drawer mobile-breakpoint="xl" color="purple" />
    <v-navigation-drawer location="end" color="primary" />
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  import { useDisplay } from '../src/composables/display'

  const { mobile, isLocalMobile, displayClasses } = useDisplay({
    mobileBreakpoint: 2156,
  }, 'v-my-class')

  console.log(mobile, isLocalMobile)
  console.log(displayClasses.value)
</script>

```
</details>

### DIFF
--- a/packages/docs/src/pages/en/features/display-and-platform.md
+++ b/packages/docs/src/pages/en/features/display-and-platform.md
@@ -242,7 +242,7 @@ The following components have built in support for the **mobile-breakpoint** pro
 
 By default, **mobileBreakpoint** is set to **lg**, which means that if the window is less than _1280_ pixels in width (which is the default value for the **lg** threshold), then the **useDisplay** composable will update its **mobile** value to `true`.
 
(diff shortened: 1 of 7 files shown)


KaelWD:
Does there need to be a distinction between mobile and localMobile?

johnleider:
> Does there need to be a distinction between mobile and localMobile?

I was looking forward to feedback on this part. I wasn't sure what to call it or how to differentiate it from **mobile**. I feel like it was a missed opportunity not originally naming it **isMobile**. I'm open to any suggestions.

KaelWD:
I was thinking just keep it always mobile, if props are provided and the prop is set then it would be local instead of global

KaelWD:
Like that

PR #18018, opened 2023-08-09
fix(VSelect): auto scroll to selected value

### LINKED ISSUE #17881: [Feature Request] v-select auto scroll to selected item
### Problem to solve
When we open the v-select item if we have many items behind a scrollbar will scroll to the preselected item.

### Proposed solution
Scroll to the selected item when the menu gets activated.

avarixe:
## Description

fixes #17881

implements the v2 functionality of scrolling to the selected value by setting `menuProps="auto"`. ~~Since this doesn't go through VMenu, prop was renamed to `autoScroll`.~~

Currently, if a user needs to select a new value in the middle of a long list, the dropdown always resets to the top of the list, even if the new value is only just after the existing value. This change takes advantage of the virtual scrolling added to VSelect to quickly scroll to the current location.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-select v-model="selected2" label="Short Select" :items="[1, 2, 3]" />

      <v-select v-model="selected3" label="Value not in items" :items="[1, 2, 3]" />

      <v-select v-model="selected" label="Longer Select" :items="items" />

      <v-autocomplete v-model="selected" label="Autocomplete" :items="items" />

      <v-combobox v-model="selected" label="Combobox" :items="items" />

      <v-select
        v-model="fruit"
        label="Select with Objects"
        :items="fruits"
        item-title="name"
        item-value="id"
        :return-object="false"
      />
    </v-container>
  </v-app>
</template>

<script lang="ts" setup>
  import { ref } from 'vue'

  const selected = ref(1000)
  const selected2 = ref(2)
  const selected3 = ref(4)
  const items = Array.from({ length: 20000 }, (_, i) => i + 1)

  const fruits = [
    { name: 'John', id: 1, fruit: 'apple' },
    { name: 'Jane', id: 2, fruit: 'banana' },
    { name: 'Joe', id: 3, fruit: 'cantelope' },
    { name: 'Doug', id: 4, fruit: 'dragonfruit' },
    { name: 'Eddy', id: 5, fruit: 'orange' },
    { name: 'David', id: 6, fruit: 'mango' },
    { name: 'Sam', id: 7, fruit: 'durian' },
    { name: 'Sally', id: 8, fruit: 'lychee' },
    { name: 'Kyle', id: 9, fruit: 'tangerine' },
    { name: 'Kevin', id: 10, fruit: 'strawberry' },
    { name: 'Daniel', id: 11, fruit: 'watermelon' }
  ]
  const fruit = ref(8)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -25,7 +25,17 @@ import { makeTransitionProps } from '@/composables/transition'
 
 // Utilities
 import { computed, mergeProps, nextTick, ref, shallowRef, watch } from 'vue'
-import { genericComponent, getPropertyFromItem, matchesSelector, noop, omit, propsFactory, useRender, wrapInArray } from '@/util'
+import {
+  genericComponent,
+  getPropertyFromItem,
+  IN_BROWSER,
+  matchesSelector,
+  noop,
+  omit,
+  propsFactory,
+  useRender,
+  wrapInArray,
+} from '@/util'
 
 // Types
 import type { PropType } from 'vue'
@@ -379,7 +389,7 @@ export const VAutocomplete = genericComponent<new <
         const index = displayItems.value.findIndex(
           item => selections.value.some(s => item.value === s.value)
         )
-        setTimeout(() => {
+        IN_BROWSER && window.requestAnimationFrame(() => {
           index >= 0 && vVirtualScrollRef.value?.scrollToIndex(index)
         })
       }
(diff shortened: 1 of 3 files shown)


websitevirtuoso:
I think would be great to make it enabled by default

avarixe:
> I think would be great to make it enabled by default

I think so as well, I just didn't want to presume that's how the maintainers want it, since I don't think `menuProps="auto"` was enabled by default in v2.

KaelWD:
Yeah this doesn't need to be a prop, auto was because it also affected menu position

johnleider (approved):

PR #18090, opened 2023-08-18
fix(VCombobox): mobile browser UX issue on focus

shiouming:
fixes #13562

## Description
1. Android Chrome user enters text **NOT from item list** into a combobox, then de-focus from the component.
2. With previous input text in place, user taps that combobox again for changing input. On focus, the component will auto selects the input text.
2. This auto input text selection triggers mobile browser's context menu for text selection (with options like copy, cut, select all, translate, etc.)
3. As this context menu popup interrupts the UI, user has to manually dismiss it before proceeding her/his action, thus poor UX.

This auto input text selection feature seems have been removed from v3. I'm unable to reproduce it in v3. This PR is only for v2.

Refer discussion #13562 for more background. Another dev reported this issue on iOS Safari.

## Markup:

```vue
<template>
  <v-container>
    <v-combobox></v-combobox>
  </v-container>
</template>

<script>
  export default {
    data: () => ({
    }),
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts
@@ -166,7 +166,6 @@ export default VSelect.extend({
     isFocused (val) {
       if (val) {
         document.addEventListener('copy', this.onCopy)
-        this.$refs.input && this.$refs.input.select()
       } else {
         document.removeEventListener('copy', this.onCopy)
         this.blur()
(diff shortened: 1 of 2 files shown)


KaelWD:
preventDefault on contextmenu seems to stop this on android

shiouming:
@KaelWD 
Indeed, this could be a workaround, though I only tested it on Android (I don't have iphone with me, hope someone from the team can test it out).
The side effect would be, unlike other input fields, users will be losing the ability to bring up context menu - probably not a concern as we're talking about combobox/autocomplete.

What is your take? Remove the auto text selection or leave the auto contextmenu popup as default behaviour?

By the way, do you know why in v3 this auto text selection behaviour is not happening? Was it intentionally removed or just pending for implementation?

KaelWD:
You could probably only prevent it if the field was just focused. 

> Was it intentionally removed 

Dunno, could have just been an oversight.

PR #18092, opened 2023-08-18
feat(VSwitch): add thumb slot

johnleider:
## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-switch
        v-model="model"
        label="Testing"
        true-icon="mdi-check"
        false-icon="mdi-close"
        inset
        color="primary"
      >
        <template #thumb="{ icon }">
          <v-scale-transition>
            <v-icon :key="icon" :color="model ? 'primary' : 'disabled'" size="12" />
          </v-scale-transition>
        </template>
      </v-switch>
    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref(false)
</script>

```
</details>

### DIFF
--- a/packages/vuetify/src/components/VSwitch/VSwitch.tsx
+++ b/packages/vuetify/src/components/VSwitch/VSwitch.tsx
@@ -3,6 +3,7 @@ import './VSwitch.sass'
 
 // Components
 import { VScaleTransition } from '@/components/transitions'
+import { VDefaultsProvider } from '@/components/VDefaultsProvider/VDefaultsProvider'
 import { VIcon } from '@/components/VIcon'
 import { makeVInputProps, VInput } from '@/components/VInput/VInput'
 import { VProgressCircular } from '@/components/VProgressCircular'
@@ -20,6 +21,7 @@ import { filterInputAttrs, genericComponent, getUid, propsFactory, useRender } f
 // Types
 import type { VInputSlots } from '@/components/VInput/VInput'
 import type { VSelectionControlSlots } from '@/components/VSelectionControl/VSelectionControl'
+import type { IconValue } from '@/composables/icons'
 import type { LoaderSlotProps } from '@/composables/loader'
 import type { GenericProps } from '@/util'
 
@@ -27,6 +29,9 @@ export type VSwitchSlots =
   & VInputSlots
   & VSelectionControlSlots
   & { loader: LoaderSlotProps }
+  & {
+    thumb: { icon?: IconValue }
+  }
 
 export const makeVSwitchProps = propsFactory({
   indeterminate: Boolean,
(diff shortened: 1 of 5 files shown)


packages/vuetify/src/components/VSwitch/VSwitch.tsx
```diff
@@ -20,13 +21,19 @@ import { filterInputAttrs, genericComponent, getUid, propsFactory, useRender } f
...
 
 export type VSwitchSlots =
   & VInputSlots
   & VSelectionControlSlots
   & { loader: LoaderSlotProps }
+  & {
+    'prepend-inner': never
+    'append-inner': never
+    thumb: { icon?: IconValue }
+  }
```
KaelWD, line 36: Add to the existing object

packages/vuetify/src/components/VSwitch/VSwitch.tsx
```diff
@@ -142,7 +149,19 @@ export const VSwitch = genericComponent<new <T>(
...
+                    >
+                      { slots['prepend-inner'] && (
+                        <div key="prepend" class="v-switch__prepend-inner">
+                          { slots['prepend-inner']() }
+                        </div>
+                      )}
+
+                      { slots['append-inner'] && (
+                        <div key="append" class="v-switch__append-inner">
+                          { slots['append-inner']() }
```
KaelWD, line 161: Should these be like track-true/track-false or something instead?

packages/vuetify/src/components/VSwitch/VSwitch.tsx
```diff
@@ -155,31 +174,44 @@ export const VSwitch = genericComponent<new <T>(
...
+                        { slots.thumb ? (
+                          <VDefaultsProvider
+                            defaults={{
+                              VIcon: {
+                                icon,
+                                size: 'x-small',
+                              },
+                            }}
+                          >
+                            { slots.thumb?.({ icon }) }
```
KaelWD, line 186: Should have `isValid` and `model`

packages/vuetify/src/components/VSwitch/VSwitch.tsx
```diff
@@ -20,13 +21,19 @@ import { filterInputAttrs, genericComponent, getUid, propsFactory, useRender } f
...
 import type { GenericProps } from '@/util'
 
 export type VSwitchSlots =
   & VInputSlots
   & VSelectionControlSlots
   & { loader: LoaderSlotProps }
+  & {
+    'prepend-inner': never
+    'append-inner': never
+    thumb: { icon?: IconValue }
```
KaelWD, line 35: icon is always present

Issue #18064, opened 2023-08-15
[Bug Report][3.3.12] Remove z-index for v-application


mkampmey:
### Environment
**Vuetify Version:** 3.3.12
**Vue Version:** 3.3.4
**Browsers:** Chrome 115.0.0.0
**OS:** Windows 10

### Steps to reproduce
Open an arbitrary vuetify app and check styles for the `v-application` div. 

### Expected Behavior
No z-index is set

### Actual Behavior
A z-index with value 1000 is set, which is not used since the div element has no position property. 
Coming from vuetify 2, this breaks my app because I have a position property set, but in vuetify 2 the `v-application` div has no z-index.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkEGOwjAMRa9Sed0kmmFWXXEPwgKIRwQat7JDEUK9O0koUlU2SCz9bf/n780dfOg7jirsen2SjqCBu6WqslNDLDRVUbI2XDDXFo4x9tIYc3CU1hy2fmBNGA31wazTmOELRR9QuS6sV3ql/4zzEueyRglqz91VkJOJhXqGif7/9gFqmiyEn98nYtJUu9tLZrx5mwQckBUjOWTkTyMt1uaxFq23aJk+WhqhhtbTWRa/Pkj58+Z15HeZi1u22mbiWEM5NKGfG7B9ANmnqgs=)

PR #18099, opened 2023-08-20
fix(VApp): remove layout styles

websitevirtuoso:
## Description

Issue - https://github.com/vuetifyjs/vuetify/issues/18064

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field v-model="msg" />
    </v-container>
  </v-app>
</template>

<script setup>
import { ref } from 'vue'

</script>
```

### DIFF
--- a/packages/vuetify/src/components/VApp/VApp.sass
+++ b/packages/vuetify/src/components/VApp/VApp.sass
@@ -11,6 +11,5 @@
   flex-direction: column
   flex: 1 1 auto
   max-width: 100%
-  min-height: 100vh
   min-height: 100dvh
   position: relative
--- a/packages/vuetify/src/components/VApp/VApp.sass
+++ b/packages/vuetify/src/components/VApp/VApp.sass
@@ -11,5 +11,6 @@
   flex-direction: column
   flex: 1 1 auto
   max-width: 100%
+  min-height: 100vh
   min-height: 100dvh
   position: relative
--- a/packages/vuetify/src/components/VApp/VApp.tsx
+++ b/packages/vuetify/src/components/VApp/VApp.tsx
@@ -23,7 +23,7 @@ export const VApp = genericComponent()({
 
   setup (props, { slots }) {
     const theme = provideTheme(props)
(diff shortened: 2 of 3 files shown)


packages/vuetify/src/components/VApp/VApp.sass
```diff
@@ -11,6 +11,5 @@
   flex-direction: column
   flex: 1 1 auto
   max-width: 100%
-  min-height: 100vh
```
KaelWD, line 14: This is for browsers without dvh

websitevirtuoso: Restore property

KaelWD:
Screenshot_20230831_021619

websitevirtuoso:
Ok I can implement and test example from johnleider. Will do soon

websitevirtuoso:
I tried to debug a little more and extra information: we really don't need z-index even in parentLayout

template

```
<template>
  <v-app>
    <v-layout class="rounded rounded-md">
      <v-app-bar title="Application bar" />

      <v-navigation-drawer>
        <v-list>
          <v-list-item title="Navigation drawer"/>
        </v-list>
      </v-navigation-drawer>

      <v-main
        class="d-flex"
        style="min-height: 300px"
      >
        <v-layout class="rounded rounded-md">
          <v-app-bar title="Application bar" />

          <v-navigation-drawer>
            <v-list>
              <v-list-item title="Navigation drawer"/>
            </v-list>
          </v-navigation-drawer>

          <v-main
            class="d-flex align-center justify-center"
            style="min-height: 300px"
          >
            <v-text-field v-model="msg" />
          </v-main>
        </v-layout>
      </v-main>
    </v-layout>
  </v-app>
</template>

<script setup>
import { ref } from 'vue'

</script>
```

ex: component v-layout has own z-index and proper position relative
image

In my own application, I tried to find an option where the z-index works and applies but didn't find it in multiple different layouts
image
I think this old code from v2 and not relevant anymore. If somebody will have issue I am able to implement it back or with parentLayout condition

johnleider (approved):

PR #18221, opened 2023-09-06
feat(IconSet): add material icons variants 

### LINKED ISSUE #16961: [Feature Request] Material Icons - Use "Rounded" style globally
### Problem to solve
Currently, it is very simple to integrate [Material Icons](https://fonts.google.com/icons) and use it with the "Filled" style.
But to use it with the "Rounded" style (same goes for "Outlined", "Sharp" and "Two tone"), a [class needs to be added](https://stackoverflow.com/a/65223943). It works for `VIcon`s, even if it is not ideal, but it doesn't for the "icon prop" (like `prepend-icon` on the [VTextField](https://vuetifyjs.com/en/components/text-fields/#icons)).

Our use case is that we want to use only the "Rounded" style in the whole app

### Proposed solution
Have a global configuration to chose the prefered md icon style.

Maybe create "by-style" icon sets for that:

```js
import { aliases, md-rounded } from 'vuetify/iconsets/md'

export default createVuetify({
  icons: {
    defaultSet: 'md-rounded',
    aliases,
    sets: {
      md-rounded,
    },
  },
})
```

I'm not familiar enough with the API to know if it is the best approach ?

PWie:
## Description
Material Icons come in various variants (outlined, rounded, etc.).
Currently there is not simple way to use them apart from creating custom iconSet.

This change introduces ready-to-use icon sets for each variant of Material Icons.
Those sets can also be used as default one.

Note that it also requires Material Icons font that includes multiple variants, like [material-icons](https://github.com/marella/material-icons#readme)

Resolves #16961

## Markup:
src/plugins/vuetify.js
```js
import 'material-icons/iconfont/material-icons.css'; // Ensure your project is capable of handling css files
import { aliases, md, mdOutlined } from 'vuetify/iconsets/md'

export default createVuetify({
  icons: {
    defaultSet: 'md',
    aliases,
    sets: {
      md,
      mdOutlined,
    },
  },
})
```

```html
<template>
  <v-icon icon="mdOutlined:home" />
</template>
```


J-Sek:
The downside is that website needs to load multiple font files - taxing the initial load.
Assuming it would be reworked for UnoCSS, the iconify collection is named "Google Material Icons" with prefix `i-ic:*` (e.g. `i-ic:outline-account-circle`). It would land too close to the `mdi` (which is better) and `ms` (that only lacks twotone variant). Overall not something users would really benefit from, while the headache when looking at the icons integration docs is potentially already an issue.

PR #18245, opened 2023-09-12
fix(VDatePicker): grid is occasionally shifting with some locales

bobodelben:
When I changed the locale of the VDatePicker component, which also altered the first day of the week, I noticed that there was a misalignment between the day numbers and the weekdays in certain cases. Upon further investigation, I found that this misalignment occurred because the firstDayWeekIndex or lastDayWeekIndex values were less than zero in those specific cases. To resolve this issue, I made some changes, and these adjustments fixed the problem for me.


johnleider:
What language are you using to test this?

bobodelben:
I am using italian locale.

Here is an example: 1st october should be sunday, but it falls under monday. On the other hand September is correct.



Not sure why otp component is failing test, I'm probably doing something wrong.

My playground:

```
<template>
  <v-app>
    <v-container>
      <v-locale-provider locale="it">
        <v-date-picker show-adjacent-months />
      </v-locale-provider>
    </v-container>
  </v-app>
</template>
```

johnleider (approved):

PR #18277, opened 2023-09-18
fix(VTabs): issue with :items parsing in v-tabs component

aauger:
The v-tab `:items` prop parsing fails to correctly shape the `TabItem` object for use in the tabs themselves, leaving the tab titles blank. This fixes that by falling back on `props.title` if `props.text` is not present.

## Description

The v-tabs component has an `items` prop which is meant to simplify the tab value/text options, but it does not work correctly because it uses `title` when the tab component itself looks for `text`. This simply adds `title` as an additional fallback which makes `items` work in the expected fashion.

## Markup:

```vue
    <v-tabs
      :items="['1', '2', '3']"
      v-model="tab"
      bg-color="primary"
    >
    </v-tabs>
```

### DIFF
--- a/packages/vuetify/src/components/VTabs/VTab.tsx
+++ b/packages/vuetify/src/components/VTabs/VTab.tsx
@@ -122,7 +122,7 @@ export const VTab = genericComponent()({
           maxWidth={ props.fixed ? 300 : undefined }
           onGroup:selected={ updateSlider }
         >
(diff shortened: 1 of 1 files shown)


johnleider:
Closing in favor of ac900991f

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

aauger:
That works as well. Thanks, John.

PR #18303, opened 2023-09-21
docs: adds button to edit prop/slots/etc description even if not missing

jacekkarczmarczyk:
image

### DIFF
--- a/packages/api-generator/src/utils.ts
+++ b/packages/api-generator/src/utils.ts
@@ -135,15 +135,15 @@ async function getSources (name: string, locale: string, sources: string[]) {
 
   return {
     find: (section: string, key: string, ogSource = name) => {
+      const githubUrl = `https://github.com/vuetifyjs/vuetify/tree/${currentBranch}/packages/api-generator/src/locale/${locale}/${ogSource}.json`
       for (let i = 0; i < arr.length; i++) {
         const source = arr[i] as any
         const found: string | undefined = source?.[section]?.[key]
         if (found) {
-          return { text: found, source: sourcesMap[i] }
+          return { text: found, source: sourcesMap[i], githubUrl }
         }
       }
-      const githubUrl = `https://github.com/vuetifyjs/vuetify/tree/${currentBranch}/packages/api-generator/src/locale/${locale}/${ogSource}.json`
-      return { text: `MISSING DESCRIPTION ([edit in github](${githubUrl}))`, source: name }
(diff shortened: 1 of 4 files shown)


packages/api-generator/src/utils.ts
```diff
@@ -133,15 +133,15 @@ async function getSources (name: string, locale: string, sources: string[]) {
 
   return {
     find: (section: string, key: string, ogSource = name) => {
+      const githubUrl = `https://github.com/vuetifyjs/vuetify/tree/${currentBranch}/packages/api-generator/src/locale/${locale}/${ogSource}.json`
```
MajesticPotatoe, line 136: This is broken for things that dont have a normal local ogSource 
eg: useDisplay ogSource is `@vue/reactivity/dist/reactivity.d.ts#L451-L459` instead of `useDisplay` or VDataTable slots that point to sources like `vuetify/src/labs/VDataTable/VDataTableHeaders.tsx#L38-L47`

packages/api-generator/src/utils.ts
```diff
@@ -133,15 +133,15 @@ async function getSources (name: string, locale: string, sources: string[]) {
...
         const source = arr[i]
         const found: string | undefined = source?.[section]?.[key]
         if (found) {
-          return { text: found, source: sourcesMap[i] }
+          return { text: found, source: sourcesMap[i], githubUrl }
         }
       }
-      const githubUrl = `https://github.com/vuetifyjs/vuetify/tree/${currentBranch}/packages/api-generator/src/locale/${locale}/${ogSource}.json`
-      return { text: `MISSING DESCRIPTION ([edit in github](${githubUrl}))`, source: name }
+      return { text: `MISSING DESCRIPTION ([edit in github](${githubUrl}))`, source: name, githubUrl }
```
johnleider, line 144:
```suggestion
      return { text: `Description Pending ([edit in github](${githubUrl}))`, source: name, githubUrl }
```

johnleider (commented):
I do want to change the button styling but it should only show up for descriptions that are missing.

jacekkarczmarczyk:
> I do want to change the button styling but it should only show up for descriptions that are missing.

My thought was that it should be easy to fix the existing description as well, not just add a missing one

johnleider:
It's visually draining

MajesticPotatoe:
> It's visually draining

what if it was only shown on hover?

PR #18316, opened 2023-09-22
fix(VDataTable): render default slot if only comments are between tags

### LINKED ISSUE #18308: [Bug Report][3.3.17] VDataTable does not render correctly if there is a single comment between the tags
### Environment
**Vuetify Version:** 3.3.17
**Vue Version:** 3.3.4
**Browsers:** Firefox 117.0
**OS:** Linux

### Steps to reproduce
Please see the reproduction link.

### Expected Behavior
Adding a single comment between `<v-data-table>` tags should not prevent the headers and rows from rendering.

### Actual Behavior
Header and rows don't render, only the footer shows up.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpNj8EOgjAMhl+l9uKFQYwmJgRNfA7xMKHI4mTLVoxKeHerEvXU9u/X9u9+wBiqbOd9eu0JcyyYLt5qpm3ZARRXVWvWivXR0ksAyFvSNYW4KXE/ABu2lMO8cW6ewJnuUw7jocRpwMjKCZeWAEcdBEjgVz9+A++7cnmmFHBrIsTW9baGzjHopqGKRSYI1IkL051AqY/V7N+rSEX294qUsQrG85ulm3eBoaZG95ZhGF/0t49jgst0ma7wExdrPDwBGjpevw==)

gitemad:
## Description

Fix the issue by writing a new helper function called isVNodesEmpty.
This function returns true if there is no VNode or all VNodes type is comment.

fixes #18308 
## Markup:

```vue
<template>
  <v-app>
      <v-data-table
        :headers="[{ title: 'foo', key: 'foo' }]"
        :items="[{ foo: 'bar' }, { foo: 'baz' }]"
      >
        <div>
          this should affect the rendering
        </div>
      </v-data-table>
      <v-data-table
        :headers="[{ title: 'foo', key: 'foo' }]"
        :items="[{ foo: 'bar' }, { foo: 'baz' }]"
      >
        <!-- this should not affect the rendering -->
        this should affect the rendering
      </v-data-table>
      <v-data-table
        :headers="[{ title: 'foo', key: 'foo' }]"
        :items="[{ foo: 'bar' }, { foo: 'baz' }]"
      >
        <!-- this should not affect the rendering -->
      </v-data-table>
      <v-data-table
        :headers="[{ title: 'foo', key: 'foo' }]"
        :items="[{ foo: 'bar' }, { foo: 'baz' }]"
      >
      </v-data-table>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/labs/VDataTable/VDataTable.tsx
+++ b/packages/vuetify/src/labs/VDataTable/VDataTable.tsx
@@ -22,6 +22,7 @@ import { makeFilterProps, useFilter } from '@/composables/filter'
 // Utilities
 import { computed, toRef } from 'vue'
 import { genericComponent, propsFactory, useRender } from '@/util'
+import { isVNodesEmpty } from '@/util/helpers'
 
 // Types
 import type { UnwrapRef } from 'vue'
@@ -203,7 +204,7 @@ export const VDataTable = genericComponent<VDataTableSlots>()({
         >
           {{
             top: () => slots.top?.(slotProps.value),
-            default: () => slots.default ? slots.default(slotProps.value) : (
+            default: () => !isVNodesEmpty(slots.default?.(slotProps.value)) ? slots.default?.(slotProps.value) : (
               <>
                 { slots.colgroup?.(slotProps.value) }
                 <thead>
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -1,5 +1,5 @@
 // Utilities
-import { camelize, capitalize, computed, Fragment, reactive, toRefs, watchEffect } from 'vue'
+import { camelize, capitalize, Comment, computed, Fragment, reactive, toRefs, watchEffect } from 'vue'
 import { IN_BROWSER } from '@/util/globals'
 
 // Types
@@ -699,4 +699,8 @@ export function matchesSelector (el: Element | undefined, selector: string): boo
   } catch (err) {
     return null
   }
+}
+
+export function isVNodesEmpty (vnodes: VNode[] | undefined): boolean {
+  return !vnodes || wrapInArray(vnodes).every(vnode => vnode.type === Comment)
 }


KaelWD:
I don't think this change is necessary, there's lots of slots that replace the default rendering if you put anything in them.

simon-lorenz:
For me (as the issue author) this behaviour was a bit unexpected. From a user perspective I think that a comment should never affect the execution of a program. But I understand that there might be reasons for this and that you don't want to make a framework wide change for a minor issue.

Thanks for looking into it though!

johnleider:
Can we test performance before / after?

KaelWD:
Hmm looks like vue already does this with template slots https://play.vuejs.org/#eNp9Ustu2zAQ/BVal7RARAJ1gQKGaqQtcmiBJkGSIy+ytHYUiw+Qy8SBoH/PknIU2XmcJO7M7gx32GW/rOUPAbJFVvjKNRaZBwx2KXWjrHHI/j/9MfSnQSNbO6PYCReTWmw+kboQQzf10QFB2bZEiCcsJuxlMctzVhml4rg8XxbTWUQvxNibnWaDhVyVlt97o8lkJzVjcg94mS1YqsQaGYlnmd0hWr8QImi73XASE2eECRc0Ngry2qizOf/Gv/8QdeNxWufgVb5y5tGDI0WZnU6GCyo+gMsd6BocuE/FjrgHgkfYG9Go2Uvd0wKOFv1OSmwfU8eCh5vWoGf9PqchGakroz01JOznSPvy9ZPUGCsif7ku23ZVVlsixmMCVi59u47NZmkor2FdhhZZT64PA0RP4utmcxQfrck2LbhLiw2ZO4iRFM3jv1RDF2BMoLqDavtO/d7vhiSuHKTNTlLD0m0AB/j85gJ29D+CytSh3T+ZD8Br8KYN0eNA+x10TbYnvOT2b4qg0Ztbf75D0P7lUtFoSjPx0yuKeX509Ve7cz4fX0H/DM6KPsU=

KaelWD:
I did have an idea to force us to make the default rendering possible with slots: #15611

Looked something like
```js
slots.default({ foo, bar, baz }, slotProps => { ...fallback... })
```

This pattern would also allow comment checks everywhere without all those ugly extra function calls. It did require re-implementing the DefineComponent interface though so maybe a helper function would be better?

```js
renderSlot(slots.default, { foo, bar, baz }, slotProps => { ...fallback... })
```

KaelWD:
- This doesn't go far enough, it really applies to every slot in the framework not just this specific one
- I'd go with some less messy syntax as mentioned above

PR #18332, opened 2023-09-25
feat(validationRules): add useRules composable

vcastro45:
## Description

This PR add, to labs for the moment, a new composable `useRules()` that add a range of common ValidationRules to be used within any field component.

The motivation is first born from a personal problem that is: I always use the same rules and I always copy/paste them from the previous project to the new one.

This is my first PR to vuetify so, I tried to respect the patterns already in place (hope I've done it good). The main "problem" I've faced is when I use the `useLocale()` composable: I can't get a global `t()` "instance" because `vuetify:locale` is not yet provided when `createRules()` is called.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field label="foo" :rules="[rules.required()]"></v-text-field>
    </v-container>
  </v-app>
</template>

<script setup>
  import { useRules } from '@/labs/rules/rules'

  const rules = useRules()
</script>
```


packages/vuetify/src/labs/rules/rules.ts
```diff
@@ -0,0 +1,108 @@
...
+  }, options)
+}
+
+export function createRules (options?: RulesOptions) {
+  // @ts-expect-error
+  const _options = parseRulesOptions(options)
+
+  const state: RulesInstance = {
+    required: (err?: string) => {
+      const { t } = useLocale()
```
johnleider, line 46: You should only need to instantiate this once

vcastro45: Are you talking about `t` ?

johnleider: Yes

vcastro45: The `locale` dependency being provided after the call of `createRules()` in `framework.ts`, it's not possible to get and share the same `t` instance using `useLocale()` by calling it at the beginning of the `createRules()`;

> [Vuetify] Could not find injected locale instance

Once this is said: Is it really a problem to get `t` in each rule function ? `t` is provided using the Vue's provide/inject (aka dependency injection pattern) system so, the same instance is re-used every time.

Do you really care about it ?

johnleider: Like I said, you *should* only need to. If it's causing an issue otherwise, maybe it's something we need to look into in addition to this change.

vcastro45: any review about last changes ?

packages/vuetify/src/labs/rules/rules.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export type RulesOptions = Partial<InternalRulesOptions>
+
+const parseRulesOptions = (options?: RulesOptions) => {
+  return mergeDeep({
+  }, options)
+}
+
+export function createRules (locale: LocaleInstance, options?: RulesOptions) {
```
johnleider, line 38:
```suggestion
export function createRules (options?: RulesOptions, locale: LocaleInstance) {
```

packages/vuetify/src/framework.ts
```diff
@@ -53,6 +56,7 @@ export function createVuetify (vuetify: VuetifyOptions = {}) {
   const icons = createIcons(options.icons)
   const locale = createLocale(options.locale)
   const date = createDate(options.date)
+  const rules = createRules(locale, options.rules)
```
johnleider, line 59:
```suggestion
  const rules = createRules(options.rules, locale)
```

johnleider (changes requested):
Looking good, a few points of feedback for now.

packages/vuetify/src/labs/rules/rules.ts
vcastro45: Looking good to you ?

johnleider:
After looking at the icon's composable today, I think we could probably glean some good practices in terms of how it defines interfaces and overwriting aliases: https://github.com/vuetifyjs/vuetify/blob/faab3326272d54ed9535d53fe0099e72372f03e8/packages/vuetify/src/composables/icons.tsx

vcastro45:
Here it is :)

vcastro45:
Any news about it ?

johnleider:
I haven't forgot about you, really busy :)

Vincere1st:
Hello, do you have a release date for this interesting feature?
Validation being essential in forms, I think this pull request should be reviewed and merge.

johnleider (approved):

laventnc:
not sure if this is the place to put this, but the typing of the composable could use some work.

```ts
const rules = useRules() // Partial<RuleAliases> | undefined
rules.required() // ValidationRuleBuilderWithoutOptions | undefined
```

in order to avoid ts issues:
```ts
const rules = useRules()
rules!.required!()
```

PR #18365, opened 2023-09-28
docs(VCard): examples

peluprvi:
## Markup:

```vue

```


packages/docs/src/pages/en/components/cards.md
```diff
@@ -110,6 +110,52 @@ Cards can be colored by using any of the builtin colors and contextual names usi
...
 <example file="v-card/prop-color" />
 
+#### Elevation
+
+The **elevation** property provides up to 24 levels of shadow depth. By default, cards rest at 2dp.
+
+<example file="v-card/prop-elevation" />
+
+#### Hover
+
```
MajesticPotatoe, line 120: Could use a description
(side note, `hover` prop seems to be bugged)

MajesticPotatoe (changes requested):

peluprvi:
@MajesticPotatoe 

VCard probable issues:
- new hover: `<v-card hover />` doesn't apply elevation
- twitter card example: `prepend-icon` with `<template v-slot:prepend><v-icon size="x-large"></v-icon></template>` doesn't work like `v-button`

packages/docs/src/pages/en/components/cards.md
peluprvi: Exactly. I thought about describing it as on the `v-card` API or as on the `v-button` docs. But I don't know how the final behavior is supposed to be when fixed. I guess it should start with some elevation and adds 2?

MajesticPotatoe: describing it similar to api is fine
However looking at the sass vars on this, it looks like base elevation is actually elevation=1, and hover is to apply elevation=8

MajesticPotatoe:
> @MajesticPotatoe
> 
> VCard probable issues:
> 
> * new hover: `<v-card hover />` doesn't apply elevation
> * twitter card example: `prepend-icon` with `<template v-slot:prepend><v-icon size="x-large"></v-icon></template>` doesn't work like `v-button`

looks like i bug tbh. according to the sass we apply elevations(0) and then hover is suppose to apply elevations(1) before and elevations(8), but no hover effect takes place.

on the twitter card example, i wouldnt expect it to work like a v-btn, because its a v-icon, if you want it to be like v-btn, make it into a v-btn, but its also missing the actual icon, so as it stands, its just an empty div (need `icon="mdi-twitter"`)
personally dont think it needs to be a button (if anything the ones in the append slot could be turned into buttons, but not sure that matters)

MajesticPotatoe:
tentative fix for the hover bug #18381

MajesticPotatoe:
pulse check

MajesticPotatoe (approved):

PR #18392, opened 2023-10-03
fix(VVirtualScroll): use a dynamic window size, cache offsets

### LINKED ISSUE #17801: [Bug Report][3.3.8] v-virtual-scroll broken

### LINKED ISSUE #18198: [Bug Report][3.3.15] v-virtual-scroll fails with items larger than viewport

KaelWD:
## Description

Help I rewrote the entire thing. 

- Varies the number of visible items based on observed item height
- Items that have not been rendered before are assumed to be the height of the smallest item that has
	- This unfortunately causes the scrollbar to resize constantly with variable height items
	- It also means over-rendering in scenarios with vastly different item sizes, but the alternatives are worse:
	- Max (master) results in under-rendering instead, small items at the top or bottom are removed too soon. It also has the opposite scrollbar problem (constantly shrinking instead of growing)
	- Average (#18392) also results in under-rendering but not as much. It also has a very stable scrollbar if item height is evenly distributed. `props.itemHeight ` is a bit of a footgun with this if it is not set to the average initially, as it will take scrolling most of the way down to converge near the actual value
	- Ideally we'd use min for rendering items and average for setting scroll height
- ~~Adds an extra buffer based on scroll velocity per frame~~
	- That actually made it slower (rendering more items). ~~Changed to offset the first element by the amount scrolled on the previous frame instead, which stops the view from going blank when scrolling faster than the browser can keep up with.~~
	- That only helped with renderless and would sometimes cause the view to go blank until scrollend, removed the entire thing instead.
- Calculates item offsets once instead of every time we want to find the offset of an item
	- Offsets are invalidated when an item height is updated (debounced)

This is seriously fast, it renders a list of 1M items better than master does with 100k. itemHeight is even more redundant than it was before, now only used for the first render.

fixes #18198
fixes #17801

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-virtual-scroll :items="items" height="400" style="margin-top: 100px; background-color: red">
        <template #default="{ item, index }">
          <div
            :class="[
              index % 2 === 0 ? 'bg-grey-lighten-2' : index % 5 === 0 ? 'bg-grey-darken-2' : 'bg-surface',
              'px-2'
            ]"
            :style="{height: `${item.height}px`}"
          >
            Dynamic item {{ item.index }} - height {{ item.height }}px
          </div>
        </template>
      </v-virtual-scroll>

      <v-list style="height: 400px; overflow-y: auto">
        <v-virtual-scroll v-slot="{ item }" height="100%" :items="selectItems">
          <v-list-item v-bind="item.props">
            Tiny item {{ item.value }}
          </v-list-item>
        </v-virtual-scroll>
      </v-list>

      <v-list style="height: 400px; overflow-y: auto">
        <v-virtual-scroll v-slot="{ item, index, itemRef }" renderless height="400" :items="selectItems">
          <v-list-item :ref="itemRef" :key="index" v-bind="item.props">
            Renderless item {{ item.value }}
          </v-list-item>
        </v-virtual-scroll>
      </v-list>

      <v-select v-model="selected" :items="selectItems" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    computed: {
      items () {
        return Array.from({ length: 10000 }, (k, v) => ({
          index: v,
          height: this.randomIntFromInterval(),
        }))
      },
    },
    methods: {
      randomIntFromInterval () {
        // min and max included
        return Math.floor(Math.random() * (600 - 15 + 1) + 15)
      },
    },
  }
</script>

<script setup>
  const selectItems = Array.from({ length: 10000 }, (k, v) => ({
    title: `Item ${v}`,
    value: v,
    props: { height: 24, minHeight: 24 },
  }))
</script>

<style>
.v-list {
  background: red !important;
  padding: 0 !important;
}
.v-list-item {
  background: white !important;
}
</style>
```


yuwu9145:
When item is beyond container, buffer should be only 1 item but redundant extra buffer items are rendered

I think the cy test case removed in d3b4d29e170acb7cc5d432332683eeea79c2e228 should be retained

Screenshot 2023-10-07 at 3 57 19 pm


```vue
<script>
  export default {
    data: () => ({
      // total of 10 - 20k images in 1k sections
      imageURLsBySection: {
        'Section 0': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx}`
        ),
        'Section 1': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 30}`
        ),
        'Section 2': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 60}`
        ),
        'Section 3': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 90}`
        ),
        'Section 4': [...Array(30).keys()].map(
          idx => `https://picsum.photos/200?image=${idx + 120}`
        ),
      },
    }),
  }
</script>

<template>
  <v-virtual-scroll
    :items="Object.keys(imageURLsBySection)"
    height="500"
    item-height="1000"
  >
    <template v-slot:default="{ item, index }">
      <div
        :style="{height: '1000px',backgroundColor: index % 2 == 0 ? 'red' : 'green'}"
      >
        <span class="text-h2">{{ item }}</span>
      </div>
    </template>
  </v-virtual-scroll>
</template>
```

KaelWD:
That's due to this check which reduces the number of DOM updates 
https://github.com/vuetifyjs/vuetify/blob/2cc5df644a93e160274c401c34662d944855ad02/packages/vuetify/src/composables/virtual.ts#L174-L177

If you're scrolling down it will only remove items from the top when it's adding items to the bottom, as a side effect of that once the last item is reached it will stop updating the top too.

yuwu9145:
For the senario of "item beyond container", fixed `BUFFER_PX` wouldn't help to render minimum items.

People could throw a lot of logics into each item, e.g. load 100 images in each section, and it's executing these logics in invisible items.

Probably this issue is unobvious for now, but I think in case we want to optimise it, this senario should be specifically handled with 1 item buffer instead of fixed pixel based buffer

KaelWD:
The buffer is more of a page size, it updates 100px worth of items at a time.

yuwu9145:
playground in #18198 doesn't work properly

Understand it renders 100px worth of items, but still, it doesn't render minimum amount of visible items when item is beyond container (which is very predictable, should be only 2).

I am not pushing it, just mention it and see if this was something worth to tackle for now.

KaelWD:
Ok `getSize(start) + viewportHeight.value` should probably just be `viewportHeight.value`, we don't need the part of the item that is overflowing. That seems to still work with smaller items and/or a 0px buffer too. 

> playground in https://github.com/vuetifyjs/vuetify/issues/18198 doesn't work properly

Removing `getSize(start)` reduces it to 4 from 5. This is part of the optimisation I mentioned above though, when the items first render they're only a couple hundred px high and aren't removed once they grow to 1300 because no new ones are added. Again this would be much more acceptable behaviour with tens or hundreds of items instead of just 5.

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
```diff
@@ -73,10 +75,22 @@ export const VVirtualScroll = genericComponent<new <T, Renderless extends boolea
...
       onMounted(() => {
         containerRef.value = getScrollParent(vm.vnode.el as HTMLElement, true)
-        containerRef.value?.addEventListener('scroll', handleScroll)
+        if (containerRef.value === document.documentElement) {
+          document.addEventListener('scroll', handleScroll, { passive: true })
+          document.addEventListener('scrollend', handleScrollend)
+        } else {
+          containerRef.value?.addEventListener('scroll', handleScroll, { passive: true })
+          containerRef.value?.addEventListener('scrollend', handleScrollend)
+        }
```
nekosaur, line 84: What's the difference here? If `containerRef.value` equals `document.documentElement`, can't you just do `containerRef.value?.addEventListener` then as well?

nekosaur (commented):
Looks good. The scrollbar does bounce a lot. That's the thing I liked most about my averaging change, that it was relatively stable.

The only thing I could find was an issue when setting v-select v-model to a large value so that it does scrollToIndex when menu open. The menu is scrolled correctly, but if you then scroll down once with mousewheel, it will teleport waay down the list. E.g. when I set v-model to ~5000, scrolling down once took me to ~9000.

I also think we can take the opportunity to better document how this thing works, with either some more inline or method comments. As an example, how scrolling on initial load works, and the pixel buffer.

packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx
KaelWD: Stolen from #17813: https://github.com/vuetifyjs/vuetify/pull/17813/files/d871cda1ae7b935f50597822b4656c8252e82d7e#r1271301766

documentElement has `scrollHeight > clientHeight` and `scrollTop` that reflects page scroll position, but it doesn't emit a scroll event.

KaelWD:
Yeah I wonder if I can make it use average for total scroll height but min for assumed offsets. One option would be to render new items one at a time so we never have to assume their height, that'd probably be pretty slow though.

johnleider (approved):
Looks really good. Only complaint is the scroll bar can feel a little jarring if you're manually dragging it.

PR #18401, opened 2023-10-05
feat(VInfiniteCarousel): add new component

Bossslime:
## Description
I am adding a new component that I think could be very useful. Don't confuse this with the infinite scroller because it is not. This component takes whatever content is inside of it and makes a slowly rotating infinite carousel with it. This is a really cool way to show info without the user having to scroll. The difference between this and the other carousel is that this will scroll through one entire div instead of scrolling between multiple divs. That means that you can have a grid of objects in one very long div and it will slowly scroll through them. Its just a different way to display data is all.

## Markup:
```Vue
<template>
  <!-- The props are optional, these values are the default values -->
  <v-infinite-carousel :reverse="false" :speed="10" :step="1" :pause="false">
    <!-- Content to scroll through, can be anything-->
  </v-infinite-carousel>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```

The props are as follows:
  - reverse   - Reverses the scroll direction, by default it scrolls to the right
  - speed     - How often the scroll steps in a direction
  - step        - How far the scroll steps in a direction (I recommend keeping this low as it looks laggy when set too high)
  - pause     - Whether the scrolling is paused or not


johnleider:
If this P.R. is going to the `master` branch, it needs to be put into Labs. If not, it should go to the `dev` branch.

J-Sek:
moving scroll in interval jitters. It needs a reboot

Issue #11243, opened 2020-04-27
[Bug Report] v-text-field v-model not updated immediately on android chrome


WIStudent:
### Environment
**Vuetify Version:** 2.2.25
**Last working version:** 2.2.22
**Vue Version:** 2.6.11
**Browsers:** Google Chrome
**OS:** Android

### Steps to reproduce
1. Open linked codepen in Android Chrome
2. Enter a word into the textfield

### Expected Behavior
The counter should increase with each character, the next button should be enabled as soon as the textfield contains at least one character.

### Actual Behavior
The counter is only updated after the autocomplete composition is finished by chrome. Likewise the v-model isn't updated immediately, causing the next button not to enable until the user either enters a blank character or closes the keyboard.

![demo](https://user-images.githubusercontent.com/2707930/80378775-77d65680-889d-11ea-9218-a32511db6156.gif)

### Reproduction Link
<a href="https://codepen.io/wistudent/pen/pojPmVw" target="_blank">https://codepen.io/wistudent/pen/pojPmVw</a>

### Other comments
This was probably introduced by the changes made in 2.2.23 and 2.2.24 concering the IME string composition:
> VTextField: skip value property update if IME composing string (#11134) (c02be82), closes #5748
> VTextField: update value on compositionend (#11194) (f51bef6), closes #11190 #5748

If this is the new intented behaviour a new property to force the immediate update would be helpful.

neenus:
I would love to see a fix for this as well

kouqon:
What I want to do is match VTextField behavior to HTMLInputElement on Vue.

https://vuejs.org/v2/guide/forms.html#Basic-Usage
> For languages that require an IME (Chinese, Japanese, Korean, etc.), you’ll notice that v-model doesn’t get updated during IME composition. If you want to cater to these updates as well, use the input event instead.

[On this patch](https://github.com/kouqon/vuetify/commit/aeb24db2862ff3c697f786dad9b3f7b710b179c7) VTextField behave as the following:
* @input event emit on every character input
* v-model and value property update if not on composing with IME
* @change event emit on blur or input enter key (not on composing)
* counter update on every character input

I'm starting to wonder if this behavior is what everyone want.
Is it better to revert to pre-v2.2.22 behavior?

WIStudent:
Interesting, I tried a new demo with HTMLInputElement and see the same issue. 
https://codepen.io/wistudent/pen/JjYyOdm

[...]

PR #18423, opened 2023-10-09
fix(VCombobox): don't trigger keydown listener during composition session

### LINKED ISSUE #18421: [Bug Report][3.3.20] Unintended Japanese input with v-combobox behavior
### Environment
**Vuetify Version:** 3.3.20
**Vue Version:** 3.3.4
**Browsers:** Chrome 117.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
1. enter text in v-combobox (e.g., "うれしい)
2. convert input into Kanji characters with space key (e.g. "嬉しい")
3. confirm with the Enter key, the same string is entered at the same time as it is registered (e.g. "嬉しい 嬉しい")

### Expected Behavior
```
嬉しい
```
Also, this is not confirmed, but is confirmed when you press Enter again

### Actual Behavior
```
嬉しい 嬉しい
```

### Reproduction Link
[<a href="https://play.vuetifyjs.com/#..." target="_blank">https://play.vuetifyjs.com/#...</a><br/>](<a href="https://play.vuetifyjs.com/#eNpNjs0OgjAMx1+l7oImCkY8mUnizTfwoB4Ah1mysWYrxITw7m4MI6e2v/w/eh+Ys3V2QUz7TrAT4yQ0qpJE8WgBeL8rEad1OmrTUilbYWc0Q12ZynxAd4okKlHw7E9/5gkt3YHEcJ4tSv3paiuRwAnqYrfUaCzBAFY0MEJjjYbE/5sEOYDPdQTaveEcFOvkKpQycDNWvVbJJhTESB/Gxi3L0zw9sjgPe/b8Aq0gVWU=)" target="_blank">https://play.vuetifyjs.com/#eNpNjs0OgjAMx1+l7oImCkY8mUnizTfwoB4Ah1mysWYrxITw7m4MI6e2v/w/eh+Ys3V2QUz7TrAT4yQ0qpJE8WgBeL8rEad1OmrTUilbYWc0Q12ZynxAd4okKlHw7E9/5gkt3YHEcJ4tSv3paiuRwAnqYrfUaCzBAFY0MEJjjYbE/5sEOYDPdQTaveEcFOvkKpQycDNWvVbJJhTESB/Gxi3L0zw9sjgPe/b8Aq0gVWU=)</a><br/>

### Other comments
I have observed the intended behavior on Windows and v2, but abnormal behavior only on mac in v3.
|vuetify|windows|mac|
|:-:|:-:|:-:|
|v2|ok|ok|
|v3|ok|ng|

yuwu9145:
fixes #18421

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-combobox
        chips
        multiple
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

</script>
```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -238,7 +238,7 @@ export const VCombobox = genericComponent<new <
       menu.value = !menu.value
     }
     function onKeydown (e: KeyboardEvent) {
-      if (props.readonly || form?.isReadonly.value) return
+      if (e.isComposing || props.readonly || form?.isReadonly.value) return
 
       const selectionStart = vTextFieldRef.value.selectionStart
       const length = model.value.length
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -26,7 +26,7 @@ import { makeTransitionProps } from '@/composables/transition'
 
 // Utilities
 import { computed, mergeProps, nextTick, ref, shallowRef, watch } from 'vue'
-import { genericComponent, IN_BROWSER, noop, omit, propsFactory, useRender, wrapInArray } from '@/util'
+import { genericComponent, IN_BROWSER, isComposingIgnoreKey, noop, omit, propsFactory, useRender, wrapInArray } from '@/util'
 
 // Types
 import type { PropType } from 'vue'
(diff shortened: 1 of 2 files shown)


KaelWD (approved):
I can't check iOS but it still works on android and doesn't break english so :+1:

johnleider:
The last time we did something with isComposing in the framework we ended up reverting it:

https://github.com/vuetifyjs/vuetify/commit/5f82fe23fe2fafe2fdb24f715346f98f17bb4639

Just pasting to make sure we're not repeating the mistake.

KaelWD:
Yeah I checked that, it still updates as you type. We actually had the same problem in v3 at one point: #11243

yuwu9145:
I think the proper fix is having an universal helper that checks whether a press key is a composition operation, so we can prevent composition operations conflicting with Vuetify logic

V2 uses `keyCode` which always returns 229 during composition sessions, so #18421 doesn't exist in v2. But it has been deprecated not being used in v3

johnleider (approved):

PR #18444, opened 2023-10-11
feat(VAutoComplete): add clearSearchOnSelection prop

### LINKED ISSUE #18428: [Feature Request] Add option to clear typed query when selecting an autocomplete item
### Problem to solve
Adding multiple items to an autocomplete via the keyboard is quite a painful user experience as the previous 'query' remains appended after the value is added (until the field is blurred).

Many users will not know to hit ESC to close the dropdown and refocus the input, and even if they do so they currently have to manually delete their previous search term.

### Proposed solution

Some form of "clear query on selection" prop that adds the selected item, removes the previously typed query, and focuses the input to automatically accept a new value.

Bonus points for an option which shows the dropdown on focus.

Nice one!

gitemad:
## Description

Add clearSearchOnSelection prop to VAutoComplete component.
Typed search query cleared after selecting an item when this prop is true.

resolves #18428 
## Markup:

```vue
<template>
  <v-autocomplete
    label="Autocomplete"
    clearSearchOnSelection
  />
</template>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VAutocomplete.json
+++ b/packages/api-generator/src/locale/en/VAutocomplete.json
@@ -1,6 +1,7 @@
 {
   "props": {
     "autoSelectFirst": "When searching, will always highlight the first option and select it on blur. `exact` will only highlight and select exact matches.",
+    "clearSearchOnSelection": "Clears search value after selecting an item.",
     "filter": "The filtering algorithm used when searching. [example](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L40).",
     "noFilter": "Do not apply filtering when searching. Useful when data is being filtered server side."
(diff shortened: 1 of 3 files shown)


hanzelkatomas:
This is exactly what I am looking for. Currently when using Autocomplete it's not very intuitive. This prop would mean Autocomplete would behave like MUI Autocomplete. I even think this prop should be `true` by default.
Do you plan to merge this PR?

amadeous:
I would also be interested in having this new feature added to this selector. As @hanzelkatomas, I do think that this prop should be `true` by default, the user experience would feel more natural.
Could someone please review this PR and merge it? :)

actionjacks:
I hope for a quick review, long awaited

captainlettuce:
This seems to exist with the prop `clearOnSelect`? (Vuetify 3.5.13)

PR #18449, opened 2023-10-12
fix: global const defaultThemeOptions moved inline

### LINKED ISSUE #18393: [Bug Report][3.3.20] Memory leak in Nuxt SSR
### Environment
**Vuetify Version:** 3.3.20
**Vue Version:** 3.3.4
**OS:** Mac OS 10.15.7

### Steps to reproduce
- yarn install
- yarn build 
- node --inspect .output/server/index.mjs
- Open localhost:3000
- Go to the Node.js devtools in Chrome (green icon in the top corner)
- Go to Memory and select Heapsnapshot (first option)
- Create a Heapsnapshot
- Collect garbage manually (trash bin icon)
- Simulate traffic: I use Apache Benchmark `ab -n 1000 -c 100 http://localhost:3000/`
- Create a Heapsnapshot
- Collect garbage manually (trash bin icon)
- Simulate traffic: `ab -n 1000 -c 100 http://localhost:3000/`
- Repeat the steps a couple of times, and you'll see that the memory goes up

Currently running Node v18.18.0, but the version doesn't seem to make a difference.

The increase in memory in the application happens a lot faster, from 30 -> 300 in about 200 users, en keeps rising. My best guess is because we have a lot more components then in the reproduction. 

### Expected Behavior
The memory should keep consistant when there is no load on the server, and the garbage should be able to be collected.

### Actual Behavior
After some debugging i can see that there are a lot of computedRefs created, with CSS theme options and kept in memory.
When we remove vuetify, the memory leaks also dissapear, so we are certain that the memory leak is in vuetify when used in SSR mode for Nuxt.

### Additional information
See below for more information on memory profiling.

image
image
image
image
image

### Reproduction Link
[https://github.com/cultuurnet/uiv-nuxt-memoryleak-reproduction](https://github.com/cultuurnet/uiv-nuxt-memoryleak-reproduction)

Any help is welcome, as this is blocking our migration to Nuxt 3. Also open to contribute if someone can point me in the right direction in the vuetify code.

memic84:
fixes #18393

## Description
In Nuxt SSR memory leaks were introduces because the defaultThemeOptions was in global scope. 

Was done:
- defaultThemeOptions and parseThemeOptions are moved inside createTheme.

## Markup:
Reproduction of the issue: [cultuurnet/uiv-nuxt-memoryleak-reproduction](https://github.com/cultuurnet/uiv-nuxt-memoryleak-reproduction)

- Run `yarn && yarn build && node --inspect .output/server/index/mjs`
- Open DevTools for Node.js
- Take a heapsnapshot
- Send some users: `ab -n 300 -c 100 http://localhost:3000/`
- Take a heapsnapshot
- Repeat this a coupe of times and you should see the memory increasing and not being released.

Now you can link this fix via yarn link, and should see that the memory leak is fixed.

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -342,22 +342,22 @@ export const VCombobox = genericComponent<new <
         vTextFieldRef.value?.focus()
       }
     }
-    function select (item: ListItem) {
+    function select (item: ListItem, add = true) {
       if (props.multiple) {
         const index = model.value.findIndex(selection => props.valueComparator(selection.value, item.value))
 
-        if (index === -1) {
-          model.value = [...model.value, item]
-        } else {
-          const value = [...model.value]
+        if (~index) {
+          const value = add ? [...model.value, item] : [...model.value]
           value.splice(index, 1)
           model.value = value
+        } else if (add) {
+          model.value = [...model.value, item]
         }
 
         search.value = ''
       } else {
-        model.value = [item]
-        _search.value = item.title
+        model.value = add ? [item] : []
+        _search.value = add ? item.title : ''
 
         // watch for search watcher to trigger
         nextTick(() => {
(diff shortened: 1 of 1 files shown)


alexan:
we tested the fix with our nuxt 3 application and it used way less memory. We still monitor a steady increase in memory but way less then without this change.
Thank you.

PR #18477, opened 2023-10-16
fix(VAutocomplete): increase selection width by using 100% calc

### LINKED ISSUE #18470: [Feature Request] Autocomplete should display as much text as it fits
### Problem to solve
Currently `v-autocomplete` cuts off selected text way too early, even if there is some space still available between text and down arrow. And if "clearable" feature is used, text gets very short.

### Proposed solution
Increase the width of text displayed in autocomplete. That would benefit users of mobile devices, where even 5 more letters make a huge difference. Also it will prevent issue, when text fits during entry, but is cut off after moving focus to other element.
Screenshot1
Screenshot2

yuwu9145:
fixes #18470

## Markup:

```vue
<template>
  <div style="width: 200px">
    <v-autocomplete
      label="Autocomplete"
      clearable
      :items="['California California California California California California California California', 'Colorado', 'Florida', 'Georgia', 'Texas', 'Wyoming']"
    />
  </div>
</template>

```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/_variables.scss
+++ b/packages/vuetify/src/components/VAutocomplete/_variables.scss
@@ -3,9 +3,9 @@
 // Defaults
 $autocomplete-content-border-radius: 4px !default;
 $autocomplete-content-elevation: 4 !default;
-$autocomplete-selection-gap: 2px !default;
 $autocomplete-focused-input: 64px !default;
 $autocomplete-line-height: 1.75 !default;
+$autocomplete-selection-gap: 2px !default;
 $autocomplete-transition: .2s settings.$standard-easing !default;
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VAutocomplete/_variables.scss
```diff
@@ -3,6 +3,7 @@
 // Defaults
 $autocomplete-content-border-radius: 4px !default;
 $autocomplete-content-elevation: 4 !default;
+$autocomplete-dirty-selection-margin-inline-end: 2px !default;
```
KaelWD, line 6:
```suggestion
$autocomplete-selection-gap: 2px !default;
```

johnleider (approved):

Issue #18494, opened 2023-10-18
[Bug Report][3.3.22] Erasing selected auto-complete option issue


DiogoRio:
### Environment
**Vuetify Version:** 3.3.22
**Last working version:** 3.3.13
**Vue Version:** 3.3.4
**Browsers:** Chrome 118.0.0.0
**OS:** Windows 10

### Steps to reproduce
Issue 1:
1. Click into the auto-complete
2. Select "John Smith" (Observe that "John Smith" is added to the autocomplete selected list)
3. Start using the back-space to delete. (Observe that to delete "J" the back-space needs to be pressed twice)

Issue 2:
1. Click into the auto-complete
2. Select "John Smith" (Observe that "John Smith" is added to the autocomplete selected list)
3. Select "John Smith" with your mouse or ctrl+A
4. Press the back-space to delete (Observe that the option is not deleted)
5. Repeat steps 3 and 4.  (Observe that the option is deleted)

### Expected Behavior
Issue 1:
 "J" should have been deleted with only one back-space press
Issue 2:
The selected option should be deleted at the first try

### Actual Behavior
Issue 1:
Pressing the back-space twice is needed to erase the last character of the selected option
Issue 2:
Selecting the whole option and deleting need to be made twice to actually delete it

### Reproduction Link
[...]

PR #15025, opened 2022-05-01
fix(VAutocomplete): update search value for mobile

tibnor:
## Motivation and Context

VAutocomplete only filters complete waits for words to be completed on Android. The input of the inner VTextField of VAutocomplete is binded with a v-model. When using an android phone with autocomplete keyboard (GBoard) the value is first emitted when a word is completed. For instance if you write "House" v-model will first be updated when the word is completed and not on "H" -> "Ho" -> "Hou"... This means that VAutoComplete only filters on completed words. 

Here is a [codepen](https://codepen.io/tibnor/pen/MWQWxwg)

## Description
This pull request changes the databinding of VTextField in VAutoComplete from v-model into `@input` and `:modelValue`. `@input` is emitted on every key, while v-model waits for a word to be completed when using autocomplete on android keyboards. 

## How Has This Been Tested?
The code has been tested visually.

## Markup:

<details>

```vue
<template>
  <v-autocomplete :items="items" />
</template>

<script>
  export default {
    name: 'Playground',
    data: () => ({
      items: [
        {
          value: 'foo',
          title: 'Foo',
        },
        {
          value: 'bar2',
          title: 'Foo',
        },
        {
          value: 'baz',
          title: 'Baz',
        },
      ],
    }),
  }
</script>

```
</details>

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [ x ] The PR title is no longer than 64 characters.
- [ x ] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [ x ] My code follows the code style of this project.
- [ x ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #18499, opened 2023-10-19
fix(VAutocomplete): update searchValue synchronous

JunIce:
## Description

fix: #18494
update searchValue synchronous

## Markup:

```vue

```


yuwu9145 (changes requested):
https://github.com/vuetifyjs/vuetify/pull/15025#issuecomment-1774436086

Please resolve with `v-model`, which is consistent with VCombobox

PR #18558, opened 2023-10-27
feat(VDatePicker): refactor structure

johnleider:
Abstract date picker components into a more reusable structure.


martseguel:
Hi, for some reason this commit made my production build stop working, but in dev works fine. 
https://github.com/martseguel/error.v-date-picker

PR #18574, opened 2023-11-01
feat(VVirtualScroll): add optional position parameter to scrollTo method

jaketig:
## Description

Adds optional `position` parameter to scrollTo method. Accepts `center` or `end`. Defaults to `start` if not provided.
This allows users to scroll an item to the start (current behaviour), center, or end of container.
Mimics `block` option from [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) 

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-virtual-scroll ref="list" :items="items" height="200" class="mb-5">
        <template #default="{ item }">
          Virtual Item {{ item }}
        </template>
      </v-virtual-scroll>

      <v-row align="center">
        <v-col>
          <v-select
            v-model="mode"
            label="Scroll Postition"
            :items="['start', 'center', 'end']"
            hide-details
          />
        </v-col>
        <v-col>
          <v-text-field
            v-model.number="idx"
            label="Scroll To Index"
            hide-details
          />
        </v-col>
        <v-col>
          <v-btn @click="scrollTo">
            Scroll
          </v-btn>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const items = Array.from({ length: 1000 }, (k, v) => v)

      const list = ref()

      const mode = ref('end')
      const idx = ref(50)

      const scrollTo = () => list.value.scrollToIndex(idx.value, mode.value)

      return {
        items,
        list,
        mode,
        idx,
        scrollTo,
      }
    },
  }
</script>

```


J-Sek:
Curious if we can make it work with items having dynamic height. rAF loop did not cover [scenario 3](https://bin.vuetifyjs.com/bins/P4gfVg).

PR #18560, opened 2023-10-29
fix(VOtpInput): Display numeric keyboard when type is number

yuwu9145:
fix #18000

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
</script>


```

PR #18585, opened 2023-11-03
fix(VOtpInput): limit length when type=number

yuwu9145:
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input />
    </v-container>
  </v-app>
</template>

```

### DIFF
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -98,11 +98,16 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
     const current = computed(() => inputRef.value[focusIndex.value])
 
     function onInput () {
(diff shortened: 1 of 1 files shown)


jacekkarczmarczyk:
shouldn't it set https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode?

yuwu9145:
> shouldn't it set https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode?

It says:

> It's important to understand that the inputmode attribute doesn't cause any validity requirements to be enforced on input.

jacekkarczmarczyk:
I'm aware of this, it's not suppose to do any validation, it's for showing proper keyboard

KaelWD:
#18560

yuwu9145:
tested on my iphone, no regression

PR #18613, opened 2023-11-07
fix(VAutocomplete): update:search to not emit when search has not changed

webdevnerdstuff:
## Description

Fixes the issue when the `@update:search` is triggered when the search value has not changed. The `focused` watcher changing the value to an empty string which triggers the search watcher. So anytime the field is focused, it is always triggering the search watcher/emit.

## Markup:

```vue
<template>
  <v-card>
    <v-container fluid>
      <v-row>
        <v-col cols="12">
          <v-autocomplete
            v-model="query"
            :items="filteredItems"
            label="Default"
            clearable
            @update:search="updateSearch"
          >
            <template #selection="{ item }"> {{ item.value }} </template>
          </v-autocomplete>
        </v-col>
      </v-row>
    </v-container>
  </v-card>
</template>

<script setup>
import { ref } from 'vue';
const query = ref('');

const items = ref(['foo', 'bar', 'fizz', 'buzz', 'foobar']);
const filteredItems = ref([]);

function updateSearch(val) {
  if (val !== '' && val.length >= 2) {
    filteredItems.value = items.value.filter((item) => item.includes(val));
  }
}
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -358,7 +358,6 @@ export const VAutocomplete = genericComponent<new <
           select(displayItems.value[0])
         }
         menu.value = false
-        search.value = ''
         selectionIndex.value = -1
       }
     })


johnleider:
I'm not seeing this issue. What am I looking for in your example?

webdevnerdstuff:
Steps to replicate

1. Type in `Foo` and select that item. 
2. Notice next to "searchVal" it shows `foo`.
3. Select any area outside of the field (focus out basically).
4. Notice the value next to "searchVal" changes to an empty string.
5. Select the field again to activate it.
6. Notice the value next to "searchVal" changes back to "foo".

Where this causes a problem is if a user is trying to debounce a search value that get's sent to the backend to find results, the `update:search` value changes every time you select/deselect the field. So what winds up happening is a query based off of the search will keep triggering when the field is focused or focus out. Both the `update:focused` and `update:search` are being triggered when just the `update:focused` should be triggered.

[Playground Example](https://play.vuetifyjs.com/#eNqVVE1v2zAM/SucB9Qp1jgf3SmIgxUdBuw8YJemB8WmF2Gy5clShtXwfx8lxa7stAV6iBXqkY/mI+mHNmpUtrir6+RkMNpEW41lLZjG3b4C2J7mGVO5++8tWWnGK1RQCMN7wEFK/h3Ms68A+jXpPlqt91EAOpgZLTNJyVBjCAGc5qXMUVDcH4Pq3z4awxtO72hZCy40Ksy/W3vqJdjBUXzFghmhp3AmkCl2EJPUX0ydU/GbhtDsSOHe/uHMKUfvXMjMNJgP3t/IxnzsPaqe6u9lho8NCsw0lxXFt2Brg47UgtYbyYkJg9B1sF2MejNQLcZahk0giDrwZltICdZYo9Tz1XLapoOCxe69Ty/eTyY2IVfbDvddd5kkvCnZb7yr+T0T4l6aSm+sGNNLkuRtFtsWzJ8Jzm26DJ7KZO3nafboeez9VtgbvxajnpDZZIrXGhrUpna+vKyl0tCCwgI6KJQsIaZViy1ItI0GN+SQWo9ZHF9bnh5yk36GHuJCyvgG4gNT9ij405MzjT8JtcgjEfThowXpaUKHoSNh+h68ENz7LMMEoaSvwBN2ixamchMP4XbNaNCvofWiDy92Hv8U6PQQL8C6woc0hTiGqysLJQKrX/oIuxTWA8nlIA1srwCfUlj1sSPxhkDXEW8l3mPmVjbd+W3lVSZMjo2rxkkBtM500OOF0v2nYlR6L9pl5aHaA/rCZV8Fpdwu/ETSLEbdTXSbfE6WcybqI0vWkbdvo8f/Q4faXQ==)

johnleider:
autocomplete

johnleider:
We talked about this last night and agreed this probably should exist as an alert on the autocomplete documentation explaining how search is manipulated.

PR #18636, opened 2023-11-09
docs(VNavigationDrawer): update location description

I-am-abdulazeez:
Location prop description missing


MajesticPotatoe (changes requested):
Description is more fluff than an actual description of what the prop does. Should be straight to the point like other descriptions.

johnleider (approved):

PR #18703, opened 2023-11-16
fix(VDataTable): improve headers value function typing

xdavidwu:
## Description

Improve typing on value functions with generics.

With markup below, before:
```
    Types of property 'value' are incompatible.
      Type '(v: Data) => string' is not assignable to type 'SelectItemKey'.
        Type '(v: Data) => string' is not assignable to type '(item: Record<string, any>, fallback?: any) => any'.
          Types of parameters 'v' and 'item' are incompatible.
            Property 'string' is missing in type 'Record<string, any>' but required in type 'Data'.

4       <v-data-table :items="items" :headers="headers" />
```
after: passes typecheck.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-data-table :items="items" :headers="headers" />
    </v-container>
  </v-app>
</template>

<script lang="ts">
  interface Data {
    string: string
  }

  export default {
    name: 'Playground',
    setup () {
      return {
        items: [
          { string: 'foo' },
          { string: 'bar' },
        ] as Data[],
        headers: [
          { key: 'foobar', title: 'String', value: (v: Data) => v.string },
        ],
      }
    },
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VDataTable/VDataTable.tsx
+++ b/packages/vuetify/src/components/VDataTable/VDataTable.tsx
@@ -24,9 +24,9 @@ import { computed, toRef } from 'vue'
 import { genericComponent, propsFactory, useRender } from '@/util'
 
 // Types
-import type { UnwrapRef } from 'vue'
+import type { DeepReadonly, UnwrapRef } from 'vue'
 import type { Group } from './composables/group'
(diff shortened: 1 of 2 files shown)


johnleider:
Can you reach out to me on Discord?

PR #18720, opened 2023-11-18
fix(VListItem): subtitle wrong font

### LINKED ISSUE #18719: [Bug Report][3.4.2] List Item subtitle font wrong when value is "0"
### Environment
**Vuetify Version:** 3.4.2
**Vue Version:** 3.3.8
**OS:** Mac OS 10.15.7

### Steps to reproduce
When setting a list item subtitle as the value '0', the font seems to be incorrect. 

### Expected Behavior
The font of subtitle should always be the same.

### Actual Behavior
The font is wrong when the value is the number 0. 

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVkDFvwjAQhf/K6y1ppZK0pUOFAlI3dgYGwhDCgSw5sWVfGED8d+wEECAxMPne8/mz3lscyLsq+7c23bVMI8qFa6tL4UnRAPluUFrbjZ2oTCOlatidrc7UystVX52BCiSIEs3jgmbtqhuhPPbsTEEY+bMXrr8KmuTZzcNXeKbhB9z3U9zFvUQK8j5TdPrIeXZTRZC+csoKPEvbN6Jqa5zgAMcbHLFxpkYSWkziOhC4XlD7LcZx4z2ZstYGc+P0+i35iB/0yACj4ycN02H6R/H8TX9oeQJnkIPr)

korpusovmax:
fixes #18719

## Description
Changed prop condition ```(props.subtitle)``` to ```(props.subtitle !== undefined)```, as subtitle can be a number (```subtitle: [String, Number, Boolean]```), and for this reason, if subtitle was ```0``` (not a string), there was a bug.

## Markup:

Code from the issue, now working fine:

```vue
<template>
  <v-app>
    <v-container>
      <v-list>
        <v-list-item title="Subtitle is zero" :subtitle="0"></v-list-item>
        <v-list-item title="Subtitle is one" :subtitle="1"></v-list-item>
      </v-list>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

</script>

```

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -180,7 +180,7 @@ export const VListItem = genericComponent<VListItemSlots>()({
     useRender(() => {
       const Tag = isLink.value ? 'a' : props.tag
       const hasTitle = (slots.title || props.title)
-      const hasSubtitle = (slots.subtitle || props.subtitle)
+      const hasSubtitle = (slots.subtitle || props.subtitle !== undefined)
       const hasAppendMedia = !!(props.appendAvatar || props.appendIcon)
       const hasAppend = !!(hasAppendMedia || slots.append)
       const hasPrependMedia = !!(props.prependAvatar || props.prependIcon)


johnleider:
I'm not sure what the reason was for making subtitle a [String, Number, Boolean], but I think this was a mistake, as everywhere else we use String.

I'd be open to changing this to [Number, String] for all components, VCard/VCardItem/VStepperItem/etc and the corresponding checks to `props.subtitle != null`.

johnleider (changes requested):
Please change the !== null comparisons to != null. Otherwise looks good.

johnleider (approved):

Issue #4000, opened 2018-05-07
[Bug Report] Selected options are removed when options are updated even when options are available in model property


akki-jat:
### Versions and Environment
**Vuetify:** 1.0.16
**Vue:** 2.5.11
**Browsers:** Chrome 66.0.3359.139
**OS:** Windows 7

### Steps to reproduce
Click on update button.
Then see in model there is but selection is removed from v-select

### Expected Behavior
Option should not removed from selection

### Actual Behavior
It is removed from selection

### Reproduction Link
[https://codepen.io/akki-jat/pen/yjPXPL](https://codepen.io/akki-jat/pen/yjPXPL)

KaelWD:
That behaviour is intentional, although it should be removed from the model too iirc.

akki-jat:
Hi @KaelWD, I want behavior like, if i remove from list after selecting it should be displayed as selected.
Is that possible?

KaelWD:
If you use `cache-items` then yes, although it will still be visible in the menu. I don't really know what your use-case is for this though.

akki-jat:
@KaelWD thanks for your answer, I can use cache-items but if it would be good if removed item is not visible in list.

akki-jat:
And if you want reference of this use-case, you can check libraries like vue-multiselect or element-ui's select component.

KaelWD:
I don't think that's good behavior. If the item is removed from the list but still selected then there will be no way to unselect it. The `el-select` seems similar to our `<v-select tags>` however, which does appear to have the behaviour you want.

akki-jat:
@KaelWD ok, thanks for answer.

akki-jat:
@KaelWD But if i make v-select autocomplete then there is problem.
It will allow user to create tag.

akki-jat:
Hi @KaelWD , can you check changes I made related to this issue.
And I am new to contributing, if there is anything I can improve you can tell me

johnleider:
You cannot have a selection that is not present in the available options. The fact that it leaves the model is completely fine as you can provide whatever value you want to the input.  I see that you provided an example of another framework working as you want, but I'd argue to say that it is invalid and not what you would expect from a regular select, so it will not be here.

akki-jat:
Hi @johnleider, I also agree with you. But in my use-case I want this behaviour, ok let me tell you how I am using v-select.
I have four v-select.
1.) Destinations
2.) Sailing Dates
3.) Sailing Duration
[...]

PR #18766, opened 2023-11-24
fix(VAutocomplete): Allow an empty string to function as a legitimate v-model when it is not part of the list items

### LINKED ISSUE #18737: [Bug Report][3.4.2] Autocomplete initial value
### Environment
**Vuetify Version:** 3.4.2
**Vue Version:** 3.3.8
**OS:** Linux x86_64

### Steps to reproduce
1. Create a v-autocomplete component, binding a model value with an initial value that does not match any item from the items.
2. The autocomplete component will render initially showing the invalid value.

### Expected Behavior
The Autocomplete should discard the non-matching initial value, as it is not contained in the items

### Actual Behavior
The Autocomplete component renders with the invalid initial value.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNkM9qwzAMxl9F+OINmgTSHUZIB91rND24mdoZ/A9bySXk3ScnC7Q99STp088f+nyaRIp9dQyhHAcUjWgJbTCK8KtzAO1YqBCWdhl670hph/FfWomBfO/5FRLCWFj/g+bQCZtunYBGs2HicaksGHVZ1gkN9gTKQd7wonrNs37dFDK72bbV8/lZWdO11V1qHlMfdSBISMMaXtvgI8EEEa8wwzV6C5I/TGYcgH0TAR8Hh0y8Sfn+INeb7rwrrKL+V7vbPbPkYOi03joBaTLYgFRyB6MyQ+65HiXMu2fm8sh8b8w5B1ujcAgx78S+3JefItePshbnP+UGo1o=)

### Other comments
This functionality was useful in order to bind an empty element to the component (such as empty string). The component would render properly without containing the empty value (the label would stay down, same as not binding any model value).

yuwu9145:
fixes #18737

As per discussion in the issue, arbitrary v-model value (except for null & undefined) in select/autocomplete should be respected.

In my opinion, the most convincible reason to support it is https://github.com/vuetifyjs/vuetify/issues/4000#issuecomment-396668322

So #18737 isn't really a bug but this PR ensures empty string is also a legit v-model.

Screen Recording 2023-12-05 at 4 53 59_pm

## Markup:

Reproduction:

1. Load playground, autocomplete presents empty string (label goes up)
2. Click and focus autocomplete 
3. Click outside and remove its focus

Outcome: 

autocomplete doesn't consistently present empty string (label goes down)

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        return-object
        v-model="msg"
        :items="items"
        label="select an item 2"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref('')
  const items = ref([
    'California',
    'Colorado',
    'Florida',
    'Georgia',
    'Texas',
    'Wyoming',
  ])
</script>


```


johnleider:
To my knowledge this is intended because of functionality that is capable in html select.

johnleider (dismissed):
Looks good, but change the title to reflect

johnleider:
Is your playground updated to reflect what this change fixes? It's not clear.

yuwu9145:
@johnleider apologise, now playground is updated with minimum code

johnleider (approved):

PR #18771, opened 2023-11-25
docs: update nuxt 3 installation

albasyir:
have tried and all works perfectly

## Description
fix wrong installation nuxt 3 installation steps

## Markup:
- `yarn create nuxt-app <project-name>` will install nuxt version 2, so change to `npx nuxi@latest init <project-name>
` as solution
- when you follow script on `plugins/vuetify.ts` you got warning because css is not loaded properly, instead i give comment to make sure follower can understand that only need when they using @mdi/font


thomasbnt (commented):
LGTM! Thanks for this edit, I also ran the wrong command, and when I saw Nuxt v2 installed instead of v3. Time to update the doc! :rocket: (Can't review here)

johnleider (approved):

PR #18785, opened 2023-11-28
types: vue types utilities

pikax:
## Description

Support for newer vue utility types https://github.com/vuejs/core/pull/9556, there was an overall work done to `defineComponent` , it requires Typescript `5.3.*` and `slots` are not currently failing to resolve the type.


KaelWD:
> requires Typescript 5.3

Not currently supported by eslint.

PR #18787, opened 2023-11-28
fix(memory): nuxt ssr memory leak fixes composable locale

memic84:
## Description
When using vuetify in Nuxt SSR, with high traffic there are memory leaks on certain objects. I was able to isolate those to the locale composable and fix them. Resolves 18393

## Markup:
This repository can be used to reproduce: https://github.com/memic84/vuetify-memory-leak-reproduction

First you should link your local vuetify code with yarn link and yarn link "vuetify"

and then you can run Nuxt.

```
yarn && yarn build && node --inspect .output/server/index.mjs
```

To test the fix, link the PR code to the vuetify reproduction code.

## Note
The leakage in locale composable led me to changing the reactive nature of isRtl and rtlClasses variables.  I am not sure if this will cause problems, and if this is the correct way. If somebody has a better way, please let me know.

### DIFF
--- a/packages/vuetify/_settings.scss
+++ b/packages/vuetify/_settings.scss
@@ -1,3 +1,0 @@
-@forward './lib/styles/settings';
-@forward './dist/component-variables';
-@forward './dist/component-variables-labs';
--- a/packages/vuetify/_styles.scss
+++ b/packages/vuetify/_styles.scss
@@ -1,1 +1,0 @@
-@forward './lib/styles/main';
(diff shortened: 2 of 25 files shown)


johnleider:
Can this be split into different PRs for theme and locale.

J-Sek:
Closing as stale. If you think it was not addressed by #18975, feel free to open issue with reproduction using latest versions (Vuetify, Vue, Nuxt or Vite).

PR #18824, opened 2023-12-05
fix(memory): defaultThemeOptions moved inline theme.ts

memic84:
## Description
When using vuetify in Nuxt SSR, with high traffic there are memory leaks on certain objects. I was able to isolate those to the theme composable and fix them. Resolves 18393

## Markup:
This repository can be used to reproduce: https://github.com/memic84/vuetify-memory-leak-reproduction

First you should link your local vuetify code with yarn link and yarn link "vuetify"

and then you can run Nuxt.

```
yarn && yarn build && node --inspect .output/server/index.mjs
```

To test the fix, link the PR code to the vuetify reproduction code.

### DIFF
--- a/packages/vuetify/_settings.scss
+++ b/packages/vuetify/_settings.scss
@@ -1,3 +1,0 @@
-@forward './lib/styles/settings';
-@forward './dist/component-variables';
-@forward './dist/component-variables-labs';
--- a/packages/vuetify/_styles.scss
+++ b/packages/vuetify/_styles.scss
@@ -1,1 +1,0 @@
-@forward './lib/styles/main';
--- a/packages/vuetify/_tools.scss
+++ b/packages/vuetify/_tools.scss
@@ -1,1 +1,0 @@
-@forward './lib/styles/tools';
--- a/packages/vuetify/src/components/VApp/VApp.tsx
+++ b/packages/vuetify/src/components/VApp/VApp.tsx
@@ -33,7 +33,7 @@ export const VApp = genericComponent()({
           'v-application',
           theme.themeClasses.value,
           layoutClasses.value,
-          rtlClasses.value,
+          rtlClasses,
           props.class,
         ]}
         style={[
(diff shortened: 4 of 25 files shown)


johnleider:
At a glance this seems very heavy handed. I'll carve out some time this week to investigate.

johnleider:
I took a different approach to this. I'm not convinced rtlClasses needs to change either. More info here https://github.com/vuetifyjs/vuetify/commit/af07b7fe8b2fb8e9b468d9c946aa5fcf2551654d

If you want to refactor this to only address rtlClasses that's fine, but I'm going to need a valid reproduction that shows the leak. Thank you!

memic84:
> I took a different approach to this. I'm not convinced rtlClasses needs to change either. More info here [af07b7f](https://github.com/vuetifyjs/vuetify/commit/af07b7fe8b2fb8e9b468d9c946aa5fcf2551654d)
> 
> If you want to refactor this to only address rtlClasses that's fine, but I'm going to need a valid reproduction that shows the leak. Thank you!

I have split the PR into two. This one contains the locale composable: https://github.com/vuetifyjs/vuetify/pull/18787

I was able te reproduce and isolate the memory leak in (the computed rtlClass): https://github.com/vuetifyjs/vuetify/blob/2c2b7deba00146f7918dfac90ea9dc31c6c86200/packages/vuetify/src/composables/locale.ts#L91C12-L91C12 (not sure why the computed is leaking, best guess is that ik keeps references)

Here is the reproduction: https://github.com/memic84/vuetify-memory-leak-reproduction

When running the nuxt build with node --inspect, and use ApacheBench or something similair (every 200 users), the memory keeps going up.

Steps are also mentioned here in the description: https://github.com/vuetifyjs/vuetify/issues/18393

johnleider:
Your P.R.s both contain the same changes. Please make sure #18787 doesn't have the same stuff as this.

J-Sek:
I am pretty sure [`e4351d4`](https://github.com/vuetifyjs/vuetify/commit/e4351d4) already covered it.

PR #18847, opened 2023-12-11
feat(dist): Components and directive names type

a3mitskevich:
## Description
Added a dts file with component and directive names for more comfortable development plugins/tools what based from vuetify.

Where it will be placed:
image
How it would be looks:
image

### DIFF
--- a/packages/vuetify/build/rollup.config.mjs
+++ b/packages/vuetify/build/rollup.config.mjs
@@ -27,6 +27,10 @@ function fixWindowsPath(path) {
   return path.replace(/^[^:]+:\\/, '\\').replaceAll('\\', '/')
 }
 
+function generateNamesTypeExport(typeName, types) {
+  return `export type ${typeName} =\n${types.map(name => `\t${name}`).join(' |\n')};`
+}
+
 export default [
   {
     input: 'src/entry-bundler.ts',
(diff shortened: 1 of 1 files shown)


KaelWD:
Why?

a3mitskevich:
> Why?

I think it will be very helpful for the vue-loader ignore property ( for example ). Instead of creating every time additional scripts to generate types or write types manually, we can create an auto-generated file with name types that to placed to dist in every release.

a3mitskevich:
If you think is a overfeature - I will close the pull request 
But I hope additional typescript types can be very helpful for other developers

KaelWD:
Yeah doesn't really clarify what you'd actually use this for or why you can't do `import type * as components from 'vuetify/components'`

a3mitskevich:
> Yeah doesn't really clarify what you'd actually use this for or why you can't do `import type * as components from 'vuetify/components'`

Your example provides a component type that is not a literal representation. To get a default Vuetify component name needs to access to name field which is not easy in typescript. 
image
For directives is not possible. Need to declare it manually in every project there it needed 😢

KaelWD:
Again, what are you actually trying to use this for?

a3mitskevich:
> Again, what are you actually trying to use this for?
example:
image

KaelWD:
```
import type * as components from 'vuetify/components'
(keyof components)[]
```

PR #18881, opened 2023-12-15
fix(VOtpInput): exclude loader slot in VField slots

### LINKED ISSUE #18809: [Bug Report][3.4.4] VOtpInput displays custom loader even when loading is false
### Environment
**Vuetify Version:** 3.4.4
**Vue Version:** 3.3.8
**Browsers:** Safari 16.6
**OS:** Mac OS 10.15.7

### Steps to reproduce
Just create a v-otp-input with a loader slot used

### Expected Behavior
Loading not rendered until loading property is set to true

### Actual Behavior
Shows immediately. And in each in of the input controls (not for the whole control as the default loader shows)

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNplj0sKwzAMRK9itK6dQrIKodBztF2IRLQGxxa2kk3p3ZtfS0x2modmNLq9IcW2uDKbcSCooRHq2aHQ5e6VakaNzMu4iDZ4QespbmiBQVhbz4P84YR/MWrUyQWpXcBuZ9usHMMzUkraTaEYlfUdCcXe+rlBUxw29ieKrOqKjm1mlteeyfpVFgGfE5SmMmeNjl9oSlh1BY8vVDRckA==)

yuwu9145:
fixes #18809

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input :loading="true">
        <template v-slot:loader>
          <v-progress-circular indeterminate ></v-progress-circular>
        </template>
      </v-otp-input>
      <v-otp-input :loading="true">
        <template v-slot:loader>
          <v-progress-linear :location="'bottom'" indeterminate></v-progress-linear>
        </template>
      </v-otp-input>
    </v-container>
  </v-app>
</template>




```

### DIFF
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -287,7 +287,7 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
             <VOverlay
               contained
               content-class="v-otp-input__loader"
-              modelValue={ !!props.loading }
+              model-value={ !!props.loading }
               persistent
             >
               { slots.loader?.() ?? (
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -212,6 +212,8 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
     useRender(() => {
       const [rootAttrs, inputAttrs] = filterInputAttrs(attrs)
 
+      const { loader: _, ...vFieldSlots } = slots
+
       return (
         <div
           class={[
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
```diff
@@ -212,6 +212,8 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
     useRender(() => {
       const [rootAttrs, inputAttrs] = filterInputAttrs(attrs)
 
+      const { loader: _, ...vFieldSlots } = slots
```
KaelWD, line 215: Use omit? https://github.com/vuetifyjs/vuetify/blob/8030d407c5767f96c0474d94828b2a79b40e7c17/packages/vuetify/src/util/helpers.ts#L257-L266

yuwu9145: `prop: _` is also being pervasively used. Was there any obvious benefit of using omit? I guess both are okay so just need to pick one of them?

https://github.com/vuetifyjs/vuetify/blob/d0765f1125b5f3555acbbcf05aeec3bfeed6de72/packages/vuetify/src/components/VTextField/VTextField.tsx#L160

PR #18975, opened 2024-01-04
fix: a bug causing rtl as a ref to cause a memory leak in ssr #18965

### LINKED ISSUE #18965: [Bug Report][3.4.9] SSR memory leak
### Environment
**Vuetify Version:** 3.4.9
**Vue Version:** 3.4.3
**OS:** Windows 10

### Steps to reproduce
- install clinic on your system to analyze the memory fo the server
- create a minimal nuxt-project with vuetify configured / or install the reproduction
- build the server and let clinic run against it / when using the reproduction -> run `npm run clinic:prep`

### Expected Behavior
Memory should stay stable

(60s test)
image

### Actual Behavior
Memory increases with every request. The more Requests you have, the bigger the leak.

(300s test)
image

### Reproduction Link
[https://github.com/Jindorf/nuxt-vuetify-leak](https://github.com/Jindorf/nuxt-vuetify-leak)

### Other comments
I originially tested the problem with Vike + Express instead of Nuxt as the ssr-engine. I was hoping that Vike was not rendering correctly, so i tested Nuxt + Vuetify. It's the same behaviour. As soon as I use the VApp-Component in a SSR-context it leaks. No theme defined and using the latest Vuetify version. This was tested in Node v20.10 aswell as in Node v18.16

Jindorf:
## Description

resolves #18965


johnleider:
What is your test process for checking the leak? Realistically the fix for this is to turn these defaults into a factory function:

image

Jindorf:
Thanks for the reply!
I think I called out every step i took inside the issue. I also provided a reproduction inside. 
I installed a plain nuxt project, installed vuetify as in the docs, then ran the clinic test i provided in the package.json. 
After that i changed the vuetify code and ran the test again. 
If you need any more information, I'll try to help as best as I can.

Jindorf:
@johnleider your suggestion also fixes the leak.

reverting my previous changes, 60s test:
image

applying your suggestions, 60s test:
image

johnleider (approved):

PR #19011, opened 2024-01-09
fix(locale): update Slovak locale

volarname:
Translated missing text for Slovak locale. Also updated some existing for better match the Slovak language


packages/vuetify/src/locale/sk.ts
```diff
@@ -56,13 +56,13 @@ export default {
...
-    today: 'Today',
+    today: 'Dnes',
   },
   input: {
-    clear: 'Clear {0}',
+    clear: 'Vymazať {0}',
     prependAction: '{0} prepended action',
     appendAction: '{0} appended action',
-    otp: 'Please enter OTP character {0}',
+    otp: 'Vložte výhradne OTP znaky {0}',
```
scscgit, line 65: Why "Enter exclusively OTP characters"?

volarname: ok then there are another posibilities valid for slovak lang, its not possible to translate it exactly and as the translation its used for aria-label, this are the best possibilities for slovak language aligned from shortest to longest:
"Vložte {0}. znak pre kód"
"Prosím vložte {0}. znak pre kód"
"Prosím vložte {0}. znak pre jednorazový kód"
"Prosím vložte {0}. znak pre jednorazový kód"
"Prosím vložte {0}. znak pre jednorazový verifikačný kód"

pick one and i will update the PR
notes: there is no term otp in slovak language or shortcut, i thing my provided possibilities are best one and at least better than before for now for aria-label

scscgit: I'd vote for `Prosím zadajte OTP znak {0}`, the abbreviation is a well known one. ["Jednorazový kód"](https://cs.wikipedia.org/wiki/Jednor%C3%A1zov%C3%A9_heslo) also seems OK, I think it'll depend on the context whether we want to use an abbreviation or not.

volarname: updated to `Prosím zadajte OTP znak {0}`

KaelWD: This is pretty verbose in english too. It's a screenreader hint for one of several fields, should probably be something a lot shorter like "OTP Character {0} of {1}".

volarname: i was working with original translation, I checked source code and lastly there was only one parameter and yes its aria-label for each input character, its the second parameter `{1}` already implemented or planned?

volarname: i am for a merge finally, now its translation 1:1 to current en version
once there will be changes/fixes to en version or parameter for totalcount will be added, i will update it again in new PR

MajesticPotatoe (approved):

PR #19037, opened 2024-01-12
feat(VStepper): pass item props from item objects

### LINKED ISSUE #19036: [Feature Request] VStepper: Pass item object properties to item components
### Problem to solve
When passing an array of objects in the `items` prop of VStepper, it automatically generates VStepperItem components from it. Currently, the only props of VStepperItem that can be set through these objects are `value` and `title`. I would like to set other props from VStepperItem too, like `completed`, `rules`, `disabled`, `icon`, etc.

### Proposed solution
VStepper creates internal items from the passed-in items, which are then bound to VStepperItem. An internal item has three properties, `value`, `title` and `raw` (the initial item - object or string). It would be easy to take over other props from the initial objects as well.

mringler:
## Description

fixes #19036 (see for problem description)

Implementation is straight-forward: If items are objects, their properties are filtered by prop names of VStepperItem and added to the internal item.

Tests are added to ensure string items still work and props are taken over as expected.
 
## Markup:

```vue
<template>
  <v-container>
    <v-card color="primary">
      <v-card-title>Stepper with items in slot</v-card-title>

      <v-stepper :model-value="items[2].value">
        <v-stepper-header>
          <v-stepper-item v-for="item in items" v-bind="item"></v-stepper-item>
        </v-stepper-header>
        <v-stepper-window />
      </v-stepper>
    </v-card>
    <v-card color="primary" class="mt-4">
      <v-card-title>Stepper with :item prop</v-card-title>

      <v-stepper :model-value="items[2].value" :items="items"
        ><template #actions
      /></v-stepper>
    </v-card>
  </v-container>
</template>
<script setup lang="ts">
    import { ref } from 'vue'
    import type { VStepperItem } from 'vuetify/components';

    const items = ref([
      {
        value: 'a',
        title: 'Completed with color',
        complete: true,
        color: 'secondary',
      },{
        value: 'b',
        title: 'Erroneous',
        rules: [() => false],
      },{
        value: 'c',
        title: 'Disabled current step',
        disabled: true
      },{
        value: 'd',
        title: 'Subtitle & custom icon',
        subtitle: 'subtitle',
        icon: 'mdi-car',
      },
    ])
</script>
```
## Notes:
- VList works similarly, but the [`item-props`](https://vuetifyjs.com/en/api/v-list/#props-item-props) prop has to be used to enable takeover whole or in parts. This makes sense for lists, which are often generated using existing data, like a list of purchases. I am not sure if it makes sense for a stepper, which is generally not created from unrelated data objects. But I can easily add it if requested.
- I would like to rename the `items` variable to `internalItems`, since the word `item` is somewhat overused in the component, but I am not sure if this is the place to do it (is it?).

This is my first contribution, please let me know what I am doing wrong!
Thank you for Vuetify! I enjoy it a lot.

### DIFF
--- a/packages/vuetify/src/components/VStepper/VStepper.tsx
+++ b/packages/vuetify/src/components/VStepper/VStepper.tsx
@@ -16,7 +16,7 @@ import { makeGroupProps, useGroup } from '@/composables/group'
 
 // Utilities
 import { computed, toRefs } from 'vue'
-import { genericComponent, getPropertyFromItem, only, propsFactory, useRender } from '@/util'
+import { genericComponent, getPropertyFromItem, isObject, only, pick, propsFactory, useRender } from '@/util'
 
 // Types
 import type { InjectionKey, PropType } from 'vue'
@@ -90,11 +90,14 @@ export const VStepper = genericComponent<VStepperSlots>()({
     const { items: _items, next, prev, selected } = useGroup(props, VStepperSymbol)
     const { color, editable, prevText, nextText } = toRefs(props)
 
+    const itemComponentPropNames = Object.keys(VStepperItem.props)
     const items = computed(() => props.items.map((item, index) => {
       const title = getPropertyFromItem(item, props.itemTitle, item)
       const value = getPropertyFromItem(item, props.itemValue, index + 1)
+      const inheritedProps = isObject(item) ? pick(item, itemComponentPropNames) : null
 
       return {
+        ...inheritedProps,
         title,
         value,
         raw: item,
--- a/packages/vuetify/src/components/VStepper/__tests__/VStepper.spec.ts
+++ b/packages/vuetify/src/components/VStepper/__tests__/VStepper.spec.ts
@@ -1,0 +1,77 @@
+// Components
+
+import { VStepper } from '../VStepper'
+import { VStepperItem } from '../VStepperItem'
+
+// Utilities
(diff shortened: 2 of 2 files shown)


mringler:
Hmm, unrelated tests fail, please let me know what to do about that.

captainlettuce:
> Hmm, unrelated tests fail, please let me know what to do about that.

I'm not a maintainer but if you rebase on main (if possible); the pipeline should trigger again and hopefully pass :) 

I'm also waiting for this to be merged, thanks for your fix @mringler

mringler:
@captainlettuce Thanks for the tip! Changed to master, which apparently does not run tests at all. Problem solved (kind of).

johnleider:
I feel this implementation should match what we do with v-list, where we have an **item-props** property https://github.com/vuetifyjs/vuetify/blob/3d9deb7db53e70e0605b76f270f7d962d38f09e4/packages/vuetify/src/composables/list-items.ts#L63

mringler:
@johnleider Thank you for the feedback, I have updated code and tests.

johnleider:
As a true feature, this will also need to go to the `dev` branch. Sorry for not mentioning this before.

mringler:
@johnleider No worries, thanks for letting me know. Rebased to dev and changed merge target.

J-Sek:
Closing in favor of #20651

PR #19048, opened 2024-01-15
feat(VDataTable/Virtual): add support for sorting raw objects

### LINKED ISSUE #11226: [Feature Request] v-data-table header:sort function should be passed entire row, not just cell
### Problem to solve
consider that you have a table of first name in one column and last name in the other column. There is going to be a high probability that multiple entries have the same first name. There is also going to be a high probability that multiple entries will have the same last name. 

A good sort for the first name column will be to sort by the first name while also doing a sort on the last name when duplicate first names are detected. This will result in a well sorted list by first name. 

The same happens when sorting the last name. A good solution is to break the tie based on the first name. 

This seems to be impossible with the header sort function of the v-data-table because the sort function seems to be passed the cell values (first name or last name) only and not the entire record for that row. 

I understand that you might come back with "use the custom sort prop" for something like this, but that seems like drastic overkill for something as simple as this problem and as prevalent as this problem.

### Proposed solution
I recognize that if you change the signature of that custom sort function now, it will break all existing code.
I propose that you add two arguments to the sort function after the a and b arguments that are there already. 
```JS
export type DataTableCompareFunction<T = any> = (a: T, b: T) => number
```
would become:
```JS
export type DataTableCompareFunction<T = any, R = any> = (a: T, b: T, ra: R, rb: R) => number
```
where R is the record type.

Since this function will be called a number of times, maybe having 4 arguments rather than 2 arguments is too much overhead. Another approach would be to add a second sort function with the  same signature, but instead of passing in T, you would be passing in R.

```JS
export interface DataTableHeader<T extends any = any> {
  text: string
  value: string
  align?: 'start' | 'center' | 'end'
  sortable?: boolean
  filterable?: boolean
  divider?: boolean
  class?: string | string[]
  width?: string | number
  filter?: (value: any, search: string | null, item: any) => boolean
  sort?: DataTableCompareFunction<T>,
  sortWithItem?: DataTableCompareFunction<T>
}
```

johnleider:
## Motivation and Context

closes #11226

TODO:

- [x] Types
- [x] Documentation

## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-data-table-virtual
        height="500"
        :headers="headers"
        :items="items"
      />
    </div>
  </v-app>
</template>

<script setup>
  function useSort (a, b) {
    if (a.email === 'email31@example.com') return -1
    if (b.email === 'email31@example.com') return 1

    return a.first.localeCompare(b.first)
  }

  const headers = [
    { title: 'First Name', key: 'first', sortRaw: useSort },
    { title: 'Last Name', key: 'last' },
    { title: 'Email', key: 'email' },
    { title: 'Age', key: 'age' },
    { title: 'Company', key: 'company' },
  ]

  const items = Array.from({ length: 500 }, (_, i) => ({
    first: `First` + (i === 4 ? 'a' : 'ab'),
    last: `Last`,
    email: `email${i}@example.com`,
    age: 20 + i,
    company: `Company${i}`,
  }))
</script>

```
</details>

### DIFF
--- a/packages/docs/src/examples/v-data-table/prop-headers-sort-raw.vue
+++ b/packages/docs/src/examples/v-data-table/prop-headers-sort-raw.vue
@@ -53,21 +53,23 @@
 </script>
 
 <script>
-  function sortRaw (a, b) {
-    if (a.location < b.location) return -1
-    if (a.location > b.location) return 1
-
-    const dateA = a.constructed.split('-').pop().trim()
-    const dateB = b.constructed.split('-').pop().trim()
-
-    return dateA.localeCompare(dateB, undefined, { numeric: true, sensitivity: 'base' })
-  }
-
   export default {
     data: () => ({
       headers: [
         { title: 'Name', key: 'name' },
-        { title: 'Location', key: 'location', sortRaw },
+        {
+          title: 'Location',
+          key: 'location',
+          sortRaw (a, b) {
+            if (a.location < b.location) return -1
+            if (a.location > b.location) return 1
+
+            const dateA = a.constructed.split('-').pop().trim()
+            const dateB = b.constructed.split('-').pop().trim()
+
+            return dateA.localeCompare(dateB, undefined, { numeric: true, sensitivity: 'base' })
+          },
+        },
         { title: 'Constructed', key: 'constructed' },
         { title: 'Description', key: 'description' },
       ],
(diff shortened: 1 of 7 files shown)


claussma:
The data in the example is mainly sorted by location, but the resulting table is not sorted correctly. I tested the function locally and got the same problem.  Maybe sortRaw did not work properly.

PR #19053, opened 2024-01-15
feat(Tooltip): add new directive

### LINKED ISSUE #9610: [Feature Request] Add tooltip as directive
### Problem to solve
Tooltips are needed very often, and every time they are needed, we need to write quite a lot of code for one (according to documentation):
```
<v-tooltip right>
    <template v-slot:activator="{on}">
        <v-btn v-on="on">Button</v-btn>
    </template
    <span>I'm a tooltip!</span>
</v-tooltip>
```
Most of this stuff is needed only for advanced cases

### Proposed solution
 It would be awesome if they were available as a directive as well, for example:
```
<v-btn v-tooltip:right="I'm a tooltip!">Button</v-btn>
```

johnleider:
## Motivation and Context

closes #9610

TODO:

- [x] Documentation

## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <div v-tooltip="'Hello there'" class="d-inline-block">
        This is long text with a tooltip
      </div>
    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
</script>

```
</details>


KaelWD:
The blocker for this was documentation, `/api/v-tooltip` is already used for the component.

Safrankuchen:
pls merge fast i want this

PR #19046, opened 2024-01-15
feat(goTo): port to v3

johnleider:
## Motivation and Context

closes #13538
closes #5732

TODO:

- [x] Horizontal scrolling
- [x] RTL
- [x] Offset
- [x] Unit tests
- [x] Documentation

## Markup:
<details>

```vue
<template>
  <v-layout>
    <div ref="container" class="overflow-x-auto" style="width: 2200px;">
      <div
        id="test2"
        style="margin-left: 2000px; width: 200px;"
      >
        Testing 2

        <v-btn @click="goTo.horizontal(0, container)">Back to left</v-btn>
      </div>

      <v-main style="min-height: 700vh">
        <v-btn text="Vertical" @click="goTo('#test')" />
        <v-btn text="Horizontal" @click="goTo.horizontal(2000, container)" />

        <div id="test" ref="btn" style="margin-top: 300vh" />

        <v-btn @click="goTo(-100)">Back to Top</v-btn>
      </v-main>
    </div>
  </v-layout>
</template>

<script setup>
  import { ref } from 'vue'
  import { useGoTo } from '../src/composables/goto.ts'

  const btn = ref(null)
  const container = ref(null)
  const goTo = useGoTo()
</script>
```
</details>

PR #19057, opened 2024-01-17
improve typing

yuwu9145:
Improve typing for #19046

### DIFF
--- a/packages/vuetify/src/composables/goto.ts
+++ b/packages/vuetify/src/composables/goto.ts
@@ -1,6 +1,6 @@
 // Utilities
 import { inject } from 'vue'
-import { mergeDeep } from '@/util'
+import { mergeDeep, refElement } from '@/util'
 
 // Types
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/composables/goto.ts
```diff
@@ -43,15 +43,15 @@ function genDefaults () {
   }
 }
 
-function getTarget (el: any) {
+function getTarget (el: ComponentPublicInstance | HTMLElement | string | undefined) {
   if (typeof el === 'string') return document.querySelector<HTMLElement>(el)
   if (el instanceof HTMLElement) return el
-  if (el?.$el instanceof HTMLElement) return (el.$el) as HTMLElement
+  if (el && '$el' in el && el?.$el instanceof HTMLElement) return (el.$el) as HTMLElement
```
KaelWD, line 49: We have a refElement helper fn

yuwu9145: Thanks! Updated to use refElement

johnleider (approved):

Issue #5479, opened 2018-10-29
[Bug Report] When the return-object prop is set to false, it interferes with item-text on v-combobox component


rgb-one:
### Versions and Environment
**Vuetify:** 1.3.1
**Vue:** 2.5.17
**Browsers:** Firefox 63.0
**OS:** Linux x86_64

### Steps to reproduce
- Select something from the dropdown
- Instead of item-text (which is set to name), the id for the item shows up in the v-combobox text field. 

### Expected Behavior
When the return-object is set to false, the item-text displays its set value instead of the id.

### Actual Behavior
When the return-object prop is set to false on a v-combobox, the text-field 'item-text' apparently gets changed to the id of the item-value.

### Reproduction Link
[https://codesandbox.io/s/xov7v23wrz](https://codesandbox.io/s/xov7v23wrz)

KaelWD:
combobox won't work without return-object, there's no way to tell if a value is supposed to be user input or an object lookup. 

See #4345, #4423, #5358

rgb-one:
If whatever is typed into the `v-combobox` is not in the items array of objects, then return what was typed.

KaelWD:
What if I type '4'?

rgb-one:
4 doesn't match any of the item-text so it would be a new entry, so just return '4'

KaelWD:
But if I click "Item 4" it would set the model to 4, which is ambiguous

rgb-one:
Why doesn't it work like `v-autocomplete` in that respect? With `v-autocomplete` the `v-model` would use the item-value (id in this case) white the text-field would display the item-text.

rgb-one:
Any reply?

KaelWD:
autocomplete doesn't mix text and value in the same input. If the user enters the ID of an existing item it would be displayed as the text of that item instead of a separate value.

SasanFarrokh:
Any progress?

sevillaarvin:
I tried setting return-object to false, it did return the property (id) in item-value, but it doesn't use the item-text for its label. In the end, I adjusted my code so that it accounts for the return-object property set to true.

![image](https://user-images.githubusercontent.com/33614172/64922364-19260600-d801-11e9-8759-5cfec71e691a.png)

stevenspinoza:
In case anybody find the same issue, **use autocomplete**, which has the correct functionality.
Here a demo/link: [https://vuetifyjs.com/en/components/autocompletes#slots](url)

[...]

PR #19059, opened 2024-01-17
fix(VCombobox): default ReturnObject props to false 

shobu13:
## Description
ReturnObject props default to false for VAutocomplete and VSelect but not for VCombobox who essentially behave similar to previous two for this functionality.
So this default behaviour seems inconsistant.  
If intended, i'll be glad to have an explanation why.  
thanks for taking the time of reading this PR and have a nice day.


johnleider:
This is because Combobox is the only select that accepts custom values not present in its items.

KaelWD:
#5479

This change also does nothing, you only touched the types.

PR #19066, opened 2024-01-17
feat(VOtpInput): add number password type to component

### LINKED ISSUE #18964: [Feature Request] "Number and password" type for v-otp-input
### Problem to solve
Besides for actual **one-time** pins, the OTP input can also be used to handle a user-chosen PIN that should remain a secret. In this situation, we would like to be able to use both the masked feature of `type="password"` and also the non-number rejection of `type="number"`. Right now, if the component is in "password" mode, it allows non-numeric characters to be entered.

### Proposed solution
I can think of three possible ways to handle this:

1. Add another type option named `numberPassword` that combines the features of the two existing types `number` and `password`.
2. Using the `password` type, also add a boolean property `numeric` that enforces using numbers.
3. Using the `number` type, also add a boolean property `masked` that chooses whether to show the text or not.

developer-diganta:
resolves #18964

## Description
As mentioned in the issue, I have added an attribute called numeric which can be used with type password. Once numeric is set to true, VOtpInput would not accept non-numeric values. 
## Markup:

```vue
eslint-disable vue/script-indent
<template>
  <v-app>
    <v-container>
      <!-- -->
      <v-otp-input v-model="otp" type="password" numeric/>
      {{ otp }}
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue';

  import { VOtpInput } from '../src/components/VOtpInput/index.ts'
  export default {
    name: 'Playground',
    components: {
      VOtpInput,
    },
    setup () {
      const otp = ref('')
      return {
        otp,
        //
      }
    },
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -47,6 +47,7 @@ export const makeVOtpInputProps = propsFactory({
     type: String as PropType<'text' | 'password' | 'number'>,
     default: 'number',
   },
+  numeric: Boolean,
 
   ...makeDimensionProps(),
   ...makeFocusProps(),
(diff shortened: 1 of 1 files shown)


J-Sek:
closing in favor of #20950

Issue #18454, opened 2023-10-12
[Bug Report][3.3.21] VAutocomplete selection slot not used while dropdown is open


MelvinFrohike:
### Environment
**Vuetify Version:** 3.3.21
**Vue Version:** 3.3.4
**Browsers:** Safari 17.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
When creating a VAutocomplete with objects as items and using the selection slot, the selection slot is only used when the automcomplete dropdown is not open. If the dropdown is open, the selection is displayed using the item-title prop.

### Expected Behavior
When the selection slot is defined it should be used whether the dropdown is open or not.

### Actual Behavior
The defined slot is used when the dropdown is closed. The item-title prop is used when the dropdown is open.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqdkjFP6zAQx7/KPb8hPKlJH5QBRQWJDRYWJBiaDia5VpYc27KdMET57pydtDh0Y3Jy97/7/e/s3cCcrdePxhR9h6xkW4+tkdzjQ6UAtn3OjYmf8afWynOh0M6hSdF5XWuqQo+nMECft7pBeV8xhxJrj80blx1W7FtSCoI5UsQzzYRA7oWXSFnF20VZTPahGSX5McmdXZGv0xzwd+ILrUg+xGoYK5ZoAYYpXlj+WQQcjCPkiyhxKJj0Xy8WdQENhWfeCozVxl1gaXlSOJ9HT33+IVRDNVFbMTjN/6PoN26D34SVWr6cI0jTOz0niLC4yzMiVCyfRuwRX84CQL+utsJ4auQ7A5KrI03oadxYJlqjrYcBLB5ghIPVLWT0MLNQSmnl0R54jfAcNjZM9LCCEl69Feq4mkI0fwkvXfuBNgTIZjjIovPQuiPcB8BV9oRSanjXVjZ/sn+pKL7IMmJ2e5LvQmrmzcDsIKzz2UyckLc3ERdjP9QOqXOzlN9tZncA+5S+3HI0uw1WHq6ir93/PZndrqdV0uLYuGKbYlPcsum8uWb7L7eLMXM=)

### Other comments
This makes for a confusing UI if the slot definition differs from the item-title. Note that this does not happen with the chip slot. There the slot definition is used whether the dropdown is open or not.

KaelWD:
Duplicate of #17573

Issue #17573, opened 2023-06-09
[Bug Report][3.3.2] v-autocomplete selection slot not shown while control has focus


drennenb:
### Environment
**Vuetify Version:** 3.3.2
**Vue Version:** 3.3.3
**Browsers:** Chrome 114.0.0.0
**OS:** Windows 10

### Steps to reproduce
Select a Building.  Then click whitespace.  Then select another building.  The selection slot is not shown when control has focus as it did in V2.

### Expected Behavior
Always show selection slot template while an item is selected.

### Actual Behavior
Selection slot template hidden while control has focus. When focused, item.title is shown instead.

### Reproduction Link
[https://stackblitz.com/edit/vuetify3-krppgl?file=README.md](https://stackblitz.com/edit/vuetify3-krppgl?file=README.md)

### Other comments
Not sure if this is different by design, but if it is, the part that makes this bad UX is the user must backspace out the current text, or must click the closeable button to enter a new search,

nemoneph:
We are facing the same issue here.
We have not found an alternative yet.

VicentSE:
Same here! 

I believe this was not always the case. As I understand this behaviour was introduced in **v3.2.2** https://vuetifyjs.com/en/getting-started/release-notes/?version=v3.2.2. I think in an attempt to fix other issues related to selected content overlapping the user input. In order to fix this `opacity: 0 ` was given to the selection when the input is focused.

Unfortunately, this change makes for a rather bumpy experience when using custom slots that involve images or a more complex layout since these will be completely hidden on focus to display only a text, and then this text will be replaced by the original selection slot content on blur.

But perhaps I am missing something, would be nice to know how to avoid this from happening since I am trying to migrate a project from v2 to v3🙂

glacierPEgli:
I am currently migrating from v2 to v3 and facing this issue as well. I also haven't found a workaround yet.

drennenb:
Im sorry if this kind of comment is unhelpful, but I wanted to re-iterate how important I think this is.  The autocomplete with templating is essentially broken, and is preventing the deployment of several updated apps we work on, and I see others have commented the same.  I am wondering if I need to find a different control.  I appreciate this is a free product, we want so badly to use it because its great.

BrahimMasmoudi:
hello, can you try adding :key="selectedBuilding?.name" to the v-autocomplete? 
that's the workaround i did in my project so that the selection slot works when i click on item in the autocomplete list.
i hope that helps

[...]

Issue #17291, opened 2023-05-04
[Bug Report][3.2.2] VAutocomplete has problem on rendering slot:selection when muliple property is falsy.


velly:
### Environment
**Vuetify Version:** 3.2.2
**Vue Version:** 3.2.47
**Browsers:** Chrome 112.0.0.0
**OS:** Windows 10

### Steps to reproduce
Switch vuetify from 3.1.6 to 3.2.2

### Expected Behavior
Selection slot has nothing to do with multiple property

### Actual Behavior
Slot template disappers when binding multiple to false;
The @update:model-value will emit twice and the final value becomes null when blur.(has something to do with wrong slot rendering)

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrdVD1v2zAQ/SssF7eAJQFugQKCEyTt0qVThwxmBls622wpkiBPagJB/71HSkoluQVSdAgQLzbfvXvvPkjvWn5rbdrUwHO+Rais2iNcC83Ytkn21saf8VDtpR5OfbBGUxhKAIQRHj/ENiWoK8GPToIuBV8ycklmnggWDElcEgqjjKP4QdWQnBw8Jkqezgg62VyS1f4Q3b6BggLZncSzqZF9rRXKP6g/tUGNjD1TzV4ZzH2UkEaTXMtClWsmdQkPrBN8mljKZnJkDOEB8ymw9Xavr9s2aKQoUUHXbbMITmSyqc42m62gh+aj/p37jzvwL7CEv26AvKshRBno6te4owA/vZrIie9pJkBHXzhpkXnAOoRlZY1D1jIHR9axozMVW9EDXQUVoQujPRIaVsquAuntuznql3C/XUJ3gt8KvmaCf+q/Pgt+P7IqfxoSV19AKcPujFPlmxXJ0ERijVQdX/O+QOrMpt+90fTH0cbShgDds5xFJGBUeDgLfka0Ps+yotSURhdTNi7VgJm2VXZDtMzVGmUFSWmqm/fpJv3wkebucYqn4Kvk4MxPD45UqImJD8rj4zO8Bma02PQOA5TQDfbB4kI6I78GXOJovODAPbelRdqsrUXsorVg3wnd0cSV1D/8YtiFj4PejVX+V89RLCjRbeh4t+Z9oWQdM/j9L1cv6Aw=)

KaelWD:
Caused by #17100

[...]

PR #19104, opened 2024-01-25
fix:(VAutoComplete/VCombobox): show selection slot when input has focus and multiple prop is false

devCodeHub-star:
**Description**
Because of #17294 on focus, the search value updated to item tile and added to input field. Selection slot has opacity: 0 and input tag has position: absolute, due that input overlap the selection slot and showing search value which cause hiding selection slot style.

**Open Issue**
VAutoComplete/VCombobox doesn't use formatted style from selection slot when multiple prop is false and dropdown open
fixes: #18454, #17573, #17291

**Markup**
```
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        :items="items"
        item-title="name"
        item-value="age"
        label="Show selection"
        :show-selection-slot="true"
      >
        <template #selection="{ item }">
          <VChip>{{ item.raw.name }} - {{ item.raw.age }}</VChip>
        </template>
        <template #item="{ item, props }">
          <v-list-item v-bind="props" title="">
            {{ item.raw.name }} - {{ item.raw.age }}
          </v-list-item>
        </template>
      </v-autocomplete>

      <v-autocomplete
        :items="items"
        item-title="name"
        item-value="age"
        label="Show chip without selection slot"
        chips
        :show-selection-slot="true"
      >
        <template #item="{ item, props }">
          <v-list-item v-bind="props" title="">
            {{ item.raw.name }} - {{ item.raw.age }}
          </v-list-item>
        </template>
      </v-autocomplete>

      <v-combobox
        chips
        label="Combobox"
        :show-selection-slot="true"
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'
  const items = [
    {
      name: 'first',
      age: 42,
    },
    {
      name: 'second',
      age: 83,
    },
  ]
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -78,6 +78,10 @@ export const makeVAutocompleteProps = propsFactory({
     type: [Boolean, String] as PropType<boolean | 'exact'>,
   },
   clearOnSelect: Boolean,
+  showSelectionSlot: {
+    type: Boolean,
+    default: false,
+  },
   search: String,
 
   ...makeFilterProps({ filterKeys: ['title'] }),
@@ -243,7 +247,7 @@ export const VAutocomplete = genericComponent<new <
         listRef.value?.focus('next')
       }
 
-      if (!props.multiple) return
+      if (!props.multiple && !props.showSelectionSlot) return
 
       if (['Backspace', 'Delete'].includes(e.key)) {
         if (selectionIndex.value < 0) {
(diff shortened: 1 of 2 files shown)


yuwu9145:
The new `show-selection` seems to be equivalent to `props.chips || slots.chip || slots.selection`. Can you explore if it's possible to internally convert them into a computed variable, as opposed to introducing a new prop?

The general strategy is input should stop displaying as absolute (overlap whole field) when either selection slot or chip is used. So it can display next to the selection

yuwu9145:
Issue has been fixed in https://github.com/vuetifyjs/vuetify/commit/1e574538282e245142571e9bdc84e342ad006d26

PR #19136, opened 2024-01-31
fix(Tooltip): set width attr,text too length,but text not wrap

wz930206:
Tooltip set width，text too length, but text not wrap

## Description

image

## Markup:

```vue

```


oracast:
Please provide markup

johnleider (approved):

PR #19141, opened 2024-01-31
feat(VDatePicker): add isBeforeYear check

Fo00oX:
## Description
Add missing IUtil interface function: isBeforeYear

## Markup:

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      {{ adapter.date() }}
    </div>
  </v-app>
</template>

<script setup>
  import { useDate } from '../src/composables/date'

  const adapter = useDate()

  console.log(adapter)
</script>
```


KaelWD:
The built-in adapter is only supposed to support what is needed for our calendar components, we might as well just pull all in of date-fns otherwise.

johnleider:
@blalan Of the new implementations, which would be useful to you to keep?

blalan05:
Dealing with Events, I'd say isAfterDay is the only one I might need, for those multi-day/allDay events.

johnleider (approved):

PR #19154, opened 2024-02-02
feat(VColorPicker): add hide-eye-dropper prop

### LINKED ISSUE #19150: [Feature Request] Ability to hide eye-dropper in `VColorPicker`
### Problem to solve
There is currently no native way to disable the eye-dropper in the color picker component.
I had to disable it for a project by using 
```css
.v-color-picker-preview__eye-dropper {
    display: none !important;
}
```

### Proposed solution
Add a `hideEyeDropper` prop on `VColorPicker` that will prevent it from being rendered.

I can make a PR for this if its wanted.

yimme:
## Description

resolves #19150 

### Why is this change required?
You currently can't disable the eye-dropper in v-color-picker if its enabled on your browser.
When you do want it hidden you have to write some custom css to force hide it.

### What problem does it solve?
The lack of hiding it using a prop on the component itself.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-color-picker hide-eye-dropper />
    </v-container>
  </v-app>
</template>

<script setup>
</script>

```

### DIFF
--- a/packages/api-generator/src/locale/en/VColorPicker.json
+++ b/packages/api-generator/src/locale/en/VColorPicker.json
@@ -4,6 +4,7 @@
     "dotSize": "Changes the size of the selection dot on the canvas.",
     "flat": "Removes elevation.",
     "hideCanvas": "Hides canvas.",
+    "hideEyeDropper": "Hides EyeDropper.",
     "hideSliders": "Hides sliders.",
     "hideInputs": "Hides inputs.",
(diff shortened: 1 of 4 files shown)


phoenix-oes:
Yes please. The eye dropper seems very buggy, and in my experience adds unexpected levels of alpha, even when the color-picker is in RGB mode.

PR #18629, opened 2023-11-08
fix(VSelect): add option role to VListItem elements

Tboodle:
# Description
This change fixes an accessibility issue where the items in a VSelect dropdown can not be navigated between by using the arrow keys when [NVDA](https://www.nvaccess.org/) is running. The VList element has a role of "listbox," and according to the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role), each child element should have a role of "option." After adding this role, the VSelect items can  be navigated between using the arrow keys while NVDA is open.

Note: You need to have NVDA running to test these changes out in the playground.

fixes #17767 
fixes #16881

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-select :items="items" label="Select an Item" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
    data: () => ({
      items: ['Apple', 'Orange', 'Pineapple']
    })
  }
</script>

```

PR #19199, opened 2024-02-08
fix(VAutocomplete): fix screenreader navigation

### LINKED ISSUE #18858: [Bug Report][3.4.2] NVDA prevents selecting options in autocomplete
### Environment
**Vuetify Version:** 3.4.2
**Vue Version:** 3.3.11
**Browsers:** Google Chrome
**OS:** Windows

### Steps to reproduce
1. Open the NVDA tool with the speech viewer.
2. Focus on the autocomplete using the tab key and open it by pressing the down arrow key. 
3. Try to select the different items in the autocomplete only using the arrow keys.
4. Check what the speech viewer from NVDA reads.

### Expected Behavior
Be able to select and read each of the items in the autocomplete by moving using the arrow keys in the keyboard.

### Actual Behavior
It reads the first one, and pressing on the arrow keys don't move it to any other option. The speech viewer also reads stuff outside of the autocomplete.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNq9ks1qwzAQhF9l2R58iW3S9BTcQDC0L1DoIc5BsbZBoGiFLJuWkHevJDdxTC899TYavtWMfnZn7Fxbbq0thp5wjZWnk9XC06YxANWQC2uTTIuWjRfKkPuxRqL33HKYIk9XG0CLA+nnBuveOTIeZhRO3FqFxC6Au6wWWn2wM0pkC8hq1uyE5KhfglQy2a/E7jgSb/Qpuijev/ikzDHbTxtvqnLeLDT+c+etlCTBsSbwDKngf1SeEm6vAA8xK0SdU40FWMe2g0uDd3Q6kVadzyMDQ35QRoaZxDaYDhKWbL1iEyYh3s2Nv48tZ68/Wr+u8WrPP0MC01+Z7YKXBa6KVbFcYhRPxSPuvwFHwcJC)

### Other comments
My form of solution was adding the `role="option"` to the `v-list-item` (like the second autocomplete in the playground) [based on this pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/#wai-ariaroles,states,andproperties).

Martin-Idel:
Screen readers need the options role in order to successfully redirect arrow navigation to the options. Without this change, keyboard navigation is broken when having a screen reader on.

fixes #18858

## Description

This is the same fix as https://github.com/vuetifyjs/vuetify/pull/18629 but now for VAutocomplete instead of VSelect.
You need to have a screen reader running to test these changes, as the screen reader changes keyboard navigation!

Tested with Orca and JAWS.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete :items="items" label="Select an Item" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
    data: () => ({
      items: ['Apple', 'Orange', 'Pineapple']
    })
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -502,7 +502,7 @@ export const VAutocomplete = genericComponent<new <
                             index,
                             props: itemProps,
                           }) ?? (
-                            <VListItem { ...itemProps }>
+                            <VListItem { ...itemProps } role="option">
                             {{
                               prepend: ({ isSelected }) => (
                                 <>
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -527,7 +527,7 @@ export const VCombobox = genericComponent<new <
                             index,
                             props: itemProps,
                           }) ?? (
-                            <VListItem { ...itemProps }>
+                            <VListItem { ...itemProps } role="option">
                             {{
                               prepend: ({ isSelected }) => (
                                 <>


johnleider:
Would this apply to VSelect or VCombobox?

Martin-Idel:
https://github.com/vuetifyjs/vuetify/pull/18629 did the exact same for VSelect, so it applies but is already done.

I missed VCombobox. I checked and yes this should also have it. I updated the PR with a second commit for this.

johnleider (approved):

PR #19324, opened 2024-03-04
fix(VSwitch): remove elevation with flat props

### LINKED ISSUE #19307: [Bug Report][3.5.6] VSwitch - flat prop doesn't remove elevation
### Environment
**Vuetify Version:** 3.5.6
**Vue Version:** 3.4.21
**Browsers:** Chrome 122.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Add `flat` prop into v-switch as described in https://vuetifyjs.com/en/components/switches/#props-flat

### Expected Behavior
Elevation removed

### Actual Behavior
Elevation presents

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqLrlYqLkrWdywo0CsrTVWyUrIpSc0tyEksSbWLyVNQsCnTTSwoADPBnOT8vJLEzLzUIqgQWLC4PLMkOUNBH4tYGtAkuISNProBIBGI+Tb6SPYCucXJRZkFJQrFqSWlBXY2+hAuUEqpVkfJWM9Ez8hQCcQw1TNTigUAA189gw==)

SonTT19:
## Description

fixes #19307 

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-switch />
      <v-switch flat />
    </v-container>
  </v-app>
</template>

<script setup></script>
```

### DIFF
--- a/packages/vuetify/src/components/VSwitch/VSwitch.sass
+++ b/packages/vuetify/src/components/VSwitch/VSwitch.sass
@@ -7,10 +7,6 @@
   .v-label
     padding-inline-start: $switch-label-margin-inline-start
 
-  .v-switch__thumb
-    background-color: $switch-thumb-background
-    color: $switch-thumb-color
-
 .v-switch__loader
   display: flex
 
@@ -58,6 +54,8 @@
 
 .v-switch__thumb
   align-items: center
+  background-color: $switch-thumb-background
+  color: $switch-thumb-color
   border-radius: $switch-thumb-radius
   display: flex
   font-size: .75rem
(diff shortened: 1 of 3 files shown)


yuwu9145 (changes requested):
inset needs elevation to be 0 as well, can you try to combine flat "elevation 0" logic with inset together?

Something like

```
&:not(.v-switch--flat):not(.v-switch--inset)
    @include tools.elevation($switch-thumb-elevation)
```

SonTT19:
> inset needs elevation to be 0 as well, can you try to combine flat "elevation 0" logic with inset together?
> 
> Something like
> 
> ```
> &:not(.v-switch--flat):not(.v-switch--inset)
>     @include tools.elevation($switch-thumb-elevation)
> ```

This is not working because `.v-switch--inset` is in parent DOM, not for `.v-switch__thumb`
Screenshot 2024-03-04 at 13 55 59

And we already have the reset elevation for inset
image
https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VSwitch/VSwitch.sass#L79

yuwu9145:
@SonTT19 https://github.com/vuetifyjs/vuetify/issues/19307#issuecomment-1975862946

johnleider (approved):

johnleider:
I made an update to how the colors are applied, and added surface-variant as default when using flat, which looks great.

PR #19378, opened 2024-03-12
fix(VDatePicker): fix range selection crossing DST

### LINKED ISSUE #19313: [Bug Report][3.5.6] vDatePicker component: Range selection is dropping a value
### Environment
**Vuetify Version:** 3.5.6
**Vue Version:** 3.4.19
**Browsers:** Firefox 123.0
**OS:** Windows 10

### Steps to reproduce
Today is March 1st. This bug is currently only happening when the month of March is selected. The year does not  seem to matter as long as it's March.

Use a vDatePicker in a template. Set the multiple prop to 'range'. Select a range that spans from a full calendar row (7 days) to another. 

### Expected Behavior
All dates are selected.

### Actual Behavior
The next-to-last date in the selection array is being dropped and does not display as selected on the calendar.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpVjlsKwjAQRbcS5tumiA9QouA6jB8hHSWaJmGaVIq4d8eqUP8uh/s6PqAjWx9Skn1B2ILK2CZvMu51EEL1lY0hGxeQRjAiindxLV1252GnoUvGYmUoltBo+LpGX8M1VXL2hiRs9JHYnci1hgYNoi0+u+SRIZlwQc6q+i/0W2TKk59DrKePVD35C88ZLORSzjfwFiu5htMLiflJqw==)

### Other comments
It appears that this bug may be an interaction between the current date and how that effects selecting a range? Maybe because this February had 29 days?

alexkunin:
## Description

fixes #19313

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      {{ data }}
      <v-date-picker v-model="data" multiple="range"></v-date-picker>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      return {
        data: ref(null)
      }
    },
  }   
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
@@ -79,15 +79,11 @@ export const VDatePickerMonth = genericComponent<VDatePickerMonthSlots>()({
           rangeStop.value = _value
         }
 
-        const diff = adapter.getDiff(rangeStop.value, rangeStart.value, 'days')
-        const datesInRange = [rangeStart.value]
+        const datesInRange = []
 
-        for (let i = 1; i < diff; i++) {
-          const nextDate = adapter.addDays(rangeStart.value, i)
-          datesInRange.push(nextDate)
+        for (let d = rangeStart.value; d <= rangeStop.value; d = adapter.addDays(d, 1)) {
+          datesInRange.push(d)
         }
-
-        datesInRange.push(rangeStop.value)
 
         model.value = datesInRange
       } else {
--- a/packages/vuetify/src/components/VDatePicker/__tests__/VDatePicker.spec.cy.tsx
+++ b/packages/vuetify/src/components/VDatePicker/__tests__/VDatePicker.spec.cy.tsx
@@ -21,4 +21,22 @@ describe('VDatePicker', () => {
         expect(model.value).to.have.length(11)
       })
   })
+
+  it('selects a range of dates crossing DST boundary', () => {
+    const model = ref<unknown[]>([])
+    cy.mount(() => (
+      <Application>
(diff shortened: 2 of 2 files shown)


KaelWD (changes requested):
This has to be compatible with https://github.com/dmtrKovalenko/date-io/blob/master/packages/core/IUtils.d.ts#L122

alexkunin:
Fair point.

So, `Unit` should not be changed as it will change public API.

Using ad-hoc date calculations in the component it not an option too, as it does not allow to use adapter of choice in cases when non-built-in adapter is provided.

Possible solutions:

1. fix `getDiff` to use `Math.round` (still not sure if this should be done for all unit, or only for `'days'`)
    a) actually, this might not be the correct fix because its correctness depends on semantics: does `'days'` mean calendar days, or just 24 hours periods? tbf it feels like the latter is the case, and thus `getDiff` does not need 'fixing'

2. or change the way `VDatePicker` works with dates, i.e. maybe use UTC
    a) we can use UTC everywhere in the component, which may lead to many imcompatibilities
    b) or, we can change that single place where `getDiff` is used -- this should fix the bug and introduce the least amount of very localized possible incompatibilities

I think I'll stick with 2b.

alexkunin:
The solution turned out to be even simpler: don't use `getDiff` at all, just iterate over range of dates.
```ts
const datesInRange = []
  
for (let d = rangeStart.value; d <= rangeStop.value; d = adapter.addDays(d, 1)) {
  datesInRange.push(d)
}
  
model.value = datesInRange
```
So, instead of counting days (incorrectly) and then iterating over that count, I just iterate from starting date, add 1 day on each iteration, stop when target date is reached.

johnleider:
I fixed this in your other P.R.

PR #19386, opened 2024-03-12
fix: make router useLink link reactive

### LINKED ISSUE #19300: [Bug Report][3.5.6] v-list-item initialized with an empty "to" property stays without an href
### Environment
**Vuetify Version:** 3.5.6
**Vue Version:** 3.4.21
**Browsers:** Chrome 122.0.0.0
**OS:** Linux x86_64

### Steps to reproduce
When using a nav v-list-item, if it is initialized with an empty "to" prop, changing its value doesn't add the associated href on the element a.v-list-item--link.

To reproduce in the repro link: 
- Try to click on "bar" (element has no pointer cursor here)
- Click on the "Toggle to" button 
- Try to click on "bar" (element now has a pointer cursor)
- Check the a.v-list-item html, no href attribute

### Expected Behavior
The href gets added to the anchor a.v-list-item--link

### Actual Behavior
The href isn't added to the anchor

### Reproduction Link
[https://jsfiddle.net/edx3rac5/](https://jsfiddle.net/edx3rac5/)

### Other comments
I did the same test using only the vue-router and it "works" as expected: https://jsfiddle.net/b9782hej/, although with vue-router the anchor starts with an href attribute even if the "to" value is an empty string. 

Note that the exact behavior I'm looking for is exactly what vuetify does (non clickble link if no "to" value), the only problem is the href not being added.

Also, if the v-list-item is first created with a valid "to" value (in the repro link, set the initial "showTo" value to "true"), and then the "to" value is removed, the behavior is correct (could be a workaround).

Antti-Palola:
## Description
If for example VCard to prop is evaluated as undefined and then changed, the changes were not be propagated to the href.

fixes #19300

Card link after enabling link (from undefined to page2)

| Before | After |
| -- | -- |
|  router-before | router-after |

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      Click button to toggle link state on list item
      <v-breadcrumbs active-color="primary">
        <v-breadcrumbs-item active>Hello</v-breadcrumbs-item>
        <v-breadcrumbs-divider>/</v-breadcrumbs-divider>
        <v-breadcrumbs-item to="/page2">Page 2</v-breadcrumbs-item>
      </v-breadcrumbs>
      <v-btn @click="setLink">Set link</v-btn>
      <v-list>
        <v-list-item :to="listLink" color="green">{{ `Link to ${listLink?.name}` }}</v-list-item>
        <!-- <v-list-item :to="{name: 'page1'}">Working link</v-list-item> -->
      </v-list>
      <v-card :to="listLink">
        <v-card-title>Linked card</v-card-title>
        <v-card-text>{{ `Link to ${listLink?.name}` }}</v-card-text>
      </v-card>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'
  export default {
    name: 'Playground',

    setup () {
      const listLink = ref(undefined)
      function setLink () {
        if (listLink.value) {
          listLink.value = undefined
        } else {
          listLink.value = { name: 'page2' }
        }
      }
      return {
        listLink,
        setLink,
      }
    },
  }
</script>


```

### DIFF
--- a/packages/vuetify/src/composables/router.tsx
+++ b/packages/vuetify/src/composables/router.tsx
@@ -64,22 +64,27 @@ export function useLink (props: LinkProps & LinkListeners, attrs: SetupContext['
       href: toRef(props, 'href'),
     }
   }
+  // vue-router useLink `to` prop needs to be reactive and useLink will crash if undefined
+  const linkProps = computed(() => ({ ...props, to: props.to ? props.to : {} }))
 
-  const link = props.to ? RouterLink.useLink(props as UseLinkOptions) : undefined
+  const routerLink = RouterLink.useLink(linkProps.value as UseLinkOptions)
+  // Actual link needs to be undefined when to prop is not used
+  const link = computed(() => props.to ? routerLink : undefined)
   const route = useRoute()
 
   return {
     isLink,
     isClickable,
-    route: link?.route,
-    navigate: link?.navigate,
-    isActive: link && computed(() => {
-      if (!props.exact) return link.isActive?.value
-      if (!route.value) return link.isExactActive?.value
+    route: link.value?.route,
+    navigate: link.value?.navigate,
+    isActive: computed(() => {
+      if (!link.value) return false
+      if (!props.exact) return link.value.isActive?.value ?? false
+      if (!route.value) return link.value.isExactActive?.value ?? false
 
-      return link.isExactActive?.value && deepEqual(link.route.value.query, route.value.query)
+      return link.value.isExactActive?.value && deepEqual(link.value.route.value.query, route.value.query)
     }),
-    href: computed(() => props.to ? link?.route.value.href : props.href),
+    href: computed(() => props.to ? link.value?.route.value.href : props.href),
   }
 }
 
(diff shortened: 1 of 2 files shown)


Antti-Palola:
The breadcrumbs cypress test is failing locally for me as well but I updated the Playground example to contain a breadcrumb too and it's working as it's supposed to.

packages/vuetify/src/composables/router.tsx
```diff
@@ -64,22 +64,28 @@ export function useLink (props: LinkProps & LinkListeners, attrs: SetupContext['
       href: toRef(props, 'href'),
     }
   }
+  // vue-router useLink `to` prop needs to be reactive and useLink will crash if undefined
+  const toPropRef = computed(() => props.to ? props.to : {})
```
johnleider, line 68: This would be fine inside of linkProps.

johnleider (changes requested):

packages/vuetify/src/composables/router.tsx
Antti-Palola: yeah, changed ✅

johnleider (approved):

PR #19408, opened 2024-03-15
fix(theme): add null check for current dark check

webdevnerdstuff:
## Description
Adding in a null check if `dark` is undefined

## Markup:

```vue
if (current.value?.dark) {
  createCssClass(lines, ':root', ['color-scheme: dark'])
}
```


KaelWD:
- Why would this be null?
- There's other usages of this in genCssVariables that are unchecked

PR #19509, opened 2024-03-28
docs: change `vue-i18n` url to point to v9 docs

BobbieGoede:
## Description
Updates the link to `vue-i18n`, currently it's pointing to the v8 docs, you probably want to point to the v9 docs instead.

## Markup:

```vue

```


userquin (approved):
thx Bobbie

johnleider (approved):

PR #19514, opened 2024-03-29
feat(router): add `NuxtLink` support for `useLink`

### LINKED ISSUE #17490: [Feature Request] Support nuxt prop for Vuetify 3 on Routable components
### Problem to solve
In Vuetify2 adding the `nuxt` to a routable component (e.g. `v-btn` or `v-card`) produced a `nuxt-link` if `to` prop or `href` prop was passed. In Vuetify 3 the `nuxt` prop does no longer exist.

### Proposed solution
This is the solution i would like to propose:
Without `nuxt` prop:
`<v-btn href="https://vuetifyjs.com"/>` => `<a href="https://vuetifyjs.com"/>`
`<v-btn to="/vuetify"/>` => `<router-link to="https://vuetifyjs.com"/>`
`<v-btn />` => `<v-btn />` (No `href` or `to`)

With `nuxt` prop:
`<v-btn nuxt href="https://vuetifyjs.com"/>` => `<nuxt-link href="https://vuetifyjs.com"/>`
`<v-btn nuxt to="/vuetify"/>` => `<nuxt-link to="https://vuetifyjs.com"/>`
`<v-btn nuxt />` => `<v-btn />`  (No `href` or `to`)

NOTE: `nuxt-link` can now handle internal (`to`) and external (`href`) links:  https://nuxt.com/docs/api/components/nuxt-link

userquin:
## Description

This PR allows use `useLink` with `NuxtLink`, this Nuxt PR should be merged https://github.com/nuxt/nuxt/pull/26522

This will work only when:
- NuxtLink is registed globally: Nuxt will resolve it on demand, it is not registered globally, check https://github.com/nuxt/nuxt/issues/13659
- NuxtLink is register with that name: Nuxt allows us to change the name

To register NuxtLink globally we can use a plugin:
```ts
import { NuxtLink } from '#components'

export default defineNuxtPlugin((nuxtApp) => {
  nuxtApp.vueApp.component('NuxtLink', NuxtLink)
})
```

`vuetify-nuxt-module` will provide an option to register NuxtLink globally: will allow also to use other name (will register both).

resolves #17490

### DIFF
--- a/packages/vuetify/src/composables/router.tsx
+++ b/packages/vuetify/src/composables/router.tsx
@@ -50,7 +50,13 @@ export interface UseLink extends Omit<Partial<ReturnType<typeof _useLink>>, 'hre
 }
 
 export function useLink (props: LinkProps & LinkListeners, attrs: SetupContext['attrs']): UseLink {
-  const RouterLink = resolveDynamicComponent('RouterLink') as typeof _RouterLink | string
+  // NuxtLink component name can be changed in nuxt config file, but we cannot access to that name
+  // NuxtLink is not registered globally, and so we cannot use resolveDynamicComponent, the user
+  // must register it globally (vuetify-nuxt-module will add an option to enable global registration)
+  let RouterLink = resolveDynamicComponent('NuxtLink') as typeof _RouterLink | string
+  if (typeof RouterLink === 'string' || !('useLink' in RouterLink)) {
+    RouterLink = resolveDynamicComponent('RouterLink') as typeof _RouterLink | string
+  }
 
   const isLink = computed(() => !!(props.href || props.to))
   const isClickable = computed(() => {


userquin:
Right now there is an alternative, but requires wrap the component, for example:
```html
<template>
  <NuxtLink to="/someroute" custom #="{ navigate }">
    <v-btn @click="navigate">Link</v-btn>
  </NuxtLink>
</template>

**Comment by KaelWD** [comment]
Router-link is available in nuxt right? How does using this instead behave differently?

**Comment by userquin** [comment]
> Router-link is available in nuxt right? How does using this instead behave differently?

Both are there, this PR should be changed to allow render `RouterLink` and `NuxtLink` (v2 behavior): https://nuxt.com/docs/api/components/nuxt-link

The linked issue explains the behavior...

**Comment by userquin** [comment]
Sorry for the confusion:

This PR was about adding `NuxtLink:useLink` support from Nuxt issue (the releated issue here was mentioned there, but it is not related): Nuxt issue is about adding `useLink`, Vuetify issue is about rendering `RouterLink` and/or `NuxtLink`.

**Comment by KaelWD** [comment]
That doesn't really answer my question. NuxtLink supports internal and external links but we already have that with to/href. Afaict the only other difference is prefetching but you haven't implemented that in useLink.

**Comment by franz-bendezu** [comment]
Hey everyone! 👋  

I was reviewing this PR and wanted to suggest an improvement to make it more flexible and configurable.  

Currently, the PR directly resolves `'NuxtLink'`, which could cause issues if the component name is changed in the Nuxt config or if it's not globally registered. Instead, we could use Vue’s `inject` API to dynamically resolve the component name, making it more adaptable.  

### **Proposed Approach:**  
Use `inject` to allow dynamic configuration, with `'RouterLink'` as the default fallback:  
```ts
const injectedRouterLinkName = inject<string>('vuetify-router-link', 'RouterLink')
const RouterLink = resolveDynamicComponent(injectedRouterLinkName) as typeof _RouterLink | string
```  

This approach also enables Nuxt users to configure the component name via a plugin while keeping it flexible for other use cases:  
```ts
export default defineNuxtPlugin((nuxtApp) => {
  nuxtApp.provide('vuetify-router-link', 'NuxtLink') // Customizable component name
})
```  

This would make the implementation more adaptable to different configurations and user preferences. 🚀

Issue #18942, opened 2023-12-28
[Bug Report][3.4.8] Bugs and problems in date picker in Solar and Lunar calendars (used for Persian/Farsi and Arabic locales)


mahozad:
### Environment
**Vuetify Version:** 3.4.8
**Vue Version:** 3.3.13
**Browsers:** Chrome 120.0.0.0
**OS:** Windows 10

### Steps to reproduce
Open the reproduction link and see the date picker in it (it is using Persian/Farsi locale which uses Solar Hijri calendar). Read on to see what problems exist in it.

### Expected Behavior

![Problems shown on an image](https://github.com/vuetifyjs/vuetify/assets/29678011/ef5b7c9f-f09b-414f-9cc5-bb3814adeaf6)

Many other related issues were closed and were referred to #1892.
The issue #1892 was closed and the authors requested a new issue for new versions of Vuetify.

These bugs and problems are for [Jalali/Solar/Persian calendar](https://en.wikipedia.org/wiki/Solar_Hijri_calendar) used for Persian/Farsi (`fa`) locale and also for Lunar calendar used for Arabic (`ar-SA`) locale.

1. The "year month" label should be concatenated in reverse order for `fa` locale (instead of `‏۱۴۰۲ دی` should be `دی ۱۴۰۲`)
2. The previous/next month buttons should be swapped (reversed) for **RTL** layouts (including `fa` locale)
3. The month labels in the month selection section are [off by one](https://en.wikipedia.org/wiki/Off-by-one_error) for `fa` locale (the default selected month is one less than what it should be and selecting a month selects the month after it)
4. The year numbers (labels) in the year selection section are off by one (one less than what they should be)

The following problems are less serious:

5. The title and header have not been localized (translated) for `fa` locale
6. The days in the picker do not start from `1` (they start from `9` or `10` or `11`) (because the first day of month in Solar Hijri calendar corresponds to `9`th or `10`th or `11`th day of a Gregorian month)
7. The months in the month selection section  do not start from `فروردین` (they start with `دی`)
(because the first month of Solar Hijri calendar corresponds to March in Gregorian calendar)

### Actual Behavior
It was hard to separate expected and actual functionality. Please see the expected section above.

### Reproduction Link
[...]

PR #19516, opened 2024-03-30
fix(VDatePicker): fixed missing adapter abstractions

SoheilHasankhani:
Fixed code to use adapter abstraction in text and header computed values

## Description
changed text and header computed values to use adapter abstraction while computing first day of month and header value.
resoves #18942

## Markup:
packages/vuetify/dev/vuetify/locale.js
```js
import { ar, en, fa, ja, sv } from 'vuetify/src/locale'

export default {
  locale: 'fa',
  messages: {
    en,
    ar,
    sv,
    ja,
    fa,
  },
}
```
packages/vuetify/dev/Playground.vue:
```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker
        v-model="date"
        show-adjacent-months
      />
      <!-- -->
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    data: () => ({
      date: new Date(),
    }),
  }
</script>

```

### DIFF
--- a/packages/vuetify/dev/vuetify/locale.js
+++ b/packages/vuetify/dev/vuetify/locale.js
@@ -1,10 +1,12 @@
-import { ar, en, ja, sv } from 'vuetify/src/locale'
+import { ar, en, fa, ja, sv } from 'vuetify/src/locale'
 
 export default {
+  locale: 'fa',
   messages: {
     en,
     ar,
     sv,
     ja,
+    fa,
   },
 }
--- a/packages/vuetify/dev/vuetify/locale.js
+++ b/packages/vuetify/dev/vuetify/locale.js
@@ -1,12 +1,10 @@
-import { ar, en, fa, ja, sv } from 'vuetify/src/locale'
+import { ar, en, ja, sv } from 'vuetify/src/locale'
 
 export default {
-  locale: 'fa',
   messages: {
     en,
     ar,
     sv,
     ja,
-    fa,
   },
 }
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
@@ -119,25 +119,22 @@ export const VDatePicker = genericComponent<new <
 
     const isReversing = shallowRef(false)
     const header = computed(() => {
-      return props.multiple && model.value.length > 1
-        ? t('$vuetify.datePicker.itemsSelected', model.value.length)
-        : model.value[0] && adapter.isValid(model.value[0])
-          ? adapter.format(
-            adapter.date(model.value[0]),
-            'normalDateWithWeekday')
-          : t(props.header)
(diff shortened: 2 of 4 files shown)


johnleider:
As a none native speaker, can you help me understand what your changes do for your locale? Does it effect other locales in any way?

SoheilHasankhani:
First and foremost, my changes will not affect other locales.

To describe the changes, I will explain them using an example. Let's consider today's date (2024-03-31) as the display date for the calendar. This date corresponds to (1403-01-12) in the Jalali calendar. In the Jalali calendar, the first day of the month is (1403-01-01), which is equivalent to (2024-03-20).

We have `year` and `month` properties which refer to the year value and month value according to the selected calendar:

```javascript
const month = ref(Number(props.month ?? adapter.getMonth(adapter.startOfMonth(internal.value))));
const year = ref(Number(props.year ?? adapter.getYear(adapter.startOfYear(adapter.setMonth(internal.value, month.value)))));
```

In our example, the `month` value would be 01 for the month and 1403 for the year.

For rendering the month selection box, the `text` computed value has been used. It computes the first day of the month and then passes it to `adapter.format` to format the date value to 'monthAndYear' format.

The current implementation creates a new Date object by passing a hardcoded value of 1 to the day value and the mentioned year and month values:

```javascript
new Date(year.value, month.value, 1)
// new Date(1403, 1, 1) ??? what is this!!!
```

Passing converted values for year, month, and day to the native Date object is incorrect and results in the wrong date. We should use the adapter functionality to compute the first day of the month properly in different calendars:

```javascript
const text = computed(() => {
  // Creating an adapter date object
  let _date = adapter.date();

  // Set the date according to the current picker values
  _date = adapter.setYear(_date, year.value);
  _date = adapter.setMonth(_date, month.value);
  _date = adapter.setDate(_date, 1);

  return adapter.format(
    _date,
    'monthAndYear',
  );
});
```

The same issue exists for the `header` computed value. With the idea of using the adapter, we should use the date object created by the adapter everywhere to ensure date operations work properly.

```javascript
const header = computed(() => {
  return props.multiple && model.value.length > 1
    ? t('$vuetify.datePicker.itemsSelected', model.value.length)
    : model.value[0] && adapter.isValid(model.value[0])
      ? adapter.format(model.value[0], 'normalDateWithWeekday')
      : t(props.header)
})
```

In the current implementation, `model.value[0]` is passed to the `adapter.format` method
while the adapter should receive a date object that is compatible with the adapter. So, I modified the code to accept an adapter-compatible date object.

```javascript
const header = computed(() => {
  return props.multiple && model.value.length > 1
    ? t('$vuetify.datePicker.itemsSelected', model.value.length)
    : model.value[0] && adapter.isValid(model.value[0])
      ? adapter.format(
        adapter.date(model.value[0]),
        'normalDateWithWeekday')
      : t(props.header);
});
```

johnleider (approved):

PR #19517, opened 2024-03-30
fix: no types for `vuetify/framework` sub-package

userquin:
## Description

This PR also moves `.esm.js` to `.mjs` files that can be interpreted as CJS when they are ESM:
- `dist/vuetify.esm.js` => `dist/vuetify.mjs`
- `dist/vuetify-labs.esm.js` => `dist/vuetify-labs.mjs`

Current framework subpackage (https://arethetypeswrong.github.io/?p=vuetify%403.5.13):

imagen

With this PR:

imagen

About the `esm.js` files you can check the warning here (or running `npx publint` in your local, install publint globally):  https://publint.dev/vuetify@3.5.13

### DIFF
--- a/packages/vuetify/build/rollup.config.mjs
+++ b/packages/vuetify/build/rollup.config.mjs
@@ -32,7 +32,7 @@ export default [
     input: 'src/entry-bundler.ts',
     output: [
       {
-        file: 'dist/vuetify.esm.js',
+        file: 'dist/vuetify.mjs',
         format: 'es',
         sourcemap: true,
         banner,
@@ -189,7 +189,7 @@ export default [
         banner,
       },
       {
-        file: 'dist/vuetify-labs.esm.js',
+        file: 'dist/vuetify-labs.mjs',
         format: 'es',
         sourcemap: true,
         banner,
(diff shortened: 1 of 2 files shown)


packages/vuetify/build/rollup.config.mjs
```diff
@@ -32,7 +32,7 @@ export default [
     input: 'src/entry-bundler.ts',
     output: [
       {
-        file: 'dist/vuetify.esm.js',
+        file: 'dist/vuetify.mjs',
```
KaelWD, line 35: This is a breaking change. The publint warning is meaningless, you would never need to import this.

userquin: ok, reverting those changes...

packages/vuetify/package.json
```diff
@@ -62,7 +62,10 @@
       "default": "./lib/styles/main.css"
     },
     "./styles/*": "./lib/styles/*",
-    "./framework": "./lib/framework.mjs",
+    "./framework": {
```
KaelWD, line 65: vuetify/framework is deprecated, it's the same as vuetify now

userquin: ok, closing the PR

PR #19536, opened 2024-04-03
fix(locale): add missing Lithuanian translations

jonasgadeikis:
This pull request adds missing translations for Lithuanian language.


KaelWD:
Doesn't look like @nesu is around to check this.

PR #19572, opened 2024-04-09
fix(VTextField): new-password of autocomplete doesn't behave correctly on chrome

### LINKED ISSUE #19559: [Bug Report][3.5.14] autocomplete="new-password" doesn't work well with chrome autofill when editing
### Environment
**Vuetify Version:** 3.5.14
**Vue Version:** 3.4.21
**Browsers:** Chrome 123.0.0.0
**OS:** Windows 10

### Steps to reproduce
```javascript
<VForm>
	<VTextField
		label="Password"
		name="password"
		autocomplete="new-password"
		type="password"
		v-model={form.password.value}
	/>
	<VTextField
		label="Confirm Password"
		name="confirmPassword"
		autocomplete="new-password"
		type="password"
		v-model={form.confirmPassword.value}
	/>
</VForm>
```

### Expected Behavior
With the form above, after using google's password suggestion, editing the first password field would subsequently update the second field without jump to the second field.

### Actual Behavior
Everytime we change the first field, focus jumps to the second one.

### Reproduction Link
[https://github.com/tongxuanbao/vue-project-example](https://github.com/tongxuanbao/vue-project-example)

Go to About route and check out the issue

tongxuanbao:
## Description

fixes #19559

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-responsive
        class="mx-auto"
        width="344"
      >
        <v-form>
          <v-label>Vuetify field</v-label>
          <v-text-field
            autocomplete="new-password"
            label="Password"
            name="password"
            type="password"
          />
          <v-text-field
            autocomplete="new-password"
            label="Confirm Password"
            name="confirm password"
            type="password"
          />
        </v-form>
        <!-- <form class="pure-form">
          <fieldset>
            <legend>Confirm password with HTML5</legend>

            <input id="password" placeholder="Password" type="password" required>
            <input id="confirm_password" placeholder="Confirm Password" type="password" required>

            <button class="mx-1" type="submit">Confirm</button>
          </fieldset>
        </form> -->
      </v-responsive>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.tsx
+++ b/packages/vuetify/src/components/VTextField/VTextField.tsx
@@ -108,10 +108,6 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
       props.active
     ))
     function onFocus () {
-      if (inputRef.value !== document.activeElement) {
-        inputRef.value?.focus()
-      }
-
       if (!isFocused.value) focus()
     }
     function onControlMousedown (e: MouseEvent) {


johnleider:
This seems like it would cause issues, why was the code you're removing put there in the first place?

tongxuanbao:
Ah i see, the code is being used to focus on the TextField when we click or clear. Let me know if reverting the focus event to the before of this PR [#14504 - feat(VSelect): port to v3](https://github.com/vuetifyjs/vuetify/pull/14504/files?file-filters%5B%5D=.tsx&show-deleted-files=true&show-viewed-files=true#diff-8414de8db55a5dd0885562c40495c4cc4a19e5662df81444f98992bb6af1def8) makes any sense.

What we're trying to do is to streamline our sign up page, and making password suggestion works better helps with that.

johnleider:
Is there a different way that we can approach it with causing a regression?

J-Sek:
@tongxuanbao, what do you even try to get in terms of Chrome functionality. AFAIK it stopped generating new passwords long time ago and the type seems to be supported only on Firefox ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#browser_compatibility)).

Maxim-Mazurok:
> AFAIK it stopped generating new passwords long time ago

Chrome still suggests new passwords, tried 127.0.6533.73 on Windows and 130.0.6723.117 on Mac, both have the same option for me:

image

J-Sek:
It's hard to argue with the screenshot :) You are right, but...
- it only works for people with active "sync" - only people logged in to their Google accounts on the browser level
- it seems to work regardless of your change (popup below was displayed in both cases)

Screenshot from 2024-11-18 10-15-13

sidenote: it is kinda weird Google uses `autocomplete="off"` for their signup password forms.

PR #19588, opened 2024-04-10
docs(Theme): theme.global.name/current shouldn't have .value

crazyrabbit0:
## Description
Removed `.value` from `theme.global.name` and `theme.global.current`,
because it broke the function `toggleTheme`.

### DIFF
--- a/packages/docs/src/pages/en/features/theme.md
+++ b/packages/docs/src/pages/en/features/theme.md
@@ -165,7 +165,7 @@ import { useTheme } from 'vuetify'
 const theme = useTheme()
 
 function toggleTheme () {
-  theme.global.name.value = theme.global.current.value.dark ? 'light' : 'dark'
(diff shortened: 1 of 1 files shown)


johnleider (approved):

carvenl:
After my testing, `theme.global.name = theme.global.current.dark ? 'light' : 'dark'` is ineffective, while `theme.global.name.value = theme.global.current.value.dark ? 'light' : 'dark'` is functional.

alihardan:
@johnleider, I believe this PR should reverted. the `.value` is necessary here (I have tested), because the example in docs written in Composition API and we interact with `ref`s here.

alihardan:
> Interesting, so `.value` is needed when in Composition API, but not in Options API ? 🤔

Yes, you have to write `.value` when accessing to `ref` reactive state only when you are in `script setup`.

crazyrabbit0:
That's new, good to know! 👌

PR #19609, opened 2024-04-15
feat(VTooltip): Implemented color prop

### LINKED ISSUE #19567: [Feature Request] Implement v-tooltip color prop
### Problem to solve
The ability to set the color of v-tooltips individually. Currently you can set the color globally with SASS variables, or another CSS-based solution. 

It is inconsistent with other components which support the color prop.
It was mentioned in #16690 that that issue should be closed and implementation should be requested instead.

### Proposed solution
Implement the color prop on v-tooltip.

sytexa-julia:
## Description

Resolves #19567 by implementing the `color` prop for VTooltip.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <template v-for="color in colors" :key="color">
        <div class="mb-2">
          <v-btn>
            {{ color.length ? color : 'Default' }}
            <v-tooltip
              v-model="show"
              :color="color"
              :text="`A ${color} tooltip!`"
              activator="parent"
              model-value
              @update:model-value=""
            />
          </v-btn>
        </div>
      </template>
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'
  export default {
    name: 'Playground',
    setup () {
      const colors = ref(['', 'primary', 'secondary', 'grey', 'red', 'grey-lighten-3'])
      return {
        colors,
      }
    },
  }
</script>
```


packages/vuetify/src/components/VTooltip/VTooltip.tsx
```diff
@@ -9,6 +9,7 @@ import { makeVOverlayProps } from '@/components/VOverlay/VOverlay'
 import { forwardRefs } from '@/composables/forwardRefs'
 import { useProxiedModel } from '@/composables/proxiedModel'
 import { useScopeId } from '@/composables/scopeId'
+import { makeVariantProps, useVariant } from '@/composables/variant'
```
johnleider, line 12: I don't think variant makes sense because there are at least 3 that come to mind that would be a bad experience, **plain**, **outlined**, and **text**.

johnleider (changes requested):

packages/vuetify/src/components/VTooltip/VTooltip.tsx
sytexa-julia: That is true. I'll look at revising this tomorrow.

PR #19629, opened 2024-04-17
feat: create VSnackbarQueue component

### LINKED ISSUE #2384: [Feature Request] Snackbar queue

KaelWD:
## Motivation and Context
closes #2384

## Markup:
<details>

```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <v-text-field v-model="field" label="Text" @keydown.enter="onKeydown" />
      <v-snackbar-queue v-model="model" color="success" location="bottom start" closable />
    </div>

    <v-fab icon="$vuetify" app />
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref([])
  const field = ref()

  function onKeydown () {
    model.value.push({
      text: field.value,
      timeout: -1,
      location: 'bottom',
    })
  }
</script>

```
</details>


johnleider (approved):

johnleider (approved):

Joebayld:
This looks quite interesting. What are your thoughts on an option to stack the notifications so you can see multiple at once?

Theres an example for Vuetify 2 under the notifications tab. 

https://iwdcf.csb.app/

blaskovicz:
Works great, thank you!
https://vuetifyjs.com/en/components/snackbar-queue/

PR #19656, opened 2024-04-20
fix(VFileInput): emit single file when multiple is false

yuwu9145:
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-file-input
        label="File input"
        v-model="msg"
      />
    </v-container>
  </v-app>
</template>

<script setup>

  const msg = ref()

  watch(msg, val => {
    console.log('isArray', Array.isArray(val))
    console.log(val)
  })
</script>


```


johnleider (approved):

svenjacobs:
Hello all,

I have to second that. This change was added to the minor version `3.5.17`. It broke our Web application because we were expecting an `Array` as a result.

PR #19689, opened 2024-04-24
feat(VTooltip): add `color` prop

### LINKED ISSUE #19567: [Feature Request] Implement v-tooltip color prop

julia-and:
resolves #19567

Using useColor directly instead of useVariant, as several variants would not be good user experience.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <template v-for="color in colors" :key="color">
        <div class="mb-2">
          <v-btn>
            {{ color.length ? color : 'Default' }}
            <v-tooltip
              v-model="show"
              :color="color"
              :text="`A ${color} tooltip!`"
              activator="parent"
              model-value
              @update:model-value=""
            />
          </v-btn>
        </div>
      </template>
    </v-container>
  </v-app>
</template>

<script>
import { ref } from 'vue'
export default {
  name: 'Playground',
  setup() {
    const colors = ref(['', 'primary', 'secondary', 'grey', 'red', 'grey-lighten-3', '#57A98B', '#CD1FB7'])
    return {
      colors,
    }
  },
}
</script>
```


sytexa-julia:
Sorry to create this as a new PR. Hopefully this is okay.

packages/vuetify/src/components/VTooltip/__tests__/VTooltip.spec.cy.tsx
```diff
@@ -0,0 +1,39 @@
+/// <reference types="../../../../types/cypress" />
+
+// Components
+import { VTooltip } from '..'
+
+// Utilities
+const colors = ['success', 'info', 'warning', 'error', 'invalid']
```
vanngoh, line 7: Besides of these, can it support hex code?

sytexa-julia: Yes, it works with hex codes too, see the markup example. I can add some hex codes to the test if you like.

J-Sek (approved):

PR #19725, opened 2024-04-30
chore(tests): fix flaky cypress tests

userquin:
There is no way, `VSlideGroup` cy test  cannot be fixed on CI ~~, but why is it failing if it is excluded (`it.skip`) ?~~.

I'll try to fix the `pointer-events` and CI issue...

### DIFF
--- a/packages/vuetify/src/components/VSlideGroup/__tests__/VSlideGroup.spec.cy.tsx
+++ b/packages/vuetify/src/components/VSlideGroup/__tests__/VSlideGroup.spec.cy.tsx
@@ -52,12 +52,9 @@ describe('VSlideGroup', () => {
         </CenteredGrid>
       </Application>
     ))
-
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VSlideGroup/__tests__/VSlideGroup.spec.cy.tsx
```diff
@@ -38,11 +38,12 @@ describe('VSlideGroup', () => {
...
   })
 
-  it('should disable affixes when appropriate', () => {
+  // TODO: fails in headless mode
+  it.skip('should disable affixes when appropriate', () => {
     cy.mount(() => (
       <Application>
         <CenteredGrid width="400px">
-          <VSlideGroup showArrows="always">
+          <VSlideGroup showArrows="always" className="test">
```
userquin, line 46: Rever this change

johnleider (approved):

PR #19568, opened 2024-04-09
feat(VCalendar): add emits and slots

blalan05:
```vue
<template>
  <v-app>
    <v-main>
      <v-container>
        <v-row>
          <v-col>
            <v-select v-model="view" :items="['month', 'week', 'day']" label="View" />
          </v-col>
        </v-row>
        <v-row>
          <v-col>
            <v-switch v-model="showHeaderSlot" label="Header Slot" />
          </v-col>
          <v-col>
            <v-switch v-model="showTitleSlot" label="Title Slot" />
          </v-col>
          <v-col v-if="view === 'month'">
            <v-switch v-model="showDayTitleSlot" label="Day Title Slot" />
          </v-col>
          <v-col v-if="view === 'month'">
            <v-switch v-model="showDayBodySlot" label="Day Body Slot" />
          </v-col>
          <v-col v-if="view === 'month'">
            <v-switch v-model="showDayEventSlot" label="Day Event Slot" />
          </v-col>
          <v-col v-if="view === 'week'">
            <v-switch v-model="showIntervalTitleSlot" label="Interval Title Slot" />
          </v-col>
        </v-row>
        <v-row>
          <v-col>
            <v-calendar
              v-model="thisDate"
              :events="events"
              :view-mode="view"
              @click:date.stop="clickedDate"
              @click:day.stop="clickedDay"
              @click:event.stop="clickedEvent"
              @click:interval-date.stop="clickedIntervalDate"
              @click:interval-event="clickIntervalEvent"
              @click:interval.stop="clickedInterval"
              @click:time.stop="clickedTime"
              @click:week-number.stop="clickedWeekNumber"
              @contextmenu:date.stop="rightClickedDate"
              @contextmenu:day.stop="rightClickedDay"
              @contextmenu:event.stop="rightClickedEvent"
              @contextmenu:interval-date.stop="rightClickedIntervalDate"
              @contextmenu:interval-event="rightClickedIntervalEvent"
              @contextmenu:interval.stop="rightClickedInterval"
              @contextmenu:time.stop="rightClickedTime"
              @contextmenu:week-number.stop="rightClickedWeekNumber"
            >
              <template v-if="showHeaderSlot" #header="{ title }">
                <span>{{ `Header ${title}` }}</span>
              </template>
              <template v-if="showTitleSlot" #title="{ title }">
                <span>Test {{ title }}</span>
              </template>
              <template v-if="showDayTitleSlot" #dayTitle="{ title }">
                {{ `Test ${title}` }}
[...]

PR #19728, opened 2024-05-01
feat(VCalendar): add day click events

musicchenko:
## Description
it would be nice to be able to track clicking on a day in the calendar (to add a new event i.e.)
I added events ```monthDayClick``` for ```VCalendarMonthDay```, ```dayClick``` for ```VCalendarDay``` and several events for ```VCalendar``` itself to handle all these clicks: ```click:monthDay```, ```click:weekDay```, ```click:day```

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-calendar
        view-mode="month"
        @click:day="click('Day', $event)"
        @click:month-day="click('Month', $event)"
        @click:week-day="click('Weekday', $event)"
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
function click (type: string, date: Date) {
  alert(`${type}: ${date}`)
}
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VCalendar/VCalendar.tsx
+++ b/packages/vuetify/src/labs/VCalendar/VCalendar.tsx
@@ -36,6 +36,9 @@ export const VCalendar = genericComponent<VCalendarSlots>()({
   emits: {
     next: null,
     prev: null,
+    'click:monthDay': (value: Date) => true,
+    'click:weekDay': (value: Date) => true,
+    'click:day': (value: Date) => true,
     'update:modelValue': null,
   },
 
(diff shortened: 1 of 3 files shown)


packages/vuetify/src/labs/VCalendar/VCalendar.tsx
```diff
@@ -36,6 +36,9 @@ export const VCalendar = genericComponent<VCalendarSlots>()({
   emits: {
     next: null,
     prev: null,
+    'click:monthDay': (value: Date) => true,
+    'click:weekDay': (value: Date) => true,
+    'click:day': (value: Date) => true,
```
userquin, line 41: I guess these events should use any since we're using Date adapter and not all adapters using Date (`onXXXClick` functions should be also changed: check VDatePicker.tsx)

In fact, `onClickToday` is wrong:
```ts
    function onClickToday () {
      model.value = [new Date()]
    }
```

should be:
```ts
    function onClickToday () {
      model.value = [adapter.date()]
    }
```

I need to check if this is working:

imagen

KaelWD:
Duplicate of #19568

JulienAramino:
Hello, could we have also click:time?

PR #19735, opened 2024-05-01
feat(VTable): add striped prop

### LINKED ISSUE #17505: [Feature Request] Tables striped and hoverable
### Problem to solve
I really miss the option to make ther rows striped and hoverable rows in the tables.

### Proposed solution
My proposed solution is please have an option to make the rows striped with a prop. In the striped prop we could add odd or even strings to activate the stripes depending on what the developer needs. 
For the hover state it will be enought if there would be a boolean hoverable prop.
It would be really awesome if the developer could change the theme color for these features, and the  feature should consider dark mode as well.

MajesticPotatoe:
## Description
Adds new prop called `striped` which accepts values `even` or `odd` and applies a background to applicable rows. Color is changeable via sass var `$table-stripe-color`. Prop passes styling through to `v-data-table` variants and works with `hover` prop.

Resolves #17505

## Markup:

```vue
<template>
  <v-table hover striped="even">
    <thead>
      <tr>
        <th class="text-left">
          Name
        </th>
        <th class="text-left">
          Calories
        </th>
      </tr>
    </thead>
    <tbody>
      <tr
        v-for="item in items"
        :key="item.id"
      >
        <td>{{ item.id }}</td>
        <td>{{ item.value }}</td>
      </tr>
    </tbody>
  </v-table>
  <v-data-table :items="items" striped="even"></v-data-table>
  <v-data-table-virtual :items="items" striped="odd"></v-data-table-virtual>
</template>

<script setup>
  const items = [
    { id: 1, value: '1' },
    { id: 2, value: '2' },
    { id: 3, value: '3' },
    { id: 4, value: '4' },
  ]
</script>
```


leebickmtu:
As this exists in v3.9.0, fixed columns are missing the striping. 

Not sure if that was intended or not, but looks a bit weird when there are multiple fixed columns that are all just white at the start of the table. Or when there are some non-fixed followed by a fixed. It's just a weird non stripped column hanging out there.

J-Sek:
> As this exists in v3.9.0, fixed columns are missing the striping.

Good catch, #21715

PR #19742, opened 2024-05-02
fix mobile header title font weight

webdevnerdstuff:
## Markup:

```vue

```


johnleider (approved):

rauhur:
Now both headers have the same origin for` font-weight`, but this is set to 500, which makes the headers no different from the cells.

Is it possible to change: `$table-header-font-weight: 500 !default; to $table-header-font-weight: 700 !default;` or is there a way to change by slots or properties of the header?

webdevnerdstuff:
I am not part of the Vuetify team, if you need help I suggest joining the Discord channel or StackOverflow and asking there. 

https://community.vuetifyjs.com/

Issue #18983, opened 2024-01-05
[Feature Request] Grid System Enhancement with Container Queries and Media Query Fallback


tlserver:
### Problem to solve
The Vuetify grid system, which currently uses "CSS media queries," is adept at creating responsive layouts by responding to changes in the viewport size. This conventional approach works well for general responsiveness across various devices. However, it encounters limitations in more dynamic web application layouts where the viewport is not the sole determinant of the layout's container size.

Web applications frequently feature a vertical navigation menu on the left side of the screen. This menu often includes a collapsibility feature to maximize the usable space of the main content area. When this menu is toggled, it significantly alters the width of the main content container without affecting the viewport size. The current grid system doesn't respond to these changes, leading to a suboptimal use of space and potentially a compromised user experience.

Furthermore, contemporary web applications sometimes require a more complex layout, such as a split-screen interface where the screen is divided into resizable panels or "windows." Users may wish to interact with these panels by adjusting their respective sizes to suit their workflow needs. With a grid system entirely reliant on viewport breakpoints, it cannot adapt to the resizing of these panels, as they are independent of the overall viewport dimensions.

In essence, while the Vuetify grid system excels in adapting to different device screens, it falls short in scenarios where the layout must respond to changes within the viewport itself. The lack of responsiveness to container size changes, not just viewport changes, can hinder the creation of highly interactive and adaptable web applications that modern users expect.

### Proposed solution
CSS has introduced [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries), which allow styling based on the size of the nearest container element, such as a `<div>`. This capability means that a grid system can now reference any DOM node and adjust its layout according to various breakpoints. As a result, grid systems have become significantly more versatile.

The majority of modern browsers have adopted this feature. As reported by [caniuse](https://caniuse.com/css-container-queries), over 90% of users were utilizing browsers that supported container queries as of January 2024. To accommodate the remaining 10% of users, Adrian Bece proposed [a fallback strategy](https://www.smashingmagazine.com/2021/05/css-container-queries-use-cases-migration-strategies/#migrating-from-media-queries-to-container-queries).

For backwards compatibility, grid systems can employ media queries as the default strategy while offering the option to switch to container queries. This can be done by setting a boolean flag in the configuration. This dual-strategy approach ensures that layouts remain functional and responsive across all browsers, providing a seamless experience while leveraging the benefits of container queries where available.

KaelWD:
[...]

PR #19759, opened 2024-05-06
feat(display): reduce default breakpoint sizes

KaelWD:
https://m3.material.io/foundations/layout/applying-layout/window-size-classes

-55px for edge's stupid sidebar
reduced xxl to 4k at 175% scaling


J-Sek:
I always though there is something off about those defaults.
Hopefully Google guidelines are backed by some real science.

My only concern is the impact on `$container-max-widths`:
- md: `900` » `787.5`
- lg: `1200` » `1073.4375`
- xl: `1800` » `1448.4375`
- xxl: `2400` » `2004.375`

With rounding down to nearest 100px we would get: `700`, `1000`, `1400`, `2000`

```scss
100 * math.floor(map.get($grid-breakpoints, 'md') * 0.009375)
// or with math.div() to make it more obvious:
// 100 * math.floor(math.div(map.get($grid-breakpoints, 'md') * 0.9375, 100))
```

KaelWD:
My biggest problem with the current ones is that they exactly match common screen resolutions (especially xl and xxl). Any sort of sidebar like edge's or firefox vertical tabs bumps you down a breakpoint for no real reason.

> $container-max-widths

Yeah that seems better actually. Many screen sizes will still get a wider container with that change eg. 1080p goes from 1200 to 1400px.

KaelWD:
We could actually use container queries now, any thoughts on these questions? https://github.com/vuetifyjs/vuetify/issues/18983#issuecomment-2287864177

J-Sek:
> _sidebar like edge's or firefox vertical tabs bumps you down a breakpoint_

I think that ~55px margin is going to be perfect to account for this.

johnleider (approved):

donalmurtagh:
@KaelWD Does this change mean that a Vuetify v3 responsive application will potentially have to re-test and modify every page's layout? 

Is there a way to revert to the v3 breakpoints, other than by overriding the `$grid-breakpoints` properties in `_variables.scss`?

J-Sek:
> _Does this change mean that a Vuetify v3 responsive application will potentially have to re-test and modify every page's layout?_

yes

> _Is there a way to revert to the v3 breakpoints, other than by overriding the `$grid-breakpoints` properties in `_variables.scss`?_

If you are not using Sass, you would need to append custom CSS for `.v-container`, `.hidden-*`utilities + a bit more to patch VTabs and VBottomSheet.

donalmurtagh:
> If you are not using Sass

@J-Sek If I am using Sass, is there an easier way to revert to the v3 behaviour?

KaelWD:
I've added instructions to the upgrade guide with changes needed to keep the v3 values.

PR #19771, opened 2024-05-07
fix(VMenu): min width property

### LINKED ISSUE #16987: [Bug Report][3.1.11] v-menu min-width property does not work
### Environment
**Vuetify Version:** 3.1.11
**Vue Version:** 3.2.47
**Browsers:** Chrome 111.0.0.0
**OS:** Linux x86_64

### Steps to reproduce
Click on the button to open the menu

### Expected Behavior
The menu should have 500px width because it has min-width="500"

### Actual Behavior
min-width property is not respected

### Reproduction Link
[https://codepen.io/kasheftin/pen/dyqQgJV?editors=1010](https://codepen.io/kasheftin/pen/dyqQgJV?editors=1010)

elkofy:
## Description

I switched the dimensionsStyle and contentStyle since the first one gets overrided by the second.Hence I provided a much simpler solution that what was proposed originally.
fixes #16987

## Markup:

```vue
<template>
  <v-app>
    <div class="text-center">
      <v-menu :min-width="500">
          <v-btn v-bind="props">
            Dropdown
          </v-btn>
        </template>
        <v-list>
          <v-list-item>
            <v-list-item-title>Here</v-list-item-title>
          </v-list-item>
        </v-list>
      </v-menu>
    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
</script>

```


johnleider:
When no min-width is supplied it's no longer calculated properly.

elkofy:
> When no min-width is supplied it's no longer calculates properly.

fixed I hope

johnleider:
Valiant effort but we're going to have to figure out a different way.

elkofy:
> Valiant effort but we're going to have to figure out a different way.

how about this method ?

johnleider:
I was thinking more along the lines of, if a value is undefined, dimensionStyles wouldn't even register the key.

johnleider (approved):

PR #19786, opened 2024-05-08
feat(VConfirm): add new directive

kieuminhcanh:
## Confirm
```vue
<template>
  <v-app>
    <div class="ma-4 pa-4">
      <VBtn
        prepend-icon="mdi-delete-outline"
        text="Delete"
        v-confirm="{
          text: `Are you sure?`,
          onSubmit: () => console.log(`Record has been deleted`),
          onCancel: () => console.log('Canceled'),
        }"
      />
    </div>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
</script>

```

## Confirm with input
```vue
<VBtn text="Change Username" 
    v-confirm="{ text: `Enter new name`, input, onSubmit}"/>

const input = ref('')
```


packages/vuetify/src/labs/VConfirm/VConfirm.tsx
```diff
@@ -0,0 +1,120 @@
...
+            lines="one"
+            v-slots={{
+              title: () => (
+                <VListItemTitle>
+                  { !showInput.value ? (
+                    props.text
+                  ) : (
+                    <VTextField
+                      v-model={ input.value }
+                      variant="solo-filled"
```
yuwu9145, line 76: Should we add/allow text-field props to VConfirm? So it's not hardcoded

yuwu9145: Oops, it has `inputProps`, but label is partially outside

Screenshot 2024-05-10 at 4 22 55 pm

J-Sek:
VList is unnecessary overhead. Other than this, the idea is nice, assuming we can have slots and support `rules` for validation. Possible revival in v5.x

PR #19819, opened 2024-05-15
feat(VField): support border prop

yuwu9145:
## Description

When border prop applied, VField actually uses outlined variant under the hood but override outline styles with border styles

May-16-2024 14-19-32

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        border="md"
        placeholder="test"
      />
      <v-text-field
        label="Label"
        border="success lg"
      />
      <v-text-field
        label="Label"
        variant="outlined"
      />
    </v-container>
  </v-app>
</template>
<script setup>
</script>



```

### DIFF
--- a/packages/vuetify/src/components/VField/VField.sass
+++ b/packages/vuetify/src/components/VField/VField.sass
@@ -325,12 +325,6 @@
     right: 0
     width: 100%
     
-    @include tools.border($field-border-prop...)
-
-    &.v-field__outline--border
-      --v-field-border-width: 0
-      --v-field-border-opacity: 0
-
 
     @media (hover: hover)
       .v-field:hover &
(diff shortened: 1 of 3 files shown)


johnleider:
Solo variants don't work with the border prop:

```html
<v-text-field
        placeholder="test"
        variant="underlined"
        border
      />
```

J-Sek:
I have a use case for detecting changed values (against a baseline object) in a big form. Currently my workaround requires custom class and a lengthy override. It would be much better to get this from the framework :)

```sass
.my-number-input--changed ::v-deep(.v-field:not(.v-field--error))
  .v-field__outline__start,
  .v-field__outline__notch::before,
  .v-field__outline__notch::after,
  .v-field__outline__end
    --v-field-border-opacity: 1
    border-color: orange !important
```

PR #19828, opened 2024-05-16
fix(VDivider): visiblity in container with dynamic height

### LINKED ISSUE #19827: [Bug Report][3.6.6] VDivider vertical not visible due to height `100%` resulting in `0px`
### Environment
**Vuetify Version:** 3.6.6
**Vue Version:** 3.4.27
**OS:** Linux x86_64

### Steps to reproduce
1. create flexbox container where height is controlled by it's children
2. use `<v-divider vertical>` inside flexbox container

### Expected Behavior
Divider should be visible and take 100% of containers height.

### Actual Behavior
Divider is not visible, because `height: 100%` results in `0px` anyway unless the container has explicit height.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqdVM1u2zAMfhXOKJAOqKp0PxcvLbbbdhkGrNhlHlDVlmOtsixItJsi8LuPkpvUP9lh88E2yY8URX7kz33iXc67VqIqny5/+yRNVG0bh7CH3EmB8sdggx5K19SwesauMpMZuYvQvDEe4dkA11PH831mALCStUwh/gMUshStxttBuSqEe1hdDCbOw7cnqX+dmeQi5vfJ2ksKT8ltUNZWU/SbANt0TFgbf6NAiaBQRrpnFSkL1UGuhffXWVKwUstdlhytZBdeFfKIsI69yxLw+KQliZVU2wpTuFqvu+oDPKoCqxTertd2GiUerhUV4RDofsvQCeOtcNLgDEzwwzWgY2XjyEGBMvCezk4f5FOQFz7xELoO5evITZXR6wauCLnhR9Mpr5AaU3Tm3AbHhF3TmkIWTG+zZIn6mGuVPxDMS1l8c7JUO+oz5biEpiJH1YXqjbHXJ9GLXKfZMlSo5c1X0UEQYb9XfR+uOkcsrjzGzEvPJxSau4y5wSM5SDPSvWIMbj9/+Q6HTigPpiH2K6/utSSOYNW0CKIoFKrGCD2wyQNj49ijVkqHKj/gRqwTLTZ/7W1MhP3PM01kFLGm0XmZhF2YhDnHXfM4L2eYOk0rQAevMDz/yOhcuOIEi1QttqEYdxWi9SnnVuW+rS9t1WDjeeAWP9u/UKzvzogd/M16zWk+704xM44vhSTMKfNQdrKT/wmqhjaEXBd0igVYcGlaqQ0PxX3ZSpwaelhbMcB4bwXNsNYmXCXR505ZBC+xHbbecVdTDcYbOm5nGu64mCczS9/zK9qsGz4EozBJf2FarePr1x+rbbyJ)

> Remove `style="height: auto"` to make `<v-divider horizontal>` disappear

### Other comments
I think vertical VDivider's height should be `auto` so that `align-self: stretch` applies correctly. The change does not seem to have any negative effects, however I consider this variant to always be used in the flexbox container.

J-Sek:
## Description

Assuming `<v-divider vertical>` makes sense only inside flexbox container, the `height: auto` allows `align-self: stretch` to take effect when container does not have explicit height.

Example in [docs](https://vuetifyjs.com/en/components/dividers/#usage) does not expose the issue, because `d-flex` gets additional explicit height, but in real life, we usually leave height to be dynamic and determined based on the children height.

After the change I did not notice any negative effects, but I may not know about use cases outside flexbox.

fixes #19827

### DIFF
--- a/packages/vuetify/src/components/VDivider/VDivider.sass
+++ b/packages/vuetify/src/components/VDivider/VDivider.sass
@@ -16,7 +16,7 @@
       align-self: stretch
       border-width: $divider-vertical-border-width
       display: inline-flex
-      height: 100%
+      height: auto
       margin-left: $divider-vertical-margin-left
       max-height: 100%
       max-width: 0px


johnleider:
This ended up breaking vertical dividers that use "align-self-center" within contains that have a specified height. https://play.vuetifyjs.com/#eNptUMFOwzAM/RUTCRUkSjlwQt0kbvwBB8IhpG6J5CaR4/Yy7d+Xpd3USbtE7z3H79n+OajEtvmM8XWeUH2oVnCMZAT32gO0c21iLPBC6j/Dq1CkFI1F3rfNFWq/KXdudh0yzMjirCEg9IP877R6f3vUCiyZlDIz5AZfJ6S+tugFWaviufbfmt7LLKUsbEdcecZts9kr02TZRYGEMi3ruTEGFjgAYw9H6DmMUOWTVEuwDT4JjGmA3fnHU/WFRAG+A1P3UD2fAxbLbKaOL34iKs/vCWOPdNg=

J-Sek:
I can see that translating `length` prop into `max-height` or `max-width` (depending on orientation) is a 3 year old decision, but if we could change it to `height` and `width`, the example use case would work. Are there any arguments against this idea?

johnleider:
I feel having to use the utility classes is unintuitive. Whatever changes are made, we should probably add some e2e tests to make sure this doesn't happen again. For now, I'm not sure the best path, however, I'm leaning towards changing to width/height from max width/height.

J-Sek:
I have spent a bit of time reviewing paid themes based on Vuetify and found 2 cases where going back to `height: 100%` would make the divider disappear [[1]](https://demos.icreatorstudio.com/demo-1/ecommerce/jobs/details), [[2]](https://spike-vue-main.netlify.app/ecommerce/products-one). So the most sensible option seems to be an update to `length` prop. The PR includes couple of test cases, let me know if they are an overkill for such a small component. They run pretty fast on my machine, but I don't want to impact CI/CD processing times more than it is necessary.

johnleider:
Want to take this under your wing @J-Sek ?

PR #19861, opened 2024-05-22
fix(VCheckboxBtn): reset checked state when readonly is true

### LINKED ISSUE #19137: [Bug Report][3.5.2] VCheckboxBtn - double click is required when @update:model-value does not update model-value
### Environment
**Vuetify Version:** 3.5.2
**Vue Version:** 3.4.15
**OS:** Windows

### Steps to reproduce
Some notes to the playground:
* `Checkbox #1` is where the problem will occur.
* `Checkbox #2` is used to control the behaviour of the `Checkbox #1`

Steps:
1. Click on `Checkbox #2`
2. Click on `Checkbox #1`
    * state should not change, as `@update:model-update` event handler is doing nothing
3. Click on `Checkbox #2`
4. Click on `Checkbox #1`

### Expected Behavior
After step 4, the `Checkbox #1` should be checked

### Actual Behavior
After step 4, the `Checkbox #1` is unchecked.

Additional click is required on the checkbox to correctly update the value.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNp1UstugzAQ/JWteyiRElDS5kJJ1KrXnlq1l9KDgSVYMbYFhlRC/HuNgbxabt7R7M7OeL8aUhax96yUW1dIfBJozBWnGrehAAjqBVXKPm0RS6EpE1gMEMBLhvE+kj9wu/RHrCMO8CLSYoQB/FwmyBc15RVuQjKS3rXRC8mJ91SpxEBX9IyKhOMoaJs+LO/U6pm9jltEha376g1jZDUmYKf50DRQDNBnh0DbTrWO1V/LK3BYCtYGJnPQGQqwqzOxO08GDoxziBASVtKIYzKbygrqhTVt7JaZrHjyKuP9OKl3W4bkuFzgXf9Jh/RfFnhnX2nKMi6Y0lCirhRwKnZGQ5thto3lShYaulBSaCEtZA535iLuettGo9S90TF62HRcJ6W8xNnjiTS59lTD5TdYUhBJyZGKrdMRO2paiVgzKWDyCByBBzvCh6F7Bk2f0oWCaw/A6Iz8QcFkkIJzM7l+3ze7DOG/YQDmkgKvj9uES9o5uXcf3OWadI+1uyLfvzG4Jk4=)

### Other comments
Observations:
* if step 4 is executed **ODD** number of times, then the bug happens
* if step 4 is executed **EVEN** number of times, then there everything is ok

Probably some internal state in VCheckboxBtn is updated. But if the `update:model-value` handler does not update the `model-value`, then there is mismatch with internal state and modelValue.

### Workarounds

I have found the following workaround - using `@click.prevent` instead of the `@update:mode-value` (as can be seen below), so it's not a big problem, but the issue still remains, that it can happen that internal state of checkbox component is not aligned with the modelValue

```vue
<v-checkbox-btn
  :model-value="checkboxState"
  @click.prevent="handleCheckboxStateUpdate(!checkboxState)"
/>
```

zepa19:
## Description

fixes #19137

How to reproduce the bug is described in the referenced issue, but shortly, the `input` internal checked state was out of sync with modelValue when the `readonly` flag is `true`.

That caused the bug, when the checkbox becomes interactive again, and sometimes it would not update the model value on the first click on the checkbox.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-checkbox-btn
        label="Checkbox #1:"
        v-model="checkboxState"
        :readonly="shouldLockCheckboxUpdates"
      />

      <v-checkbox-btn
        v-model="shouldLockCheckboxUpdates"
        label="Checkbox #2:"
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
import { ref } from 'vue'

const checkboxState = ref(false);
const shouldLockCheckboxUpdates = ref(false);
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx
+++ b/packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx
@@ -209,10 +209,10 @@ export const VSelectionControl = genericComponent<new <T>(
           // model value is not updated when input is not interactive
           // but the internal checked state of the input is still updated,
           // so here it's value is restored
-          input.value.checked = model.value;
+          input.value.checked = model.value
         }
 
-        return;
+        return
       }
 
       if (props.readonly && group) {
--- a/packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx
+++ b/packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx
@@ -204,7 +204,16 @@ export const VSelectionControl = genericComponent<new <T>(
     }
 
     function onInput (e: Event) {
-      if (!isInteractive.value) return
+      if (!isInteractive.value) {
+        if (input.value) {
+          // model value is not updated when input is not interactive
+          // but the internal checked state of the input is still updated,
+          // so here it's value is restored
+          input.value.checked = model.value;
+        }
+
+        return;
+      }
 
       if (props.readonly && group) {
         nextTick(() => group.forceUpdate())
--- a/packages/vuetify/src/components/VCheckbox/__tests__/VCheckboxBtn.spec.cy.tsx
+++ b/packages/vuetify/src/components/VCheckbox/__tests__/VCheckboxBtn.spec.cy.tsx
@@ -51,7 +51,7 @@ describe('VCheckboxBtn', () => {
   it('should not update input checked state when it is readonly', () => {
     const model = ref(false)
     cy.mount(() => (
-      <VCheckboxBtn v-model={ model.value } readonly={ true } />
+      <VCheckboxBtn v-model={ model.value } readonly />
     ))
 
     cy.get('.v-checkbox-btn').click(20, 20)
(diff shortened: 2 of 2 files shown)


packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx
```diff
@@ -204,7 +204,16 @@ export const VSelectionControl = genericComponent<new <T>(
...
+          // model value is not updated when input is not interactive
+          // but the internal checked state of the input is still updated,
+          // so here it's value is restored
+          input.value.checked = model.value;
+        }
+
+        return;
+      }
 
       if (props.readonly && group) {
```
zepa19, line 218: I just want to point out, that this condition is unreachable, so probably it can be removed? Not sure, why it is here.

johnleider (approved):

PR #15728, opened 2022-09-01
fix(VAutocomplete): divider/subheader not rendering

nekosaur:
## Description

closes #15721

## Markup:

<details>

```vue
<template>
  <div>
    <v-select
      :items="people"
      filled
      chips
      closable-chips
      color="blue-grey-lighten-2"
      label="Select"
      multiple
    ></v-select>

    <v-autocomplete
      :items="people"
      filled
      chips
      closable-chips
      color="blue-grey-lighten-2"
      label="autocomplete"
      multiple
    ></v-autocomplete>

    <!-- <v-list :items="people" /> -->
  </div>
</template>

<script>
  const srcs = {
    1: 'https://cdn.vuetifyjs.com/images/lists/1.jpg',
    2: 'https://cdn.vuetifyjs.com/images/lists/2.jpg',
    3: 'https://cdn.vuetifyjs.com/images/lists/3.jpg',
    4: 'https://cdn.vuetifyjs.com/images/lists/4.jpg',
    5: 'https://cdn.vuetifyjs.com/images/lists/5.jpg',
  }

  export default {
    data: () => ({
      people: [
        {
          title: 'Group 1',
          subheader: true,
          divider: true,
          children: [
            { title: 'Jane Smith', group: 'Group 1', avatar: srcs[5] },
            { title: 'Britta Holt', group: 'Group 1', avatar: srcs[4] },
          ],
        },
        {
          title: 'Group 2',
          subheader: true,
          children: [
            { title: 'Trevor Hansen', group: 'Group 2', avatar: srcs[3] },
            { title: 'Ali Connors', group: 'Group 2', avatar: srcs[2] },
          ],
        },
      ],
    }),
  }
</script>



```
</details>

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [ ] The PR title is no longer than 64 characters.
- [ ] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [ ] My code follows the code style of this project.
- [ ] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #19912, opened 2024-05-27
fix(VAutocomplete): divider/subheader not rendering

func0der:
Original PR: https://github.com/vuetifyjs/vuetify/pull/15728

## Description
This is just an updated version of https://github.com/vuetifyjs/vuetify/pull/15728, since the functionality is still missing and the PR was closed as stale.
@nekosaur I hope this was okay?

You are welcome to update and fix accordingly.

I am not 100% in the Vuetify game and I might have made mistakes merging `master`. I am especially insecure about the changes in `VTreeView`, since the component was not present when the original PR was created.

## Markup:
As described here: https://vuetifyjs.com/en/getting-started/contributing/#for-docs-language

```vue
<template>
  <v-card class="mx-auto" max-width="300">
    <v-list :items="items" />
    <v-select :items="items" />
    <v-autocomplete :items="items" />
    <v-treeview :items="items" />
  </v-card>
</template>

<script>
  export default {
    data: () => ({
      items: [
        {
          title: 'Group (Dropdown in List, old behavior)',
          value: 'group',
          children: [
            {
              title: 'Child',
              subtitle: 'Subtitle',
              value: 'child',
            },
          ],
        },
        {
          props: {
            subheader: true,
            divider: true,
          },
          title: 'Group #1 (subheader in list, additional new behavior)',
          value: 'foo',
          children: [
            {
              title: 'Item #1',
              value: 1,
            },
            {
              title: 'Item #2',
              value: 2,
            },
            {
              title: 'Item #3',
              value: 3,
            },
          ],
        },
        {
          props: {
            subheader: true,
          },
          title: 'Group #2 (choose to have no divider)',
          children: [
            {
              title: 'Item #4',
              value: 4,
            },
            {
              title: 'Item #5',
              value: 5,
            },
            {
              title: 'Item #6',
              value: 6,
            },
          ],
        },
      ],
    }),
  }
</script>

```

--- update 1:

Updated the playground items array to better reflect how the items diff from each other.


func0der:
Known issues for now are:
* Failing test because the rending of groups is somehow broken (have not found out why yet)
* The syntax to define subheaders and dividers is unique to this usecase and different from the way of VListItem ,which would result in a breaking change and not a feature
* There might still be unnecessary code in the like the `!slots.default` lines in `VListItem` that I can not quite place in the context of this issue

packages/vuetify/src/components/VList/VListChildren.tsx
```diff
@@ -43,65 +45,53 @@ export const VListChildren = genericComponent<new <T extends InternalListItem>(
...
+              default: () => (
+                <VListChildren
+                  items={ children }
+                  v-slots={ slots }
+                />
+              ),
+            }}
+          </VListGroup>
+        ) : (
+          slots.item ? slots.item({ item: internalItem, props: itemProps, index }) : (
```
func0der, line 87: is this kind of change breaking or do you not need to use ALL available properties when redefining slots?

func0der:
@yuwu9145 @johnleider @KaelWD 
Any chance we can get a review on this, so I can work on possible change requests and we can continue to work on this? :)

packages/vuetify/src/components/VList/VListItem.tsx
```diff
@@ -294,13 +294,13 @@ export const VListItem = genericComponent<VListItemSlots>()({
           )}
 
           <div class="v-list-item__content" data-no-activator="">
-            { hasTitle && (
+            { hasTitle && !slots.default && (
```
johnleider, line 297: What is the reason for this change?

packages/vuetify/src/components/VList/VListItem.tsx
```diff
@@ -294,13 +294,13 @@ export const VListItem = genericComponent<VListItemSlots>()({
...
               <VListItemTitle key="title">
                 { slots.title?.({ title: props.title }) ?? props.title }
               </VListItemTitle>
             )}
 
-            { hasSubtitle && (
+            { hasSubtitle && !slots.default && (
```
johnleider, line 303: What is the reason for this change?

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -85,9 +86,12 @@ export const makeSelectProps = propsFactory({
     default: '$vuetify.noDataText',
   },
   openOnClear: Boolean,
+  readonly: Boolean,
```
johnleider, line 89: Is this related?

packages/vuetify/src/composables/list-items.ts
```diff
@@ -63,7 +63,7 @@ export function transformItem (props: Omit<ItemProps, 'items'>, item: any): List
   const itemProps = props.itemProps === true
     ? typeof item === 'object' && item != null && !Array.isArray(item)
       ? 'children' in item
-        ? omit(item, ['children'])
+        ? omit(item as any, ['children'])
```
johnleider, line 66: This use of "any" here is probably a red flag for @KaelWD

johnleider (changes requested):
What is the reasoning from switching from InternalListItem to ListItem?

func0der:
@nekosaur Would love for you to chip in on some of those review comments/question, if you can :)

packages/vuetify/src/components/VSelect/VSelect.tsx
func0der: No. It is a relic of a commit I have removed, since it was unrelated to the issue at hand: https://github.com/vuetifyjs/vuetify/commit/6d6c1dc43eb3d9018420ebdbee57ee2b8ff51e42

I will remove it.

packages/vuetify/src/composables/list-items.ts
func0der: I can remove it, I think, since `transformItem` already types it as `any`.

packages/vuetify/src/components/VList/VListChildren.tsx
func0der: My question was about the definition of a custom slot.
So for example, if user has this in their template:

```
<template #item="{ props }">

</template>
```

Would adding new available parameters for the slot here, break their code.

But since we are not changing the naming of `props`, it should work fine :)

packages/vuetify/src/components/VList/VListItem.tsx
func0der: Actually. I have no real clue. This is something I just copied over from the other PR. It came with this commit, which I think is not really related to this change: https://github.com/vuetifyjs/vuetify/commit/e6792c1b79b0e0a5cab35db6720dd5b62dfc2d0a

If it is not needed or is bad practice, I will remove it.

func0der:
> What is the reasoning from switching from InternalListItem to ListItem?

I think that the `InternalListItem` was more of a helper construct inside the component and not something that was actually used.
Since the whole list flattening is now inside `list-items.ts` it is not needed anymore. Inside `list-items.ts` the type is not called `FlatListItem`.

packages/vuetify/src/components/VList/VListItem.tsx
func0der: Actually. I have no real clue. This is something I just copied over from the other PR. It came with this commit, which I think is not really related to this change: https://github.com/vuetifyjs/vuetify/commit/e6792c1b79b0e0a5cab35db6720dd5b62dfc2d0a

If it is not needed or is bad practice, I will remove it.

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -87,7 +88,9 @@ export const makeSelectProps = propsFactory({
   openOnClear: Boolean,
   itemColor: String,
 
-  ...makeItemsProps({ itemChildren: false }),
+  ...makeItemsProps({
+    itemProps: () => (item: any) => ({ subheader: item.subheader, divider: item.divider }),
```
gduliscouet-ubitransport, line 92: I tested the branch with the provided markup and the dividers are missing for the VSelect (and VAutocomplete). It seems like this is expecting `subheader` and `divider` to be at the root, but it is not the case.

image


Looking into `item.props?.subheader` works. It also works like that:

```suggestion
    itemProps: 'props',
```

image

gduliscouet-ubitransport: Unless the markup you provided is not the good one @func0der ? I see on the [original PR](https://github.com/vuetifyjs/vuetify/pull/15728) that `subheader` and `divider` were at the root of the item

gduliscouet-ubitransport:
We are quite interested in this feature and I would love to help moving this PR forward. But it is, I think, too much to handle safely by people unfamilliar with the Vuetify codebase. Would a core member be able to help on this ? @johnleider @KaelWD ?

johnleider:
> We are quite interested in this feature and I would love to help moving this PR forward. But it is, I think, too much to handle safely by people unfamilliar with the Vuetify codebase. Would a core member be able to help on this ? @johnleider @KaelWD ?

I've asked @nekosaur if he could take a look at some point when he has time.

packages/vuetify/src/components/VSelect/VSelect.tsx
func0der: Thanks for the pointer. 
I could not make out any occurence where this change would be needed.
I guess, it is just a leftover from the time this features was experimented with.

I changed this to have only the `itemChildren` removed.

func0der:
Sorry for the late reaction. I have addressed your remarks. :) 

> But it is, I think, too much to handle safely by people unfamilliar with the Vuetify codebase. Would a core member be able to help on this ?

I'll try my best anyway so they only have to check it afterwards. After all it is just an adoption of @nekosaur's code :)

packages/vuetify/src/components/VSelect/VSelect.tsx
gduliscouet-ubitransport: It looks good with your change @func0der, thank you !

egeersoz:
Thank you for this PR. This functionality is really important IMO and I hope we get it back soon!

egeersoz:
Just wanted to inquire about the status of this PR. Thanks!

egeersoz:
Any chance this can be reviewed and merged? We had to add a lot of conditionals inside our slots to make it work, and miss the way it worked seamlessly in v2.

Issue #19898, opened 2024-05-27
[Feature Request] Strict precision for VNumberInput


J-Sek:
### Problem to solve
In my large web app I have multiple forms with many different use cases for number inputs. When I want user to put a value in USD it should display nicely using expected formatting and enforce up to 2 decimal digits. Having "$" prefix is OK, but actually formatting and enforcing precision is a level-up in terms of UX.

[...]

PR #19913, opened 2024-05-27
fix(VNumberInput): avoid showing NaN

### LINKED ISSUE #19798: [Bug Report][3.6.5]  (VNumberInput): interpretes negative numbers as NaN
### Environment
**Vuetify Version:** 3.6.5
**Vue Version:** 3.4.27
**OS:** Windows 10

### Steps to reproduce
1. Enter a minus sign (≒hyphen `-`)
    (Copy&Pasting a minus also causes same behavior)

image

### Expected Behavior
The input field should accept negative numbers if the `min` prop is set to a value less than zero

### Actual Behavior
As soon as you enter a minus, the input will be replaced with a literal string of `NaN`.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptkM1OwzAQhF9l8SUgkQauUVqJG7wAHAgHN9mApfWP1nYuVd8d26E0kXqxvLOjGfv7PAnPQ/Pi3G6OKFrRBdSOZMBDbwC6uZbOlWsZBmuCVAb5TyqiifqIXCvjYrjIAC3jjOxx34tJksdeXHc5hi29S1bShOQYcZKRwtrT+oAurZ52z2uZ5BEpyRvrjxrxLdffKmuV8XhzM//3e0u2nhQRjlfDoWu2n7twSPIWRFYWTl2z4pdGP7ByAdID4oJRaWc5wAkYJzjDxFZDldBX2V7I+ADaf8M+O+6rVySy8GGZxrvqIRcskSlMnB9NJCrH1y8u0JCN)

### Other comments
This may relate to the following issue: 
https://github.com/vuetifyjs/vuetify/issues/2359

J-Sek:
## Description
In order to avoid `NaN` we need to decouple "text" model for VTextField and the "exposed" model representing value for VNumberInput's consumer. I don't see it any other way and this approach served me many years in my custom wrapper arount VTextField back in Vuetify 2, so I am pretty sure it's the only way forward.

Each time we sync between these two
- "text" model becomes valid number or an empty string – thanks to `typeof model.value === 'number' && !isNaN(...)`
- "exposed" model becomes valid number – thanks to custom aggressive parsing in `extractNumber(...)`

Note: there are many ways to break this implementation, so we will need a bunch of test cases as a safety net for future changes. I plan on introducing them in the following days.

fixes #19798

## Markup:
```vue
<template>
  <v-app theme="dark">
    <v-container>
      <v-card class="mx-auto pa-6 my-4" style="max-width: 1200px">
        <v-row>
          <v-col cols="4">
            <small class="d-block mb-2">(clearable)</small>
            <v-number-input v-model="emptyValue" clearable />
            <code>value: {{ emptyValue }}</code>
          </v-col>
          <v-col cols="4">
            <small class="d-block mb-2">(:step=".25")</small>
            <v-number-input v-model="value1" :step=".25" />
            <code>value: {{ value1 }}</code>
          </v-col>
          <v-col cols="4">
            <small class="d-block mb-2">(:max="20" :min="-20")</small>
            <v-number-input v-model="value2" :max="20" :min="-20" />
            <div class="d-flex align-center justify-space-between">
              <code>value: {{ value2 }}</code>
              <div class="text-right">
                <v-btn @click="value2 = 5.125">Set to 5.125</v-btn>
                <v-btn @click="value2 = NaN">Set to NaN</v-btn>
              </div>
            </div>
          </v-col>
        </v-row>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'

  const emptyValue = ref<number>(null)
  const value1 = ref<number>(15)
  const value2 = ref<number>(-10.25)
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
@@ -98,7 +98,7 @@ describe('VNumberInput', () => {
             class="disabled-input-2"
             v-model={ value4.value }
             min={ 0 }
-  max={ 10 }
+            max={ 10 }
             disabled
           />
         </>
@@ -156,6 +156,99 @@ describe('VNumberInput', () => {
         .click()
         .get('.v-number-input input').should('have.value', '0')
         .then(() => expect(numberInputValue.value).to.equal(0))
+    })
+  })
+
+  describe('parsing pasted values', () => {
+    it('should parse valid numbers correctly', () => {
+      const scenarios = [
+        { text: '-0', expected: '0' },
+        { text: '-00123', expected: '-123' },
+        { text: '.250', expected: '0.25' },
+        { text: '000.321', expected: '0.321' },
+        { text: '100.000', expected: '100' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput />)
+        cy.window().its('navigator.clipboard').then((clip: Clipboard) => clip.writeText(text))
+        cy.get('.v-number-input input').focus()
+        cy.document().invoke('execCommand', 'paste')
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+
+    it('should retrieve digits when pasted value contains some mixed characters', () => {
+      const scenarios = [
+        { text: ' 1,250.32\n', expected: '1250.32' },
+        { text: '1\'024.00 meters', expected: '1024' },
+        { text: '- 1123.', expected: '-1123' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput />)
+        cy.window().its('navigator.clipboard').then((clip: Clipboard) => clip.writeText(text))
+        cy.get('.v-number-input input').focus()
+        cy.document().invoke('execCommand', 'paste')
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+
+    it('should use as much characters as possible to display value containing only received digits', () => {
+      const scenarios = [
+        { text: '0.000010', expected: '0.00001' },
+        { text: '0.012340', expected: '0.01234' },
+        { text: '099999990.000000010', expected: '99999990.00000001' },
+        { text: '99999999999999999999999999999', expected: '999999999999999' },
+        { text: '0.999999999999999999999999999', expected: '0.999999999999999' },
+        { text: '3.9999999999999995', expected: '3.999999999999999' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput />)
+        cy.window().its('navigator.clipboard').then((clip: Clipboard) => clip.writeText(text))
+        cy.get('.v-number-input input').focus()
+        cy.document().invoke('execCommand', 'paste')
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+  })
+
+  describe('typing values', () => {
+    it('should ignore invalid and duplicated characters', () => {
+      const scenarios = [
+        { text: '+1\'234.12 meters', expected: '1234.12' },
+        { text: '-12 34 55', expected: '-123455' },
+        { text: '1,123..50', expected: '1123.5' },
+        { text: 'abc--', expected: '' },
+        { text: '00.123.40', expected: '0.1234' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput />)
+        cy.get('.v-number-input input').focus().realType(text)
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+
+    it('should enforce range by replacing text typed by user', () => {
+      const scenarios = [
+        { text: '9999', expected: '50' },
+        { text: '-99', expected: '-25' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput min={ -25 } max={ 50 } />)
+        cy.get('.v-number-input input').focus().realType(text)
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+
+    it('should not interrupt typing float numbers', () => {
+      const scenarios = [
+        { text: '0.12{backspace}{backspace}89', expected: '0.89' },
+        { text: '-.{backspace}25.{backspace}0.50{backspace}{backspace}99', expected: '-250.99' },
+      ]
+      scenarios.forEach(({ text, expected }) => {
+        cy.mount(() => <VNumberInput />)
+        cy.get('.v-number-input input').focus().realType(text)
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
     })
   })
 })
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -427,10 +427,9 @@ export function clamp (value: number, min = 0, max = 1) {
 }
 
 export function getDecimals (value: number) {
-  const trimmedStr = value.toString().trim()
-  return trimmedStr.includes('.')
-    ? (trimmedStr.length - trimmedStr.indexOf('.') - 1)
-    : 0
+  const fractionDigits = (value.toString().trim().match(/\.\d+/g)?.at(0) ?? '.').length - 1
+  const correction = Number(value.toString().split('e')[1] ?? '0')
+  return Math.max(0, fractionDigits - correction)
 }
 
 export function padEnd (str: string, length: number, char = '0') {
@@ -759,4 +758,38 @@ export function templateRef () {
   })
 
   return fn as TemplateRef
+}
+
+export function extractNumber (text: string): number | null {
+  let cleanText = text.split('')
+    .filter(x => /[\d\-.]/.test(x))
+    .filter((x, i, all) => (i === 0 && /[-]/.test(x)) || // sign allowed at the start
+        (x === '.' && i === all.indexOf('.')) || // decimal separator allowed only once
+        /\d/.test(x))
+    .join('')
+
+  const hasAnyDigit = /\d/.test(cleanText)
+  if (!hasAnyDigit) return null
+
+  if (!/\d\./.test(cleanText) && cleanText.includes('.')) {
+    cleanText = cleanText.replace('.', '0.') // inject missing zero before separator
+  }
+
+  cleanText = cleanText.replace(/^(-)?0*(\d)/, '$1$2') // remove leading zeros except one
+  cleanText = cleanText.replace(/\.(\d*[1-9])?0+$/, '.$1') // remove trailing zeros
+  cleanText = cleanText.replace(/\.$/, '') // remove trailing dot
+
+  if (cleanText.length > 17) {
+    cleanText = cleanText.substring(0, 17)
+  }
+
+  let decimalDigits = (cleanText.split('.')[1] ?? '').length
+  while (cleanText.length > 1 && cleanText !== Number(cleanText).toFixed(decimalDigits)) {
+    cleanText = cleanText.substring(0, cleanText.length - 2)
+    if (decimalDigits > 0) decimalDigits--
+  }
+
+  if (cleanText === '-0') return 0
+
+  return Number(cleanText)
 }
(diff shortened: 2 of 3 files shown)


J-Sek:
While writing tests I have noticed many edge cases related to `.toString()` producing totally different text that was used to parse the number. Initially I just wanted to avoid situation when component "misbehaves" in a quite critical manner - by interrupting user when he is in the middle of typing. But I learned about handful of other scenarios, most of which apply to pasting text from clipboard and raw typing.

Explanation of scenarios (I am not convinced E2E tests are self-explanatory)

- typing `-` is immediately replaced with `NaN` (but it later works when you clear text and try again)
- typing `-0` is immediately replaced with `0` (I wanted to type `-0.2` and I am basically unable to do it)
- typing `0.0000001` is immediately replaced with `1e-7` and it only gets worse the higher the precision
- once the mentioned problems were fixed I had to be very careful and disallow typing when one more digit would result in rounding the whole value (e.g. from `3.99...` to `4`) as I think it is not the intention of the user. He can still move cursor and type digits earlier, however effective digits are cut off as we prioritize digits from the left. I am OK with this tradeoff although it would be the best to display validation error stating that the field gets more than it can handle.
   - the core of this functionality is a little buried inside `extractNumber(...)`. This function now removes digits until the condition `cleanText !== Number(cleanText).toFixed(decimalDigits)` is satisfied.

> reasoning: accepting all digits them would affect user's input due to JS number format limitation. I would rather cut the text instead of replacing whole input with something unexpected.

The issue is related to precision. I think we would benefit from strict `precision` prop (i.e. fractional digits limit) that would be `0` by default. Feature request is already open #19898.

Along the way I was forced to rewrite `getDecimals(...)`. As mentioned `.toString()` produces exponential notation after some point. New implementation just counts digits after the dot (if any are present) and includes value after `e`. I don't love the internal naming and lack of comments inside, but opted for brevity until someone complaints.

yuwu9145 (commented):
Did an initial testing, when typing "-" the first time, it's ignored, but when typing second time, it shows up

J-Sek:
> Did an initial testing, when typing "-" the first time, it's ignored, but when keep typing the second time, it shows up.

Could you share the code you've used? I cannot reproduce this behavior with the simplest case I can think of:

```
// value = ref(null)
<v-number-input v-model="value" />
```

Input behaves correctly on Chrome and Firefox (on Linux). And there are test cases in the `typing values` group that should probably fail if such bug exists.

---

> Also, could you provide a simplified example that demonstrate the reason to justify introducing the "text" model for VTextField?

1. Type `0.001`, `Backspace` - results in `0`
   - intention was to replace `1` with non-zero character, but user is interrupted and forced to type `.00` again
   - if there is a single decimal digit, replacement is interrupted as well but in more subtle way which might be even more frustrating
3. Type or paste `0.0000001` - results in `1e-7`

There are only 2 ways to solve it that I see:
- defer `onModelUpdate` until the field is blurred
- split `text` model from exposed numeric model and synchronize them as suggested in this PR

J-Sek:
Hey @yuwu9145, how can I help push this PR forward? This component will soon be out of labs and I think it would not look good if we let it show "NaN" or ship suboptimal UX (i.e. interruptions while typing).

yuwu9145:
> Hey @yuwu9145, how can I help push this PR forward? This component will soon be out of labs and I think it would not look good if we let it show "NaN" or ship suboptimal UX (i.e. interruptions while typing).

I'll be available to check it again tomorrow

yuwu9145 (changes requested):
+/- buttons become useless when exponent number is displayed.

Instead, numberInput should only support a certain max/min/fractionDigits range allowed in javascript. If users want to use number as advance level (e.g. bigInt), they should use string with v-text-field. This can be shown in the doc once we figure out what are the exact ranges it supports.


Screenshot 2024-07-18 at 1 40 08 pm



NumberInput should focus on basic use cases:

- Only allow numbers, +, - , .
- +,-,. are only allowed once, AND +,- only allowed at the start


@johnleider could you confirm if this was the right direction for numberInput?

I think solution to resolve NaN issue should be around `onModelUpdate`

Internal model split opens the door for more advanced usages which I personally don't think suitable in number input

J-Sek:
> I think solution to resolve NaN issue should be around `onModelUpdate`

I am trying to follow... we need to suppress emits that lead to `NaN`. I have replaced text field model with `textToSync` that will be temporarily used when `onModelUpdate` had to skip `model` change. Synchronization is executed on blur or when user attempts to use up/down arrow keys.

The downsides:
- removing characters except the sign (e.g. `-2` → `-`) is does not emit the value outside of the component (until blur or ↓/↑)
- having any trailing decimal zeros (e.g. `52.10`) is does not emit the value outside of the component (until blur or ↓/↑)

image

Users copy/paste from all the different sources that might format numbers in their locale. If we would like to support it, the `extractNumber` should rather stay.

All the relevant E2E tests still pass.

J-Sek:
@yuwu9145, I have polished the PR a little after following your guide. I don't necessarily like the fact we are closing the doors for grouping separators and formatting locales, but maybe it's better to ship a scalper than swiss army knife (for the first version out of labs).

packages/vuetify/src/util/helpers.ts
```diff
@@ -760,3 +759,37 @@ export function templateRef () {
 
   return fn as TemplateRef
 }
+
+export function extractNumber (text: string): number | null {
```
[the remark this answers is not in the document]
yuwu9145: I see the reason from cy test

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
```diff
@@ -158,4 +158,97 @@ describe('VNumberInput', () => {
...
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+
+    it('should use as much characters as possible to display value containing only received digits', () => {
+      const scenarios = [
+        { text: '0.000010', expected: '0.00001' },
+        { text: '0.012340', expected: '0.01234' },
+        { text: '099999990.000000010', expected: '99999990.00000001' },
+        { text: '99999999999999999999999999999', expected: '999999999999999' },
```
yuwu9145, line 200: Why `99999999999999999999999999999` becomes `999999999999999` ?

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
```diff
@@ -158,4 +158,97 @@ describe('VNumberInput', () => {
...
+        cy.document().invoke('execCommand', 'paste')
+        cy.get('.v-number-input input').blur().should('have.value', expected)
+      })
+    })
+  })
+
+  describe('typing values', () => {
+    it('should ignore invalid and duplicated characters', () => {
+      const scenarios = [
+        { text: '+=1234... abc e12', expected: '1234.12' },
```
yuwu9145, line 217: Invalid input should be enforced to empty `undefined`

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
yuwu9145: JS number has a quirk:

Precision loss from using numeric literals with more than 15 significant digits.

Screenshot 2024-07-19 at 11 48 20 pm

johnleider:
> @johnleider could you confirm if this was the right direction for numberInput?

Confirmed.

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
J-Sek: I think the end goal is the same. We want to respect whatever user typed and avoid replacing already typed characters because of JS quirks. Using integer limit/range with inputs that can be float values seems like something that will leave tons of scenarios uncovered.

Current implementation shortens input string (removing characters from the end) until condition is satisfied
```ts
cleanText !== Number(cleanText).toFixed(decimalDigits)
```

Original question was about a very long number. You can paste or type to any example from documentation. Pasted it becomes `1e+29`, typed... well breaks in more spectacular way and can lead to `NaN`, hence test case seems a great fit.

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
J-Sek: I wanted a test case for number extraction, but this example does not convey that we wish to accept numbers copy/pasted with any format (as long as decimal separator is dot). I will update this part.

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
yuwu9145: > Using integer limit/range with inputs (that can be float values) seems like something that will leave tons of scenarios uncovered.

The primary purpose of number-input is **BASIC NUMERICAL INPUTS**, for example, money/weight etc. It should AVOID JS number quirks rather than RESOLVE all precision & unsafe JS quirks.

As per my understanding, there are three number challenges:

- Unsafe integer (`9007199254740993` -> `9007199254740992`) - use `MAX_SAFE_INTEGER` & `MIN_SAFE_INTEGER` range
- JS floating error (`0.1 + 0.2 === 0.30000000000000004 `) - almost impossible to resolve perfectly, so we make it clear in doc we use `toFixed` to cope with it
- Exponent format - not supported because they invalidate +/- buttons

These would cover the most of the use cases.

If users want to deal with crazy numbers (bigInt or long decimals), they should not use number-input but operate on STRING in text-field. These cases do not require +/- buttons anyway

J-Sek: I always considered my implementation as a way to avoid JS quirks rather than resolving them :)

**Scenario**: For whatever reason user types (or pastes) `555555...` (very long string of digits). What would we like to show:
- `NaN` --- exposing JS quirk
- `9007199254740991` (MAX_SAFE_INTEGER) --- exposing JS quirk
- `5555555555555555` (digits from user, but only up until they can be displayed)

I am leaning towards the last and I still did not hear any argument against it. In my opinion it is not contrary to any of your points.

Imagine a developer that just placed `v-number-input` in the form and never expected end-user to have a "heavy-finger" or a slightly broken keyboard (I am used to both). But in some of the scenarios, he might unnecessarily surprise end-user. I think the unsurprising result is the best outcome and it it's implementation is quite unsophisticated.

I also don't want to fight over it. It is much more important to close this PR and proceed to strict precision implementation. If you feel like `Math.max(Math.min(Number(cleanText), Number.MAX_SAFE_INTEGER), Number.MIN_SAFE_INTEGER)` should replace the while-loop, I won't have anything against adjusting the PR.

yuwu9145: MAX_SAFE_INTEGER will be used as the max `prop.max`, so any input bigger than it will be restricted/changed to MAX_SAFE_INTEGER.

yuwu9145:
Thanks for your time, but it isn't heading to the right direction:

JS built-in numbers do have quirks, but what we should do is clearly figure out a safe range they can work within and document floating limitations to prevent confusing users. `MAX_SAFE_INTEGER` is large enough to allow the number-input to serve basic numeric uses, isn't supported by the number input.

Therefore, there is no value in having extractNumber that performs custom string-to-number conversion. Additionally, a new text model isn't necessary as the [beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/Element/beforeinput_event) event works well for preprocessing and pre-validating input.

I'm going to push forward https://github.com/vuetifyjs/vuetify/pull/20211 for your reference

J-Sek:
> MAX_SAFE_INTEGER is large enough to allow the number-input to serve basic numeric uses, isn't supported by the number input.

After playing a bit with `MAX_SAFE_INTEGER` I think you will get have a hard time making it work with floats. The more fractional digits, the less decimal digits you can use to reliably display the value.

| Fraction digits | Example with issues | Decimal part limit |
|-:|:-:|-:|
| 0 | `9_999_999_999_999_999` » `10000000000000000` | 15 |
| 1 | `1_999_999_999_999_999.1` » `1999999999999999` | 15 |
| 2 | `99_999_999_999_999.01` » `99999999999999.02` | 13 |
| 3 | `9_999_999_999_999.001` » `9999999999999.002` | 12 |
| 4 | `1_999_999_999_999.0001` » `1999999999999` | 12 |

`MAX_SAFE_INTEGER` has 16 digits, so it becomes useless once there are any floats. It will be better in preventing NaN than `Infinity`, but that's basically all we can say about it. Which is good enough to resolve the issue :)

I think the docs could include something like "_this field supports numbers up to 14 digits long. Use explicit min/max and precision to avoid JS-related issues_".

PR #19928, opened 2024-05-31
feat(VAutocomplete): new prop disableSelectOnBackspace

### LINKED ISSUE #19926: [Feature Request] VAutocomplete add option to disable removing of already selected items on Backspace keydown when search si focused
### Problem to solve
Several users that are using our product based on vuetify have problems with VAutocomplete (chips, multiple):
When you search/filter something and you want to remove/clear the search text an you hit Backspace more times it will also remove data. It will be great to have possibility to disable UX jump from search field to latest item using Backspace.

### Proposed solution
New prop to disable option to jump from focused search input to latest selected item on backspace press. Hitting backspace key in focused search field will only remove the text, the remove can be done for example by arrows selection or close icon on chip for example. It will be completely optional, so user can choose to rnable/disable this feature

volarname:
## Description
In some cases hitting Backspace that will remove selected items from VAutocomplete is unwanted. New prop `disable-select-on-backspace` will prevent to blur from search field and activate the latest selected item from selected listm and additional hitting of Backspace that will remove data. This will prevent for unwanted removing.

resolves #19926

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        v-model="model"
        label="Autocomplete"
        chips
        closable-chips
        hide-selected
        multiple
        disable-select-on-backspace
      ></v-autocomplete>
    </v-container>
  </v-app>
</template>

<script>
import { ref } from 'vue'
export default {
  name: 'Playground',
  setup () {
    const model = ref(['California', 'Colorado'])

    return {
      model,
    }
  },
}
</script>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VAutocomplete.json
+++ b/packages/api-generator/src/locale/en/VAutocomplete.json
@@ -5,7 +5,8 @@
     "itemChildren": "This property currently has **no effect**.",
     "filter": "The filtering algorithm used when searching. [example](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L40).",
     "items": "Can be an array of objects or strings. By default objects should have **title** and **value** properties, and can optionally have a **props** property containing any [VListItem props](/api/v-list-item/#props). Keys to use for these can be changed with the **item-title**, **item-value**, and **item-props** props.",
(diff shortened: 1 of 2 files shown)


J-Sek:
Thank you for the PR. Closing as outdated, leaving this [workaround](https://github.com/vuetifyjs/vuetify/issues/19926#issuecomment-5285068330) as a solution.

PR #19938, opened 2024-06-03
fix(V-Table): set $table-header-font-weight: 700

rauhur:
## Description
Change in the value of $table-header-font-weight from 500 to 700 in order to visually clearly differentiate the header from the rest of the table.

## Markup:
```
$table-header-font-weight: 700 !default;
```


johnleider:
500 font weight is matching spec https://m2.material.io/components/data-tables

PR #19946, opened 2024-06-04
fix(Docs): Quick remedy to contrast issues for dark themes

J-Sek:
## Description

When I open docs in Dark theme (which can be the default System theme) contrast of "name", "type" and "default" varies between `1` and `2.9` making crucial elements of component API table almost unreadable. It get's worse on other dark themes especially the one named "High Contrast" where gold text on nearly white background gets contrast score of `1.12`.

I wish I could add a `bg-opacity-10` like in TailwindCSS, but we don't have such capabilities in Vuetify. This `rgba(0,0,0,.1)` is my current override (pasted to Amino browser extension), but it sucks that other developers need to experience this.. hence this quick PR.

Note: Maybe [themes](https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/plugins/vuetify.ts#L121) should be adjusted to have better `surface-bright` color, but I don't know the overall impact of it and there are only 3 out of 7 themes available for docs, so... well... I'm confused a bit on how it works.

### Before

image

image

image

image

### After

image

image

image

image

### DIFF
--- a/packages/docs/src/components/api/ApiTable.vue
+++ b/packages/docs/src/components/api/ApiTable.vue
@@ -22,7 +22,7 @@
             name="row"
             v-bind="{
               props: {
-                class: 'bg-surface-bright'
+                style: 'background: rgba(0,0,0,.1)'
               },
               item,
             }"


johnleider (approved):

johnleider:
Thank you for the P.R.. I hadn't considered making the theme docs part of Vuetify as a whole, they are currently just a perk of Vuetify One. Also, I'm completely open to adding bg-opacity utility classes. If you're interested making a P.R. to dev for them, I'd be on board.

PR #19947, opened 2024-06-04
fix(VOverlay): App mounted in shadow dom

### LINKED ISSUE #19943: [Bug Report][3.6.8] v-overlay-container problem with Shadow DOM.
### Environment
**Vuetify Version:** 3.6.8
**Vue Version:** 3.4.27
**Browsers:** Safari 17.5
**OS:** Mac OS 10.15.7

### Steps to reproduce
If you run an http-server in the folder repro you can find the thing that is not working.
We use bootstrap in an ASP.NET MVC Web app. But the CSS interferes between vuetify and bootstrap. I have partly solved it by mounting it on shadow dom.

### Expected Behavior
Expect v-overlay-container to be attached in the shadow dom so it would work with css isolation.

### Actual Behavior
The v-overlay-container is attached to document.body instead. And you can find the items for v-select below on the page when you scroll down after clicking v-select.

### Reproduction Link
[https://github.com/joelmandell/shadow_dom](https://github.com/joelmandell/shadow_dom)

### Other comments
I might fix an PR for this if I find the time and can find out why it behaves like this.
For the moment this is a show stopper of our usage and progressive migration over to vuetify 3.
We need it to work in shadow dom.

joelmandell:
## Description
This checks if the target of VOverlay is an shadow dom element and attaches it to that one. This fixes issues when mounting app on shadow dom. This fixes #19943

## Markup:
Playground is not provided because this is setup in a different way, it can be examined by the repro here:
https://github.com/joelmandell/shadow_dom

Without applying this PR, v-overlay-container will be attached outside the shadow dom and not working correctly.

### DIFF
--- a/packages/vuetify/src/components/VOverlay/VOverlay.tsx
+++ b/packages/vuetify/src/components/VOverlay/VOverlay.tsx
@@ -140,7 +140,6 @@ export const VOverlay = genericComponent<OverlaySlots>()({
         if (!(v && props.disabled)) model.value = v
       },
     })
-    const { teleportTarget } = useTeleport(computed(() => props.attach || props.contained))
     const { themeClasses } = provideTheme(props)
     const { rtlClasses, isRtl } = useRtl()
     const { hasContent, onAfterLeave: _onAfterLeave } = useLazy(props, isActive)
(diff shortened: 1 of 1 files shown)


joelmandell:
@MajesticPotatoe I might need a poke in right direction here to perfect this one. Because there is some css-issues still. For example in my repro I use a v-select. The items in the v-select gets dark background color and with darkened text. So there is some more to throw at this one.

johnleider (approved):

SandroPreuss1234:
> @MajesticPotatoe I might need a poke in right direction here to perfect this one. Because there is some css-issues still. For example in my repro I use a v-select. The items in the v-select gets dark background color and with darkened text. So there is some more to throw at this one.

@joelmandell 
I use `:menu-props="{ attach: getAppRoot() }"` on the `v-select`, to attach the menu to my Shadow Root. I realized the same issue with the darkened text and Chrome dev tools gave me a hint. The CSS variable `--v-theme-overlay-multiplier` is not defined, which is used in some calculations. Not sure why it is lost and if it has to do with the teleport.
But I was able to solve this by defining some custom CSS override, which works for my use case.
```
.v-menu .v-list .v-list-item__overlay {
  --v-theme-overlay-multiplier: 1
}
```
Just want to let you know. :wink:

Issue #19953, opened 2024-06-05
[Bug Report][3.6.8] VNumberInput does not retain aria-autocomplete attribute


Tomdanizer:
### Environment
**Vuetify Version:** 3.6.8
**Vue Version:** 3.4.27
**Browsers:** Edge 125.0.0.0
**OS:** Windows 10

### Steps to reproduce
Adding attributes
        - aria-autocomplete="none"
        - autocomplete="off"
to the VNumberInput component does not actually apply the attributes. This functionality works for other inputs like VTextField etc.

### Expected Behavior
The attributes        
 - aria-autocomplete="none"
 - autocomplete="off"
to be attached to the input element for VNumberInput

### Actual Behavior
Attributes are missing.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVUcGKwjAQ/ZWS83Z7L11hjwt79mI9jO0EA+MkpJMiiP9uEpWmIIKXJPPmMfNe3u6iJj80v859zwFVqzrBkyMQ3PRcVd1cg3P5mYvBsoBh9A8og4JnqbVBGp9gVREckH569Z/uXi0N8AZqCGIHG/egYCSxZVxzQlGsuVbrhbrpmnJ9KYrD6YC+NuyCLMNajzP6KQ3SQNNqa/LmLW2TQpbIGFFDIPlY/RvBy8eUWHs0I/4lpa90tYYnfNHJ5kubz5givM4pIfcYu6aIN3B0GEkxNnX94kCUj/0NI2qliw==)

PR #19959, opened 2024-06-06
fix(VNumberInput): attributes fall through to root

babu-ch:
fix https://github.com/vuetifyjs/vuetify/issues/19953

## Description

I fixed it because attr was not used.

## Markup:

```vue
  <v-app>
    <v-container>
      <v-text-field
        aria-autocomplete="none"
        autocomplete="off"
        label="Label"
        aru
      ></v-text-field>
      <v-number-input
        :hideInput="false"
        :inset="false"
        :reverse="false"
        aria-autocomplete="none"
        autocomplete="off"
        controlVariant="default"
        label=""
      ></v-number-input>
    </v-container>
  </v-app>
```

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -303,6 +303,7 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
             props.class,
           ]}
           { ...textFieldProps }
+          { ...attrs }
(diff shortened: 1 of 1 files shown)


yuwu9145 (changes requested):
Remove `inheritAttrs: false` to let attributes fall through

babu-ch:
@yuwu9145 
Thank you for your help.
I fixed it like that

yuwu9145 (approved):

PR #19990, opened 2024-06-11
fix(VDatePicker): select same date as start/end in date range picker

### LINKED ISSUE #19989: [Bug Report][3.6.8] Date picker with range: Can't select the same date as start/end
### Environment
**Vuetify Version:** 3.6.8
**Vue Version:** 3.4.27
**Browsers:** Firefox 126.0
**OS:** Mac OS 10.15

### Steps to reproduce
We really need the user to be able to pick the same date as the start and the end in the v-date-picker (with multiple="range").

### Expected Behavior
The user can click twice on the same date and get `[sameDate, sameDate]` as a value

### Actual Behavior
The date selection is reset on the second click

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptj8GKwjAQhl9lyKUKZquy7EHqwsK+gUfroaSjBJN0SKb1UPruJqmIiJcw8/En35/jKIJX5R/R19Cj2ImK0ZJpGH9rB1ANsiHKY15U57jRDv0DZdjGtCStruhhkLZr0exrkWioBdjesCaDEfnGXTCi8nn7gAYVYws5vYNxzMM0PYzluzKRuVFVvjSNa1BeE0NA7ufC2lLnGUbweIYJzr6zUMRPFikOEN8NPHthnzKLo8Mb/EewKLbr7bdc/8jNpliu4DM/LVOJWRuFYlq53ph8nO79w2+b)

bulatgab:
fixes 19989

## Description

fixes #19989

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker v-model="dates" multiple="range" />
      Selected dates: {{ dates }}
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const dates = ref([new Date('2024-06-11'), new Date('2024-06-11')])
</script>

```

### DIFF
--- a/packages/api-generator/src/locale/en/VDatePickerMonth.json
+++ b/packages/api-generator/src/locale/en/VDatePickerMonth.json
@@ -2,6 +2,7 @@
   "props": {
     "hideWeekdays": "Hide the days of the week letters.",
     "transition": "The transition used when changing months into the future",
-    "reverseTransition": "The transition used when changing months into the past"
+    "reverseTransition": "The transition used when changing months into the past",
+    "allowOneDayRange": "Allow selecting the same day as start and end of the range"
   }
 }
--- a/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
@@ -40,6 +40,10 @@ export const makeVDatePickerMonthProps = propsFactory({
     type: String,
     default: 'picker-reverse-transition',
   },
+  allowOneDayRange: {
+    type: Boolean,
+    default: false,
+  },
 
   ...makeCalendarProps(),
 }, 'VDatePickerMonth')
@@ -98,7 +102,17 @@ export const VDatePickerMonth = genericComponent<VDatePickerMonthSlots>()({
         rangeStart.value = _value
         model.value = [rangeStart.value]
       } else if (!rangeStop.value) {
-        if (adapter.isBefore(_value, rangeStart.value)) {
+        if (adapter.isSameDay(_value, rangeStart.value)) {
+          if (props.allowOneDayRange) {
+            rangeStop.value = adapter.endOfDay(_value)
+            model.value = [rangeStart.value]
+            return
+          } else {
+            rangeStart.value = undefined
+            model.value = []
+            return
+          }
+        } else if (adapter.isBefore(_value, rangeStart.value)) {
           rangeStop.value = adapter.endOfDay(rangeStart.value)
           rangeStart.value = _value
         } else {
--- a/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePickerMonth.tsx
@@ -98,11 +98,7 @@ export const VDatePickerMonth = genericComponent<VDatePickerMonthSlots>()({
         rangeStart.value = _value
         model.value = [rangeStart.value]
       } else if (!rangeStop.value) {
-        if (adapter.isSameDay(_value, rangeStart.value)) {
-          rangeStart.value = undefined
-          model.value = []
-          return
-        } else if (adapter.isBefore(_value, rangeStart.value)) {
+        if (adapter.isBefore(_value, rangeStart.value)) {
           rangeStop.value = adapter.endOfDay(rangeStart.value)
           rangeStart.value = _value
         } else {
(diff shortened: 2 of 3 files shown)


johnleider:
How do you clear a value then?

bulatgab:
Thank you for your question!

In my understanding, if the user clicks on the wrong first date, they can click on any other date to complete the range selection, and then click again to select the right first date. This UX seems intuitive enough.

If we are talking about clearing the input completely - yes, it's not possible. But it's not possible for a single date picker either (`:multiple="false"`). And it makes sense - after all, we show the form to get the user's input and we are probably not interested in an empty value. (And if the developer _is_ interested in this functionality, they can implement a custom 'clear' icon to clear the state)

bulatgab:
If I can improve anything in my PR, please let me know! It is my first one in this repo.

johnleider:
> Thank you for your question!
> 
> In my understanding, if the user clicks on the wrong first date, they can click on any other date to complete the range selection, and then click again to select the right first date. This UX seems intuitive enough.
> 
> If we are talking about clearing the input completely - yes, it's not possible. But it's not possible for a single date picker either (`:multiple="false"`). And it makes sense - after all, we show the form to get the user's input and we are probably not interested in an empty value. (And if the developer _is_ interested in this functionality, they can implement a custom 'clear' icon to clear the state)

Fair arguments. @vuetifyjs/core-team thoughts?

johnleider:
> and we are probably not interested in an empty value. (And if the developer _is_ interested in this functionality, they can implement a custom 'clear' icon to clear the state)

We still run into the case that people more than likely already have applications that depend on or expect this behavior, so we can't simply remove it, especially in a patch.

RagingRoosevelt:
> We still run into the case that people more than likely already have applications that depend on or expect this behavior, so we can't simply remove it, especially in a patch.

Counterpoint is that people upgrading from Vuetify2 expect the original

bulatgab:
For the context, other libraries also have the proposed behavior in their date-range pickers:
https://primevue.org/datepicker/#range
https://vue3datepicker.com/props/modes/#range
https://quasar.dev/vue-components/date#range-selection
(although in Quasar, the output format is different for the same-date range compared to the different-date range. But that's another story. The main point is that it allows you to select the same date).

bulatgab:
The range with the same start/end dates is a <b>valid</b> date range and not a bizarre edge case. And so it must be covered by the date selector out of the box.  

We can think of the date range as a time period that starts at 12:00 AM, startDate and finishes at 11:59 PM, endDate.  
The current behaviour allows you to select a minimal period of 48h only.  

But what about a 24h period?  
You can imagine a use case for booking something, e.g. a venue for an event. How can you choose to book it for one day only?

bulatgab:
If clearing the state is critical, it could also be done when the user clicks on the existing range selection (the same approach as in Quasar). Or a dedicated “x” icon can be added for clearing.
I suggest this be done in a separate PR though.

bulatgab:
> We still run into the case that people more than likely already have applications that depend on or expect this behavior, so we can't simply remove it, especially in a patch.

Hello,

I understand the need to maintain backward compatibility. Would you consider adding an additional prop, like `allowSameDayRange`, to enable this feature for those who need it?

Other suggestions from ChatGPT include:
`enableSingleDaySelection`
`allowIdenticalRange`
`sameDayRange`
`singleDayRange`
`permitSameDaySelection`

bulatgab:
> We still run into the case that people more than likely already have applications that depend on or expect this behavior, so we can't simply remove it, especially in a patch.

I have added an `allowOneDayRange` prop for backward compatibility :tada:  
The new behaviour is enabled only if the prop is set to `true`.

bulatgab:
### Problem

Unfortunately, there is another problem. `modelValue` becomes `[startDate]` after the first click already, and doesn't change after the second click. This is to comply with the existing contract for `modelValue`, which is a _list_ of dates in the selected range, so for the same-day date range, `modelValue` should be a one-item array.

But in the parent component, we might want to do something when the user completes the range selection. When we receive `[startDate]` value, we cannot know if that is the whole range (the same day as start/end) or an incomplete range.

### Proposal
Ideally, we could change _when_ we emit `update:modelValue`.
We could _not_ emit it when only the `startDate` is selected (after all, the range selection is not yet incomplete then).
And emit it only when both `startDate` and `endDate` get selected.
Would this be acceptable?

### Alternative proposal
Alternatively, we could emit another `update:range` event on the second click, maybe with `[startDate, endDate]` payload instead of the list of all dates.

johnleider:
I like the alternate proposal.

PR #20018, opened 2024-06-16
fix(VTimePicker): it won't trigger update Value when user only update hour 

advancedor96:
## Description
There's an obvious bug in Labs/VTimePicker. When the user first selects the hour and minute, it works fine.
However, when the user changes the hour and keeps the minute the same, it does not emit the `update:modelValue` function.
It only emits `update:hour`.

For example, if you first select "11:50 am" and then try to change it to "12:50 am", you will see that the `v-model` doesn't change to 12. Because the `update:modelValue` does not emit.

I remove the `emitChange` part.  Because it is for "update after user change the second". But at that moment, the following `lazyInputHour` would already changed.

### DIFF
--- a/packages/vuetify/src/labs/VTimePicker/VTimePicker.tsx
+++ b/packages/vuetify/src/labs/VTimePicker/VTimePicker.tsx
@@ -264,7 +264,6 @@ export const VTimePicker = genericComponent<VTimePickerSlots>()({
           break
       }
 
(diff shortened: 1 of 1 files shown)


advancedor96:
Could somebody check my PR?

blalan05:
We still need the check if we don't have minutes at least, so I made a PR as such. Thank you for your help. #20179

PR #20038, opened 2024-06-21
Feat(VTreeview): port open-on-click prop to v3

yuwu9145:
## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/labs/VTreeview/VTreeview.tsx
+++ b/packages/vuetify/src/labs/VTreeview/VTreeview.tsx
@@ -160,6 +160,7 @@ export const VTreeview = genericComponent<new <T>(
             'v-treeview',
             props.class,
           ]}
+          open-strategy="multiple"
           style={ props.style }
           v-model:activated={ activated.value }
           v-model:selected={ selected.value }
--- a/packages/vuetify/src/labs/VTreeview/VTreeviewItem.tsx
+++ b/packages/vuetify/src/labs/VTreeview/VTreeviewItem.tsx
@@ -16,7 +16,7 @@ import { genOverlays } from '@/composables/variant'
 
 // Utilities
 import { computed, inject, ref } from 'vue'
-import { genericComponent, propsFactory, useRender } from '@/util'
+import { genericComponent, omit, propsFactory, useRender } from '@/util'
 
 // Types
 import { VTreeviewSymbol } from './shared'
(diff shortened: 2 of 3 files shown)


uccmen:
omg I need this PR merged by Monday 🥲😂🫡🙌🏽

packages/vuetify/src/labs/VTreeview/VTreeview.tsx
```diff
@@ -35,7 +36,8 @@ export const makeVTreeviewProps = propsFactory({
     collapseIcon: '$treeviewCollapse',
     expandIcon: '$treeviewExpand',
     slim: true,
-  }), ['nav']),
+    color: 'primary',
```
johnleider, line 39: No default colors.

packages/vuetify/src/labs/VTreeview/VTreeviewChildren.tsx
```diff
@@ -28,7 +31,22 @@ export const makeVTreeviewChildrenProps = propsFactory({
...
+    type: IconValue,
+    default: '$checkboxIndeterminate',
+  },
+  falseIcon: IconValue,
+  trueIcon: IconValue,
+  returnObject: Boolean,
   selectable: Boolean,
+  selectedColor: {
+    type: String,
+    default: 'accent',
```
johnleider, line 48: No default colors.

packages/vuetify/src/labs/VTreeview/VTreeviewItem.tsx
```diff
@@ -85,97 +72,26 @@ export const VTreeviewItem = genericComponent<VListItemSlots>()({
...
-        ) : (
+      return (
         <VListItem
-          ref={ vListItemRef }
           { ...listItemProps }
+          active={ isActivated.value }
           class={[
             'v-treeview-item',
             {
+              'v-treeview-item--activetable-group-activator': isActivatableGroupActivator.value,
```
johnleider, line 88: Should it be **activatable** or is **activetable** correct?

johnleider (changes requested):
Looking really good.

johnleider (approved):

kvngrnr:
Any update on when this will be merged/released?

stephane303:
I did some tests, especially around selected, this seems to work but only if return-object is activated.

yuwu9145:
> I did some tests, especially around selected, this seems to work but only if return-object is activated. If return-object is not there :
> 
> * I get [ "Symbol(6)", "Symbol(9)" ] in the selection, and I guess the ids should be there.
> * the behaviour is different than if return-object is activated

I can't replicate on dev branch, did you have `item-value` applied or can you provide a sample code?

stephane303:
> > I did some tests, especially around selected, this seems to work but only if return-object is activated. If return-object is not there :
> > 
> > * I get [ "Symbol(6)", "Symbol(9)" ] in the selection, and I guess the ids should be there.
> > * the behaviour is different than if return-object is activated
> 
> I can't replicate on dev branch, did you have `item-value` applied or can you provide a sample code?

Yes I did the tests again today, can't see it either again, I am new in running vuetify locally, that was probably the reason why, sorry.

yuwu9145:
>  I am new in running vuetify locally

I don't think `dev` branch is live anywhere so you were playing around live `master` branch which doesn't have changes in this PR yet.

https://vuetifyjs.com/en/getting-started/contributing/#local-development

PR #20061, opened 2024-06-26
docs(typescript): add new page

johnleider:
## Motivation and Context
Improve the information regarding typescript and Vuetify.

### DIFF
--- a/packages/docs/src/pages/en/features/typescript.md
+++ b/packages/docs/src/pages/en/features/typescript.md
@@ -16,11 +16,11 @@ Vuetify supports Typescript out of the box. This means that you can use Typescri
 
 ## Installation
 
-To get started with a new project using typescript, following the instructions in the [Getting Started](/getting-started/installation/) guide and select the Typescript option when prompted.
+To get started with a new project using Typescript, following the instructions in the [Getting Started](/getting-started/installation/) guide and select the Typescript option when prompted.
 
 ## Types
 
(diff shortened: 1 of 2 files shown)


packages/docs/src/pages/en/features/typescript.md
```diff
@@ -0,0 +1,27 @@
...
+  - /getting-started/contributing/
+---
+
+# Typescript
+
+Vuetify supports Typescript out of the box. This means that you can use Typescript in your Vuetify project without any additional configuration.
+
+## Installation
+
+To get started with a new project using typescript, following the instructions in the [Getting Started](/getting-started/installation/) guide and select the Typescript option when prompted.
```
pano9000, line 19: admittedly this is a bit nitpicky, but `typescript` should be capitalized here

packages/docs/src/pages/en/features/typescript.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Types
+
+Vuetify provides types for all of its components and seemlessly integrates with Typescript. If you're using an IDE such as VSCode, with the [Vue - Official extension](https://marketplace.visualstudio.com/items?itemName=Vue.volar) and `<script lang="ts">`, you'll get automatic type checking and code completion.
```
pano9000, line 23: `seemlessly` -> contains a typo -> should be `seamlessly`

J-Sek:
I hoped it would cover the props "pass-through" for custom components augmenting native-Vuetify components, but it has many rough edges and it would be better to revisit the subject of dedicated docs page once we migrate to SFC and/or introduce dedicated exports for the prop types.

Closing since the PR never took off.

PR #20123, opened 2024-07-08
fix(VField): always initialize expand transition

### LINKED ISSUE #19778: [Bug Report][3.6.4] clear-icon not animating when binding boolean value to :clearable prop
### Environment
**Vuetify Version:** 3.6.4
**Vue Version:** 3.4.27
**OS:** Windows 10

### Steps to reproduce
Create any VField
add prop :clearable="someBoolean"

### Expected Behavior
Clear icon should animation from the right as with static clearable="true"

### Actual Behavior
Clear icon is not animating

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdkDFPxDAMhf+KL0tBInQ/tQgmWJgYGChDSV0U4TRR4lZIp/vvOAmnO92QKH7vk/2cj4NK0bRPIdxvK6q96hhdoJHxYVgAuk2PIZRnKYxfeLQLxn+piIy/rGeLNJ1EgE07PyH1g3Lpe1BnY28Ixzh+EYr3+vYMu76HK+bRkDU/lawdoAdhL5liaiuBMjFZHT1Lak0485lrT9Hb6+xZqat17cXKUiYTbWBIyGu2M2xd8JHhABFnOMIcvYNG/qvJZj7SO3GOKEGbFyTy8O4jTbtCVLeuIR1uhLsto9o6S6ao492yEpXr8w+/zn+Q)

### Other comments
This is a relatively small issue which is beginner friendly

samschurter:
## Description

When a VField is rendered, it only initializes the VExpandXTransition for the clear-icon if the `clearable` prop is `true`. When the `clearable` prop is bound to a reactive value this causes a case where the VExpandXTransition is initialized in the already visible state, resulting in the icon appearing instantly with no animation. 

To fix this, the VExpandXTransition and the div wrapping the clear-icon are always initialized regardless of the value of `clearable`, but the icon is still only shown when `clearable` becomes `true`.

This affects all clearable components using `VField`. This PR fixes `VTextField` from the issue mentioned below, but I was also able to replicate the issue with the following components. These components are also fixed:

- VAutocomplete
- VCombobox
- VFileInput
- VSelect
- VTextarea

fixes #19778

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        v-model="inputMsg"
        :clearable="canClear"
      />
    </v-container>
  </v-app>
</template>

<script>

  export default {
    name: 'Playground',
    setup () {
      const matchMsg = 'match'
      const inputMsg = ref('')

      const canClear = ref(false)

      watch(inputMsg, newVal => {
        canClear.value = newVal === matchMsg
      })
      return {
        inputMsg,
        canClear,
        matchMsg,
      }
    },
  }
</script>
```


johnleider:
This is still reproduceable by removing and adding back a character.
clear

samschurter:
Ok, it should be working for the hide transition as well as the show transition now. I also updated the markup in the PR for easier testing.

johnleider:
The animation works but you can see it shifts right about 12px before the transition begins. I've been looking into why that is, but it's very noticeable.

samschurter:
Ok, I will look into it as well.

samschurter:
It appears that it comes down to the `v-field--appended` class which is dynamically added to the top-level `VField` div. This  sets a padding which is outside the `VExpandXTransition`, so it does not get animated. 

Would you prefer a solution that keeps the CSS classes as-is and figure out how to animate that property, or rework the CSS so that padding is applied to the icon instead of the wrapper?

johnleider:
Whichever one involves the least amount of code to implement.

samschurter:
I think this works. I first moved the end padding from the wrapper to the icons so that the padding was inside the VExpandXTransition. That mostly worked for animating out, but when animating back in there was still some jank. The transition animates the movement based on `offsetWidth` calculations, but that does not include padding. I then extended the VExpandXTransition to include inline padding. I tried a pure CSS solution in the VField Sass, but since it's a JS transition I couldn't force it to be smooth in all circumstances without causing side effects. Someone more clever with CSS could probably figure it out though. 

If modifying the VExpandXTransition is too much, then this may need to either go to someone better at CSS animations or we may need to stop using the VExpandXTransition and write a custom transition for this component.

I used this for testing multiple variants at once to try and make sure things looked right everywhere `VField` is used. 
```vue
<template>
  <v-app>
    <v-container>
      <div class="border border-error border-md mt-3">
        VAutocomplete
        <v-autocomplete
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :items="['match']"
          :variant="variant"
          clearable
        />
      </div>

      <div class="border border-error border-md mt-3">
        VCombobox
        <v-combobox
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :items="['match']"
          :variant="variant"
          clearable
        />
      </div>

      <div class="border border-error border-md mt-3">
        VFileInput
        <v-file-input
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :items="['match']"
          :variant="variant"
          clearable
        />
      </div>

      <div class="border border-error border-md mt-3">
        VSelect
        <v-select
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :items="['match']"
          :variant="variant"
          clearable
        />
      </div>

      <div class="border border-error border-md mt-3">
        VTextarea
        <v-textarea
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :items="['match']"
          :variant="variant"
          rows="2"
          clearable
        />
      </div>

      <div class="border border-error border-md mt-3">
        VTextField
        <v-text-field
          v-model="inputMsg"
          :append-inner-icon="iconOn"
          :clearable="canClear"
          :variant="variant"
        />
      </div>

      <v-btn class="me-3 my-3" @click="toggleIcon">Toggle Icon</v-btn>
      <v-btn @click="cycleVariant">Cycle Variant</v-btn>
      <br>
      iconOn: {{ iconOn }}
      <br>
      variant: {{ variant }}
      <v-row
        justify="center"
        style="min-height: 160px;"
      >

        <v-col class="shrink">
          <v-btn
            class="ma-2"
            color="secondary"
            @click="expand2 = !expand2"
          >
            Expand X Transition
          </v-btn>

          <v-expand-x-transition>
            <v-card
              v-show="expand2"
              class="mx-auto bg-secondary"
              height="100"
              width="100"
            />
          </v-expand-x-transition>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>
<script>

  export default {
    name: 'Playground',
    setup () {
      const matchMsg = 'match'
      const inputMsg = ref('matc')

      const canClear = ref(inputMsg.value !== matchMsg)

      watch(inputMsg, newVal => {
        canClear.value = newVal !== matchMsg
      })

      const iconOn = ref('mdi-check')
      const toggleIcon = () => {
        iconOn.value = iconOn.value === 'mdi-check' ? undefined : 'mdi-check'
      }

      const variants = ['outlined', 'solo', 'filled', 'plain', 'underlined', 'solo-inverted', 'solo-filled']
      const variant = ref('outlined')
      const cycleVariant = () => {
        const idx = variants.indexOf(variant.value)
        variant.value = variants[(idx + 1) % variants.length]
      }

      const expand2 = ref(false)

      return {
        inputMsg,
        canClear,
        matchMsg,
        iconOn,
        toggleIcon,
        variant,
        cycleVariant,
        expand2,
      }
    },
  }
</script>
```

johnleider:
@KaelWD can I get your thoughts?

samschurter:
I discovered a bug in the transition. It only animates smoothly on enter, on leave it is still janky at the end. My fix also only works for padding, if you apply margin to the element being animated, it breaks again. While I was playing with it I realized a different solution I had overlooked for the `VField` component.

If I remove the padding from the immediate child of the transition and add a new div with the padding to wrap the icon, then everything animates smoothly with no changes at all to the transition. It does break some of the icon styling, but I think that should be easy to fix. That would still leave the transition janky if the immediate child has either margin or padding applied in the direction of the transition, but that is probably better dealt with in a separate issue and PR. 

What do you think, continue down the road of updating the expand transitions to work in all cases with padding and margin, or backup and try this other approach that is limited to only modifying the `VField`?

Here is the markup that demonstrates the issue with VExpandXTransition when combined with margin or padding:
```vue
<template>
  <v-app>
    <v-container>
      <v-row justify="center" style="min-height: 160px">
        <v-col class="shrink">
          <v-btn class="ma-2" color="secondary" @click="expand2 = !expand2">
            Expand X Transition
          </v-btn>

          <div
            style="
              width: fit-content;
              height: 200px;
              background-color: red;
              border: 1px solid black;
              overflow: hidden;
            "
          >
            <v-expand-x-transition>
              <div
                v-show="expand2"
                style="padding: 2rem; background: blue; margin: 1rem"
              >
                <v-card
                  class="mx-auto bg-secondary"
                  height="100"
                  style="display: inline-block"
                  width="100"
                />
                <v-card
                  class="mx-auto bg-secondary"
                  height="100"
                  style="display: inline-block"
                  width="100"
                />
                <v-card
                  class="mx-auto bg-secondary"
                  height="100"
                  style="display: inline-block"
                  width="100"
                />
              </div>
            </v-expand-x-transition>
          </div>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>
<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const expand2 = ref(false)

      return {
        expand2,
      }
    },
  }
</script>
```

samschurter:
Hey @KaelWD, just wondering your thoughts on fixing the margin/padding issue for x-transitions only in the context of a `VField` icon, or fixing it more broadly by accounting for margin/padding in the `VExpandXTransition`.

johnleider:
> Hey @KaelWD, just wondering your thoughts on fixing the margin/padding issue for x-transitions only in the context of a `VField` icon, or fixing it more broadly by accounting for margin/padding in the `VExpandXTransition`.

We've never accounted for padding in Expand transition, so focus on VField first. If it's something that can easily translate we can debate that in the future.

johnleider:
Same as #20534, I'm putting this on hold until I move these components off of grid. I appreciate you going the extra mile but we shouldn't have to animate padding.

PR #20144, opened 2024-07-11
fix: augment GlobalComponents in multiple vue modules

guibulator:
## Description
Starting with the VScode Volar extension `v2.0.20`, Vuetify components when using auto imports will show as `Unknown` by IntelliSense. 
Screenshot 2024-07-11 at 8 41 43 AM

This module augmentation has been done in many projects so far, i.e: https://github.com/nuxt/nuxt/pull/26541

The idea is to augment modules `vue`, `@vue/runtime-dom` and `@vue/runtime-core`
After augmenting all the proper modules, IntelliSense starts to work again.
Screenshot 2024-07-11 at 8 43 17 AM

### DIFF
--- a/packages/vuetify/src/shims.d.ts
+++ b/packages/vuetify/src/shims.d.ts
@@ -16,9 +16,12 @@ declare global {
     }
   }
 }
-
+interface _GlobalComponents {
+  // @generate-components
+}
 declare module 'vue' {
   export type JSXComponent<Props = any> = { new (): ComponentPublicInstance<Props> } | FunctionalComponent<Props>
+  export interface GlobalComponents extends _GlobalComponents {}
 }
 
 declare module '@vue/runtime-dom' {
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/shims.d.ts
```diff
@@ -16,9 +16,12 @@ declare global {
     }
   }
 }
-
+export interface _GlobalComponents {
+  // @generate-components
+}
```
KaelWD, line 21:
```suggestion

interface _GlobalComponents {
  // @generate-components
}

```

trc-mathieu: oh yeah you're right, I forgot to remove the export, thanks

trc-mathieu: @KaelWD Let me know if you need something else before merging. Thank you!

johnleider (approved):

PR #20178, opened 2024-07-16
fix(VTimePicker): don't show buttons if ampmInTitle is not true

blalan05:


johnleider (approved):

KaelWD:
There's no other am/pm button, so without setting `ampmInTitle` or `format="24hour"` you can't use pm times at all. 

https://github.com/vuetifyjs/vuetify/pull/21595

PR #20179, opened 2024-07-16
fix(VTimePicker): emit when changing hour

blalan05:
```vue
<template>
  <v-time-picker v-model="myTime" format="24hr" label="Pick A Time" />

  {{ myTime }}
</template>
<script setup lang="ts">
  import { ref } from 'vue'
  const myTime = ref(null)
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VTimePicker/VTimePicker.tsx
+++ b/packages/vuetify/src/labs/VTimePicker/VTimePicker.tsx
@@ -264,8 +264,8 @@ export const VTimePicker = genericComponent<VTimePickerSlots>()({
           break
       }
 
-      const emitChange = selecting.value === (props.useSeconds ? SelectingTimes.Second : SelectingTimes.Minute)
-
+      const emitChange = selecting.value === (props.useSeconds ? SelectingTimes.Second : SelectingTimes.Minute) ||
+        (SelectingTimes.Hour && (inputMinute.value === 0 || !!inputMinute.value))
       if (selecting.value === SelectingTimes.Hour) {
         selecting.value = SelectingTimes.Minute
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/labs/VTimePicker/VTimePicker.tsx
```diff
@@ -264,8 +264,8 @@ export const VTimePicker = genericComponent<VTimePickerSlots>()({
           break
       }
 
-      const emitChange = selecting.value === (props.useSeconds ? SelectingTimes.Second : SelectingTimes.Minute)
-
+      const emitChange = selecting.value === (props.useSeconds ? SelectingTimes.Second : SelectingTimes.Minute) ||
+        (SelectingTimes.Hour && (inputMinute.value === 0 || !!inputMinute.value))
```
SonTT19, line 268: adding `SelectingTimes.Hour` is unnecessary cause it's a constants

`(inputMinute.value === 0 || !!inputMinute.value)` => can this be changed to `inputMinute.value !== null`?

Also, i assume that we need to emitValue whenever all of the `inputHour`, `inputMinute`, `inputSecond` is not null, so could this condition be changed to 
`const emitChange = inputHour.value !== null && inputMinute.value !== null && (props.useSeconds ? inputSecond.value !== null : true)`?

blalan05: Changed.

NickdeK:
Can this be reviewed and merged already? Seems like a small enough change the could help a lot of users.

YournameITManoah:
+1

johnleider (approved):

PR #20192, opened 2024-07-18
fix(VTreeview): don't expose open

userquin:
## Description

https://github.com/userquin/vuetify-nuxt-module/issues/243

## Markup:

```vue

```


yuwu9145 (changes requested):
No need to expose `opened` as it's available in `v-model:opened`.

Probably rebase it to dev and just remove `open`

PR #20202, opened 2024-07-19
feat: add new SSR and client features composables

userquin:
## Description

This PR tries to remove all global.ts usages including 2 new composables:
- useSSRHandler: by default will expose `isClient` and `isServer`, for Nuxt we can detect if missing from vuetify options (for example when not using my [vuetify-nuxt-module](https://github.com/userquin/vuetify-nuxt-module/tree/main)) and provide also the callback via hook registration (my nuxt module will include the SSR handler using `import.meta.client` and `import.meta.server`)
- useClientFeatures: moved all static constants in `globals.ts` module to expose them via SSRHandler

Maybe we also need to expose a global `isHydrated` in the SSRHandler to detect when `app:suspense:resolve` is called in Nuxt.

This PR in draft since I also need to remove the global.ts module and do some local tests.

Kael, John: `useHydration` composable should be changed to use also the new missing `isHydrated`.


KaelWD:
I don't really understand what the purpose of this change is.

PR #20211, opened 2024-07-22
fix(VNumberInput): prevent NaN & properly handle js number quirks

### LINKED ISSUE #19798: [Bug Report][3.6.5]  (VNumberInput): interpretes negative numbers as NaN

### LINKED ISSUE #20171: [Bug Report][3.6.12] VNumberInput range validation too eager
### Environment
**Vuetify Version:** 3.6.12
**Vue Version:** 3.4.31
**Browsers:** Edge 126.0.0.0
**OS:** Windows 10

### Steps to reproduce
1. Set `min` as 5, `max` as 125
2. Type 100

bug

### Expected Behavior
100 in VNumberInput

### Actual Behavior
The number in input becomes 125

Validation is too eager. It's hard to input a number actually allowed.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptUdtOg0AQ/ZUJMaHGFmyNPjTQqG+NH+CDNekWBrvpsmz2glbCvzuwLZakIWFnzpzZmXP2owmMzuIXpaLaYbAMEoulEsziaiMBknrGlOrDPskqaRmXqE9QD0pX7lDPuFTOnmEAwXYo0k1A34At9zzHdcejQsGEwVGVS4PXKyWXhD+OMfZD2HwxQmsmeE7bz6quQbDfI+yE0+NGNIZ9oSHC9n3PLKzBHhXmS7hp+qDdXvKfD3jMq+/uvkq++fi/vkrisQNnswgeu9Uh3swkvjCZUpNpriyQeOe95qWqtIUGNBbQQqGrEkJ6n7ArFk5mllcShm0mWKO0t9D40bwAj0S0OaRpCuEryw5GsQzDgQVedESOOYT0MouM4BlO7qcwm996dgtIb3K99y6FYdyJ3R39jyww1rNpBsmZhCFdmcReM6kN2ql0QkyDh+gpmi+Czz+7Sc0Q)

yuwu9145:
fixes #19798
fixes #20171

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <h1>{{ typed }}</h1>
      <v-number-input
        label=""
        clearable 
        v-model="typed"
        :hideInput="false"
        :inset="false"
        :min="5"
        :max="125"
        :step="0.32"
        validate-on="lazy blur"
        :messages="`What I typed: ${typed}`"
      ></v-number-input>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  // function onKeydown(event) {
  //   if (event.key === 'Backspace') {
  //     typed.value = typed.value.slice(0, -1)
  //   } else {
  //     typed.value += event.key
  //   }
  // }
  const typed = ref(-5)
</script>





```


J-Sek:
I can see number of test cases failing. In particular min/max enforcement. Did you meant to move it inside `onBeforeinput`? If so, the tests need to be updated as they check range enforcement even without user interaction.

There is also a bit of inconsistency in the emitted value type. Up/Down emit `number` while typing emits `string`. I am not sure which way is intended for this component.
- `string` model makes it somehow consistent with `<input type='number' />` when interacting through vanilla JS
- once native `<input type='number' />` is used with Vue `v-model` the value is of type `number` and that's what developers may expect when binding to `v-number-input`

The trajectory of the changes suggest we cannot go back to emitting `number`.
- users can use `v-model.number="..."` to transform values. It is not ideal though, as I get `-` as string out of it despite the modifier
- some `onBlur` action to clear leading and trailing zeros could be welcomed

Issues:
- `-` gets removed after typing `0` right after (when I use `v-model.number=...`)
- leading and trailing `0` get removed if model changes (when I use `v-model.number=...`) (maybe irrelevant, but caught my eye regardless)
- `0` is required to type `.` (maybe irrelevant)
- user cannot paste any value that has even the smallest divergence from underlying regex, so numbers copied with locale formatting, or spaces surrounding the value - may lead to "_this input is broken_" from user's perspective

J-Sek:
I am glad to see some progress :) We don't need the `.number` modifier, but it is a bit of a trade nonetheless.

Couple of weird behaviors I think are not exactly obvious, but may be important:
- (observation) component does not emit while typing - only on blur or when using up/down arrows
- (bug) `null` values get replaced with `0` - `clearable` does not work
- (bug) when I change value and use up/down arrows right after, the step is modifying last emitted value (e.g. `55{up}` » `1`)

yuwu9145:
> typing emits string

Fixed, it should only emit `number` or `null`

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -64,11 +69,20 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
...
 
-  setup (props, { attrs, emit, slots }) {
-    const model = useProxiedModel(props, 'modelValue')
+  setup (props, { slots }) {
+    const _model = useProxiedModel(props, 'modelValue')
+
+    const model = computed({
+      get: () => _model.value,
+      set (val) {
+        if (typeof val !== 'string') _model.value = val
```
KaelWD, line 78: Why aren't you using proxiedModel's transformOut argument?

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -39,18 +39,23 @@ const makeVNumberInputProps = propsFactory({
...
   },
   step: {
     type: Number,
     default: 1,
   },
 
-  ...omit(makeVTextFieldProps(), ['appendInnerIcon', 'prependInnerIcon']),
+  ...omit(makeVTextFieldProps({
+    modelValue: {
+      type: [Number, null],
```
KaelWD, line 55: `type: null` means any. Vue always allows props to be null but doesn't include it in the types for some reason.

KaelWD: Oh yeah propsFactory also only lets you set the default, this is basically
```
modelValue: {
  type: null,
  default: {
    type: [Number, null],
    default: null,
  }
}
```

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
```diff
@@ -8,6 +8,38 @@ import { VForm } from '@/components/VForm'
 import { ref } from 'vue'
 
 describe('VNumberInput', () => {
+  it('should prevent NaN from arbitrary input', () => {
+    const scenarios = [
+      { text: '---', expected: '-' }, // "-" is only allowed once
+      { text: '1-', expected: '1' }, // "-" is only at the start
+      { text: '.', expected: '' }, // "." isn't allowed at the start
```
KaelWD, line 15: `.5` == `.5`?

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
```diff
@@ -8,6 +8,38 @@ import { VForm } from '@/components/VForm'
...
+      { text: '1...0', expected: '1.0' }, // "." is only allowed once
+      { text: '123.45.67', expected: '123.4567' }, // "." is only allowed once
```
KaelWD, line 17: I guess this doesn't handle the european reversal of `,` and `.`?

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
yuwu9145: Coz I don't want it to emit anything when the value is a string, can `transform` conditionally skip an emit?

KaelWD: It skips if the new value is the same as the old value, idk how reliable that is though: https://github.com/vuetifyjs/vuetify/blob/8030d407c5767f96c0474d94828b2a79b40e7c17/packages/vuetify/src/composables/proxiedModel.ts#L58-L60

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
yuwu9145: Good catch, will tweak the regex

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
yuwu9145: At this stage, number-input only supports basic numeric input. Trying to get the basic function right.

In future, might consider integrating it with masks or a new `stringMode` prop to be able to support some advanced things

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
yuwu9145: They are definitely different when typing, it emits for sure

packages/vuetify/src/labs/VNumberInput/__tests__/VNumberInput.spec.cy.tsx
J-Sek: Or we can accept `locale` prop and extract separators ([example](https://github.com/J-Sek/vuetify/blob/feature-19898/packages/vuetify/src/util/helpers.ts#L805))

J-Sek:
All VTextField's `v-model={ model.value }` emits are strings, so the setter for computed `model` seems rather pointless. The intent could be more clear with going back to `modelValue={ model.value }`, because only `onChange` handler and `toggleUpDown` actually emit anything.

Regardless, I think internals will change yet again, but I don't want to stress about uncovered use case that bounces between _broken_ and _working_ states. Could you include following test scenarios? 😃 

```
{ text: '123{uparrow}', expected: '124' },
{ text: '521{downarrow}', expected: '520' },
```

yuwu9145:
> All VTextField's `v-model={ model.value }` emits are strings, so the setter for computed `model` seems rather pointless. The intent could be more clear with going back to `modelValue={ model.value }`, because only `onChange` handler and `toggleUpDown` actually emit numbers.
> 
> Regardless, I think internals will change yet again, but I don't want to stress about uncovered use case that bounces between _broken_ and _working_ states. Could you include following test scenarios? 😃
> 
> ```
> { text: '123{uparrow}', expected: '124' },
> { text: '521{downarrow}', expected: '520' },
> ```

The reason of double way binding is: in the case of `v-model=5`, `min=5` & `max=10` and they type "-1", I want input to be enforced back to `5`. Single way binding lose that enforcement because external v-model isn't mutated

https://play.vuetifyjs.com/#eNpdUEFugzAQ/MrIlyRSKOIakaj9Q3sqPbiwBKvGtsyaNkL8vTYkUZTLyjs7np3Zz0kMvs7fnHsZA4mDKJl6pyXTqTJAOWbSueW5NLU1LJUhf4Ui2BWn94sjSHPhTpkzLHfkwZ00KPZQjF+lNVoZy7esf8AWRZnHb3eJMWP646xVpBscetuQzkapAx0r0Q/nSuA1uCZ6ep6l7mOZRE5+s5k/+0zIGqPMH+LFdqi9coyBOKwpVe+sZ0zw1GJG622PTbzMJtGBqDswoiccE2Nb7Fa8DaZmZQ0eLG13mOa0cl0S5cW8N0HrpXz9A1y2f1E=

J-Sek:
> The reason of double way binding is: in the case of `v-model=5`, `min=5` & `max=10` and they type "-1", I want input to be enforced back to `5`.

And you achieve this with `onChange` handler. Check the example below. By removing `@change='toNumber'` you will get back to the same behavior as `:model-value='...'`, proving it is not `v-model` that helps you run `clampModel`, but `onChange` handler.

https://play.vuetifyjs.com/#eNpdkMFy4jAQRH+l4wumAri4spjd/YDltLeQg2LGoIosqeSxkhTlf8/ICgnJRR71jHqe++FS9KGp/nq/igMVm2LL1HmjmHYHC2zjUnk/ldOlcZaVthQ+JBHP693/N09Q9o3P2p7g+EwBfFYW6wU040Ubg1bJ8aSaZ7DDelvJs0+LuGR65WWryRwRl507kqkPRdef/qXyUOBPI3YnEpHdfuieKIhYXbmqn2BJydzb6uZ/5No3QXtGTzx4GPFMlv2hmJ7pzrvAuCBQu0DjOj8wHTGiDa7DTAKaJRNIy/YM4UOdZsv1/Js6UUvr6lBeMuiJeINyjnqXplZRmYEWuSVAZZzjY1BQWpQssboWEXd1jVnPQdKd3czgy0WWxas85mKcnEchS992sA1rZyX9nF9JkT+9MnkUF1FXrIKQZuPcv91zp/u92pf3Qvsb9xEb2MGYX9OylHdOWPIsxkXqTMfjO2KHxeo=

yuwu9145:
@J-Sek feel free to make a tweaking PR to `fix-19798-origin`, I will check it out tomorrow in my time

yuwu9145:
If no more comments, ready to merge in

J-Sek:
`min` and `max` are not enforced in some scenarios because model can become out-of-sync with VTextField's value. It does not seem to be specific to anything suggested in PR #20230. I may be wrong, but I wish somebody could take a look at my comment in that rejected PR.

johnleider (approved):

PR #20230, opened 2024-07-27
Fix 19798 tweaks

J-Sek:
PR consists of 2 independent parts:
- refactoring to make it clear **none** of the `update:model-value` emits from VTextField are accepted (because all of them provide type `string`, they were already filtered out by the `_model` wrapper)
- test cases (mostly failing) to guide further implementation tweaks
  - scenarios for mixed usage of typing (adding/removing digits) and incrementing value (up/down arrows)
  - additional scenarios to ensure consistent state after blur

There is one more advanced scenario that works around auto-cleanup on blur (initial typing for `-0.21` is actually optional)
1. focus input with **existing value** and change text without changing effective number
   - e.g. `-.21`, `-0000.21`, `-0.2100`
2. after blur the text stays the same
   - it might not be a critical bug, but it is inconsistent with behavior when we type a new value, developers might be bothered by their QA team about this

---
Supports #20211

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -69,14 +69,7 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
   },
 
   setup (props, { slots }) {
-    const _model = useProxiedModel(props, 'modelValue')
-
-    const model = computed({
-      get: () => _model.value,
-      set (val) {
-        if (typeof val !== 'string') _model.value = val
-      },
-    })
+    const model = useProxiedModel(props, 'modelValue')
 
     const vTextFieldRef = ref<VTextField | undefined>()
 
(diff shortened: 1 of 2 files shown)


yuwu9145:
As it has been mentioned before, this change breaks external v-model value enforcement

In this example, expected behaviour should be input value fall back to min 5

number-input-v-model


```vue
<template>
  <v-app>
    <v-container>
      <h1>{{ typeof typed }}</h1>
      <h1>{{ typed  }}</h1>
      <v-number-input
        label=""
        clearable 
        v-model="typed"
        :hideInput="false"
        :inset="false"
        :min="5"
        :max="125"
        :step="0.32"
      ></v-number-input>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const typed = ref(-5)
</script>

```

J-Sek:
@yuwu9145  It does, you are right... weird I missed it. My focus was on the behavior of unrestricted field.
- shouldn't we have some test cases for it?
- wouldn't it be more clear to put clam in onBlur to enforce it rather than relying on some magic side-effect of `@update:model-value`?

PR #20249, opened 2024-07-30
fix(VMenu): check if element still exists before closing parent

### LINKED ISSUE #20248: [Bug Report][3.6.13] VMenu closeParents error
### Environment
**Vuetify Version:** 3.6.13
**Vue Version:** 3.4.34
**Browsers:** Chrome 126.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
In the reproduction repo:
- click on the v-btn to open the menu
- then open the submenu
- lastly click on the `router-link` "leave" to trigger the error

### Expected Behavior
To not get an error

### Actual Behavior
`TypeError: Cannot read properties of null (reading 'contentEl')`
at https://github.com/vuetifyjs/vuetify/blob/3cc2763ec7638a3018f3d01ce491c22935eb67c2/packages/vuetify/src/components/VMenu/VMenu.tsx#L81

### Reproduction Link
[https://github.com/Arinono/vuetify-vmenu-bug](https://github.com/Arinono/vuetify-vmenu-bug)

Arinono:
## Description

fixes #20248 

Instead of using the `!` operator, I just added some null checks before using the `overlay.value.contentEl`.
I don't think it should cause any issue since the issue occurs when the view is completely destroyed (through the router in the reported case), so there is no cleanup to do anyways

## Markup:

I used the same reproduction repo I provided in the issue to validate the changes:
https://github.com/Arinono/vuetify-vmenu-bug
```vue
<v-menu v-if="open" :close-on-content-click="false">
      <v-btn v-bind="props">Open</v-btn>
    </template>
    <v-sheet class="pa-4">
      <div class="d-flex">
        <v-menu>
            <v-btn style="display: block" v-bind="props">Sub menu</v-btn>
          </template>
          <v-sheet class="pa-4"> Click back in the input</v-sheet>
        </v-menu>
        <v-text-field style="width: 400px"></v-text-field>
      </div>
    </v-sheet>
  </v-menu>

  <router-link to="other-route" style="margin-left: 48px"> leave </router-link>
```

and used [this section of the docs](https://vuetifyjs.com/en/getting-started/contributing/#vuetify) to test it

### DIFF
--- a/packages/vuetify/src/components/VMenu/VMenu.tsx
+++ b/packages/vuetify/src/components/VMenu/VMenu.tsx
@@ -78,7 +78,7 @@ export const VMenu = genericComponent<OverlaySlots>()({
         setTimeout(() => {
           if (!openChildren.value &&
             !props.persistent &&
-            (e == null || (e && !isClickInsideElement(e, overlay.value!.contentEl!)))
+            (e == null || (e && overlay.value && overlay.value.contentEl && !isClickInsideElement(e, overlay.value.contentEl)))
           ) {
             isActive.value = false
             parent?.closeParents()
--- a/packages/vuetify/src/components/VMenu/VMenu.tsx
+++ b/packages/vuetify/src/components/VMenu/VMenu.tsx
@@ -78,7 +78,7 @@ export const VMenu = genericComponent<OverlaySlots>()({
         setTimeout(() => {
           if (!openChildren.value &&
             !props.persistent &&
-            (e == null || (e && overlay.value && overlay.value.contentEl && !isClickInsideElement(e, overlay.value.contentEl)))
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VMenu/VMenu.tsx
```diff
@@ -78,7 +78,7 @@ export const VMenu = genericComponent<OverlaySlots>()({
         setTimeout(() => {
           if (!openChildren.value &&
             !props.persistent &&
-            (e == null || (e && !isClickInsideElement(e, overlay.value!.contentEl!)))
+            (e == null || (e && overlay.value && overlay.value.contentEl && !isClickInsideElement(e, overlay.value.contentEl)))
```
KaelWD, line 81:
```suggestion
            (e == null || (overlay.value?.contentEl && !isClickInsideElement(e, overlay.value.contentEl)))
```

PR #20266, opened 2024-08-01
docs: replace svg icons with css icons

userquin:
## Description

This PR will replace `@mdi/js` in the docs, but will still use the library and `@mdi/svg` for the mdi autocomplete in font-icons page.

There are a lot of svg icons on every page, using css icons will remove all svg icons from every page and vue component: since we're using same icons on every page (navbar, menu, code), UnoCSS will include only used icons once (svg solution will include the same icon multiple times in the dom if included multiple times).

Additionally, with css icons we can include any other iconify json library adding the corresponding `@iconify-json/<collection>` dev dependency and using it with vuetify components or directly with html markup :
```ts
<v-icon icon="i-logos:discord"></v-icon>
<div class="i-logos:discord"></div>
```

✔️ We need to review export examples logic, maybe we should change the icons back to `mdi-*`.

FIXED: Right now the docs cannot be built on my local, I'm checking what's happening (vue error without the source):
<details>
<summary>Error</summary>

image
</details>

## Markup:

NA

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -41,6 +41,7 @@
     "vuetify": "workspace:*"
   },
   "devDependencies": {
+    "@iconify-json/mdi": "^1.1.67",
     "@emailjs/browser": "^4.3.3",
     "@intlify/unplugin-vue-i18n": "^4.0.0",
     "@types/lodash-es": "^4.17.12",
@@ -69,6 +70,7 @@
     "markdown-it-meta": "^0.0.1",
     "markdown-it-prism": "^2.3.0",
     "markdownlint-cli": "^0.39.0",
+    "unocss": "^0.61.9",
     "unplugin-auto-import": "0.17.5",
     "unplugin-fonts": "1.0.3",
     "unplugin-vue-components": "^0.26.0",
--- a/packages/docs/src/components/Alert.vue
+++ b/packages/docs/src/components/Alert.vue
@@ -24,10 +24,10 @@
   })
   const icon = computed(() => {
     switch (type.value) {
-      case 'error': return 'mdi-close-circle-outline'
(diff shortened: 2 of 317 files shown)


userquin:
The entry point still huge 7.9MB, we'll need my other PR merged here. The resulting css is about 809KB, the original 599KB.

johnleider:
The only issue I see is that icon's won't work when you open an example in the Vuetify Playground.

userquin:
Using UnoCSS we can install VSCode or JetBrains extensions for icon preview and suggestions/autocomplete, for example, adding these 2 extensions to VSCode:
- [UnoCSS](https://marketplace.visualstudio.com/items?itemName=antfu.unocss)
- [Iconify IntelliSense](https://marketplace.visualstudio.com/items?itemName=antfu.iconify)

imagen

There is also UnoCSS extension for [JetBrains IDE](https://plugins.jetbrains.com/plugin/22204-unocss), for example, on IntelliJ Ultimate Edition (suggestions/autocomplete not working, I'm going to file in issue):

imagen

KaelWD:
> The resulting css is about 809KB, the original 599KB.

So it replaces 100K of js strings with 200k of CSS?

userquin:
Yes and no, it is adding only used icons, the optimal solution should be to include only in css navbar and playgroubd icons. The mdijs svg lib is huge.

Maybe this PR is not optimal with vuetify docs as a spa client app.

I Will check it using vue-scoped.

J-Sek:
> _The only issue I see is that icon's won't work when you open an example in the Vuetify Playground._

Not worth it in v3.x. Closing in favor of #22117 (only the iconset support)

The idea is not bad and I don't mind the PR being restarted in v5.x when we might adopt UnoCSS and rework big chunks of the framework.

PR #20293, opened 2024-08-07
chore: fix vuetify package build on Windows

userquin:
## Description

This PR fixes building vuetify package on Windows.

## Markup:

NA


userquin:
I need to fix this:

imagen

packages/vuetify/build/rollup.config.mjs
```diff
@@ -290,10 +288,11 @@ export default [
             source: variables.map(id => {
               return `@forward '` + path.join(
                 '../lib',
-                path.relative(fileURLToPath(new URL('../src', import.meta.url)), id),
+                // Normalize Windows Paths
```
userquin, line 291: remove this or move it

packages/vuetify/build/rollup.config.mjs
```diff
@@ -171,10 +168,11 @@ export default [
...
```
userquin, line 171: remove this or move it

KaelWD:
Does `upath` remove the need for all these `.replace(/\\/g, '/')`?

userquin:
> Does `upath` remove the need for all these `.replace(/\\/g, '/')`?

I guess no, we are using fs paths functions like join or relative. I can check it later.

PR #20317, opened 2024-08-12
refactor(Loader.tsx): Add optional chaining for slots

mrRajopadhyaya:
## Description

This solves the error `cannot read properties of undefined (reading 'default')` when compiling vue in compat mode with webpack.

```vue

```


KaelWD:
https://vuetifyjs.com/en/getting-started/frequently-asked-questions/#questions:~:text=Is%20Vuetify%203%20compatible%20with%20%40vue/compat

PR #20323, opened 2024-08-13
chore(playground): fix datepicker playground

userquin:
## Description

`VDateRangeInput` doesn't exist

imagen

## Markup:

Paste `Playground.datepicker.vue` code into `Playground.vue` and run dev script.


johnleider:
@iNSaNiA0821 can you add playgrounds for every major component? Either that, or we need to remove these because having some but not all is pointless.

johnleider (approved):
I'm opting to just remove these. Maybe we can add them for all components in the future.

PR #20327, opened 2024-08-13
chore: use pnpm shell emulator

userquin:
## Description

This PR enables [pnpm shell emulator](https://pnpm.io/cli/run#shell-emulator) and removes `cross-env` dev dependency and its usage from all package.json scripts: we only need to review the `dev:prod` script in `vuetify` package since we're using `concurrently` (using `NODE_ENV=production` for both, vite build and preview commands).

## Markup:

NA


KaelWD:
Turns out you don't even need `NODE_ENV=production`, `vite build` is already production by default.

PR #20339, opened 2024-08-15
Fix labelling

### LINKED ISSUE #19155: [Bug Report][3.5.2] V-Textarea Screenreaders get only partial information about content
### Environment
**Vuetify Version:** 3.5.2
**Vue Version:** 3.4.15
**Browsers:** Firefox 121.0
**OS:** Linux

### Steps to reproduce
- Use Screen Reader Orca (Linux) or JAWS (Windows)
- tab to or click to Textarea and listen

### Expected Behavior
In Browse mode:
- Label gets read
- Placeholder gets only read if no content is available (it's a placeholder after all)

### Actual Behavior
- The label NEVER gets read
- The placeholder ALWAYS gets read

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdj81uwkAMhF/F3UtaqSTqD5cqVOLWS2+Veuj2sCQGIu2fvCYCId4dbxYkxM0zY3k+/x1Voq5ZxliPO1QfqmV00RrGT+0B2nFmYpzGSXTBsxk80sWaTMY9G0ID48yFHu1CK5c2WoE1q0l9H8oolpzucBtsj1SCG0Pi5lrV3Hdlp6C0zQ2iyNTREBkS8q6QDi4GYjgC4RpOsKbgoJLvqrwOIHcTgxDCIm88Vl9obYDfQLZ/0Nr/oMSBtwKk/dL3YDaCUT3l5tIlLer0rN7q9/plrvIwr1/V/xkW9G3G)

### LINKED ISSUE #19156: [Bug Report][3.5.2] Screenreader never gets information about the label of a V-TextField / V-Select / etc.
### Environment
**Vuetify Version:** 3.5.2
**Vue Version:** 3.4.15
**Browsers:** Google Chrome, Mozilla Firefox
**OS:** Linux, Windows

### Steps to reproduce
Open playground link with Screenreader
Tab to v-select and listen. This also affects at least V-Textarea and V-Autocomplete

### Expected Behavior
The label is read and indicated as "name" of the combobox. 

See e.g. when reading out the example box here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

### Actual Behavior
Label is never read.
Instead, I only get something like "Open editable combobox 'Hello selected'" <-- Open is the label if I see correctly. The "real" label is not indicated as such for the screenreader.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdT8sKwjAQ/JVlL1HQFl8XqYI3L549WA+1rlJImpCkBSn9dzdpRfGUeSwzk0uHzpbpwZikbQi3mHlSRhae9nkNkLXzwpgIIyl17YuqJjtKUXQkqfTQzpW+k9zlqNwzR9hWHOWYxpcFWdyifXoNkKX0E53+ZwdlqM7Sn0lMXWkr48GRb4ZllTLaeujA0gN6eFitQPBvRDgH4FzngTfBLlxMxJGk1GL69eJAdi+jNQNx1lbexTW0D33chP0MV8k6WWwwgE2yxOsb6uxnXA==)

### Other comments
Partially related to https://github.com/vuetifyjs/vuetify/issues/19155

Martin-Idel:
Closes #19155
Closes #19156

## Content

This PR solves accessibility problems in selection components. The current behaviour of a select field is the following:
- The screen reader never reads the label
- When in multi-select mode, the screen reader never announces which fields are actually selected in the list

The two separate commits solve these two issues: The first commit actually correctly attaches the lable of the selectbox to its input so that screen readers can pick this up and associate the label with the input, while the second commit puts a label to the Multiselect boxes (VSelect and VAutocomplete) which screen readers can use to read out whether that entry is currently selected or not.

This PR should introduce no change in behaviour for anybody not using a screenreader. It was tested with the screenreaders orca and JAWS (some of the latest versions).

## Reproduction

Bug: Not reading out the label. Playground link for reproduction:

https://play.vuetifyjs.com/#eNpdT8sKwjAQ/JVlL1HQVlBQpArevHj2YD3UukohaUKSFqT0392kFcVT5rHMTC4dOlumB2OStiHcYuZJGVl42uc1QNbOC2MijKTUtS+qmuwoRdGRpNJDO1f6TnKXo3LPHGFbcZRjGl8WZHGL9uk1QJbST3T6nx2UoTpLfyYxdaWtjAdHvhmWVcpo66EDSw/o4WG1AsG/EeEcgHOdB94Eu3AxEUeSUovp14sD2b2M1gzEWVt5F9fQPvRxE/YzXCarZLnBANbJAq9v6ylnYQ== 
Using a screen reader, one can verify, that the label is not read out

Using the same Playground file with these fixes, the screen reader is now able to read it out. 

Playground file for convenience:
```vue
<template>
  <v-app>
    <v-container>
      <v-select v-model="msg" :items="items" label="My label" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref('Hello')
  const items = ['Hello', 'World']
</script>
```

Bug: Not reading out the label in text areas. Playground link for reproduction:

https://play.vuetifyjs.com/#eNpdj81uwkAMhF/F3UtaqSTqD5cqVOLWS2+Veuj2sCQGIu2fvCYCId4dbxYkxM0zY3k+/x1Voq5ZxliPO1QfqmV00RrGT+0B2nFmYpzGSXTBsxk80sWaTMY9G0ID48yFHu1CK5c2WoE1q0l9H8oolpzucBtsj1SCG0Pi5lrV3Hdlp6C0zQ2iyNTREBkS8q6QDi4GYjgC4RpOsKbgoJLvqrwOIHcTgxDCIm88Vl9obYDfQLZ/0Nr/oMSBtwKk/dL3YDaCUT3l5tIlLer0rN7q9/plrvIwr1/V/xkW9G3G

Using the same Playground file with these fixes, the screen reader is now able to read it out. 

Playground file for convenience:

```vue
<template>
  <v-app>
    <v-container>
      <v-textarea v-model="msg" label="My label" placeholder="My placeholder" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref('Hello World!\nTest other\nAnd again')
</script>
```

Bug: No information about which boxes are clicked in multiselect. Playground link for reproduction:

https://play.vuetifyjs.com/#eNpdkEFrwzAMhf+K8MUbtMmgg42SDnbbZecdmhyyVB0GOTa2Ehil/72y05C2J0vvE3pP3p9UDF356X0xDqi2qmK0nlrGj7oHqMZ1630uc9O5nlvTY7hKWYxI2PEsAIxr6w5Iu1rZ+FerBWztQGw8oSAOA94xI8ZRQH5vCbW/edn3/1QurJyDlY/JkjIFr8qbg6SNXTCeISIP013GehcYThDwCGc4BmdBy1/oNA4geyODHAK7NPG0119I5HTzvNAcWfjMVqB/XKCDbpL/5Che6rxSm+K12LyrVLwVL6q5AE2jduo=

Using the same Playground file with these fixes, the screen reader is now able to read it out. 

Playground file for convenience:

```vue
<template>
  <v-app>
    <v-container>
      <v-select
        v-model="msg"
        :multiple="true"
        :items="items"
        label="My label"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref(['Hello'])
  const items = ['Hello', 'World']
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -520,6 +520,7 @@ export const VAutocomplete = genericComponent<new <
                                       modelValue={ isSelected }
                                       ripple={ false }
                                       tabindex="-1"
+                                      aria-label={ isSelected ? 'checked' : 'not checked' }
                                     />
                                   ) : undefined }
 
--- a/packages/vuetify/src/components/VField/VField.tsx
+++ b/packages/vuetify/src/components/VField/VField.tsx
@@ -73,6 +73,7 @@ export const makeVFieldProps = propsFactory({
     type: Boolean,
     default: null,
   },
+  labelId: { type: String, default: undefined },
   error: Boolean,
   flat: Boolean,
   label: String,
@@ -303,7 +304,7 @@ export const VField = genericComponent<new <T>(
               </VFieldLabel>
             )}
 
-            <VFieldLabel ref={ labelRef } for={ id.value }>
+            <VFieldLabel id={ props.labelId } ref={ labelRef } for={ id.value }>
               { label() }
             </VFieldLabel>
 
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -439,6 +439,7 @@ export const VSelect = genericComponent<new <
                                         modelValue={ isSelected }
                                         ripple={ false }
                                         tabindex="-1"
+                                        aria-label={ isSelected ? 'checked' : 'not checked' }
                                       />
                                     ) : undefined }
 
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -375,7 +375,6 @@ export const VSelect = genericComponent<new <
           onMousedown:control={ onMousedownControl }
           onBlur={ onBlur }
           onKeydown={ onKeydown }
-          aria-label={ t(label.value) }
           title={ t(label.value) }
         >
           {{
--- a/packages/vuetify/src/components/VTextField/VTextField.tsx
+++ b/packages/vuetify/src/components/VTextField/VTextField.tsx
@@ -198,6 +198,7 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
                 role={ props.role }
                 { ...fieldProps }
                 id={ id.value }
+                labelId={ `${id.value}-label` }
                 active={ isActive.value || isDirty.value }
                 dirty={ isDirty.value || props.dirty }
                 disabled={ isDisabled.value }
@@ -226,6 +227,7 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
                         type={ props.type }
                         onFocus={ onFocus }
                         onBlur={ blur }
+                        aria-labelledby={ `${id.value}-label` }
                         { ...slotProps }
                         { ...inputAttrs }
                       />
--- a/packages/vuetify/src/components/VTextarea/VTextarea.tsx
+++ b/packages/vuetify/src/components/VTextarea/VTextarea.tsx
@@ -253,6 +253,7 @@ export const VTextarea = genericComponent<VTextareaSlots>()({
                 { ...fieldProps }
                 id={ id.value }
                 active={ isActive.value || isDirty.value }
+                labelId={ `${id.value}-label` }
                 centerAffix={ rows.value === 1 && !isPlainOrUnderlined.value }
                 dirty={ isDirty.value || props.dirty }
                 disabled={ isDisabled.value }
@@ -287,6 +288,7 @@ export const VTextarea = genericComponent<VTextareaSlots>()({
                         name={ props.name }
                         onFocus={ onFocus }
                         onBlur={ blur }
+                        aria-labelledby={ `${id.value}-label` }
                         { ...slotProps }
                         { ...inputAttrs }
                       />


packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -520,6 +520,7 @@ export const VAutocomplete = genericComponent<new <
                                       modelValue={ isSelected }
                                       ripple={ false }
                                       tabindex="-1"
+                                      aria-label={ isSelected ? 'checked' : 'not checked' }
```
johnleider, line 523: Does this need to use i18n?

packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -440,6 +439,7 @@ export const VSelect = genericComponent<new <
                                         modelValue={ isSelected }
                                         ripple={ false }
                                         tabindex="-1"
+                                        aria-label={ isSelected ? 'checked' : 'not checked' }
```
johnleider, line 442: Does this need to use i18n?

johnleider (changes requested):
Are there any other inputs that we need to consider this for?

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
Martin-Idel: Probably makes sense. I tried to add i18n (added English for most languages, I hope that would be okay).

Martin-Idel:
Regarding other parts where this might be needed: It needs to be multiselect checkboxes within a listbox and I could only find those in those two components - it works correctly for regular checkboxes.

KaelWD:
> Bug: Not reading out the label.

VFieldLabel already has `for`, it shouldn't need `id` and `aria-labeledby` as well if VSelect didn't also add `aria-label` to the input element (#17786)

> Bug: Not reading out the label in text areas.

#21025, there's two labels with `for` but both are hidden when the field is active.

> Bug: No information about which boxes are clicked in multiselect.

NVDA at least already reads "Hello not selected 1 of 2", adding the label to the checkbox seems redundant because now it says "Not checked hello not selected 1 of 2". Does JAWS support `aria-selected` and if so why isn't our current implementation of it working?

J-Sek:
I don't see VCombobox among changes. Is it just an oversight or intentional?

Martin-Idel:
@KaelWD 

> > Bug: Not reading out the label.
> 
> VFieldLabel already has for, it shouldn't need id and aria-labeledby as well if VSelect didn't also add aria-label to the input element (https://github.com/vuetifyjs/vuetify/pull/17786)

So what you are suggesting is that removing the aria-label for the input element should be enough here? That sounds correct, I'll give that a try.

> > Bug: Not reading out the label in text areas.
>
> https://github.com/vuetifyjs/vuetify/issues/21025, there's two labels with for but both are hidden when the field is active.

I guess what you are suggesting is that the labels aria-hidden or visibility: hidden are not switched correctly between variants so your suggestion is to fix that bug instead of my approach here. Will you do this and I just remove my approach here?

> >  Bug: No information about which boxes are clicked in multiselect.
> 
> NVDA at least already reads "Hello not selected 1 of 2", adding the label to the checkbox seems redundant because now it says "Not checked hello not selected 1 of 2". Does JAWS support aria-selected and if so why isn't our current implementation of it working?
On android talkback I can't even get to the list items, the menu closes as soon as I try to focus it.

Both JAWS and orca should in principle support "aria-selected". Checking my own reproduction link above I can now no longer reproduce it with orca, so maybe it was fixed on the screenreader's end. I'll ping my colleague to see whether they can still reproduce it with JAWS.

packages/vuetify/src/components/VField/VField.tsx
```diff
@@ -299,7 +300,7 @@ export const VField = genericComponent<new <T>(
             )}
 
             { hasLabel.value && (
-              <VFieldLabel key="label" ref={ labelRef } for={ id.value }>
+              <VFieldLabel key="label" id={ props.labelId } ref={ labelRef } for={ id.value }>
```
johnleider, line 303: What does this solve if we're not using aria-labelledby?

johnleider (dismissed):

J-Sek (changes requested):
It seems like we need just a few lines to include VCombobox in scope of these changes

robertzak:
Is there a timeline on when this might be merged?

KaelWD:
Nothing had changed since my review, now the third example reads "Not checked world not selected not checked editable combobox not checked" in orca.

J-Sek:
Orca glitches so much lately I slowly lose my trust in this tool. For me it only says "_not checked world not selected_" when selecting unchecked value in scenario 3. This is already redundant enough to get us bug ticket from somebody if it repeats in other screen readers.

Qualifies for revert, unless the authors can quickly share the recording proving it improved anywhere @Martin-Idel ,  @ikushum

KaelWD:
Focus field
Open the menu
Navigate to the first item
Navigate to the second item
Select the second item

### Orca

#### 3.11.2
My label clickable (does nothing)
Open open opens popup
Expanded
My label dash list list with 2 items hello hello editable combobox checked
World not selected world editable combobox not checked
Selected

#### 3.11.2-master.2025-12-08
My label clickable (does nothing)
My label my label open opens popup
Expanded
My label dash list list with 2 items checked hello checked editable combobox checked
Not checked world not selected not checked editable combobox not checked
Checked world checked editable combobox checked


### NVDA
#### 3.11.2
Combobox collapsed clickable opens list
Close combobox expanded editable opens list hello
My label list list hello 1 of 2
World not selected 2 of 2
Selected

#### 3.11.2-master.2025-12-08
Combobox collapsed clickable opens list
My label combobox expanded editable opens list selected hello
My label list list checked hello 1 of 2
Not checked world not selected 2 of 2
Checked my label list selected

### Android talkback
#### 3.11.2
My label my label double tap to activate
Expanded hello editbox hello expanded actions available
Selected hello 1 of 2 in listbox my label list 2 items double tap to select
World 2 of 2 double tap to select
World ticked

#### 3.11.2-master.2025-12-08
My label my label double tap to activate
Expanded hello editbox hello expanded actions available
Selected checked hello 1 of 2 in listbox my label list 2 items double tap to select
Not checked world 2 of 2 double tap to select
Checked world ticked

PR #20343, opened 2024-08-15
fix(VTextField): Use "appearance: none" instead of "opacity: 0" to not break a11y of VTextField

segevfiner:
## Description

By making the field invisible, a11y software, and password managers, might consider it hidden, and fail to allow autofill.

See https://github.com/keepassxreboot/keepassxc-browser/issues/2311

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <VTextField label="Email" autocomplete="username" />
      <VTextField type="password" label="Password" autocomplete="current-password" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```


johnleider:
Does the unit test need to be updated?

segevfiner:
Fixed the tests. (Didn't do so originally as I had a problem running the tests locally for whatever reason)

johnleider (approved):

xqqp:
The `appearance` CSS property does not work as the PR author expects. From MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/appearance#none): "If the element is a widget (native form control), it will be forced to use a standardized primitive appearance instead of a platform-native or operating system specific appearance, supporting the usual rules of CSS."

So the appearance property controls the styling, not the visibility. See also in this example, both inputs are visible despite `appearance: none` being set: https://play.vuetifyjs.com/#eNqNkM9OwzAMxl/lI5eC1LUqExIqHRI33oAD5ZCmLqpIk8hJJ6Zp706yIu3PaTf7s/2zP3/uhWdVvjlXbGcStWgCTU7LQK+tAZrtSjp3DI+JsibI0RD/S1EcjZsDfNhp2rSis9wT16jcLzot1Q+81WP/goghydIoqmGsoVagPEE6htLS+0iYdqvqsnjThrORprw+NCmLj6Y88xdTr3h0EU5hXmyOk7McsAfTgAMGthOy+JostQOR6wMm/41N6rjP3klriw/Lur/LHtKCBRlh4pCLdfFUVGuRgufiUeSD1J7yi58v2tcfj1t/Kg==

 The reason why this PR still works is because it removes `opacity: 0`. So instead of replacing opacity with appearance, the opacity modifier should just be removed without replacement to make it work with password managers.

segevfiner:
Without `appearence: none` you might need more CSS to make the native widget not show up AFAIK. As the code draws over it using div and other CSS rather than customizing it's base appearance. With `appearance: none` you can just use normal background and border to make it invisible.

But it's been quite a while and I don't remember the details of what I tried when I made this.

segevfiner:
Hmmm. Probably since removing appearance does remove some UA style rules that will need to be compensated for...

J-Sek:
~~Could `opacity: 0.0001` be a good-enough solution?~~

Bitwarden expects at least `0.1`

PR #20366, opened 2024-08-20
feat(VNumberInput): emit events when using controls

### LINKED ISSUE #20342: [Feature Request] [VNumberInput] Click event support for increment / decrement buttons
### Problem to solve
I tried to used this component to send API requests when on value update by typing and on clicking on the increment and decrement buttons.

According to my use case I wanted to throttle the keyup event separately and emit right after clicking on the increment / decrement buttons.

But as per the current implementation increment / decrement value change fetched under `@update:model-value`

### Proposed solution
I propose to add something like below where we can separately handle the click event on increment / decrement buttons

`@click:increment` / `@click:decrement` or one event for both

J-Sek:
## Description

resolves #20342

## Markup

```vue
<template>
  <v-app>
    <v-container>
      <v-number-input
        v-model="someValue"
        :loading="saving"
        clearable
        @blur="save"
        @click:clear="save"
        @click:control="save"
      />
      <pre class="mt-3">lastSavedValue: {{ lastSavedValue }}</pre>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const someValue = ref(0)
  const lastSavedValue = ref(0)
  const saving = ref(false)

  // simplified example - it would be debounced and guarded against race conditions in real life
  async function save() {
    const v = someValue.value
    if (v !== lastSavedValue.value) {
      saving.value = true
      await new Promise(r => setTimeout(r, 1000))
      lastSavedValue.value = v
      saving.value = false
    }
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -30,6 +30,7 @@ type VNumberInputSlots = Omit<VTextFieldSlots, 'default'> & {
 }
 
 type ControlVariant = 'default' | 'stacked' | 'split'
+type ControlClickEvent = { type: 'increase' | 'decrease' }
 
 const makeVNumberInputProps = propsFactory({
   controlVariant: {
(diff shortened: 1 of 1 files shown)


KaelWD:
#20342 was closed by the author.

PR #20379, opened 2024-08-23
docs(VTooltip): fix usage example button text

santarosalia:
fix button text

### DIFF
--- a/packages/docs/src/examples/v-tooltip/usage.vue
+++ b/packages/docs/src/examples/v-tooltip/usage.vue
@@ -45,7 +45,7 @@
   const slots = computed(() => {
     return `
   <template v-slot:activator="{ props }">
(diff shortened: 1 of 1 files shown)


packages/docs/src/examples/v-tooltip/usage.vue
```diff
@@ -45,7 +45,7 @@
   const slots = computed(() => {
     return `
   <template v-slot:activator="{ props }">
-    <v-btn v-bind="props">Tooltip</v-btn>
+    <v-btn v-bind="props">HOVER ON ME</v-btn>
```
pano9000, line 48: just commenting on the grammar here:
This should rather be "Hover over me"

santarosalia: thats good point

Issue #20389, opened 2024-08-24
[Bug Report][3.7.0] Using the v-treeview directory tree to duplicate node and parent node names leads to an endless loop


zulezhe:
### Environment
**Vuetify Version:** 3.7.0
**Vue Version:** 3.3.4
**Browsers:** Edge 128.0.0.0
**OS:** Windows

### Steps to reproduce
I want to use the v-treeview to display my project directory. I also have the src directory under my src directory. When I click src to expand it, I get stuck in a loop

### Expected Behavior
available

### Actual Behavior
not available

### Reproduction Link
[...]

Issue #12286, opened 2020-09-25
[Bug Report] v-treeview UI doesnt update correctly when selecting/deselecting identical objects


TJBrunson:
### Environment
**Vuetify Version:** 2.3.10
**Vue Version:** 3.0.0
**Browsers:** Chrome 85.0.4183.121
**OS:** Mac OS 10.15.6

### Steps to reproduce
Make a selectable tree view with the "vaccine" data object in the JSFiddle provided. Play with selecting different rows. For instance, selected the first instance of "2008-2009 trivalent influenza vaccine". Both instances will become selected as expected and the v-model is updated accordingly. Then try to deselect the second instance of "2008-2009 trivalent influenza vaccine" and notice the first instance is not de-selected, but the v-model is updated correctly. Seems like a UI bug that doesn't re-render all checkboxes correctly. This issue is present in Chrome, Brave, and Safari. I have not checked other browsers.

### Expected Behavior
When identical objects are present in different parts of the hierarchy, selecting and deselecting one of them should update the ui for every instance of the object.

### Actual Behavior
When selecting the first instance of an object, all other instances of the object become selected. Then, when deselecting a later instance of the object, previous instances are not de-selected.

### Reproduction Link
<a href="https://jsfiddle.net/hugbtknz/" target="_blank">https://jsfiddle.net/hugbtknz/</a>

nekosaur:
The fiddle does not work

TJBrunson:
Yeah, for some reason fiddle doesn't recognize the v-treeview component. This should be everything required to replicate it in a dev environment. I just have no experience with JSFiddle, so I don't know how to correctly include the component. Vuetify does it automatically in the dev environment.

nekosaur:
https://template.vuetifyjs.com

TJBrunson:
https://codepen.io/tjbrunson/pen/RwRbKGY

Try this

johnleider:
Leaves must have a unique ID. Removing the dup values solves the issue: https://codepen.io/johnjleider/pen/MWeWGjm

This Pull Request is being closed due to inactivity.

If you have any additional questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com).

TJBrunson:
Sure, thats one way to solve the problem, but what about trees like mine where objects can be duplicated? It's a real world use case that people may run into. Shouldn't it be supported?

johnleider:
[...]

PR #20390, opened 2024-08-25
fix(nested): Prevent infinite loops when resolving path

J-Sek:
## Description

While loops used by components with nesting need some safety check to prevent freezing end-users tab. Performance overhead of the checks is close to none, while it makes a big difference in the field.

Error does not necessarily need to be developers fault. Data from the API may contain duplicates or it may be a result of some drag & copy functionality built on top of Vuetify's components.

I was advised to log a warning, but I just don't think it is better than throwing error
- warning inside while loop does not help – tabs usually freeze DevTools as well and devs might work with simplified data, leaving potential for errors in the field
- warning after validating `path`/`parents` collection – unnecessary performance overhead if we can have cheaper solutions
- warning + breaking from the loop – risky... potentially misleading end-users if they click things fast and don't pay attention, and won't get logged in monitoring (like Sentry or Bugsnag)

mitigates #20389

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container>
      <h4>Tree view</h4>
      <v-treeview :items="items" item-value="title" open-on-click />
      <v-divider class="my-6" />
      <h4>List</h4>
      <v-list v-model:opened="opened" v-model:selected="selected" select-strategy="classic">
        <v-list-group value="src-group">
            <v-list-item v-bind="props" title="src">
              <template #prepend="{ isSelected }">
                <v-list-item-action start>
                  <v-checkbox-btn :model-value="isSelected" />
                </v-list-item-action>
              </template>
            </v-list-item>
          </template>
          <v-list-group value="renderer-group">
              <v-list-item v-bind="props" title="renderer">
                <template #prepend="{ isSelected }">
                  <v-list-item-action start>
                    <v-checkbox-btn :model-value="isSelected" />
                  </v-list-item-action>
                </template>
              </v-list-item>
            </template>
            <v-list-group value="src-group">
                <v-list-item v-bind="props">
                  <template #title>src ––– <span class="text-red">toggle selection here on one level below</span></template>
                  <template #prepend="{ isSelected }">
                    <v-list-item-action start>
                      <v-checkbox-btn :model-value="isSelected" />
                    </v-list-item-action>
                  </template>
                </v-list-item>
              </template>
              <v-list-group value="assets-group">
                  <v-list-item v-bind="props">
                    <template #title>assets ––– <span class="text-red">expand me or the parent</span></template>
                    <template #prepend="{ isSelected }">
                      <v-list-item-action start>
                        <v-checkbox-btn :model-value="isSelected" />
                      </v-list-item-action>
                    </template>
                  </v-list-item>
                </template>
                <v-list-item title="styles.sass" value="4 - Leaf">
                  <template #prepend="{ isSelected }">
                    <v-list-item-action start>
                      <v-checkbox-btn :model-value="isSelected" />
                    </v-list-item-action>
                  </template>
                </v-list-item>
              </v-list-group>
            </v-list-group>
          </v-list-group>
        </v-list-group>
      </v-list>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const opened = ref(['src-group', 'renderer-group', 'assets-group'])
  const selected = ref([])

  const items = ref([
    {
      title: 'src',
      children: [
        {
          title: 'renderer',
          children: [
            {
              title: 'src',
              children: [
                {
                  title: 'assets',
                  children: [
                    { title: 'styles.sass', isFileNode: true },
                  ],
                },
              ],
            },
          ],
        },
      ],
    },
  ])
</script>
```

### DIFF
--- a/packages/vuetify/src/composables/nested/nested.ts
+++ b/packages/vuetify/src/composables/nested/nested.ts
@@ -18,6 +18,7 @@ import {
   leafSingleSelectStrategy,
 } from './selectStrategies'
 import { getCurrentInstance, getUid, propsFactory } from '@/util'
+import { preventLoops } from '@/util/nested'
 
 // Types
 import type { InjectionKey, PropType, Ref } from 'vue'
@@ -182,6 +183,7 @@ export const useNested = (props: NestedProps) => {
     let parent: unknown = id
 
     while (parent != null) {
+      preventLoops(path, parent)
       path.unshift(parent)
       parent = parents.value.get(parent)
     }
--- a/packages/vuetify/src/composables/nested/nested.ts
+++ b/packages/vuetify/src/composables/nested/nested.ts
@@ -17,8 +17,7 @@ import {
   leafSelectStrategy,
   leafSingleSelectStrategy,
 } from './selectStrategies'
-import { getCurrentInstance, getUid, propsFactory } from '@/util'
-import { preventLoops } from '@/util/nested'
+import { consoleError, getCurrentInstance, getUid, propsFactory } from '@/util'
 
 // Types
 import type { InjectionKey, PropType, Ref } from 'vue'
@@ -77,6 +76,7 @@ type NestedProvide = {
     activate: (id: unknown, value: boolean, event?: Event) => void
     select: (id: unknown, value: boolean, event?: Event) => void
     openOnSelect: (id: unknown, value: boolean, event?: Event) => void
+    getPath: (id: unknown) => unknown[]
   }
 }
 
@@ -99,6 +99,7 @@ export const emptyNested: NestedProvide = {
     activated: ref(new Set()),
     selected: ref(new Map()),
     selectedValues: ref([]),
+    getPath: () => [],
   },
 }
 
@@ -183,7 +184,6 @@ export const useNested = (props: NestedProps) => {
     let parent: unknown = id
 
     while (parent != null) {
-      preventLoops(path, parent)
       path.unshift(parent)
       parent = parents.value.get(parent)
     }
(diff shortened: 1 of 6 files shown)


J-Sek:
Note: I searched for `while (parent`. Hopefully it has fund all the places.

I wish they all had unified conditions `if (parent) { ...`. Some implementation are straight copy paste, but I don't feel like I am in a position to make a call for deeper refactoring.

KaelWD:
I moved this to register instead which also detects duplicate siblings.

Should we skip this check for leaf nodes? #12286 seems to work fine now if the only duplicates are leaves, getPath will be returning the wrong value for these still though and if only one of the duplicates is unmounted it will break the others because unregister cleans up all the references.

yuwu9145 (commented):
I think this PR by this far is good enough, duplicate leaf nodes should be warned as well because it confuses selected/activated values.

Generally speaking, duplicating item-value shouldn't be allowed/encouraged (it's almost impossible to properly handled from our end), if developers don't have confidence of avoiding item-value duplication in items, then using `return-object` is a graceful way to cope with the issue.

I think we should also make this clear as a "Ceveat" in document in addition to the warning

packages/vuetify/src/composables/nested/nested.ts
```diff
@@ -209,6 +213,15 @@ export const useNested = (props: NestedProps) => {
         return arr
       }),
       register: (id, parentId, isGroup) => {
+        if (nodeIds.has(id)) {
+          const path = getPath(id).join(' -> ')
```
CrazyOrr, line 217: `getPath(id)` can be a `Symbol`, which throws `TypeError: Cannot convert a Symbol value to a string`.

KaelWD: 2728a10244f4d9ae4268621e0cf310d8f29dab44

PR #20419, opened 2024-09-03
docs: fix PWA

userquin:
## Description

This PR includes:
- remove old `service-worker.js` and all the caches: self destroying (for context check [Custom selfDestroying Service Worker](https://vite-pwa-org.netlify.app/guide/unregister-service-worker.html#custom-selfdestroying-service-worker))
- new `sw.js` with `Skip-Waiting Service Worker-based solution`: `self.skipWaiting()` + `activate` sw listener to reload all the clients
- the new `sw.js` will allow work offline: current `service-worker.js` using weird logic for the fallback and same origin assets (css, js...); the new logic will use `Stale-While-Revalidate`, checking first in the `MANIFEST` and in the caches to serve the fallback or the corresponding asset from  the cache, if missing then fetch + store response in caches
- custom sw registration in `plugins/pwa.ts` with periodic sync for updates: we can change the interval (right now every hour will check for `sw.js` update)
- update pwa plugin to `0.23.0`
- removed `vuetify:ssg` vite plugin: we don't use generated html files
- using `index.html` as navigate fallback and removed/updated `_fallback.html` usages
- change some `injectManifest` options in the pwa plugin options to use the default ones excluding the old  `service-worker.js` and all `*.map` assets from the sw injection point

Pending tasks:
- [ ] apply `Stale-While-Revalidate` for assets from the same origin
- [ ] review `install` listener to create the external cache (external fonts, css...)
- [ ] check if the caches are correctly updated: remove old entries and insert the new ones
- [ ] check if we can switch to `workbox` with some pnpm patch for `async-es/eachLimit`: iirc there is some comment from Kael in workbox repo
- [ ] Review `vercel.json` cache headers: any navigation page will redirect/forward to `index.html` (SPA), shouldn't be cached and should be revalidated (same as `sw.js` or old `service-worker.js`); any asset from `assets` folder should be inmmutable with long max-age and public => check response headers for [Old Vuetify Nuxt Module docs](https://vuetify-nuxt-module.netlify.app/)  deployed at Netlify and compare with the [New Vuetify Nuxt Module docs](https://nuxt.vuetifyjs.com/) deployed at Vercel (both sites using VitePress, we've all pages generated with SSG) or current Vuetify docs
- [x] Add the new `sw.js`  to `vercel.json`: copy/paste the `service-worker.js` and rename the url
- [ ] Change `vercel.json` rewrites to use the `index.html` in the redirection/forward
- [ ] Review if we want to add api calls even to external origin to the caches
- [ ] cleanup old code and some `console.log`  from the codebase

The current version using Node 22.6.0 in my Windows laptop doesn't end, the `buildEnd` hook in the Vite PWA build plugin not being called, I need to stop the build via `Ctrl + C` (maybe `pnpm 9.6.0`, we should update to 9.9.0 and check)

### DIFF
--- a/packages/docs/vite.config.mts
+++ b/packages/docs/vite.config.mts
@@ -20,7 +20,7 @@ import MagicString from 'magic-string'
 import { configureMarkdown } from './build/markdown-it'
 import Api from './build/api-plugin'
 import { Examples } from './build/examples-plugin'
-import { genAppMetaInfo } from './src/utils/metadata'
+// import { genAppMetaInfo } from './src/utils/metadata'
 import { MdiJs } from './build/mdi-js'
 import { frontmatterBuilder, getRouteMeta, scriptFixer } from './build/markdownBuilders'
 
@@ -206,15 +206,20 @@ export default defineConfig(({ command, mode, isSsrBuild }) => {
       // https://github.com/antfu/vite-plugin-pwa
       VitePWA({
         srcDir: 'src',
-        filename: 'service-worker.js',
+        filename: 'sw.js',
         strategies: 'injectManifest',
         includeAssets: ['favicon.ico'],
+        injectRegister: false,
         injectManifest: {
-          globIgnores: ['**/*.html'],
-          additionalManifestEntries: [
-            { url: '/_fallback.html', revision: Date.now().toString(16) },
-          ],
-          dontCacheBustURLsMatching: /assets\/.+[A-Za-z0-9]{8}\.(js|css)$/,
+          minify: false,
+          enableWorkboxModulesLogs: true,
+          // globIgnores: ['**/*.html'],
+          globIgnores: ['assets/*.map', 'service-worker.js'],
+          // globPatterns: ['**/*.{js,css}', 'index.html'],
+          // additionalManifestEntries: [
+          //   { url: '/_fallback.html', revision: Date.now().toString(16) },
+          // ],
+          // dontCacheBustURLsMatching: /assets\/.+[A-Za-z0-9]{8}\.(js|css)$/,
           maximumFileSizeToCacheInBytes: 24 * 1024 ** 2,
         },
         manifest: {
@@ -301,7 +306,7 @@ export default defineConfig(({ command, mode, isSsrBuild }) => {
           }
         },
       },
-
+/*
       {
         // lightweight head-only ssg
         name: 'vuetify:ssg',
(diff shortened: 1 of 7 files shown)


packages/docs/vite.config.mts
```diff
@@ -206,15 +206,20 @@ export default defineConfig(({ command, mode, isSsrBuild }) => {
...
         includeAssets: ['favicon.ico'],
+        injectRegister: false,
         injectManifest: {
-          globIgnores: ['**/*.html'],
-          additionalManifestEntries: [
-            { url: '/_fallback.html', revision: Date.now().toString(16) },
-          ],
-          dontCacheBustURLsMatching: /assets\/.+[A-Za-z0-9]{8}\.(js|css)$/,
+          minify: false,
+          enableWorkboxModulesLogs: true,
```
userquin, line 215: we're not using  `workbox`

MajesticPotatoe: VitePWA is built on top of workbox

userquin: no, you can use it without using any workbox module in the sw and in the client

userquin: check the 4th task in pending task list

userquin: we only use here `injectManifest` module in the build to add the injection point (MANIFEST) in the sw.js

MajesticPotatoe: sorry, the comment seems conflicting

userquin: I'll try to explain, there are 3 `workbox` packages types:
- runtime
  - client logic for sw registration and interaction: `workbox-window` 
  - sw packages with some recipes to simplify sw logic like `workbox-expiration`, `workbox-precaching`, there are more
- build time (node): `workbox-build` package

We're using here only `workbox-build` since we need the injection point for manual sw precaching logic. If you disable the `injectionPoint` in the `injectManifest` pwa option (`injectionPoint: undefined`), the Vite PWA plugin will not use any `workbox` module.

MajesticPotatoe: yeah, get that, i mean the initial comment about `enableWorkboxModulesLogs: true` -> "we're not using `workbox`" on the surface seemed to contradict each other. thanks for the clarification.

userquin: Both, `workbox-build::generateSW` and `workbox-build::injectManifest` loaded with dynamic import and when required:
- `injectManifest` strategy in pwa options: the custom Vite build will omit the `workbox-build::injectManifest` call if no injection point https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/vite-build.ts#L61-L81
- `generateSW` strategy in pwa options: https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/modules.ts#L80-L92

KaelWD:
#20425

PR #20425, opened 2024-09-04
docs: make the service worker optional, improve precache logic

KaelWD:
- Caching is moved from activate to install to prevent blocking requests
- Replaced urlsToCacheKeys with just adding a query string to assets with a revision
- Added update indicator


userquin:
Check https://vercel.com/docs/edge-network/headers#cache-control

KaelWD:
I think that's if you only specify `s-maxage`

packages/docs/vite.config.mts
```diff
@@ -209,12 +209,13 @@ export default defineConfig(({ command, mode, isSsrBuild }) => {
...
         strategies: 'injectManifest',
         includeAssets: ['favicon.ico'],
+        injectRegister: false,
         injectManifest: {
           globIgnores: ['**/*.html'],
           additionalManifestEntries: [
             { url: '/_fallback.html', revision: Date.now().toString(16) },
           ],
-          dontCacheBustURLsMatching: /assets\/.+[A-Za-z0-9]{8}\.(js|css)$/,
+          dontCacheBustURLsMatching: /assets\/.+-[A-Za-z0-9_-]{8}\.(js|css)$/,
```
userquin, line 218: cannot just use `/assets\//` adding `**/*.map`  to `globIgnores`?
```js
globIgnores: ['**/*.html', '**/*.map'],
dontCacheBustURLsMatching: /assets\//,
```

PR #20483, opened 2024-09-16
fix(VDateInput):  cancel action button should always be active

### LINKED ISSUE #20226: [Bug Report][3.6.13] VDateInput cancel action button should always be active.
### Environment
**Vuetify Version:** 3.6.13
**Vue Version:** 3.4.15
**Browsers:** Firefox 128.0
**OS:** Linux x86_64

### Steps to reproduce
Simply click in the text field and the date-picker popus up with an <ok> and <cancel> button but cancel is disabled. Cancel should always be active and close the date picker menu/date-picker

### Expected Behavior
Cancel should always be active and should close the date picker

### Actual Behavior
Cancel and Ok are not active when date picker opens. User has to press escape to get date picker to go away

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdjs0KwjAQhF9lycUWrLlLKwi+hfVQ2q0E8rMkm3oofXeTVEG8LLvDzDd7X0Xwo7wSnZaI4ixaRkN6YLz0FqBdmoGorOUYneVBWfQfqYhTcjfKUmRYGuMm1F0vTHj2AuQ3Kv+zWdnRrfypTGcYvSKGgBz3ZmXIeYYVPM6wweydgUP69pDtAIkbGFIfdNlRWXzBLdGqus7wHZdAYjvaqHUZjzen+lL7)

vikash8813:
fixes #20226 

**Description**

This pull request fixes an issue where the cancel and ok button of date-input not enables, clearable does not clear the model value.

**Changes Made**

Add a prop in VConfirmEdit to show the action buttons and make it true from the VDateInput which enables the action buttons for the date input, VConfirmEdit for other inputs works same as before.

**Testing:**

Verified the fix locally by using multiple conditions in VDateInput.
All relevant tests are passing.

**Recording**

### DIFF
--- a/packages/vuetify/src/components/VConfirmEdit/VConfirmEdit.tsx
+++ b/packages/vuetify/src/components/VConfirmEdit/VConfirmEdit.tsx
@@ -101,7 +101,7 @@ export const VConfirmEdit = genericComponent<new <T> (
           />
 
           <VBtn
-            disabled={ isPristine.value }
+            disabled={ props.enableActions ? false : isPristine.value }
             variant="text"
             color={ props.color }
             onClick={ save }
--- a/packages/vuetify/src/components/VConfirmEdit/VConfirmEdit.tsx
+++ b/packages/vuetify/src/components/VConfirmEdit/VConfirmEdit.tsx
@@ -56,6 +56,7 @@ export const VConfirmEdit = genericComponent<new <T> (
     cancel: () => true,
     save: (value: any) => true,
     'update:modelValue': (value: any) => true,
+    closeMenu: () => true,
   },
 
   setup (props, { emit, slots }) {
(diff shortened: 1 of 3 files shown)


johnleider:
I understand the sentiment behind the issue, but this approach feels a bit heavy handed. If we did make a change here, I would probably want to start with an implementation that we've already done, example [VStepperActions](https://github.com/vuetifyjs/vuetify/blob/1bf8a2d8d4fe37d5d42d607aae0bdb3ceb891e3f/packages/vuetify/src/components/VStepper/VStepperActions.tsx#L25).

ikushum:
The issue https://github.com/vuetifyjs/vuetify/issues/20226 is already fixed by https://github.com/vuetifyjs/vuetify/pull/21154

PR #20535, opened 2024-09-29
feat(theme): add utility functions

johnleider:
### Description:
This pull request introduces new methods to `theme.ts` for enhanced theme management and updates `theme.md` to document these changes. The added methods will allow developers to better manage and switch themes at runtime.

#### `theme.ts`:
- **`get`**: Retrieve the current or a specified theme by name.
- **`update`**: Update the current color within a theme or a specified theme by name.
- **`register`**: Register a new theme with custom properties.
- **`set`**: Change the global theme.

### DIFF
--- a/packages/docs/src/pages/en/features/theme.md
+++ b/packages/docs/src/pages/en/features/theme.md
@@ -351,3 +351,71 @@ export const vuetify = createVuetify({
 ## Implementation
 
 Vuetify generates theme styles at runtime according to the given configuration. The generated styles are injected into the `<head>` section of the DOM in a `<style>` tag with an **id** of `vuetify-theme-stylesheet`.
+
+## Utility functions
+
+### `get`
+
+The `get` method allows you to retrieve a theme by its name. If no name is provided, the current global theme will be returned.
+
+```js { resource="src/App.vue" }
+import { useTheme } from 'vuetify'
+
+const theme = useTheme()
+
+// Get the currently active theme
+const currentTheme = theme.get()
+
+// Get a specific theme by name
+const customTheme = theme.get('myCustomLightTheme')
+```
+
+### `update`
+
+The `update` method is used to update the color of a specific theme.
+
+```js { resource="src/App.vue" }
+import { useTheme } from 'vuetify'
+
+const theme = useTheme()
+
+// Update the 'primary' color of the active theme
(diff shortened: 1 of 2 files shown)


KaelWD:
I feel like this doesn't really add anything more useful than just making useTheme return a reactive(theme) instead would.

johnleider:
> I feel like this doesn't really add anything more useful than just making useTheme return a reactive(theme) instead would.

Open to suggestions. The current process for working with themes directly is quite cumbersome; some form of update is necessary.

4444TENSEI:
Currently, to persist the theme color and avoid reverting to the default theme on page refresh, I'm using watchEffect to monitor localStorage or Cookie values. Is there a simpler built-in solution in Vuetify for theme persistence?

Here's my current implementation:
```
import { THEME_KEY } from '@/hooks/useKey'

const themeRef = ref('light')
const storedTheme = localStorage.getItem(THEME_KEY)

if (!storedTheme) {
  localStorage.setItem(THEME_KEY, themeRef.value)
} else {
  themeRef.value = storedTheme
}

const getCurTheme = () => themeRef.value

const toggleTheme = () => {
  const newTheme = getCurTheme() === 'light' ? 'dark' : 'light'
  localStorage.setItem(THEME_KEY, newTheme)
  themeRef.value = newTheme
}

export { toggleTheme, getCurTheme }
```
And in the component:
```
const theme = useTheme()
watchEffect(() => {
  theme.global.name.value = getCurTheme()
})
```

J-Sek:
> Currently, to persist the theme color and avoid reverting to the default theme on page refresh, I'm using watchEffect to monitor localStorage or Cookie values. Is there a simpler built-in solution in Vuetify for theme persistence?

There is nothing built-in, but you can use `useLocalStorage` from VueUse and set watcher on the ref you get from it. There are some references to SSR in `useStorage` (used underneath), so it could make your composable portable between different kinds of Vue projects.

J-Sek:
@johnleider does the following look interesting?

```ts
const { syncWith } = useTheme()
const storedThemeName = useLocalStorage('theme', 'dark')
syncWith(storedThemeName)
```

BTW. I would not mind explicit `register(name, spec)` and `update(name, spec)` methods.

PR #20549, opened 2024-10-03
fix(router): check $router.beforeEach exists before using

sgpinkus:
Make use of $router global safer.

Code breaks if something that isn't a vue-router inhabits the global Vue var "$router". For example, I'm using a different router and it installed itself to the global name "$router" but it doesn't have a beforeEach method. I don't think "$router" has been reserved specifically for vue-router or any router implementation.

## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/composables/router.tsx
+++ b/packages/vuetify/src/composables/router.tsx
@@ -113,7 +113,7 @@ export function useBackButton (router: Router | undefined, cb: (next: Navigation
   let removeBefore: (() => void) | undefined
   let removeAfter: (() => void) | undefined
 
-  if (IN_BROWSER) {
+  if (IN_BROWSER && router?.beforeEach) {
     nextTick(() => {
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/composables/router.tsx
```diff
@@ -113,7 +113,7 @@ export function useBackButton (router: Router | undefined, cb: (next: Navigation
   let removeBefore: (() => void) | undefined
   let removeAfter: (() => void) | undefined
 
-  if (IN_BROWSER) {
+  if (IN_BROWSER && router?.beforeEach) {
     nextTick(() => {
       window.addEventListener('popstate', onPopstate)
       removeBefore = router?.beforeEach((to, from, next) => {
```
johnleider, line 119:
```suggestion
      removeBefore = router.beforeEach((to, from, next) => {
```

johnleider (approved):

PR #15229, opened 2022-06-08
fix(layout): use suspense to delay render of layout items

nekosaur:
## Description

closes #15202

same principle as in https://github.com/vuetifyjs/vuetify/pull/15215

## Markup:

<details>

```vue
<template>
  <v-app>
    <v-navigation-drawer permanent>
      nav drawer

    </v-navigation-drawer>
    <v-app-bar>
      <v-toolbar-title>Title</v-toolbar-title>
    </v-app-bar>

    <v-main app>
      <v-container>
        Something!
      </v-container>
    </v-main>
  </v-app>

</template>

```
</details>

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Improvement/refactoring (non-breaking change that doesn't add any features but makes things better)

## Checklist:

- [x] The PR title is no longer than 64 characters.
- [x] The PR is submitted to the correct branch (`master` for bug fixes and documentation updates, `dev` for new features and backwards compatible changes and `next` for non-backwards compatible changes).
- [x] My code follows the code style of this project.
- [x] I've added relevant changes to the documentation (applies to new features and breaking changes in core library)

PR #20563, opened 2024-10-10
fix(VAppBar): vertical main layout shift on SSR request

### LINKED ISSUE #15202: [Bug Report][3.0.0] V-App-Bar Collapse On Reload Page
### Environment
**Vuetify Version:** 3.0.0-beta.3
**Vue Version:** 3.2.36
**Browsers:** Chrome 102.0.0.0
**OS:** Windows 10

### Steps to reproduce
0. You should run server (npm install && npm run dev OR npm install && npm run build && npm run start)
1. Open `/` page
2. Reload page
3. Look at app bar
4. Bar collapse for a few seconds (in run dev mode this collapse a bit longer)

### Expected Behavior
App bar didn't collapse or something, duo page reload

### Actual Behavior
Bar collapse on every page reload. After a few seconds, bar return to valid position. If you disable javascript, app bar didn't return to required position.

### Reproduction Link
<a href="https://gitlab.com/BlackYuzia/nuxt-vuetify" target="_blank">https://gitlab.com/BlackYuzia/nuxt-vuetify</a>

### Screenshots

Collapse example:
image

Un Collapsed after a few seconds:
image

userquin:
## Description

Since we need to await to render the VToolbar from VAppBar, this PR just uses the props to calculate the computed height in VAppBar when VToolbar is not still present: on SSR page request this will prevent the vertical shift layout.

~~https://streamable.com/llbms5~~

resolves #15202 partially, now there is a flickering in the navigation drawer (we should also review the navigation drawer and the hydration missmatch): ~~https://streamable.com/6pjzvc~~

related PR #15229

### DIFF
--- a/packages/vuetify/src/components/VAppBar/VAppBar.tsx
+++ b/packages/vuetify/src/components/VAppBar/VAppBar.tsx
@@ -2,7 +2,7 @@
 import './VAppBar.sass'
 
 // Components
-import { makeVToolbarProps, VToolbar } from '@/components/VToolbar/VToolbar'
+import { calculateHeight, makeVToolbarProps, VToolbar } from '@/components/VToolbar/VToolbar'
 
 // Composables
 import { makeLayoutItemProps, useLayoutItem } from '@/composables/layout'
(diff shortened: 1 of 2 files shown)


johnleider:
Is this still relevant since @KaelWD reverted the suspense layout changes in 3.7.

imduchy:
@johnleider is there any plan to merge this? Could this solve the problem described here? https://github.com/vuetifyjs/vuetify/issues/15202#issuecomment-3227986462

I'm on version `3.9.6` of Vuetify.

BessetThomas:
Hello, any update on this issue ?

J-Sek:
Only rebased.
I don't have time to verify it.

It would be great if community members could step in and provide a repo or a playground setup to verify this.

J-Sek:
It does not seem to make any difference. `ssr: true` is the only thing that actually resolves hydration and CLS.

userquin:
> It does not seem to make any difference. `ssr: true` is the only thing that actually resolves hydration and CLS.

Not sure if this is true, the problem is the CLS when using `ssr: false`  since the size is not precalculated, we should precalculate the data before rendering.

Videos at streamable removed arggg.

J-Sek:
My mistake. I can see it having an impact on the [baseline nuxt+ssr app](https://github.com/J-Sek/vuetify-nuxt-ssr). VMain is fine, but the VAppBar shifts because of the sidebar. We might need to prepare a robust/complex layout example and make sure the fix helps avoid shift (as much as we can eliminate with reasonable changes).

PR #20594, opened 2024-10-19
feat(VDataTable): Screenreader suport for sorting

viktorkratz:
## Description
For screenreader users there was no possibility to dtermine wether a table column is sortable or not. The sort status was not readable too. This commit fixes this problem. In theory aria-sort can only be set to one table header 
(see https://www.w3.org/TR/wai-aria-1.2/#aria-sort), 
but in most screen readers it's even working with multiple columns containing aria-sort. For others it's only anouncing the first one, so i think setting it on every column which has an order set is the best option.
## Markup:
<template>
  <v-app>
    <v-container>
      <v-data-table :headers="[{key:'name',title:'name', sortable:true},{key:'test', title:'test', sortable:true}]" :items="items" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        items: [{ name: 'bla', test: 'test1' }, { name: 'blu', test: 'test2' }],
      }
    },
  }
</script>


J-Sek:
already resolved

Issue #19447, opened 2024-03-21
[Bug Report][3.5.11] Select All in VDataTable is freezing on selection-strategy="all"


AANGHEL1:
### Environment
**Vuetify Version:** 3.5.11
**Vue Version:** 3.3.4
**Browsers:** Google Chrome
**OS:** Mac OS 10.15.7

### Steps to reproduce
Create a Data Table that contains more than 6000 items. 
Add property show-select.
Add select-strategy="all" instead of "page".

### Expected Behavior
This should not freeze the browser. It should work fast.

### Actual Behavior
It freezes the browser.

### Reproduction Link
[...]

PR #20595, opened 2024-10-19
fix(VTreeviewChildren): return-object mess up ListGroup id

yuwu9145:
## Markup:

```vue
<template>
  <v-app>
    <h1>{{ selectedModel }}</h1>
    <v-container>
      <v-treeview
        selectable
        select-strategy="independent"
 :items="items" return-object v-model="selectedModel" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selectedModel = ref([])
  const items = ref([
    {
      id: 1,
      title: 'title 1',
      children: [
        {
          id: 2,
          title: 'title 2',
        },
      ],
    },
  ])
</script>

```

### DIFF
--- a/packages/vuetify/src/labs/VTreeview/VTreeviewChildren.tsx
+++ b/packages/vuetify/src/labs/VTreeview/VTreeviewChildren.tsx
@@ -128,7 +128,7 @@ export const VTreeviewChildren = genericComponent<new <T extends InternalListIte
       return children ? (
         <VTreeviewGroup
           { ...treeviewGroupProps }
-          value={ props.returnObject ? item.raw : treeviewGroupProps?.value }
+          value={ treeviewGroupProps?.value }
         >
           {{
             activator: ({ props: activatorProps }) => {


yuwu9145:
@KaelWD @johnleider Could you briefly check the browser.spec.tsx file being migrated?

johnleider (dismissed):

KaelWD:
I think we need to require a primitive item-value anyway to fix return-object matching performance and #19447.

yuwu9145:
> I think we need to require a primitive item-value anyway to fix return-object matching performance and #19447.


- When `return-object` is applied, the item object must be passed to VListGroup to be able to return the item object to `opened`.
- `nested` composable doesn't have `return-object` state, so it is unable to do primitive item-value matching and return object. But it fully supports item object used as item id.

primitive item-value is cleaner but atm it requires refactoring `nested` to internally support `return-object`

yuwu9145:
> I think we need to require a primitive item-value anyway to fix return-object matching performance and #19447.

@KaelWD If pass primitive item-value, it ends up with three props to be able to get the real "item-key" as well as return the object to v-model:opened, so I came up with a new more straightforward prop "rawId"

image

johnleider (approved):

KaelWD (approved):
Still not perfect if we have multiple nodes with the same value or have non-primitive values but it's better than before.

PR #20615, opened 2024-10-24
feat(VDatatable): column sort modifier to start descending first

### LINKED ISSUE #11117: [Feature Request] Ability to sort descending on first click of DataTable header
### Problem to solve
For some table columns it makes more sense to sort rows descending rather than ascending. For instance a product listing with a product revenue column, most of the time you want to sort on that column you would be interested in getting the most profitable products first.

It would be very useful to be able to specify that this column should be sorted descending first, then ascending and (optionally) none.

### Proposed solution
A new attribute on TableHeader:
sortDescendingFirst?: boolean

If set to true the sort order when clicked should be descending -> ascending -> none.

berrywhj:
## Description

Allow user to pass a param named sortDescendingFirst, so that the column can be sorted descending first, then ascending and (optionally) none.
resolves #11117
## Markup:

packages/vuetify/dev/Playground.vue:

```vue
<template>
  <v-app>
    <h1>header rowspan/colspan</h1>
    <v-data-table
      :headers="complex"
      :items="items"
      :multi-sort="true"
      :sortDescendingFirst="true"
    >
    </v-data-table>

  </v-app>
</template>

<script setup>

  const items = Array(20).fill(0).map((_, i) => ({ id: i, name: `Name ${i}`, one: i, two: Math.random() > 0.5 ? 'foo' : 'bar', three: Math.random() > 0.5 ? 'hello' : 'world', four: i }))

</script>

```


Revadike (commented):
Looks good to me. Looking forward for this to be merged.

PR #20626, opened 2024-10-28
docs(VDataTable): remove unused mobile prop from headers type

J-Sek:
## Description

`mobile` should not appear on [API docs for v-data-table](https://vuetifyjs.com/en/api/v-data-table/#props-headers), because it is not utilized anywhere.

~~I wish I could fully verify the change locally, but the project does not build at the moment due to chai TS issues. Hopefully it is the only place to change.~~ Verified locally.


J-Sek:
Ouch.. sorry @MajesticPotatoe, if you still think your title is more precise feel free to change it back :)

PR #20637, opened 2024-10-29
feat(VInfiniteScroll): add reset method

### LINKED ISSUE #19935: [Feature Request] Change load status in v-infinite-scroll  from parent component
### Problem to solve
I want connect v-infinite- loading to to some filter. such country and city. 

If the user changes the filter and before that the loading has reached the empty state. It is no longer called by changing the filter of other pages

### Proposed solution
I think we need a props to can change status programmatically

### LINKED ISSUE #20308: [Feature Request] VInfiniteScroll - reset method?
### Problem to solve
When my component loads for the first time, VInfiniteScroll calls `onLoad` and loads the first page - and then subsequent pages as scrolling occurs.  So far so good but if the underlying result-set changes, e.g., through altered search parameters or whatever, there does not seem to be a method of having VInfiniteScroll call onLoad once more to start showing a fresh set of results.

### Proposed solution
I can hack the behavior I want by caching the `done()` callback in the `onLoad` handler and calling it later with `"ok"` when I want to nudge the VInfiniteScroll component into action again.  This seems a bit of a fragile way to go though, so a method I could call from a reference to the component would seem more satisfactory.

    const infiniteScroll = ref<typeof VInfiniteScroll>();
    // ... new search...
    items.value = [];
    infiniteScroll.value.reset(); // <-- what I want :-)

Thanks!

giftimie:
## Description

When the VInfiniteScroll reaches the 'empty' state, it is not possible to get out of it without doing any tricks such as caching the `done` method.
Many users are using this component to paginate search results, and when a specific search returns zero items, the component will be stuck in the `empty` state

closes #20308
closes #19935

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-infinite-scroll ref="infiniteScrollRef" height="300" side="end" @load="load">
        <template v-for="(item, index) in items" :key="index">
          <div :class="['px-2', index % 2 === 0 ? 'bg-grey-lighten-2' : '']">
            Item number {{ item }}
          </div>
        </template>
      </v-infinite-scroll>
      <v-btn @click="reset">Call reset</v-btn>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const infiniteScrollRef = ref(null)

  const items = ref([])

  function reset() {
    items.value = []
    infiniteScrollRef.value.reset()
  }

  function load({ side, done }) {
    setTimeout(() => {
      if (items.value.length >= 40) {
        done('empty')
        return
      }

      const arr = Array.from({ length: 10 }, (k, v) => (items.value.at(-1) ?? 0) + 1 + v)
      items.value = [...items.value, ...arr]

      done('ok')
    }, 1000)
  }
</script>

```


packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -292,6 +292,29 @@ export const VInfiniteScroll = genericComponent<VInfiniteScrollSlots>()({
         </Tag>
       )
     })
+
+    function reset () {
+      setStatus(props.side, 'ok')
```
J-Sek, line 297: It won't work for `both`. I guess `setStatus` should be updated to accept 3rd option to set both `start` and `end`.

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -292,6 +292,29 @@ export const VInfiniteScroll = genericComponent<VInfiniteScrollSlots>()({
         </Tag>
       )
     })
+
+    function reset () {
```
J-Sek, line 296: I have played with your example - merging it with virtualized infinite scroller demo from docs, but also trying to have more realistic example, where server response decides whether there are more elements or not (typical with GraphQL offset-based pagination). Due to lack of reset, my [playground](https://play.vuetifyjs.com/playgrounds/_2AcSw) includes workaround, but it also shows that this use case would benefit from `reset()` accepting `side` as argument.

```ts
function reset(side?: InfiniteScrollSide) {
  setStatus(side ?? props.side, 'ok')
  // ...
```

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
giftimie: Oh alright I see, I completely missed that. I will be revisiting this shortly and updating everything that requires it

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
giftimie: Just commited your suggestions and rebased just in case.
Let me know if I still missed anything. I double checked but since I am fairly new to the codebase I still feel like there's something that's not on my radar

laa-1:
I also encountered this problem when I tried to combine VInfiniteScroll and paging

erikvdplas:
@J-Sek Any chance this PR will get merged? Solution seems to work well and the method seems to be highly-requested by the community. I currently encounter this issue as well. If I can do anything to expedite the process, let me know.

J-Sek (approved):

packages/vuetify/src/components/VInfiniteScroll/VInfiniteScroll.tsx
```diff
@@ -292,6 +295,35 @@ export const VInfiniteScroll = genericComponent<VInfiniteScrollSlots>()({
...
+      setStatus(effectiveSide, 'ok')
+
+      nextTick(() => {
+        setScrollAmount(
+          getScrollSize() - previousScrollSize + getScrollAmount(),
+        )
+        if (props.mode !== 'manual') {
+          nextTick(() => {
+            window.requestAnimationFrame(() => {
+              window.requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
giftimie: On my end it also worked fine with and without, I just copied them from above because of #17475

J-Sek:
@giftimie, CI linting seems to be stuck for reasons unrelated to the introduced changes. I wouldn't mind if you could rebase the branch.

J-Sek:
Nevermind, linting fails for spec files.

J-Sek (approved):

johnleider (approved):

xqqp:
Any chance for this getting merged soon?

PR #20661, opened 2024-11-04
docs(installation): add ES Module setup

kingma-sbw:
## Description
Without tooling include Vuetify and Vue using import map
## Markup:
none

```vue

```


J-Sek:
I would not really advise users to use it without integrity hashes. And even if [integrity](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#integrity_metadata_map) was supported on all browsers and we would refine PR with `{version}` placeholders, single CDN and closing script tag, I personally have doubts whether it brings any value over existing instructions for CDN-based setup.

kingma-sbw:
> I would not really advise users to use it without integrity hashes. And even if [integrity](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#integrity_metadata_map) was supported on all browsers and we would refine PR with `{version}` placeholders, single CDN and closing script tag, I personally have doubts whether it brings any value over existing instructions for CDN-based setup.

the big advantage apart from strict-typing is that a ES Module issolatates its content so it is even possible to have more than one (mini) Vue app on a page.

J-Sek:
> _...possible to have more than one (mini) Vue app on a page_

This sounds like a very niche use case through.

Anyway, assuming we include this section, I wonder if it would be more beneficial to have a full working example (e.g. [[1]](https://jsbin.com/dihemasaga/edit?html,output) or [[2]](https://jsfiddle.net/4yo6cqwr/latest/) as well as some links to MDN, SHA generator, etc.

BTW. While working on these examples I noticed that CDN suggested URLs with `*/+esm`  and `https://esm.run/{package}` which both did not work, making the case for full demo even stronger.

kingma-sbw:
> Anyway, assuming we include this section, I wonder if it would be more beneficial to have a full working example (e.g. [[1]](https://jsbin.com/dihemasaga/edit?html,output) or [[2]](https://jsfiddle.net/4yo6cqwr/latest/)) as well as some links to MDN, SHA generator, etc.
> 
> BTW. While working on these examples I noticed that CDN suggested URLs with `*/+esm` or shortcut `esm.run/{package}` which both did not work, making the case for full demo even stronger.

Valid points. But they would also apply to other sections ("Using CDN").

packages/docs/src/pages/en/getting-started/installation.md
```diff
@@ -273,6 +273,29 @@ const app = createApp()
...
+<script type="importmap">
+  {
+    "imports": {
+      "vue": "https://unpkg.com/vue@3.2.47/dist/vue.esm-browser.js",
+      "vuetify": "https://unpkg.com/vuetify@3.1.10/dist/vuetify.esm.js"
+    }
+  }
+</script>
+<script type="module">
+  const apiUrl = 'http://zeugnis-admin.localhost/'
```
MajesticPotatoe, line 291:
```suggestion
```

packages/docs/src/pages/en/getting-started/installation.md
```diff
@@ -273,6 +273,29 @@ const app = createApp()
...
+
+To import Vuetify (and Vue) using an import map you can use the same CDN but contain it in a ES module without tooling
+
+```html
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/vuetify@3.7.3/dist/vuetify.min.css">
+<script type="importmap">
+  {
+    "imports": {
+      "vue": "https://unpkg.com/vue@3.2.47/dist/vue.esm-browser.js",
+      "vuetify": "https://unpkg.com/vuetify@3.1.10/dist/vuetify.esm.js"
```
MajesticPotatoe, line 286:
```suggestion
      "vuetify": "https://unpkg.com/vuetify@{{ version }}/dist/vuetify.esm.js"
```

MajesticPotatoe:
feels like these instructions are pretty much 1:1 with the regular cdn install instructions, just with how to use an script/link tag sprinkled in, and an esm import instead of .min/js|.css. Would probably refine this to just include both (update the section above to have the links for the css/js and esm versions, and add a note about it instead of making it its own section

J-Sek:
@MajesticPotatoe , Including `@{{ version }}` for Vue, `integrity` part and non-google link to Roboto are all my personal preferences and I won't mind you adjusting it for simplicity and consistency.

KaelWD:
The whole page is like that really, all it needs to say is

- load materialdesignicons.css somehow
- load vuetify/styles or dist/vuetify.css somehow
- `app.use(createVuetify())`

but apparently people need a copy-paste guide for every little thing these days.

packages/docs/src/pages/en/getting-started/installation.md
```diff
@@ -273,6 +273,38 @@ const app = createApp()
...
+
+```html
+<head>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/vuetify@{{ version }}/dist/vuetify.min.css" />
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@mdi/font@latest/css/materialdesignicons.min.css" />
+  <link rel="stylesheet" href="https://fonts.bunny.net/css?family=roboto:400,500,700" />
+  <script type="importmap">
+  {
+    "imports": {
+      "vue": "https://cdn.jsdelivr.net/npm/vue@{{ version }}/dist/vue.esm-browser.js",
```
KaelWD, line 288: version is the vuetify version

MajesticPotatoe: should be fine to just do `@latest`, also is on line 292

J-Sek (approved):

MajesticPotatoe (approved):

PR #20686, opened 2024-11-10
fix(VNumberInput): allow scientific notation

### LINKED ISSUE #20639: [Bug Report][3.7.3] VNumberInput doesn't allow scientific notation
### Environment
**Vuetify Version:** 3.7.3
**Vue Version:** 3.5.12
**Browsers:** Edge 130.0.0.0
**OS:** Windows 10

### Steps to reproduce
Try to type 'e' in VNumberInput, fail

### Expected Behavior
Users should be able to type number like `1e7` in VNumberInput just like HTML number input `<input type="number" />`

image

### Actual Behavior
Cannot type scientific notation number.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdjz0LwkAMhv9KuMVFLTpKFRxdHF2sQ1sjHNxHyOWKpfS/e70qikvI++TrzXVQgdviSLTuIqqdKgUtmVrwUDmAslvVRDnNovVOau2Q3yhB7SgKtKYOYV8p26y2lQLpCZNy0TbISQfpzQQaz3fkHWzoCcEbfU+14rvrcs4Dp7zyw8vi/+5EZltl8WM3ydCyJoGAEmfX2pJngQEYHzDCg72FRfp0Mc3O3alPjUsXjcnh9gJhFVus)

kingyue737:
## Description

Enable the use of scientific notation in VNumberInput to ensure compatibility with the HTML number input.

close #20639

## Markup:

```vue

```


kingyue737:
That's interesting. The test succeeded before but failed now. I also experienced this issue before. I typed `1e7` in a `<v-text-field type="number"/>`, and it becomes `10000000` directly in the latest Electron, but in the latest Edge no such phenomenon. It may be related to the chromium version?

yuwu9145:
It's too complex to properly support exponential notation, it makes +/- completely useless or it's unrealistic to properly do +/- to exponential numbers.

Or we can introduce a new prop called "stringMode", which allows exponential numbers and users can implement their own +/- event listeners

kingyue737:
> It's too complex to properly support exponential notation

I believe behaving like `<input type="number" />` is already proper. Not complex.

> it makes +/- completely useless or it's unrealistic to properly do +/- to exponential numbers.

It's still useful, user can modify the `step` prop. Let's see the behavior of  `<input type="number" />`:

number-input

> Or we can introduce a new prop called "stringMode"

We already have `<v-text-field type="number" />`. But `<v-number-input>` looks more beautiful (arrow button style is consistent with other Vuetify components) and it has IntelliSense for props `max`, `min` and `step`.

yuwu9145:
`<input type="number" />` is deliberately avoided in number input because it has a lot of wierd behaviours

https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

PR #20729, opened 2024-11-24
fix(VNumberInput): clearing v-number-input with backspace resets to 0

### LINKED ISSUE #20607: [Bug Report][3.7.3] clearing v-number-input with backspace resets to 0
### Environment
**Vuetify Version:** 3.7.3
**Vue Version:** 3.5.12
**Browsers:** Chrome 129.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Type any number for example 3
Delete number by pressing backspace
Input rewrites value to 0 (i consider it as not wanted behaviour)
Delete zero by pressing backspace (validation is not called and no required message is visible)
Click outside of v-number-input and validation is called and required is displayed

### Expected Behavior
When i delete value by backspace or delete validation is called immediately on input

### Actual Behavior
When i delete value by backspace or delete validation is not called immediately on input

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVUk1znDAM/Ssqh8DOsEvSmVwysNNObz20h/aW5MAuYuMZY7vG3jZD+O+VPwgszaG5GKQn6UlPuh+SXh+Lz0rtzhaTu6Q02CleG9w/CIDyvK2V8r/eOEphaiZQR5d3CtsdUG+ZUNZMbgCNbfWQBGcyu+862SDfnmtukXBvXeDacuwJ8d8lQimsoca2UrxV+JNVDl3Vl+LLUy1OOEcWc+sHvbSUxv0wgC/8U3798f0bjGNZOHecv1gL4DxBn7JY6EZmf9RMGejR2CAf65TUBganSw5H2RELNjBCq2UHKYmfukQgSPQmdAGVi842S8CP9xbg9YrAfWi4KMj6ZZkmoqgekz7Fq2kRqj0MkwCujOS44/KUpVNemodIoprWaqwWkM3CZ6HUh6oCYTmHqyswzwplGzkckFrRYEu6NanDF0C6gZeXudhMPPki85i/ztQx8R/jxEYDsK/glngg7Z+k5Q0cEE4aaVsapAairDncRsolU/3nvUxlBTfXKyoufy+Jbq4vmB4v1jgdLG1yzXSxoClutSB/HruYGRAHjP/cVrzw6vUUs2yz4IozuZhdbzQTJ9Y+Zz4zVM/9snP46IlHessiHD2dezLmHnXP419GtWEc)

EvgenyWas:
## Description

The `useValidation` composable doesn't perform validation when the input value changes to `null`. When resetting the input, it sets the value to an empty string. Instead of assigning 0, the model should be reset to match this behavior. Using `null` for resetting is not viable because validation won't be triggered. Therefore, the model value should remain as the empty string in such cases to ensure proper validation behavior.
fixes #20607 
## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-number-input
        ref="input"
        :model-value="model"
        :rules="rules"
        validate-on="input"
        @update:model-value="onChange"
      />
      <br />
      <pre>{{ inputToJSON }}</pre>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref, computed } from 'vue'

  const input = ref()

  const model = ref()

  const rules = ref([
    // required validation
    value => {
      console.log('required', value)
      return (
        (value !== null && typeof value !== 'undefined' && value !== '') ||
        'required'
      )
    },
    // min validation
    value => {
      return value >= 5 || 'should be greater or equal 5'
    },
    // max validation
    value => {
      return value <= 10 || 'should be lower or equal 10'
    },
  ])

  const onChange = value => {
    console.log('onChange', value)
    model.value = value
  }

  const inputToJSON = computed(() => {
    return JSON.stringify(input.value, null, 2)
  })
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -75,13 +75,14 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
     const model = computed({
       get: () => _model.value,
       set (val) {
-        if (val === null) {
-          _model.value = null
+        if (val === null || (typeof val === 'string' && !val)) {
+          _model.value = val
           return
         }
 
-        if (!isNaN(+val) && +val <= props.max && +val >= props.min) {
-          _model.value = +val
+        const value = Number(val)
+        if (!isNaN(value) && value <= props.max && value >= props.min) {
+          _model.value = value
         }
       },
     })
@@ -294,9 +295,9 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
 
               { incrementControlNode() }
             </div>
-          ) : (!props.reverse
-            ? <>{ dividerNode() }{ controlNode() }</>
-            : undefined)
+          ) : (props.reverse
+            ? undefined
+            : <>{ dividerNode() }{ controlNode() }</>)
 
       const hasAppendInner = slots['append-inner'] || appendInnerControl
 
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -74,9 +74,11 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
 
     const model = computed({
       get: () => _model.value,
-      set (val) {
-        if (val === null || (typeof val === 'string' && !val)) {
-          _model.value = val
+      // model.value could be empty string from VTextField
+      // but _model.value should be eventually kept in type Number | null
+      set (val: Number | null | string) {
+        if (val === null || val === '') {
+          _model.value = null
           return
         }
 


J-Sek (changes requested):

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -75,13 +75,14 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
     const model = computed({
       get: () => _model.value,
       set (val) {
-        if (val === null) {
-          _model.value = null
+        if (val === null || (typeof val === 'string' && !val)) {
```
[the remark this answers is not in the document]
EvgenyWas: I fully agree it looks exhaustive, but the value type is number or undefined and if we compare it without typeof there is a TS error. 
We can extend the initial type with string as well, but it will bring a bunch of refactoring.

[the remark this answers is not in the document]
J-Sek: Oh.. I see it now. Sure, let's leave it

J-Sek (approved):

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
yuwu9145, line 78: Clearing number via backspace currently results in warning:

```
[Vue warn]: Invalid prop: type check failed for prop "modelValue". Expected Number with value 0, got String with value "". 
  at <VNumberInput ref="input" model-value="" rules= (3) [ƒ, ƒ, ƒ]  ... > 
  at <VContainer> 
  at <VApp> 
  at <Playground> 
  at <App>
```

`modelValue` here should be strictly enforced with type `number | null`, probably try to set `_model.value` to `null` when backspace to empty string.

EvgenyWas: Hey @yuwu9145, thanks for the review!

As I described in the PR description - the useValidation composable doesn't perform validation when the input value changes to null. When resetting the input, it sets the value to an empty string. Instead of assigning 0, the model should be reset to match this behavior. Using null for resetting is not viable because validation won't be triggered. Therefore, the model value should remain as the empty string in such cases to ensure proper validation behavior.

I think it might be worth of refactoring to extend the model type with string. WDYT?

yuwu9145: I still believe model value should be set to `null` in number input.

As for the issue of `null` doesn't trigger validation, it's a separate issue that should be resolved in validation composable [demo](https://play.vuetifyjs.com/#eNqFkk1vnDAQhv/KhAuJFJZDc0Lsqh9STz01Ui8hBwOTXavGuPaYbpXsf+/YfCxtssoFhvE7j1+/5uE5cbbJPxmzGTwmRVISdkYJwl2lAcohs+hMr50cEBolnNtWSXfMhKe+SqATx+y3bOnA3Q93d1USp3jO7L6jQwLCI8GTRNXCIJRHoB60VwraHp1OWWDlfo82rMpWkOx1mZuZUlvIz8TPnoDdXKCybfoDjnl6H+kX0RNwyAImi5ixBVBYrzAc8SEWG4u/vLTYPlbJLFGiRsWKb8IRaNHheano+hZVFi2xIuB/hPqsaBQKK2qFY2NX5msX81GHrCYNHxslm5/MqT1Rr7+ELw74ns8vteEs/gk0kHjqAiKGc0/27fl1dGtOqM/Xz60yX/0d/OkaKw1FLR5NbwlafBJeETyPPjh2UcD1DWx3cD31AGK2xayJnSnnYjLF8qursXx5gfRrvGrpFl16O8+elmqJu4CU0NGsOd1MRYd06Nv1xqtk2ePKDx2k2yxA2MaIX285x/rOcJr+Pzq+TyHRJcPkdBt2iY/Hv5dWKig=)

yuwu9145: #20765

This PR should aim to make number input work consistent with text field, but model value type is kept with `number | null`

EvgenyWas: It happens because of [this condition in `useValidation` composable](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/composables/validation.ts#L157), but it was added over two years ago and I assume it can affect many places if we change it.

yuwu9145: Yeah, it's not a concern in VNumberInput so should be carefully resolved/tested in #20765

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -294,9 +297,9 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
 
               { incrementControlNode() }
             </div>
-          ) : (!props.reverse
```
yuwu9145, line 297: Is this change just a code format? Or was there any logically concrete reason behind it?

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
EvgenyWas: Then we can close this PR or just solve the first part of the issue regarding resetting the value?

yuwu9145: Just need to resolve the issue regarding resetting the value to null, and ignore validation.

I have pushed a commit, see if you had any comment

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
EvgenyWas: Only code format to remove the exclamation mark. I believe it's easier to read especially when there is the same condition with another case.

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
EvgenyWas: @yuwu9145 it's fine for me, thanks for your participation and prompt reply! :)

yuwu9145 (approved):

Issue #20517, opened 2024-09-25
Replace cypress with vitest


KaelWD:
- [x] composables/display
- [x] composables/display-components
- [x] composables/goto
- [x] composables/resizeObserver
- [x] composables/scroll
- [x] directives/intersect
- [x] directives/scroll
- [x] directives/touch
- [x] framework

Tests with percy snapshots:

- [x] VAlert
- [x] VAutocomplete
- [x] VBadge
- [x] VBtn
- [x] VCombobox
- [x] VFileInput
- [x] VInput
- [x] VSelect
- [x] VSwitch
- [x] VTextField
- [x] VVirtualScroll

Everything else:

- [x] VAppBar
- [x] VBottomNavigation
- [x] VBottomSheet
- [x] VBreadcrumbs
- [x] VBtnGroup
- [x] VCheckboxBtn
- [x] VChip
- [x] VColorPicker
- [x] VConfirmEdit
- [x] VDataIterator
- [x] VDataTable
- [x] VDataTableServer
- [x] VDataTableVirtual
- [x] VDatePicker
- [x] VDefaultsProvider
- [x] VDialog
- [x] VDivider
- [x] VExpansionPanels
- [ ] VForm
- [x] VHover
- [x] VIcon
- [x] VInfiniteScroll
- [x] VItemGroup
- [x] VLayout
- [x] VList
- [x] VListGroup
- [ ] VListItem
- [x] VMain
- [x] VNavigationDrawer
- [x] VNumberInput
- [x] VOtpInput
- [ ] VOverlay
- [x] VPagination
- [x] VProgressLinear
- [x] VRating
- [x] VSelectionControl
- [ ] VSlideGroup
- [x] VSlider
- [x] VTabs
- [x] VTextarea
- [x] VThemeProvider
- [x] VTimeline
- [x] VToolbar
- [x] VTreeview
- [x] VWindow

PR #20740, opened 2024-11-27
chore: Added cypress tests for VMain

iNSaNiA0821:
## Markup:

```vue

```


KaelWD:
We no longer use cypress, and this is already covered by VLayout tests. If you want to help with testing there's a todo list here: #20517

PR #20742, opened 2024-11-27
fix(VSelect): recognize empty string as no value

### LINKED ISSUE #20660: [Bug Report][3.7.3] v-combobox placeholder is not consistent with v-textbox
### Environment
**Vuetify Version:** 3.7.3
**Vue Version:** 3.5.12
**Browsers:** Google Chrome, Microsoft Edge
**OS:** Windows, Linux

### Steps to reproduce
A placeholder bound to an empty but non null model value

### Expected Behavior
The combobx placeholder should show up as the v-text-field placeholder does

### Actual Behavior
The placeholder remains hidden instead and strictly requires a null 

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkM0KgzAQhF9lycUWar0XW+ihT1F70LhqIH8kqwjiuzeJpUivvSy7w/Axs8+FeceLu7XnaUR2YSWhsrImvFUaoJzy2tq0poMbTbXQ6D5SEglnyjuBsoUpV6ZFea2Y8n3FIJA4Dka26IL20IQOaEBQ6H3dY3AUX9KOyI1qTGPmP3hl8Zs2KluZstiVDKfnTlgCjzRuXYWyxhEs4LCDFTpnFGThP1m0AwSuJwiJ4Bodhyw7RuiGCQC2nvQoZRqvN49udbg=)

J-Sek:
## Description

Selection controls should correctly reflect empty state. And most of the times empty string is perceived by end users as "no value". There is an edge case if `""` is among options' values - it results in selection in VSelection and VAutocomplete, while VCombobox will not recognize it as selection because it does not care about item value (thus has hardcoded `['']` and won't consume `emptyValues` from `useItems`)

fixes #20660

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container>
      <v-row>
        <v-col>
          <h4>Typical options</h4>
          <v-divider class="my-3" />
          <h5>VTextField</h5>
          <v-text-field v-model="msg1" :items="options1" clearable placeholder="Enter the message" />
          <h5>VSelect</h5>
          <v-select v-model="msg1" :items="options1" clearable placeholder="Select from the options" />
          <h5>VCombobox</h5>
          <v-combobox v-model="msg1" :items="options1" clearable placeholder="Select or type custom value" />
          <h5>VAutocomplete</h5>
          <v-autocomplete v-model="msg1" :items="options1" clearable placeholder="Type to find an item" />

          <pre>{{ JSON.stringify({ msg1 }) }}</pre>
        </v-col>
        <v-col>
          <h4>Options with empty string value</h4>
          <v-divider class="my-3" />
          <h5>VTextField</h5>
          <v-text-field v-model="msg2" :items="options2" clearable placeholder="Enter the message" />
          <h5>VSelect</h5>
          <v-select v-model="msg2" :items="options2" clearable placeholder="Select from the options" />
          <h5>VCombobox</h5>
          <v-combobox v-model="msg2" :items="options2" clearable placeholder="Select or type custom value" />
          <h5>VAutocomplete</h5>
          <v-autocomplete v-model="msg2" :items="options2" clearable placeholder="Type to find an item" />

          <pre>{{ JSON.stringify({ msg2 }) }}</pre>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script setup>
import { ref } from 'vue'
const msg1 = ref('')
const msg2 = ref('')

const options1 = ref([
  { title: 'Alabama', value: 'AL' },
  { title: 'Alaska', value: 'AK' },
  { title: 'American Samoa', value: 'AS' },
  { title: 'Arizona', value: 'AZ' },
  { title: 'Arkansas', value: 'AR' },
])
const options2 = [
  { title: "(I don't want to disclose)", value: '' },
  { title: 'California', value: 'CA' },
  { title: 'Colorado', value: 'CO' },
  { title: 'Connecticut', value: 'CT' },
  { title: 'Delaware', value: 'DE' },
]

setTimeout(() => {
  options1.value.push({ title: 'Empty', value: '' })
}, 2000)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -126,14 +126,14 @@ export const VAutocomplete = genericComponent<new <
     const vMenuRef = ref<VMenu>()
     const vVirtualScrollRef = ref<VVirtualScroll>()
     const selectionIndex = shallowRef(-1)
-    const { items, transformIn, transformOut } = useItems(props)
+    const { items, transformIn, transformOut, emptyValues } = useItems(props)
     const { textColorClasses, textColorStyles } = useTextColor(() => vTextFieldRef.value?.color)
     const search = useProxiedModel(props, 'search', '')
     const model = useProxiedModel(
       props,
       'modelValue',
       [],
-      v => transformIn(v === null ? [null] : wrapInArray(v)),
+      v => transformIn(v === null ? [null] : wrapInArray(v, emptyValues.value)),
       v => {
         const transformed = transformOut(v)
         return props.multiple ? transformed : (transformed[0] ?? null)
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -139,7 +139,7 @@ export const VCombobox = genericComponent<new <
       props,
       'modelValue',
       [],
-      v => transformIn(wrapInArray(v)),
+      v => transformIn(wrapInArray(v, [''])),
       v => {
         const transformed = transformOut(v)
         return props.multiple ? transformed : (transformed[0] ?? null)
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -143,12 +143,12 @@ export const VSelect = genericComponent<new <
     const vTextFieldRef = ref<VTextField>()
     const vMenuRef = ref<VMenu>()
     const vVirtualScrollRef = ref<VVirtualScroll>()
-    const { items, transformIn, transformOut } = useItems(props)
+    const { items, transformIn, transformOut, emptyValues } = useItems(props)
     const model = useProxiedModel(
       props,
       'modelValue',
       [],
-      v => transformIn(v === null ? [null] : wrapInArray(v)),
+      v => transformIn(v === null ? [null] : wrapInArray(v, emptyValues.value)),
       v => {
         const transformed = transformOut(v)
         return props.multiple ? transformed : (transformed[0] ?? null)
--- a/packages/vuetify/src/composables/list-items.ts
+++ b/packages/vuetify/src/composables/list-items.ts
@@ -122,7 +122,8 @@ export function transformItems (
 
 export function useItems (props: ItemProps) {
   const items = computed(() => transformItems(props, props.items))
-  const hasNullItem = computed(() => items.value.some(item => item.value === null))
+  const allValues = computed(() => items.value.map(item => item.value))
+  const hasNullItem = computed(() => allValues.value.includes(null))
 
   const itemsMap = shallowRef<Map<Primitive, ListItem[]>>(new Map())
   const keylessItems = shallowRef<ListItem[]>([])
@@ -204,5 +205,7 @@ export function useItems (props: ItemProps) {
       : value.map(({ value }) => value)
   }
 
-  return { items, transformIn, transformOut }
+  const emptyValues = computed(() => ['', null, undefined].filter(v => !allValues.value.includes(v)))
+
+  return { items, transformIn, transformOut, emptyValues }
 }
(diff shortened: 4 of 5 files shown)


yuwu9145 (commented):
I guess whether empty string is real empty is pretty subjective, different people take it differently, so I guess it's important for us to agree on an opinionated direction in Vuetify and make it clear in docs:

Current implementation can be found in this [demo (empty string is legit and non-empty)](https://play.vuetifyjs.com/#eNqdk11qwzAMx68i8pIN2gXat9IOugPsAm0fnERpDY4dbDkQSu8+2e5nKGzrS2JJf0v6CXlzzJytinXXffQes0W2JGw7JQg/txpg2U9F18VjNCqjSUiN9uxi52F2PQOsNfB1GsCRlXoP0oGAXihZQz9tTY0qWB4ngD1qkA1ICiJt+Kcr5Wus+XBLSAdkCbbuWq+4Kxja82Qqwy0j4aXGapu1br/NYBGvsplSZKBEGcMOFVYEglvgCAcKznnNWtpoP7XGuMKWkqywQwJ7QvwLa0C8ZYyNfvwfdvZ3Wgja13i/vVKcpgava2x4D2rmR+CtcLJGyyZZr4bzDsSBONYqdC5gDkF9x/o4hJfZ5y+wpwrjhQ6etO/L4u4dsOkqKzsCh+TTc5BtZyzBESw2cILGmhZyfkJ5mmsYCQE3B6ugeMvz9wf37OLXRk9bQdWBn8tIMz9rNE+dI7dYZOTgJnEcgSQpXEAu8kmaOp/5v87hNBlrykfN10WzC9AJkwGz0ySUjZ/dD+3ITVo=).

Visually, now it looks to me that makes sense to treat empty string to be real empty, for both v-text-field & all select components.

Need an opinionated agreement here, then implement and put it into docs. @johnleider @KaelWD

johnleider:
I think we should try to normalize this if we can. I agree with @yuwu9145

yuwu9145:
@johnleider by looking at this [demo](https://play.vuetifyjs.com/#eNqdk11qwzAMx68i8pIN2gXat9IOugPsAm0fnERpDY4dbDkQSu8+2e5nKGzrS2JJf0v6CXlzzJytinXXffQes0W2JGw7JQg/txpg2U9F18VjNCqjSUiN9uxi52F2PQOsNfB1GsCRlXoP0oGAXihZQz9tTY0qWB4ngD1qkA1ICiJt+Kcr5Wus+XBLSAdkCbbuWq+4Kxja82Qqwy0j4aXGapu1br/NYBGvsplSZKBEGcMOFVYEglvgCAcKznnNWtpoP7XGuMKWkqywQwJ7QvwLa0C8ZYyNfvwfdvZ3Wgja13i/vVKcpgava2x4D2rmR+CtcLJGyyZZr4bzDsSBONYqdC5gDkF9x/o4hJfZ5y+wpwrjhQ6etO/L4u4dsOkqKzsCh+TTc5BtZyzBESw2cILGmhZyfkJ5mmsYCQE3B6ugeMvz9wf37OLXRk9bQdWBn8tIMz9rNE+dI7dYZOTgJnEcgSQpXEAu8kmaOp/5v87hNBlrykfN10WzC9AJkwGz0ySUjZ/dD+3ITVo=), should empty string be treated as real empty like null & undefined here?

johnleider:
My instinct is saying it should mimic the native select functionality / behavior since that it's direct replacement.

J-Sek:
On the contrary I get the feeling native controls are usually inferior to what frameworks provide and aligning with them means giving up UX improvements. In case of `<select>` it does not provide placeholder and the only workaround is to include an option with empty value (e.g. `<option value="">--choose an option--</option>`) that does not show up when the value is set to `null`. So I am not sure which implementation would mimic the native control as current one is directly opposite – shows placeholder for `null` and does not show it for empty string.

Giulio's argument was pretty strong – user can type something and clear it with backspace, ending up with `null` value and placeholder showing up. This is an inconsistency we should resolve instead of debating native behavior that everyone is trying to escape from.

johnleider:
Not trying to increase the scope of this P.R., but what would you thoughts be on making this configurable? So each input component could define what it considers an empty value?

J-Sek:
I like this idea. It could help `VNumberInput` recognize zero as empty value (#20365).

PR #20768, opened 2024-12-07
fix(VAutocomplete/VCombobox): consistent open/close transition

J-Sek:
## Description

At the moment only VSelect has nice expand transition. I wish to make it a default for all selection components.

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="600">
      <v-card>
        <v-container fluid>
          <v-select label="v-select" :items="shortList" />
          <v-combobox label="v-combobox" :items="shortList" />
          <v-autocomplete label="v-autocomplete" :items="shortList" />
        </v-container>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup>
  const shortList = [0, 1, 2, 3, 4, 5, 6]
</script>
```


KaelWD:
- You could just omit the override, that's already the default for VMenu.
- This was here because combobox and autocomplete clear the filter when the menu closes which causes the full list to appear for a moment if there is a transition.

J-Sek:
> * This was here because combobox and autocomplete clear the filter when the menu closes which causes the full list to appear for a moment if there is a transition.

I see.. maybe `useDelay` would help here.

KaelWD:
This has the opposite problem:


https://github.com/user-attachments/assets/bce8457f-3ab4-4ff9-9e49-575f060dc8a5

J-Sek:
> ...the opposite problem:

Arguably it is still a progress, but let me take another look.

J-Sek:
Aligned VAutocomplete with VCombobox. The trick to fix selection was to `emit('update:search', ...)` instead of applying it immediately.

johnleider (dismissed):
Looks good besides broken test.

J-Sek:
> Looks good besides broken test.

Some kind of accident when switching branches. Rebased and tested locally.
Works fine (except for other outstanding issues related to VAutocomplete).

J-Sek:
Need to step away and look again in 2 days with a fresh mind, but it feels ready to merge.

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -139,13 +137,36 @@ export const VAutocomplete = genericComponent<new <
         return props.multiple ? transformed : (transformed[0] ?? null)
       }
     )
+    const form = useForm(props)
+
+    const hasChips = computed(() => !!(props.chips || slots.chip))
+    const hasSelectionSlot = computed(() => hasChips.value || !!slots.selection)
```
johnleider: These could probably be "toRef" instead of "computed".

KaelWD:
> This has the opposite problem:

Still does this. VCombobox with chips also doesn't re-open the menu if it has a custom selection not in items.

I've reverted this and ffe05014 for release.

J-Sek:
> _Still does this_

Too much last-minute changes. Thanks for verification 👍🏼

PR #20776, opened 2024-12-12
fix(v-ripple): avoiding more and more

dgmpk:
In some cases, when calling show continuously, ripples will increase.

## Description

If import @vant/touch emulator, ripples will increase when trigger mousedown.

## Markup:

```vue
<script>
import '@vant/touch-emulator' // for vant.js
</script>

<template>
<v-btn>Click here</v-btn>
</template>
```

### DIFF
--- a/packages/vuetify/src/directives/ripple/index.ts
+++ b/packages/vuetify/src/directives/ripple/index.ts
@@ -130,9 +130,9 @@ const ripples = {
     const ripples = el.getElementsByClassName('v-ripple__animation')
 
     if (ripples.length === 0) return
-    const animation = ripples[ripples.length - 1]
+    const animation = Array.from(ripples).findLast(ripple => !ripple.dataset.isHiding);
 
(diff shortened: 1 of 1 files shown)


johnleider:
What is the correlation to the import and the bug?

dgmpk:
> What is the correlation to the import and the bug?

After importing @vant/touch-emulator, when a mouse event is triggered, a touch event will be triggered, so this bug will be exposed.

johnleider:
> > What is the correlation to the import and the bug?
> 
> After importing @vant/touch-emulator, when a mouse event is triggered, a touch event will be triggered, so this bug will be exposed.

Do I have to install vant.js?

J-Sek (approved):

J-Sek:
Verified. That simulator duplicates events

<details>
<summary>relevant code</summary>

image

</details>

...so 2 ripples appear at the same time and the 2nd one is switched to hide twice.

PR #20789, opened 2024-12-18
feat(theme): add stylesheetId and scope options

### LINKED ISSUE #4065: [Feature Request] Add option to scope vuetify-theme-stylesheet styles
### Problem to solve
Webpack + postcss has been working pretty well for us, giving us the ability to scope any and all css in any given app to a specific selector, including the main Vuetify stylesheet. We noticed, however, that vuetify generates a stylesheet at run-time with an id of "vuetify-theme-stylesheet." Unfortunately this bypasses our postcss config at build-time and thus overwrites any existing styles with the same name.

### Proposed solution
Add the ability to scope the "vuetify-theme-stylesheet" stylesheet to a specific selector. Perhaps an option during...
```
Vue.use(Vuetify, {
   options: { ...
      scope: '#my-app'
    }
})
```
Ideally we wouldn't need to declare the selector twice (once in our postcss config and another in Vue.use()).

DerYeger:
## Description

This PR adds two new optional properties to `ThemeOptions` that enable a conflict-free usage of multiple Vuetify instances, e.g., in an app with microfrontends using Vuetify with different themes.

- `stylesheetId`: Customizable `id` for generated `<style>` element to allow for multiple Vuetify instances accros microfrontends.
- `scope`: Customizable scope for classes in generated `<style>` element.

closes #4065

### DIFF
--- a/packages/docs/src/pages/en/features/theme.md
+++ b/packages/docs/src/pages/en/features/theme.md
@@ -350,4 +350,10 @@ export const vuetify = createVuetify({
 
 ## Implementation
 
-Vuetify generates theme styles at runtime according to the given configuration. The generated styles are injected into the `<head>` section of the DOM in a `<style>` tag with an **id** of `vuetify-theme-stylesheet`.
+Vuetify generates theme styles at runtime according to the given configuration. The generated styles are injected into the `<head>` section of the DOM in a `<style>` tag with a default **id** of `vuetify-theme-stylesheet`.
+
+### Microfrontends
+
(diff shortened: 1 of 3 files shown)


KaelWD:
Maybe call it scope instead of prefix and always use `:where`, then we can switch to `@scope` when browser support improves?

DerYeger:
> Maybe call it scope instead of prefix and always use `:where`, then we can switch to `@scope` when browser support improves?

Sounds good to me! I pushed this change and updates the docs.

Issue #20724, opened 2024-11-22
[Bug Report][3.7.4] Wrong number week with locale-provider "fr"


SamsGuamejy:
### Environment
**Vuetify Version:** 3.7.4
**Vue Version:** 3.5.13
**Browsers:** Firefox 128.0, Google Chrome
**OS:** Windows 10

### Steps to reproduce
When I use the v-locale-provider with "fr" around the v-date-picker, the weekNumber for the first week of 2025, the date-picker write the week 53.
The weekNumber will be week 01 for the December 30, 2024 to January 5, 2025.

### Expected Behavior
Display the week 01 for the December 30, 2024 to January 5, 2025. ISO 8601 right

### Actual Behavior
Display the weekNumber 53 for the December 30, 2024 to January 5, 2025.

### Reproduction Link
[Original Playground](https://play.vuetifyjs.com/#eNplUMsOwiAQ/BXCWezFk6lN/A7rgdDVYilsePVg/HeRqqV6IczMzmZ2TnfqrKiOiNsYgO5p7WFExT00rSakjowj5m8GwmjPpQb7pjKpjOAKGFoTZQeWzPjQ0ott6Xcwj3ZpMUMpBrALT4jrzcQmgKEkx6C8xLzIcn2Flv5ZeHfjArRnY8rVu1IHBZF7aXSy70prGahaJVpuSvzPUZ8KkrLu4MXMFdVVUV2CTliJnjjwAZu6mmGS6GOjg1L5OT8BVDN3lw==)

[...]

PR #20794, opened 2024-12-19
fix(VDatePicker): show correct week number for selected locale

### LINKED ISSUE #20490: [Bug Report][3.6.7] Wrong calculation of the calendar week in date composable
### Environment
**Vuetify Version:** 3.6.7
**Vue Version:** 3.4.23
**Browsers:** Chrome 128.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
use a v-date-picker, display calendar week and navigate to December 2024

### Expected Behavior
The week starting the 30th of December should be week 1 not week 53

### Actual Behavior
Week 53 is dispayed instead of week 1

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpNjc0KwyAQhF9FPNdKrsEW+hy1BzEbIjG6+JNQSt+9W6XQy7LfzOzO/cVzsvKGeN4r8JGrAht6U+CqA2NqFwaxrQ0mMgQ6u0JieYmHOABWNm4xlOWi+TBozny0xgPRBESzS7nQ3VPEuaW/MdJl/y9/BUr+FRNmmxwWlqHUbneBVv4+hep9G48PCGZBHw==)

### Other comments
I guess it's a bug in the getWeek function in https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/composables/date/date.ts 

I tried this function also and it returns 53. 

According to the ISO standard it should be week 1: https://en.wikipedia.org/wiki/ISO_week_date

"... 
First week
The ISO 8601 definition for week 01 is the week with the first Thursday of the Gregorian year (i.e. of January) in it. The following definitions based on properties of this week are mutually equivalent, since the ISO week starts with Monday:

It is the first week with a majority (4 or more) of its days in January.
Its first day is the Monday nearest to 1 January.
It has 4 January in it. Hence the earliest possible first week extends from Monday 29 December (previous Gregorian year) to Sunday 4 January, the latest possible first week extends from Monday 4 January to Sunday 10 January.
It has the year's first working day in it, if Saturdays, Sundays and 1 January are not working days.
..."

Also performed a Wolfram Alpha search: https://www.wolframalpha.com/input?input=what+is+the+date+of+the+first+week+of+2025

There is a working implementation here: https://gist.github.com/vincentorback/8edffeca20e7a9e5e2a6?permalink_comment_id=1769808 and some more info here https://stackoverflow.com/questions/58708417/why-is-1-added but I don't know whether you can use this code (licence?).

simon-tackstrand:
## Description
Current getWeek implementation does not work for different locales.
Uses @date-io adapter to calculate week number.

resolves #20490, #20724

Requires @date-io version >= 3.2.0 for any project referencing vuetify

The previous getWeek implementation has been moved to the VuetifyDateAdapter, meaning the bug will still exist for those not using any @date-io implementation.

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-row>
        <v-col cols="12">
          <div style="height: 600px">
            <v-locale-provider locale="sv">
              <v-date-picker show-week />
            </v-locale-provider>
          </div>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```


johnleider:
Are you saying the only way to resolve this is to require a dependency?

simon-tackstrand:
> Are you saying the only way to resolve this is to require a dependency?

The dependency '@date-io' is already included in vuetify 3 so it is already required. With this PR we have just updated to a version which has a method for "getWeek". This enables anyone to get a correct getWeek implementation for all locales using any @date-io adapter compared to currently where a vuetify specific method is always used, even when a @date-io adapter is provided.

The VuetifyDateAdapter will still work for 'en-US', but this PR does NOT resolve the issue for locales where it calculates the week incorrectly. I will leave it for you to decide on how to proceed going forward with updating the VuetifyDateAdapter, in my point of view it is rather complex to get it correct for all locales since it requires information about each locale. An easier alternative could be an vuetify 2 inspired approach where the API allows the user to specify [locale-first-day-of-year](https://v2.vuetifyjs.com/en/api/v-date-picker/#props-locale-first-day-of-year).

The approach in this PR will allow anyone using a @date-io adapter to get a correct locale specific getWeek implementation. For those choosing to use the VuetifyDateAdapter the functionality will remain unchanged.

I would suggest to include this in the next release of Vuetify to give an option for those having issues and being without any alternative than to disable 'show-week'. The issues with the VuetifyDateAdapter can then be addressed later.

johnleider:
Disregard my previous comment, I think I was reviewing on my phone and it looked like we were adding a new dependency.

johnleider (approved):
Very clean, thank you.

simon-tackstrand:
Seems to be out of sync with vuetifyjs:master, can you help me?

KaelWD:
```
git merge upstream/master
git checkout upstream/master -- pnpm-lock.yml
pnpm i
git add .
git commit
git push
```

ikushum (approved):

johnleider (approved):

johnleider:
Closing in favor of https://github.com/vuetifyjs/vuetify/pull/21199.

PR #20824, opened 2025-01-02
export the type for header

davidmeirlevy:
## Description
the interface is missing.
we need the ability to use the interface in our codebase, and then to pass the headers dynamically, and we can't because the interface is only internal.

## Markup:

```vue
<script lan="ts">
const headers: MyInterface[] = {}
</script>
<template>
<v-data-table
 :headers="headers" <-- TS errors here -->
/>.   
</template>

```

### DIFF
--- a/packages/vuetify/src/components/VDataTable/index.ts
+++ b/packages/vuetify/src/components/VDataTable/index.ts
@@ -5,3 +5,5 @@ export { VDataTableRows } from './VDataTableRows'
 export { VDataTableRow } from './VDataTableRow'
 export { VDataTableVirtual } from './VDataTableVirtual'
 export { VDataTableServer } from './VDataTableServer'
+
+export type { DataTableHeader } from './types'; 


J-Sek:
I believe it was resolved in v3.8.0 ([`44be603`](https://github.com/vuetifyjs/vuetify/commit/44be603))

```ts
import type { DataTableSortItem, DataTableHeader } from 'vuetify'

const defaultSort = [{ key: 'startDate', order: 'asc' }] as DataTableSortItem[]

const headers = [
  { title: 'ID', value: 'id' },
  { title: 'Name', value: 'name' },
  { title: 'Size', value: 'size' },
  { title: '', key: 'actions' },
] satisfies DataTableHeader[]
```

Can be extended to be more strict on `value`s:

```ts
// utils/types.ts
import type { DataTableHeader as IDataTableHeader } from 'vuetify'
export type { DataTableSortItem } from 'vuetify'
export type DataTableHeader<T> = IDataTableHeader<T> & { value?: keyof T }
```

```ts
import type { DataTableHeader } from '~/utils/types'

const headers = [
  { title: 'ID', value: 'id' },
  { title: 'Name', value: 'name' },
  { title: 'Size', value: 'size' }, // <-- will underline here if "size" does not exist on the row item type
  { title: '', key: 'actions' },
] satisfies DataTableHeader<YourDataRowType>[]
```

PR #20874, opened 2025-01-17
feat(VSlider): introduce track disabled

### LINKED ISSUE #20872: [Feature Request] V-slider: Disable clicking on the slider track
### Problem to solve
We have a use case where we want to disable clicking on the v-slider's track. The only interaction allowed with the slider should be through moving the thumb around. 

### Proposed solution
Introduce a new disabled property that allows disabling the mouse down event on the slider track

mrooding:
## Description

resolves #20872

The current implementation keeps the overall structure and event listeners in place. It simply ignores click events if they do not come from the thumb element. 

I'm happy to discuss alternative options but would like to get some feedback first.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-slider :track-disabled="true" @end="onChange" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
    methods: {
      onChange (value) {
        console.log(value)
      }
    }
  }
</script>

```

### DIFF
--- a/packages/api-generator/src/locale/en/VSlider.json
+++ b/packages/api-generator/src/locale/en/VSlider.json
@@ -2,7 +2,8 @@
   "props": {
     "alwaysDirty": "When used with the **thumb-label** prop will always show the thumb label.",
     "inverseLabel": "Reverse the label position. Works with **rtl**.",
-    "vertical": "Changes slider direction to vertical."
+    "vertical": "Changes slider direction to vertical.",
+    "trackDisabled": "Disables clicking on the slider track"
   },
   "slots": {
     "thumb-label": "Slot for the thumb label.",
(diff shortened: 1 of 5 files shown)


mrooding:
Another solution could be to modify when we add `pointer-events: none;` using a css class. If we remove that from the overall container but keep it on the thumb it should also suffice. We could then remove the guard check inside the click event.

J-Sek:
Thank you for the interest in improving Vuetify. After careful consideration, if feels like too much trouble in exchange for 2 lines of straightforwards CSS and we will keep VSlider without this prop.

If you have any questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com/).

PR #20891, opened 2025-01-21
docs(VDataTable): remove obsolete reference to page.sync

zostay:
## Description
Makes a small change to the documentation for External Pagination of the Data and Display page of `v-data-table` docs. The docs momentarily confused me when it referred to the need to set a `.sync` attribute, which no longer seems to be part of the component. I just removed the offending sentence.


KaelWD:
Looks like there's actually about 30 references to this still.

PR #20937, opened 2025-01-29
fix(VSelect): improve selection performance

### LINKED ISSUE #20703: [Bug Report][3.7.4] v-autocomplete laggy when using "select all"
### Environment
**Vuetify Version:** 3.7.4
**Vue Version:** 3.5.13
**Browsers:** Chrome 130.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Create a simple v-autocomplete with large number of items (here 5K)

### Expected Behavior
On "Select all" click, all items should be active immediately.

### Actual Behavior
App freezing for several seconds before doing so.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/playgrounds/_GJSQQ)

### Other comments
It works well with a virtual-scroller and v-list-items inside. Is there any reason it's laggy with autocomplete ? Is it due to component intern logic ?
If so, is there any workaround for large sets of items like this one ?

Thank you in advance !

KaelWD:
## Description

fixes #20703

Time to run `transformIn`, primitive values:
| Item # | Before | ~~After~~ | After |
|--------|--------|--------|--------|
| 1k | 320ms | ~~56ms~~ | 0.5ms |
| 10k | 40sec | ~~780ms~~ | 1ms |
| 100k | &infin; | ~~75sec~~ | 8ms |
| 1M | &infin; | ~~&infin;~~ | 125ms |

Object values (`{ key: i }`):
| Item # | Before | After |
|--------|--------|--------|
| 1k | 900ms | 33ms |
| 10k | 88sec | 2.5sec |
| 100k | &infin; | &infin; |
| 1M | &infin; | &infin; |

return-object with item-value set appropriately uses primitive value lookup

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete
        v-model="selectedItems"
        :items="items"
        label="test"
        multiple
      >
        <template #prepend-item>
          <v-list-item @click="selectAll()">
            <v-list-item-title> Select all </v-list-item-title>
          </v-list-item>
        </template>
        <template #selection="{ item, index }">
          <template v-if="selectedItems.length > 5">
            <span v-if="!index">{{ selectedItems.length }}</span>
          </template>
          <template v-else>
            <span v-if="index" class="hint mr-1"> or </span>
            <v-chip
              :style="{ height: '24px' }"
              size="small"
              variant="flat"
              label
            >
              {{ item.title }}
            </v-chip>
          </template>
        </template>
      </v-autocomplete>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selectedItems = ref([])
  const items = Array.from({ length: 10000 }, (_, i) => ({ title: `Item ${i}`, value: i }))

  function selectAll () {
    selectedItems.value =
      items.length === selectedItems.value.length ? [] : items.map(c => c.value)
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -28,6 +28,7 @@ import { makeTransitionProps } from '@/composables/transition'
 import { computed, mergeProps, nextTick, ref, shallowRef, watch } from 'vue'
 import {
   checkPrintable,
+  deepEqual,
   ensureValidVNode,
   genericComponent,
   IN_BROWSER,
@@ -346,7 +347,7 @@ export const VAutocomplete = genericComponent<new <
       if (!item || item.props.disabled) return
 
       if (props.multiple) {
-        const index = model.value.findIndex(selection => props.valueComparator(selection.value, item.value))
+        const index = model.value.findIndex(selection => (props.valueComparator || deepEqual)(selection.value, item.value))
         const add = set == null ? !~index : set
 
         if (~index) {
--- a/packages/vuetify/.eslintrc.js
+++ b/packages/vuetify/.eslintrc.js
@@ -12,6 +12,7 @@ module.exports = {
   rules: {
     'no-console': 'error',
     'no-debugger': 'error',
+    'no-labels': 'off',
 
     // 'vue/html-self-closing': 'off',
     // 'vue/html-closing-bracket-spacing': 'off',
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -29,6 +29,7 @@ import { makeTransitionProps } from '@/composables/transition'
 import { computed, mergeProps, nextTick, ref, shallowRef, watch } from 'vue'
 import {
   checkPrintable,
+  deepEqual,
   ensureValidVNode,
   genericComponent,
   IN_BROWSER,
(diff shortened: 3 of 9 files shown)


johnleider:
- Is there a reason for valueComparator vs _valueComparator?
- For selectStrategies, how did you test the differences?
- It seems that most of the performance boost comes from caching the valueComparator, the other changes don't seem to do much

KaelWD:
- There was no valueComparator variable in scope already
- There's no behaviour difference, the map is created in strategy.in in the first place so cloning it for every iteration is just a waste of performance
- Caching hasNullItem and returnObject is worth 100ms with a million items. `new Map(map)` takes 24ms with just 100 items. The double filter is pretty quick but might as well save the gc on an array that's going to be thrown away instantly.

KaelWD:
There's still a O(n^2) deepEqual to find the first selected item index: https://github.com/vuetifyjs/vuetify/blob/c4ba8bf3e3a4d3a6ecb6fa00b8e725b3ff16c313/packages/vuetify/src/components/VSelect/VSelect.tsx#L317-L319

And a linear search to figure out if we're adding or removing a selection: https://github.com/vuetifyjs/vuetify/blob/c4ba8bf3e3a4d3a6ecb6fa00b8e725b3ff16c313/packages/vuetify/src/components/VSelect/VSelect.tsx#L266

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -348,7 +349,7 @@ export const VAutocomplete = genericComponent<new <
       if (!item || item.props.disabled) return
 
       if (props.multiple) {
-        const index = model.value.findIndex(selection => props.valueComparator(selection.value, item.value))
+        const index = model.value.findIndex(selection => (props.valueComparator || deepEqual)(selection.value, item.value))
```
johnleider, line 352: Would setting the default in makeItemsProps do the same?

PR #20960, opened 2025-02-05
feat(styles): cut down CSS reset

KaelWD:
90% of this is for ancient browsers and the rest can be done in component styles instead of affecting the entire page.


packages/vuetify/src/styles/generic/_reset.scss
```diff
@@ -1,122 +1,29 @@
...
     ::before,
     ::after {
-      text-decoration: inherit; /* Inherit text-decoration and vertical align to ::before and ::after pseudo elements */
+      text-decoration: inherit;
       vertical-align: inherit;
     }
 
-    * {
-      padding: 0; /* Reset `padding` and `margin` of all elements */
-      margin: 0;
```
J-Sek, line 39: This one maybe risky. I just toggled it off on my website and some things moved (`<body>` and `<hr>` were the most obvious). Tested on Brave.

image

image

KaelWD:
It looks like the only component this actually affected was VField, everything else still looks identical.
Should we keep some sensible defaults for headings and lists? Reset to 0 is kinda awful but the browser defaults have huge margins.

packages/vuetify/src/styles/generic/_reset.scss
J-Sek: `VDivider.sass` needs `margin-block: 0`, while the `body` might need to replace `*` here.. at least to get rid of the margin.

KaelWD:
I think typography can be separate, maybe bring back the pre-0.17 styles as an optional import after #22396?

PR #20962, opened 2025-02-05
fix(nested): don't toggle disabled children

### LINKED ISSUE #20386: [Bug Report][3.7.0] VTreeView - Disabled items selected when parent item is selected
### Environment
**Vuetify Version:** 3.7.0
**Vue Version:** 3.4.38
**Browsers:** Edge 127.0.0.0
**OS:** Windows 10

### Steps to reproduce
Select an item with child items. ('Applications' in the reproduction link) All child items are selected, including items that are disabled and should not be selectable.

### Expected Behavior
Disabled items can not be selected.

### Actual Behavior
Disabled items are selected if the parent is selected.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNUUFuwjAQ/MrKF0AKoS1IrayAVPURPQAH4yzFkpNY9iZthfL3OjEBp2olfFnP7nhmbG/PzFm5eDUmbWpknGWEhdGCcLMrAbJmThaxUfjZQQCu/Nytd6yvOxa6HZgbWxkXcGWwnAutA3KoUZI4aIymjqw3+fj2UkWtSRk/vagFfsyQWjinZCBsssUtlU+ZLaLIHjpplaE+Pn6ZyhLkeBTeA85BPxckOExnsN7A9NILd3ActgOGgR5WI3SNHB6TuKny3x1SpD1t4h9UKylIVaUDPhlxcuW6x/Bnj0I7HM3kSencYjkK0ocZEjwlva0vg9eb0FjmwgIHYcwkiQzI1gjtyCGSWgYpX65SJ1sV+IdQn/R/pVVQ8mVQeseDo8oWd2rtI3SbDN121m/a7rOv38vahC3TVbp8Yd3mOX1g+x+AH9Km)

johnleider:
fixes #20386

## Markup:

```vue
<template>
  <v-treeview
    v-model="model"
    :items="items"
    select-strategy="classic"
    item-props
    open-all
    selectable
  />
</template>

<script>
  export default {
    data: () => ({
      model: [],
      items: [
        {
          value: 1,
          id: 1,
          title: 'Applications :',
          disabled: false,
          children: [
            {
              value: 2,
              id: 2,
              title: 'Calendar : app',
              disabled: false,
              children: [
                { value: 12, id: 12, title: 'Calendar : app2', disabled: false },
                { value: 13, id: 13, title: 'Chrome : app2', disabled: true },
                { value: 14, id: 14, title: 'Webstorm : app2', disabled: false },
              ],
            },
            { value: 3, id: 3, title: 'Chrome : app', disabled: false },
            { value: 4, id: 4, title: 'Webstorm : app', disabled: false },
          ],
        },
      ],
    }),
  }
</script>

```


KaelWD:
Wondering if this should be indeterminate instead, not all the children are actually selected.

Screenshot_20250206_165921

3.8 adds `select-strategy="trunk"` which is classic with a different out function, does that need to account for this too?

packages/vuetify/src/composables/nested/__tests__/selectStrategies.spec.ts
```diff
@@ -731,5 +765,51 @@ describe('selectStrategies', () => {
         ['6', 'on'],
       ]))
     })
+
+    it('should not select disabled children when parent is selected', () => {
```
yuwu9145, line 769: return-object is usually vulnerable when nested composable has changes, so probably better add a test case to cover return-object as well

johnleider:
> Wondering if this should be indeterminate instead, not all the children are actually selected.

I fiddled with this and was ultimately unsuccessful in getting the clicked on element that has the disabled children to default to the indeterminate state.

johnleider:
> Wondering if this should be indeterminate instead, not all the children are actually selected.
> 
> Screenshot_20250206_165921
> 
> 3.8 adds `select-strategy="trunk"` which is classic with a different out function, does that need to account for this too?

I can't figure out how to make this work in all scenarios, namely when clicking on a parent. The problem is clicking a checkbox of a parent and having it go straight to indeterminate instead of checked.

PR #20973, opened 2025-02-10
fix(validation): add validation watcher for rules

### LINKED ISSUE #18875: [Bug Report][3.4.7] Changing :rules of Vuetify Inputs does not affect the validation
### Environment
**Vuetify Version:** 3.4.7
**Last working version:** 2.3.0
**Vue Version:** 3.4.0-beta.2
**OS:** Windows 10

### Steps to reproduce
Warning: possible duplicate, could be closed because of high complexity, but i've found no workaround to make it work securely in my big application, read other comments

- Input "aa"
- Enable the checkbox (replaces the rules array with different rules)
- Focus Input
- Blur Input

### Expected Behavior
Changing the Rules should immediately change the valid status of the form to not allow invalid inputs/confusion of the user. The error message should be immediately displayed.

### Actual Behavior
Dynamically replacing the rules array of an input does not trigger a revalidation. This leaves the form valid, until the user revisits the input. 

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNVE2L2zAQ/SuzOmwc2CSlKRRS23QplPZW9tAe4lBkW07EypIry2Yhm//e0UdiO5vDXoz09ObN08zI2yNpdbF6bJpl3zGyIbFhdSOoYWkmAeJ+QZvGLd2mUNJQLpkOkAcPrHjO1csZAhA0ZyLJCDUgGG3tV+7NAdYZGUj9olalp5UlN1xJKnQn2MBJ49WgPs5YKV2DZhXG2mVGRmI9FbzMyIXuAgx7MYuKM1EOMAQFSWv2UzadGZsb26vb/fRoY222ePLXLaKp//mYvBps412sWQQuUMn79Lf1u4Hj0Rk/neKVRSeUx4v+E+pb7jTjzaBvqpMGVAXOow267XbpezPVcIUf9foyALmRoWyatQxLBl8LwYvn0ImlQyMsQfpkV/AdQSuGcX6gcO0nKl6NJg23baF5YwCDugYHSO5R0rShkbxulDZwtLnhBJVWNcxwYGc21BLQLI4ZdgoSy4lbo7ncp9HsBxNCwR+lRXk3mw9UV+1AzpXCMZWvshMijeZfBta0VlN6GlVUtGxMd3PpSfbiU63LnAWGazxen+aCeaa7qTRMV7RgMD4HOPoWONsIRi3HvpkNBDPzDfzCovCWhatvd65wp2tV6+ss5poV0uAlIxTpFS+d53A6wYJY1cnC8oG3T+xfxzUro34DPi/y/QJewdbznItXEPWQJAng8yDzM2zTmE5LRHXQykgwgOlGBCs2djHyQY37y/hBXr/Xy13iwft76MMjgBjWt6zVHfbvQHuGqcIfbQ3FgWqcz3eYnRTt5isc+hiyW4tXD/WNr+1Q/4erGuy8K2/qLd0f42m88s8OZ4WcHsh6+Wn5YZEzQ5cfid9+Jrv/t6gDFQ==)

### Other comments
The behavior of showing/hiding the error message on first form input of the user is already very complex. Dynamic rules make this even more complex. Should a changing rule always make the Input immediality show the error or only if its already dirty/has some input? I could'nt test some combinations, because i failed to find/access the `isDirty`/`isPrestine` state of the input.

I've found https://github.com/vuetifyjs/vuetify/issues/8698#issuecomment-921092838. And tried the `validate(silent: true)` option, which didn't help me because i couldn't find a isDirty state. The  `<v-validation :rules="[rules]">` is not usable to me because i've found no documentation on how to use it with all the existing variants of inputs, having to write the inputs on my own would make me question the use of the framework for my size of application.

---

Edit: The following snippet is my current wip workaround in my Compontent that wraps `v-text-field` that does not completly work because the way of getting the silent value is not robust/consistent on prisitine/dirty inputs. 

```ts
const vtextfield = ref<
  | (ComponentPublicInstance & { validate: (silent: boolean) => Promise<string[]>,})
  | null
>(null);

watch(() => props.rules, () => {
  nextTick(() => validate());
}, {deep: true})

async function validate(): Promise<void> {
  const silent = props.modelValue === '' || props.modelValue === null || props.modelValue === undefined;
  await vtextfield.value?.validate(silent);
}
```

---

Edit2: Another Workaround i've found looks like this (⚠️ didn't work consistent for me):

```html
<v-text-field ref="vtextfield">
    <template v-slot:append="{isDirty}">
       {{_workaroundDirty(isDirty.value)}}
    </template>
</v-text-field>

<script lang="ts">

watch(() => props.rules, () => {
  nextTick(() => validate());
}, {deep: true})

const dirty= ref<boolean>(false);

function _workaroundDirty(isDirty: boolean): string {
  dirty.value = isDirty;
  return "";
}

async function validate(): Promise<void> {
  const silent = !dirty.value;
  await vtextfield.value?.validate(silent);
}
</script>
```

---

Edit 3: Because we had to move on in our project we decided to replace the rules evaluation and v-form with our own typescript code and just pass errors to the inputs. Therefore the issue is no longer relevant to us. I'll leave it open to document that the migration was quite effortful and others might arrive at this problem.

johnleider:
After extended consideration and testing, I think this is a risky change.

fixes #18875

```vue
<template>
  <v-app>
    <v-container>
      <v-checkbox
        label="at least length 3"
        v-model="additionalrule"
      ></v-checkbox>
      <v-form ref="form" v-model="valid">
        <v-text-field
          ref="nameInput"
          v-model="msg"
          :rules="_rules(additionalrule)"
        />
      </v-form>

      <div>Valid: {{valid}}</div>
      <div>AdditionalRule: {{additionalrule}}</div>
      <div>Count of rules: {{_rules(additionalrule).length}}</div>
    </v-container>

    <v-btn ref="reset" @click="form.reset()">Reset Form</v-btn>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'


  const msg = ref<string>('aa')
  const valid = ref<boolean|null>();
  const additionalrule = ref<boolean>(false);
  const form = ref<Form>();
  const nameInput = ref<Validateable>();

  interface Validateable  {
    validate(silent: boolean): Promise<string[]>
  }

  interface Form {
    resetValidation(): void;
    reset(): void;
  }

  function isRequired(v: string): string | true {
    if (v === "") {
      return "required";
    }
    return true;
  }

    function atleastlength3(v: string): string | true {
    if (v != null && v.length < 3) {
      return "must have at least 3 chars";
    }
    return true;
  }


  function _rules(additionalrule: boolean) {
   if (additionalrule) {
      return [isRequired, atleastlength3];
   }
   return [isRequired];
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/composables/validation.ts
+++ b/packages/vuetify/src/composables/validation.ts
@@ -6,7 +6,7 @@ import { useToggleScope } from '@/composables/toggleScope'
 
 // Utilities
 import { computed, nextTick, onBeforeMount, onBeforeUnmount, onMounted, ref, shallowRef, unref, watch } from 'vue'
-import { getCurrentInstance, getCurrentInstanceName, getUid, propsFactory, wrapInArray } from '@/util'
+import { deepEqual, getCurrentInstance, getCurrentInstanceName, getUid, propsFactory, wrapInArray } from '@/util'
 
 // Types
 import type { PropType } from 'vue'
@@ -175,6 +175,8 @@ export function useValidation (
   watch([isValid, errorMessages], () => {
     form.update?.(uid.value, isValid.value, errorMessages.value)
   })
+
+  watch(() => props.rules, () => validate())
 
   async function reset () {
     model.value = null
--- a/packages/vuetify/src/composables/__tests__/validation.spec.ts
+++ b/packages/vuetify/src/composables/__tests__/validation.spec.ts
@@ -173,4 +173,19 @@ describe('validation', () => {
 
     expect(wrapper.vm.isValid).toBe(true)
   })
+
+  it.only('should rerun validation when rules changes', async () => {
+    const wrapper = mountFunction({
+      modelValue: 'bar',
+      rules: [(v: any) => v === 'foo' || 'Invalid'],
+    })
+
+    await wrapper.vm.validate()
+
+    expect(wrapper.vm.isValid).toBe(false)
+
+    await wrapper.setProps({ rules: [(v: any) => v === 'bar' || 'Invalid'] })
+
+    expect(wrapper.vm.isValid).toBe(true)
+  })
 })


J-Sek:
I was hoping that it would handle numeric range (split into fields rather than using `v-range-slider`), and it kinda does, but also requires sideeffect (that makes both ESLint and Vue angry with "_Maximum recursive updates exceeded_")

When the lines are kept commented, computed does not reevaluate the rules. It may be more of a Vue problem than something for Vuetify to cover. I am fine with my typical workarounds (forcing error on 2nd field only and/or using `:key=`). Just wanted to share in case anyone else would expect this fix to handle field pairs.

```vue
<template>
  <v-app>
    <v-container>
      <v-number-input v-model="min" :rules="minRules" label="Min" clearable />
      <pre>Min rules change counter: {{ minRulesChanges }}</pre>

      <v-number-input v-model="max" :rules="maxRules" label="Max" clearable />
      <pre>Max rules change counter: {{ maxRulesChanges }}</pre>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  const min = ref<number | null>(0)
  const max = ref<number | null>(100)

  const minRulesChanges = ref(0)
  const maxRulesChanges = ref(0)

  const minRules = computed(() => {
    // minRulesChanges.value++
    console.log('minRules changed')
    return [
      (v: number | null) => v != null || 'Value is required',
      (v: number | null) => v == null || max.value == null || v <= max.value || 'Has to be smaller than "max"',
    ]
  })

  const maxRules = computed(() => {
    // maxRulesChanges.value++
    console.log('maxRules changed')
    return [
      (v: number | null) => v != null || 'Value is required',
      (v: number | null) => v == null || min.value == null || v >= min.value || 'Has to be larger than "min"',
    ]
  })
</script>
```

packages/vuetify/src/composables/validation.ts
```diff
@@ -176,6 +176,8 @@ export function useValidation (
     form.update?.(uid.value, isValid.value, errorMessages.value)
   })
 
+  watch(() => props.rules, () => validate())
```
J-Sek, line 179: `watch(..., { deep: 1 })` worth considering, although I am not sure.. maybe would not matter.

PR #19709, opened 2024-04-27
feat(VTimeInput): add new component

johnleider:
## Motivation and Context

- typing from keyboard should enforce format
  - ideally similar to native input
- minimal picker (only clock and period toggle)

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="400">
      <v-code>{{ String(value1) }}</v-code>
      <v-time-input v-model="value1" hide-actions />
      <v-code>{{ String(value2) }}</v-code>
      <v-time-input v-model="value2" format="24hr" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const value1 = ref(null)
  const value2 = ref(null)
</script>
```

PR #20983, opened 2025-02-13
fix(VTimePickerControls): disable in disabled forms

mtdvlpr:
## Markup:

```vue
<template>
    <v-form disabled>
        <v-time-picker />
    </v-form>
</template>
```


KaelWD:
This does nothing as VTimePicker is not form aware. It can be revisited if we ever do a VTimeInput.

mtdvlpr:
You're already working on it: #19709, but it doesn't include this fix yet.

PR #20987, opened 2025-02-15
feat(VNumberInput): support control holding down

yuwu9145:
## Markup:

```vue
<template>
  <v-container>
    <v-row>
      <v-col cols="12" md="4" sm="4">
        <h5>Default</h5>

        <v-number-input 
          v-if="showNumberInput"
          control-variant="default"
        ></v-number-input>
      </v-col>

      <v-btn
        @click="remove"
      >Remove Number Input</v-btn>
    </v-row>
  </v-container>
</template>
<script setup>
import { ref } from 'vue'

  const showNumberInput = ref(true)
  function remove (event) {
    showNumberInput.value = false
  }
</script>

```


packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -121,12 +121,35 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
...
+    // Control holding down states
+    let hldDwnReqNextTime = 0
+    let hldDwnReq = -1
+    const hldDwnReqDelay = 100
+    const hldDwnDelay = 500
+    let hldDwn: null | 'up' | 'down' = null
+    let hldDwnTimeout = -1
+
+    function controlHoldingDownWatcher (time: number) {
+      hldDwnReq = requestAnimationFrame(controlHoldingDownWatcher)
```
J-Sek, line 133: - `hldDwnReq` » `holdDownRaf` or `holdActionRaf`

I actually don't care much about the suffix. It can be `*Raf`,  `*Frame`,  `*Req`. My issue is with saving 2 vowels for no obvious gain.

J-Sek (commented):
I only see potential for improvement around variable naming and maybe avoiding boolean as direction argument. Other than this, it delivers the functionality and I was not able to break it or come up with edge case scenario.

packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
```diff
@@ -121,12 +121,35 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
...
+      if (time < hldDwnReqNextTime) return
+      hldDwnReqNextTime = time + hldDwnReqDelay
+
+      if (hldDwn) {
+        toggleUpDown(hldDwn === 'up')
+      }
+    }
+
     onMounted(() => {
+      hldDwnReq = requestAnimationFrame(controlHoldingDownWatcher)
```
KaelWD, line 143: Why is this running constantly instead of only while the button is held?

yuwu9145: Thanks, refactored

KaelWD:
@yuwu9145 how's that?

yuwu9145:
> @yuwu9145 how's that?

Thanks, `setInterval` has less executions than `requestanimationframe`, it's more performant

Issue #21005, opened 2025-02-21
[Bug Report][3.7.13] VNumberInput doesn't apply precision when disabled/readonly


mathiasrando:
### Environment
**Vuetify Version:** 3.7.13
**Vue Version:** 3.5.13
**Browsers:** Chrome 132.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
See the rendered value in the three inputs.

### Expected Behavior
All three inputs should apply the precision.

### Actual Behavior
The inputs where `controlsDisabled` resolves to `true`(so effectively being `disabled` or `readonly`) does not apply the precision.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlkMsKgzAQRX9lyKYtVKV1J1rod9Quoo4QyGPIQyjivzeaUkq3bkLmMHOYuY+ZOdsXd6J8CsgqVntUJLnHW6sB6injRNt3K3qjPRca7QdtUAfVoc2EpuBhypQZUDYtS7hlUJHFXjhhdKRlBMWu6UE43kkcdmos8sFo+fpq6uL/vpWk8+viJ5ZYut4K8uDQh5SOUGSshzlqR1hgtEbBISZ6WNsBotd5SMtAszYdL/m1PK3m5IoWtpx1kHJ7nm/4nYhl)

crystalfp:
**Cannot create new issues!!**
Pushing new issue gives: "No template available"
Please fix

mathiasrando:
> **Cannot create new issues!!** Pushing new issue gives: "No template available" Please fix Sorry for polluting this issue, but don't know how to signal this problem.

Hey 🙂 

You can create issues here: https://issues.vuetifyjs.com/
This way the issues will follow a specific template and ensures that a reproduction is provided as well 👍

J-Sek:
It was on purpose – to display value as-is – but I am not sure if should really work this way in the first place. It would make more sense for all states to behave the same. Just need a second pair of eyes to review the change.

PR #21010, opened 2025-02-23
fix(VNumberInput): apply precision even when disabled/readonly

babu-ch:
## Description

fixes https://github.com/vuetifyjs/vuetify/issues/21005

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-number-input v-model="number" :precision="3" />
      <v-number-input v-model="number" :precision="3" disabled />
      <v-number-input v-model="number" :precision="3" readonly />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const number = ref(1.23)
</script>

```

### DIFF
--- a/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/labs/VNumberInput/VNumberInput.tsx
@@ -103,8 +103,8 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
     watchEffect(() => {
       if (isFocused.value && !controlsDisabled.value) {
         // ignore external changes
-      } else if (model.value == null || controlsDisabled.value) {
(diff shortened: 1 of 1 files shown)


J-Sek:
Simplified the code a bit. It will also fix another problem when readonly/disabled fields go empty for `0` value. [playground](<template>
  <v-app>
    <v-container>
      <v-text-field v-model="number" />
      <v-number-input v-model="number" :precision="3" />
      <v-number-input v-model="number" :precision="3" disabled />
      <v-number-input v-model="number" :precision="3" readonly />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">

  const external = ref('1.22293')
  const number = computed({
    get: () => Number(external.value),
    set: (v: number) => { external.value = String(v) },
  })
</script>
)

babu-ch:
@J-Sek 
oh, thank you! 😄

PR #21035, opened 2025-02-27
refactor(VField): convert display from grid to flex

johnleider:
In order to remove inline margin on clearable, we need to use gap. Grid gap always renders space defined in the grid template areas, so in order to account for this, we would need to handle all permutations available (prepend, prepend+append, prepend+append+clear, etc). Because of how many different combinations of elements a VField can contain, I'm opting to starting moving some things back to flex box.

https://github.com/w3c/csswg-drafts/issues/5813

## Markup:

```vue
<template>
  <v-responsive class="mx-auto" max-width="444">
    <v-text-field
      v-model="model"
      append-inner-icon="$vuetify"
      hide-details="auto"
      label="Last name"
      clearable
    />

    <v-text-field
      v-model="model"
      append-inner-icon="$vuetify"
      hide-details="auto"
      label="Last name"
      clearable
      persistent-clear
    />
  </v-responsive>

</template>

<script setup>

  const model = shallowRef('cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc')
</script>

```


KaelWD:
Ok but now we have a huge gap for *every* icon not just clear
image

PR #21038, opened 2025-02-28
fix(VFileUpload): remove value from input file correctly

### LINKED ISSUE #21029: [Bug Report][3.7.14] File upload: select previous selected file not shown
### Environment
**Vuetify Version:** 3.7.14
**Vue Version:** 3.5.13
**Browsers:** Chrome 133.0.0.0
**OS:** Windows

### Steps to reproduce
- Select a file
- Remove item
- Select file again

### Expected Behavior
Item should be shown as before

### Actual Behavior
Item not be shown

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdjs0KwjAQhF9lyUUFa+7SCj6H9RDTLQQ2yZKfXkrf3SRVEC/L7jD7zTxWEYOWd+bLklFcRZ/QMqmEt9EB9EunmNvaDu1dUsZh+EhNnA1hl5m8mmDprJ+QhlFUdRSgCVVQL0KQX4z851Rlj+nlT3w5ow6GE0RMeW9hLPuQYIWAM2wwB2/hUJofqh2gcGOCGg1DtRxPlblTyr/Yzi4TtfF8A94kVIY=)

SonTT19:
## Description

fixes #21029 

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-file-upload v-model="file" clearable />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const file = ref()
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VFileUpload/VFileUpload.tsx
+++ b/packages/vuetify/src/labs/VFileUpload/VFileUpload.tsx
@@ -163,9 +163,10 @@ export const VFileUpload = genericComponent<VFileUploadSlots>()({
     }
 
     function onClickRemove (index: number) {
-      model.value = model.value.filter((_, i) => i !== index)
+      const newValue = model.value.filter((_, i) => i !== index)
+      model.value = newValue
 
-      if (model.value.length > 0 || !inputRef.value) return
+      if (newValue.length > 0 || !inputRef.value) return
 
       inputRef.value.value = ''
     }


johnleider:
This works but I don't understand why it's needed.

SonTT19:
> This works but I don't understand why it's needed.

```
// model.value.length before = 1
model.value = model.value.filter((_, i) => i !== index)
// after filter if you access into model.value, it does not return newly set value, but still return the old one,
// in this case model.value.length still = 1
// even though model.value.filter((_, i) => i !== index) already clear all the data
```
Not really sure why model.value isn't up to date, maybe related to either proxiedModel or how component rendered

KaelWD:
Yeah because proxiedModel uses props and emit as the state when controlled, write then read has to wait for the parent to update.

```js
emit('update:modelValue', 'new')
log(props.modelValue) // 'old'
await nextTick()
log(props.modelValue) // 'new'
```

KaelWD (approved):

PR #21044, opened 2025-03-01
feat(VDatePicker): add "click:date" event

### LINKED ISSUE #21037: [Feature Request] Add "click:day" event to VDatePicker
### Problem to solve
Currently, `VDatePicker` does not provide an event specifically for detecting when a user clicks on a day. While `update:modelValue` captures changes to the selected date(s), it does not allow distinguishing between different types of clicks. This makes it difficult to implement custom behaviors, such as advanced date selection logic or showing a tooltip.

In my case, I have custom logic for selecting a range of dates. When clicking on a date within the selected range, I want to start a new range from that date instead of simply removing it from the selection. However, since there's no dedicated `click` event for days, achieving this behavior requires complex workarounds.

### Proposed solution
Introduce a new `click:day` event that emits whenever a user clicks on a day in `VDatePicker`. This event should include:  

- The clicked date  
- Whether the date is currently selected  
- Whether the date is disabled  

With this event, developers will have greater flexibility in handling custom interactions, such as modifying selection logic or enhancing UI behavior based on clicks.

luizhenriquerdias:
## Description
resolves #21037

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker
        @click:date="clickedDate = $event"
      />

      <p v-if="!clickedDate">Click on any date</p>
      <p v-else>Clicked date: {{ clickedDate }}</p>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const clickedDate = ref(null)
</script>
```


packages/vuetify/src/components/VDataTable/VDataTableRow.tsx
```diff
@@ -143,7 +143,7 @@ export const VDataTableRow = genericComponent<new <T>(
                       <VCheckboxBtn
                         disabled={ !item.selectable }
                         modelValue={ isSelected([item]) }
-                        onClick={ withModifiers(event => toggleSelect(item, props.index, event as MouseEvent), ['stop']) }
+                        onClick={ withModifiers((event: Event) => toggleSelect(item, props.index, event as MouseEvent), ['stop']) }
```
luizhenriquerdias, line 146: This was causing the build to fail

packages/vuetify/src/components/VDataTable/composables/select.ts
```diff
@@ -2,7 +2,7 @@
 import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utilities
-import { computed, inject, provide, ref, shallowRef } from 'vue'
```
luizhenriquerdias, line 5: This was causing the build to fail

packages/api-generator/src/locale/en/VDatePickerMonth.json
```diff
@@ -16,6 +16,6 @@
   "events": {
     "update:month": "Fired when the month changes.",
     "update:year": "Fired when the year changes.",
-    "day": "Fired when a day is clicked."
```
luizhenriquerdias, line 19: This event doesn't exist in the component

Issue #19998, opened 2024-06-12
[Bug Report][3.6.9] V-inputs counter with hide-details auto don't fully hide details


Yanis-Riani:
### Environment
**Vuetify Version:** 3.6.9
**Vue Version:** 3.4.27
**Browsers:** Chrome 125.0.0.0
**OS:** Windows 10

### Steps to reproduce
N/A

### Expected Behavior
When using `v-input` with the `hide-details` prop set to `auto` and `counter` enabled, the details section should automatically hide if there is nothing to display. This means that if the input meets all validation criteria or there are no errors or additional messages to show, the details area should remain collapsed, preventing any unnecessary space or blank areas under the input field.

### Actual Behavior
Despite the `hide-details` prop being set to `auto`, the `v-input` details section remains persistently visible when the `counter` is enabled. This results in a persistent blank space underneath the input field even when there are no messages or errors to display.

### Reproduction Link
[...]

PR #21050, opened 2025-03-02
fix(VTextField): hide details with counter and hide-details="auto"

babu-ch:
## Description

fixes https://github.com/vuetifyjs/vuetify/issues/19998

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <p>Counter only</p>
      <v-text-field
        v-model="msg"
        counter
        class="bg-indigo-lighten-2"
        :rules="[requiredValidator]"
        variant="outlined"
      />
      <p>With counter & hide details</p>
      <v-text-field
        v-model="msg"
        counter
        hide-details="auto"
        class="bg-indigo-lighten-2"
        :rules="[requiredValidator]"
        variant="outlined"
      />
      <p>Hide details normal behavior</p>
      <v-text-field
        v-model="msg"
        hide-details="auto"
        class="bg-indigo-lighten-2"
        :rules="[requiredValidator]"
        variant="outlined"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const requiredValidator = value => {
    return !!String(value).trim().length || 'This field is required'
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.tsx
+++ b/packages/vuetify/src/components/VTextField/VTextField.tsx
@@ -155,7 +155,7 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
 
     useRender(() => {
       const hasCounter = !!(slots.counter || (props.counter !== false && props.counter != null))
-      const hasDetails = !!(hasCounter || slots.details)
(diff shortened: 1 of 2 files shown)


johnleider:
This breaks the transition of the non hide-details variant.

PR #21071, opened 2025-03-06
fix(VDataTableRow): provide correct defaults for special slots

johnleider:
When using the special **item.data-table-select** and **item.data-table-expand** slots, would always return at:

image

and never actually get to the code that called **slots['item.data-table-expand']** or select.

Since I was fixing that, I brought the slots in line with how we handle defaults in other components, so that you can actually modify these with relative ease now.

## Markup:

```vue
<template>
  <v-container class="pa-md-12">
    <v-data-table :items="items" show-expand show-select>
      <template #item.data-table-select="{ props: selectProps }">
        <v-checkbox-btn v-bind="selectProps" />
      </template>

      <template #item.data-table-expand="{ props: expandProps }">
        <v-btn v-bind="expandProps" />
      </template>

      <template #expanded-row="{ columns, item }">
        <tr>
          <td :colspan="columns.length">
            I'm expanded for item {{ item.id }}!
          </td>
        </tr>
      </template>
    </v-data-table>
  </v-container>
</template>

<script setup>
  const items = [
    { id: 'John', age: 42 },
    { id: 'Jane', age: 36 },
    { id: 'Joe', age: 30 },
  ]
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VDataTable/VDataTableRow.tsx
+++ b/packages/vuetify/src/components/VDataTable/VDataTableRow.tsx
@@ -148,7 +148,6 @@ export const VDataTableRow = genericComponent<new <T>(
                           VCheckboxBtn: {
                             disabled: !item.selectable,
                             modelValue: isSelected([item]),
-                            onClick: withModifiers(() => toggleSelect(item), ['stop']),
                           },
                         }}
                       >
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VDataTable/VDataTableRow.tsx
```diff
@@ -136,10 +141,24 @@ export const VDataTableRow = genericComponent<new <T>(
...
-                  if (slots[slotName] && !mobile.value) return slots[slotName]?.(slotProps)
-
                   if (column.key === 'data-table-select') {
-                    return slots['item.data-table-select']?.(slotProps) ?? (
+                    return slots['item.data-table-select'] ? (
+                      <VDefaultsProvider
+                        defaults={{
+                          VCheckboxBtn: {
+                            disabled: !item.selectable,
+                            modelValue: isSelected([item]),
```
KaelWD, line 150: Is there any point having these provided if you have to use `v-bind="props"` anyway to make the component work?

J-Sek (approved):

PR #21076, opened 2025-03-06
feat(VField/VInput): add new props icon-color and glow

### LINKED ISSUE #17763: [Feature Request] prepend/append icon focus color
### Environment
**Vuetify Version:** 3.3.7
**Last working version:** 2.7.0
**Vue Version:** 3.3.4
**Browsers:** Safari 16.5.1, Google Chrome, Safari, Mozilla Firefox
**OS:** Mac OS 10.15.7

### Steps to reproduce
The VInput, icons used to reflect the component color in v2. In v3, this does not work any more.

### Expected Behavior
Icons in component color.

### Actual Behavior
Icons stay gray.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrlUstOwzAQ/BXjS0Gq4wNISFVBcOMPOBAOId6CwS+tNwFU9d9ZJwWigqpKHLnFM+MZ72bu1vI6parvQC7kksAn1xBc1kGIZa+alIbP4dDGQI0NgFtoAAneSK0sOPMJCtErHw24i1r6/FjLb4LtIBhlA5soy35FYqxqHCD9JtwjSQgHmn0p92ja6CIymbu2hZynVN+gbQIx2QUD6HgD5pvX02VkcNDS/17EUu82pSBjkZZ6UjA+5hZtIpGBurFn1qeIJNYCYSU2YoXRixl3c1bk5XEhk+BdiouiOJ7dgHNR3EZ05mh2UgJGSzaTczm6Kd+k6jnHwAVfF5d6S/BwCzEgBeOUcq7lE1HKC61bE/ga/z7bYxWAdEheX7FMYxfIelAm+qvT6rQ608ZmmsIVZK8eML5mQDap5XwSQ3b1fkDUVjkknI8JW0i55iGXiB/WmvN6bgHybwYEPHSinWvTqXaoH5OV9E0dNrxvbsRL3ll1Wzq0EHefj/zTyINZcbovgZu5HN7JycMFef8BUqmN/A==)

johnleider:
resolves #17763

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-input
        :focused="isFocused"
        append-icon="$vuetify"
        color="primary"
        prepend-icon="$vuetify"
        icon-color
      >
        <v-field
          :active="isFocused"
          :focused="isFocused"
          append-inner-icon="$vuetify"
          color="primary"
          label="Custom implementation"
          prepend-inner-icon="$vuetify"
          glow
          icon-color
          v-click-outside="() => isFocused = false"
          @click="isFocused = true"
        >
          <template #default="{ props: fieldProps }">
            <div v-bind="fieldProps" />
          </template>
        </v-field>
      </v-input>

      <v-text-field
        :glow="glow"
        append-icon="$vuetify"
        append-inner-icon="$vuetify"
        color="primary"
        label="Textfield"
        prepend-icon="$vuetify"
        prepend-inner-icon="$vuetify"
      />

      <v-select
        :glow="glow"
        append-icon="$vuetify"
        append-inner-icon="$vuetify"
        color="primary"
        label="Select"
        prepend-icon="$vuetify"
        prepend-inner-icon="$vuetify"

        icon-color
      />
      <v-file-input
        :glow="glow"
        :icon-color="iconColor"
        append-icon="$vuetify"
        append-inner-icon="$vuetify"
        color="primary"
        label="File"
        prepend-icon="$vuetify"
        prepend-inner-icon="$vuetify"
      />
      <v-date-input
        :glow="glow"
        :icon-color="iconColor"
        append-icon="$vuetify"
        append-inner-icon="$vuetify"
        color="primary"
        label="Date"
        prepend-icon="$vuetify"
        prepend-inner-icon="$vuetify"
      />
      <v-textarea
        :glow="glow"
        :icon-color="iconColor"
        append-icon="$vuetify"
        append-inner-icon="$vuetify"
        color="primary"
        label="Textarea"
        prepend-icon="$vuetify"
        prepend-inner-icon="$vuetify"
      />

      <v-btn :text="`iconColor ${iconColor}`" @click="iconColor = !iconColor" />
      <v-btn :text="`glow ${glow}`" @click="glow = !glow" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const glow = ref(false)
  const iconColor = ref(true)
  const isFocused = ref(false)
</script>
```


MatthewAry:
When I think of the prop 'bloom' I think of glow. Why is this getting called bloom?

johnleider:
> When I think of the prop 'bloom' I think of glow. Why is this getting called bloom?

Changed prop name to **glow**.

packages/vuetify/src/components/VField/VField.tsx
```diff
@@ -359,8 +362,9 @@ export const VField = genericComponent<new <T>(
 
               { props.appendInnerIcon && (
                 <InputIcon
-                  key="append-icon"
+                  key={ `append-icon-${props.glow}` }
```
KaelWD, line 365: Is this necessary?

KaelWD (changes requested):
#17763 wanted outer icons too (VInput not just VField)

MatthewAry:
> When I think of the prop 'bloom' I think of glow. Why is this getting called bloom?

⬇️

> Changed prop name to glow.

Oh man, I didn't mean that the name needed to be changed. They're synonyms. I was just trying to understand the reasoning. The names 'bloom' or 'glow' suggest a visual effect which is not happening here.

This is just a way for a color to be passed down to child VInputIcon components via an optional, boolean prop. I'm assuming that this prop is added because you're trying to avoid breaking it for users who currently expect the current behavior.

I'm sorry for the confusion.

Wouldn't it be better if the name of the prop was descriptive like `color-icon`?

johnleider:
I'm up for a conversation on prop name. @vuetifyjs/core-team

J-Sek:
I agree neither `glow` nor `bloom` really suggest what is going to happen. And I would rather expect `icon-color` to be static and the name could potentially land in other places as `string` prop (e.g. in VAlert, VSwitch or VListItem), where it would not depend on the focus state. How about `icon-color-match`?

johnleider:
The action only occurs on focus as well, not sure if that should play a role in the naming.

J-Sek:
BTW. After seeing it in action I wonder if it should aslo drop the opacity. Weird to see that washed-out blue.

packages/vuetify/src/components/VField/VField.tsx
johnleider: I tried a few different approaches to get reactivity to work when props.glow is toggled:

- Passing a reactive color prop to useInputIcon
- Wrapping <InputIcon> in <VDefaultsProvider>
- Using a dynamic key on <InputIcon />

I felt the key was the cleanest. Open to suggestions.

johnleider:
> #17763 wanted outer icons too (VInput not just VField)

Are we going to need a field-props property so that they can be configured separately?

J-Sek:
> Are we going to need a field-props property so that they can be configured separately?

We [did not in v2](https://codepen.io/J-Sek-the-selector/pen/ZYEyRbK?editors=101). Issue author seems to want to the previous behavior back.

johnleider:
Pushed changes and updated playground.

PR #21095, opened 2025-03-13
feat: move to ESM-first and fix package exports

userquin:
## Description

This PR moves the repository to ESM-first (all packages with `"type": "module"`), we need to:
- check `scripts` folder: I cannot test all the script on my local (using `import` instead `require`), some can be tested in the CI here
- `eslint` modules renamed to `.cjs`: we should update `eslint` and fix/review some plugins; for example, `vuetify eslint`  is  `CJS` only
- check if we can rename `.pnpm.cjs` at root  to `.pnpm.js`: looks like we cannot https://pnpm.io/es/pnpmfile (uhmm :thinking: , shouldn't be `.pnpmfile.cjs`?)
- rename all `.mjs` files in the `vuetify/build` folder (there is no need, it will require to review some package.json script)

I use `createRequire` to load `vuetify/dist/vuetify-labs.js` in the `api-extractor` package since it is `CJS` and so it is failing when importing the module: tried to import `vuetify/dist/vuetify-labs.esm.js` but TS complaining since TS unable to resolve the types (missing dts file in the dist folder).

This PR also fixes the types for the types at `vuetify/framework` subpackage export, it is broken when using `node16/bundler` module resolutions, check https://arethetypeswrong.github.io/?p=vuetify%403.7.16. With this PR the subpackage export is fine:

image

This PR also includes some `jsdocs` to some `js/mjs/cjs` scripts to have some autocompletion, I will check if we can add also some `jsdocs` to the `eslint/babel` stuff: ~~I suggest you to move to latest `eslint` (I use  `@antfu/eslint-config` in all my repositories, you should check it), there are types and dual ESM/CJS exports (or at least ESM, not sure if it is ESM-only).~~ (latest version also CJS only, `@antfu/eslint-config` will load ESM config file for us, no idea yet how to solve this here)

## Markup:

NA


packages/vuetify/build/run-tests.js
```diff
@@ -1,5 +1,4 @@
-'use strict'
-const spawn = require('child_process').spawn
+import { spawn } from 'child_process'
```
userquin, line 1: use default export here

scripts/confirm-npm-tag.js
```diff
@@ -1,10 +1,18 @@
...
+import lerna from  '../lerna.json' with { type: 'json' }
+// const semver = require('semver')
+// const shell = require('shelljs')
+// const inquirer = require('inquirer')
+// const version = require('../lerna.json').version
 
 if (process.env.CI) process.exit(0)
 
+console.log(lerna.version)
+process.exit(0)
```
userquin, line 13: cleanup this

packages/vuetify/build/rollup.config.mjs
```diff
@@ -2,33 +2,27 @@ import path from 'upath'
 import { mkdirp } from 'mkdirp'
 import { writeFile } from 'fs/promises'
 import { fileURLToPath } from 'url'
```
userquin, line 4: use `node:`  prefix

userquin:
Wrong branch,  should be dev ibstead main

userquin:
superseded by https://github.com/vuetifyjs/vuetify/pull/21105

PR #21101, opened 2025-03-14
docs(team.json): update core team

johnleider:
Welcome to the Team!

### DIFF
--- a/packages/docs/src/data/team.json
+++ b/packages/docs/src/data/team.json
@@ -208,10 +208,17 @@
   "ikushum": {
     "name": "Ishan Subedi",
     "team": "core",
-    "discord": "",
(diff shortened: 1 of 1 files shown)


Haviles04:
Do we need to add an avatar link? I see most of the existing members don't have an avatar property

johnleider:
> Do we need to add an avatar link? I see most of the existing members don't have an avatar property

It's based off of your Github profile. It won't show if your account is set to private. If that's the case for you, I can manually upload an avatar for you.

johnleider:
The key of your entry must correlate to your github profile, and your profile must be public. If someone wants to keep their profile private and still display an avatar, we can just upload and manually specify it for them.

PR #21114, opened 2025-03-17
feat(VIconBtn): create new component

johnleider:
## Motivation and Context
- A new lightweight component (~~50%~~ 30% faster than VBtn) that will be used to eventually replace VBtn usage throughout
- Currently VBtn is used in a variety of components such as VTreeview, VDataTable, VDatePicker, VCarousel where it contains infinitely more functionality than what will ever be needed
- Wanted to have a way to start bringing back toggle icons rotating for expansion components like VList, VTreeview, VExpansionPanels
- ~~Will also eventually replace closable buttons~~

## Markup:
<details>

```vue
<template>
  <v-container class="pa-md-12 text-center">
    <v-sheet
      class="d-inline-flex ga-1 pa-4 align-center justify-center mx-auto"
      color="surface-light"
      rounded="pill"
    >
      <v-sheet
        :rounded="!mic ? 'lg' : 'xl'"
        class="overflow-visible d-flex align-center"
        color="surface"
        height="48"
        width="88"
        flat
      >
        <v-icon-btn
          v-model:active="micOptions"
          :active-variant="!mic ? 'flat' : 'text'"
          :base-variant="!mic ? 'flat' : 'text'"
          :color="!mic ? 'error' : ''"
          :rounded="!mic ? 'lg' : 'circle'"
          active-icon="mdi-chevron-down"
          icon="mdi-chevron-up"
          size="48"
          hide-overlay
        />

        <v-icon-btn
          v-model:active="mic"
          :rounded="!mic ? 'lg' : 'circle'"
          active-color=""
          active-icon="mdi-microphone"
          active-variant="tonal"
          base-variant="flat"
          class="ms-n2"
          color="#f9dedc"
          icon="mdi-microphone-off"
          size="48"
          v-ripple
        />
      </v-sheet>

      <v-sheet
        :rounded="!video ? 'lg' : 'xl'"
        class="overflow-visible d-flex align-center"
        color="surface"
        height="48"
        width="88"
        flat
      >
        <v-icon-btn
          v-model:active="videoOptions"
          :active-variant="!video ? 'flat' : 'text'"
          :base-variant="!video ? 'flat' : 'text'"
          :color="!video ? 'error' : ''"
          :rounded="!video ? 'lg' : 'circle'"
          active-icon="mdi-chevron-down"
          icon="mdi-chevron-up"
          size="48"
          hide-overlay
        />

        <v-icon-btn
          v-model:active="video"
          :rounded="!video ? 'lg' : 'circle'"
          active-color=""
          active-icon="mdi-video-outline"
          active-variant="tonal"
          base-variant="flat"
          class="ms-n2"
          color="#f9dedc"
          icon="mdi-video-off"
          size="48"
        />
      </v-sheet>

      <v-icon-btn
        v-model:active="caption"
        :class="!caption ? 'mx-1' : undefined"
        :rounded="!caption ? 'circle' : 'lg'"
        :width="caption ? 56 : 48"
        active-color="#9bbbef"
        active-icon="mdi-closed-caption"
        active-variant="flat"
        height="48"
        icon="mdi-closed-caption-outline"
      />

      <v-icon-btn
        v-model:active="emoji"
        :class="!emoji ? 'mx-1' : undefined"
        :rounded="!emoji ? 'circle' : 'lg'"
        :width="emoji ? 56 : 48"
        active-color="#9bbbef"
        active-icon="mdi-emoticon"
        height="48"
        icon="mdi-emoticon-outline"
      />

      <v-icon-btn
        v-model:active="share"
        :class="!share ? 'mx-1' : undefined"
        :rounded="!share ? 'circle' : 'lg'"
        :width="share ? 56 : 48"
        active-color="#9bbbef"
        active-icon="mdi-arrow-up-bold-box"
        height="48"
        icon="mdi-arrow-up-bold-box-outline"
        @click="onClick"
      />

      <v-icon-btn
        v-model:active="raised"
        :class="!raised ? 'mx-1' : undefined"
        :rounded="!raised ? 'circle' : 'lg'"
        :width="raised ? 56 : 48"
        active-color="#9bbbef"
        active-icon="mdi-hand-back-right"
        height="48"
        icon="mdi-hand-back-right-outline"
      />

      <v-icon-btn
        height="48"
        icon="mdi-dots-vertical"
        rounded="xl"
        variant="tonal"
      >
        <v-icon />

        <v-menu activator="parent" offset="4" origin="bottom" position="top">
          <v-list rounded="lg" slim>
            <v-list-item
              prepend-icon="mdi-radiobox-marked"
              title="Manage recording"
              link
            />

            <v-divider class="my-2" />

            <v-list-item
              prepend-icon="mdi-view-grid-outline"
              title="Change layout"
              link
            />

            <v-list-item
              prepend-icon="mdi-fullscreen"
              title="Full screen"
              link
            />

            <v-list-item
              prepend-icon="mdi-share-variant-outline"
              title="Share screen"
              link
            />

            <v-divider class="my-2" />

            <v-list-item
              prepend-icon="mdi-cog-outline"
              title="Settings"
              link
            />
          </v-list>
        </v-menu>
      </v-icon-btn>

      <v-icon-btn
        v-model:active="hangup"
        active-variant="outlined"
        base-variant="flat"
        color="error"
        height="48"
        icon="mdi-phone-hangup-outline"
        rounded="xl"
        width="72"
      />
    </v-sheet>
  </v-container>
</template>
<script setup>

  const caption = shallowRef(false)
  const emoji = shallowRef(false)
  const raised = shallowRef(false)
  const hangup = shallowRef(false)
  const mic = shallowRef(true)
  const micOptions = shallowRef(false)
  const share = shallowRef(false)
  const video = shallowRef(true)
  const videoOptions = shallowRef(false)

  function onClick () {
    console.log('clicked', share.value)
  }
</script>
```
</details>


packages/vuetify/src/components/VIconBtn/VIconBtn.tsx
```diff
@@ -0,0 +1,148 @@
...
+  text: {
+    type: [String, Number, Boolean],
+    default: undefined,
+  },
+
+  ...makeBorderProps(),
+  ...makeComponentProps(),
+  ...makeElevationProps(),
+  ...makeRoundedProps(),
+  ...makeTagProps(),
```
AndreyYolkin, line 52: makeTagProps() creates tag prop with `div`. Might be better set it as button?

packages/docs/src/examples/v-btn-icon/prop-active.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <v-toolbar class="px-3" title="Toolbar">
+    <template #append>
+      <div class="d-flex ga-2">
+        <v-icon-btn icon="mdi-magnify">
+        <v-icon />
```
eesayas, line 6: should `<v-icon/>` be baked into `<v-icon-btn />` especially if `icon` prop is defined.

johnleider: It is but we typically use this format when there's also something else in the slot. If default slot is rendered we don't render the icon.

packages/docs/src/examples/v-icon-btn/usage.vue
oumoussa98, line 1: Missing variant prop?

packages/vuetify/src/components/VIconBtn/VIconBtn.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ...makeComponentProps(),
+  ...makeElevationProps(),
+  ...makeRoundedProps(),
+  ...makeTagProps({ tag: 'button' }),
+  ...makeThemeProps(),
+  ...makeVariantProps({ variant: 'text' } as const),
+}, 'VIconBtn')
+
+export const VIconBtn = genericComponent<VIconBtnSlots>()({
+  name: 'VIconBtn',
```
ikushum, line 59: Since our button-related components have names such as VBtnToggle, VBtnGroup, it would make more sense to name this component `VBtnIcon` rather than `VIconBtn`.

johnleider: What does everyone think? I chose the name for 2 reasons:

- VCheckboxBtn
- That's what Google calls it

packages/docs/src/pages/en/components/icon-buttons.md
```diff
@@ -0,0 +1,101 @@
...
+The `v-icon-btn` component is a lightweight reusable button that displays icons and text in various states.
+
+### Using with vue-router
+
+The `v-icon-btn` component does not have built in support for vue-router. The following example demonstrates how to use the `v-icon-btn` component with the `v-router-link` component:
+
+```html
+<template>
+  <RouterLink to="/path" custom v-slot="{ navigate }">
+    <v-icon-btn icon="$vuetify" @click="navigate" />
```
KaelWD, line 60: Also needs `tag="a" :href="href"`

packages/docs/src/pages/en/components/icon-buttons.md
```diff
@@ -0,0 +1,101 @@
...
+| - | - |
+| 1. Container | The container element that holds the icon and text |
+
+## Guide
+
+The `v-icon-btn` component is a lightweight reusable button that displays icons and text in various states.
+
+### Using with vue-router
+
+The `v-icon-btn` component does not have built in support for vue-router. The following example demonstrates how to use the `v-icon-btn` component with the `v-router-link` component:
```
KaelWD, line 55: What is v-router-link?

KaelWD:
Sample size of 1 but looks like 1000x v-btn = 380ms and 1000x v-icon-btn = 270ms (29% faster)

packages/vuetify/src/components/VIconBtn/VIconBtn.tsx
eesayas: I'm okay with `VIconBtn`. I think the last word should be what it is, a btn.

eesayas (dismissed):
I think it looks good! I don't think a density prop is needed since icon will remain the same size and btn size can be adjusted via `size` prop.

packages/vuetify/src/components/VIconBtn/VIconBtn.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const { roundedClasses } = useRounded(props)
+
+    const { colorClasses, colorStyles, variantClasses } = useVariantFast(() => ({
+      color: props.disabled ? undefined
+      // Use an inline fallback as opposed to setting a default color
+      // because non-toggle buttons are default flat whereas toggle
+      // buttons are default tonal and active flat. The exact use
+      // case for this is a toggle button with no active color.
+      : isActive.value !== undefined && isActive.value ? props.activeColor ?? props.color ?? 'surface-variant'
+      : props.color,
```
KaelWD, line 94: Was this supposed to be

```suggestion
      color: props.disabled ? undefined
      : isActive.value === undefined ? props.color
      // Use an inline fallback as opposed to setting a default color
      // because non-toggle buttons are default flat whereas toggle
      // buttons are default tonal and active flat. The exact use
      // case for this is a toggle button with no active color.
      : isActive.value ? props.activeColor ?? props.color ?? 'surface-variant'
      : props.color,
```

because they're both redundant, if `isActive.value === undefined` then the second case would be false and it would end up as `props.color` anyway.

packages/vuetify/src/components/VIconBtn/VIconBtn.tsx
```diff
@@ -0,0 +1,192 @@
...
+      variant: isActive.value === undefined ? props.variant
+      : isActive.value ? props.activeVariant ?? props.variant
+      : props.baseVariant ?? props.variant,
+    }))
```
KaelWD, line 98: Can make it IIFE if that helps

```suggestion
    const { colorClasses, colorStyles, variantClasses } = useVariantFast(() => {
      return {
        color: (() => {
          if (props.disabled) return undefined
          if (isActive.value === undefined) return props.color

          // Use an inline fallback as opposed to setting a default color
          // because non-toggle buttons are default flat whereas toggle
          // buttons are default tonal and active flat. The exact use
          // case for this is a toggle button with no active color.
          if (isActive.value) return props.activeColor ?? props.color ?? 'surface-variant'
          return props.color
        })(),
        variant: (() => {
          if (isActive.value === undefined) return props.variant
          if (isActive.value) return props.activeVariant ?? props.variant
          return props.baseVariant ?? props.variant
        })(),
      }
    })
```

MatthewAry:
Might be out of scope but would love to see a loading state. 🥹

KaelWD:
Like this one? https://github.com/vuetifyjs/vuetify/blob/d2be0d256febdad3b7898c9bbc4429e73acd944b/packages/vuetify/src/components/VIconBtn/VIconBtn.tsx#L219-L230

MatthewAry:
I looked at the prop list too!

J-Sek (approved):

emelendez:
I miss what exactly this VIconBtn "lite" is different from  VBtn + icon="". Simpler DOM? no RouterLink wrapper?

why don't create a VBtnLite with no RouterLink wrapper either? (most buttons are for internal actions...) so to show a text/label in the button. 

VIconBtn can show "text" instead of "icon" ???

thx in advanced!!!

emelendez:
ok, I see playing with "text" and "rounded" you can make a lite Btn.

I would rather call this component "VBtnLite" because it can be display as normal btn and rounded icon btn....

thnks!!

PR #20536, opened 2024-09-29
feat: add `unplugin-vue-components` and `unimport` integration

userquin:
## Description

This PR includes:
- `unimport` presets for Vuetify composables and directives ~~and Vuetify One composables~~
- `unplugin-vue-components` resolvers for Vuetify components and directives
- update `unimport` and `unplugin-auto-import` to latest versions (3.13.1 and 0.18.3 respectively)
- docs: replace inlined auto import presets for Vuetify ~~and Vuetify One composables~~
- docs: add Vuetify Directives preset to allow Volar suggest them: right now we need https://github.com/unplugin/unplugin-auto-import/pull/534 merged and released
- ~~add `unbuild`  to build the new new subpackages exports (node stuff): I have no idea how to configure/call `unbuild` with custom `build.config.mts` configuration file, right now using `build.config.mts` in vuetify package root, should be moved to `build` folder: should be fixed with this PR https://github.com/unjs/unbuild/pull/440 (it should be ported to v2 or update here to use v3-rc version)~~
- add some jsdocs to Rollup configuration files and extract some constants to its own module: now we have types inside the mjs modules.
- add `vuetify/unimport` and `vuetify/unplugin-vue-components` to package exports and `typesVersions`
- remove auto import logic in vuetify playground: use `unplugin-vue-components` to auto import components, lab components and directives: **reverted since the CI test process hangs (I need to review Vite playground configuration)**

We need to add these 2 new integrations somewhere in the docs.

>[!NOTE]
> We need to review the lint script, right now I excluded `packages/vuetify/src/unplugin` from the `tsconfig.dist.json`  file to allow build vuetify.

>[!NOTE]
> Maybe we should add a new package or move this PR to another repository.

>[!TIP]
> - `vuetify/unplugin-vue-components`: will resolve components and directives: the directives will be missing from the Vue module augmentation and so Volar will not suggest Vuetify directives in the Vue templates. `vuetify/unplugin-vue-components` won't be able to auto import Vuetify composables.
> - `vuetify/unimport` via `unplugin-auto-import`: will auto import Vuetify composables and directives and won't be able to auto import Vuetify components. `vuetify/unimport` will add Vuetify directives in the Vue module augmentation to allow Volar suggest them in the Vue templates (when https://github.com/unplugin/unplugin-auto-import/pull/534 merged and released).
> - `vite-plugin-vuetify`: same behavior `vuetify/unplugin-vue-components`. `vite-plugin-vuetify` adds Vuetify SASS variables support.
[...]

PR #21129, opened 2025-03-19
feat(package): add `unimport` presets and `unplugin-vue-components` resolvers

userquin:
## Description

Last PR to include `node` stuff: this PR can be closed once merged https://github.com/vuetifyjs/vuetify/pull/20536

Both subpackage exports ESM-only, if you want to use dual ESM/CJS ping me...

I need to do some tests... 
- ✔️ Nuxt working, check https://github.com/userquin/vuetify-nuxt-3.16-lazy-hydration/pull/3 => `dev-pr-21129` branch
- ✔️ Vite + `unplugin-vue-components` working: https://github.com/userquin/check-vuetify-unplugin-vue-components-2 (the directives works only via unimport preset)

## Markup:

NA


packages/vuetify/src/node/unimport-presets.ts
```diff
@@ -0,0 +1,147 @@
...
+    map.set(component, {
+      from: `vuetify/${resolveVuetifyComponentFrom(entry)}`,
+    })
+  })
+
+  for (const [component, entry] of map.entries()) {
+    info.push({
+      pascalName: component,
+      kebabName: toKebabCase(component),
+      export: component,
```
userquin, line 134: should we add component prefix support here?

userquin:
At this point I want to create also custom vuetify styles plugin using unplugin, this way we can also support other bundlers "for free" (🤞 ). I guess at this point this PR should be moved to loader repo as a new package with 3/4 sub-package exports, unimport preset, unplugin-vue-components resolvers, unplugin styles plugin (default) and maybe utils with some exports from the utils.ts module here.

/cc @KaelWD @AndreyYolkin

userquin:
New [@unvuetify/monorepo](https://github.com/userquin/unvuetify-monorepo) repository created and initial packages versions published, we'll have some more Vite/Nuxt utilities in this new repository

Issue #19446, opened 2024-03-21
[Bug Report][3.5.8] Text input hint overlaps with the underline


daenamkim:
### Environment
**Vuetify Version:** 3.5.8
**Last working version:** 3.5.7
**Vue Version:** 3.4.21
**Browsers:** Chrome 122.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
- Click a text input
- Click outside of the text input
- Repeat 2 steps above

### Expected Behavior
The hint fades away as quickly as possible

### Actual Behavior
It remains until it overlaps with the underline and disappears (from 3.5.8 ~ 3.5.11)

This caused e2e tests of our product to fail since both hints displayed during validation 😅 

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpNkE1PwzAMhv+K5TNtVVYuqEziwJ074ZA1Ro2UJlHstJ0Q/52s29TdrOf1I398/SKnoXmPsZ4z4Sv2QlN0WuioPEA/V4k4Bs92JhicZn5TOK2VzhIUwqTXarFGxkIPXadwszZPaJXqx5IzVwQw62S1l9KavaHkrCej8J6Odos+vFCCc8gJYpm2hGRAAuhhIGaQ0TIsdGIrtKtOn8gV9/Mm7ImcI5XA+pjlTo9987jd9c6C9kML6puHP+DfEx7qrn5u8VK81G2L3/9sLGYK)

PR #21138, opened 2025-03-22
fix(VWindow, VTabs): don't override cursor in VWindowItem

### LINKED ISSUE #20090: [Bug Report][3.6.10] VWindowItem always sets cursor to default
### Environment
**Vuetify Version:** 3.6.10
**Vue Version:** 3.4.31
**Browsers:** Edge 126.0.0.0
**OS:** Windows

### Steps to reproduce
* Create a VWindow containing a VWindowItem
* Add any content to the VWindowItem

### Expected Behavior
The cursor displayed should depend on the element that is hovered over.

### Actual Behavior
The cursor stays as the default cursor unless it hovers over an element that explicitly set a different cursor.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptkEFLxDAQhf/Ks5cq2O1d6oI3f4DgQT2EdNYGkkxIJl1h2f9umrpaYSGEzDfDe2/ydmpS1P1TCLs5U/PQDEIuWCW0f/fAMHcqhPqshWYvyniKP6jCo/EjH3/JhnWmqG0apRX2L5NJKCexIwh9CWRSgjRxtiMmNRNU5V0iS1oMe+gcE8fd0F+yVK3+us8fvwQv4H/yhayLDf1m4VImHU0oaUjy6mVc4Cg4IdIBZxwiO7Tlr9plHCi6SeDSJx6Xidv2maxlvHK04017txiskkWsOd/7bG29Pr4BdqV7fQ==)

### Other comments
The `MaybeTransition` element inside of VWindowItem seems to incorrectly output a `disabled` attribute into the rendered markup (`disabled="false"`). Since this is a boolean attribute, it's presence indicates `true` and it should not be rendered at all in the `false` case.

The presence of the attribute makes the CSS rule at https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/styles/generic/_reset.scss#L127 apply that sets the cursor for any children to the default one (unless a child explicitly overrode its cursor).

J-Sek:
## Description

`disabled` attribute is handled by Vue differently when used on actual `<input>` or `<button>`. It would be great to make it consistent (i.e. make it upstream Vue.js concern), but for now a simple enhancement extracted into reusable helper function should suffice.

CSS rule `[disabled] { cursor: default; }` is part of the `_reset.scss` and I'd rather keep it.

fixes #20090

- [x] verified against regression (#19446)
- includes `checked` to make helper function future-proof (see [HTML spec](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes))

## Markup:

```vue
<template>
  <v-app>
    <v-container class="d-flex flex-column ga-5 py-16">
      <p>
        <b>Outside <v-code>&lt;v-window-item&gt;</v-code></b>
        : This is some text that has a text-selection cursor.
      </p>
      <v-window>
        <v-window-item>
          <p>
            <b>Inside <v-code>&lt;v-window-item&gt;</v-code></b>
            : This text should have a text-selection cursor too.
          </p>
        </v-window-item>
      </v-window>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/composables/transition.ts
+++ b/packages/vuetify/src/composables/transition.ts
@@ -1,6 +1,6 @@
 // Utilities
 import { h, mergeProps, Transition, TransitionGroup } from 'vue'
-import { propsFactory } from '@/util'
+import { onlyDefinedProps, propsFactory } from '@/util'
 
 // Types
 import type { Component, FunctionalComponent, PropType, TransitionProps } from 'vue'
@@ -35,7 +35,7 @@ export const MaybeTransition: FunctionalComponent<MaybeTransitionProps> = (props
         : customProps as any,
       typeof transition === 'string'
         ? {}
-        : Object.fromEntries(Object.entries({ disabled, group }).filter(([_, v]) => v !== undefined)),
+        : onlyDefinedProps({ disabled, group }),
       rest as any,
     ),
     slots
--- a/packages/vuetify/src/util/helpers.ts
+++ b/packages/vuetify/src/util/helpers.ts
@@ -767,3 +767,9 @@ export type Primitive = string | number | boolean | symbol | bigint
 export function isPrimitive (value: unknown): value is Primitive {
   return typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean' || typeof value === 'bigint'
 }
+
+export function onlyDefinedProps (props: Record<string, any>) {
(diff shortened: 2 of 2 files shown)


J-Sek:
- rebased
- updated playground

PR #21141, opened 2025-03-22
fix(VDateInput): prevent mobile keyboard showing

### LINKED ISSUE #20580: [Bug Report][3.7.2] VDateInput Opens Mobile Keyboard
### Environment
**Vuetify Version:** 3.7.2
**Vue Version:** 3.5.12
**Browsers:** Google Chrome, Safari
**OS:** Android, iOS

### Steps to reproduce
1. Open the DateInput page from Vuetify on a mobile device
2.  Open a date input and it will cause the keyboard to appear
3. I have to close the keyboard to see the calendar input

### Expected Behavior
Should not open the keyboard

### Actual Behavior
Opens the keyboard

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdjssKwjAQRX9lzCYK1uylCu78AxfqorSpBPIYJpNuSv/dJFUQN8PM5XDu3GcRqVcXxMOUtDiKlrVD27E+PzxAOzUdYl3r0QfPnfGaPlENh0w3xmNiUF9U/bMlWVWt+qnIZ+zJIEPUnNYm4zAQwwykR1hgpOBA5u9kwQGyNzK4+IJTIbbyqq0NcAtkh43clYJVmWVi2ftkbR3PN+XyTvk=)

ikushum:
## Description
In mobile, when the VDateInput is clicked, the date picker and mobile keyboard both show up at the same time and as a result, the date picker is hidden under the keyboard. To fix this, I've made the VDateInput `readonly` initially, and only on the first click, it will be editable. This will ensure the keyboard appears only on second click. This behaviour exists only on mobile.

fixes #20580

### DIFF
--- a/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
+++ b/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
@@ -6,6 +6,7 @@ import { makeVTextFieldProps, VTextField } from '@/components/VTextField/VTextFi
 
 // Composables
 import { useDate } from '@/composables/date'
+import { useDisplay } from '@/composables/display'
 import { makeFocusProps, useFocus } from '@/composables/focus'
 import { forwardRefs } from '@/composables/forwardRefs'
 import { useLocale } from '@/composables/locale'
@@ -62,6 +63,7 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
   setup (props, { slots }) {
     const { t } = useLocale()
     const adapter = useDate()
+    const { mobile } = useDisplay()
     const { isFocused, focus, blur } = useFocus(props)
     const model = useProxiedModel(
       props,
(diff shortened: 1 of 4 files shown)


johnleider (dismissed):

packages/vuetify/src/labs/VDateInput/VDateInput.tsx
```diff
@@ -95,7 +100,21 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
...
+    })
+
     const isInteractive = computed(() => !props.disabled && !props.readonly)
+    const isReadonly = computed(() => !(mobile.value && isEditingInput.value) && props.readonly)
+
+    watch(menu, val => {
+      if (val) return
+
+      isEditingInput.value = false
+    })
```
ikushum, line 117: This is redundant because we already have `onUpdateMenuModel` which does the same thing

Issue #21098, opened 2025-03-14
[Bug Report][3.7.16] v-select menu flickering on chrome (in a very specific condition)


AdrienMonpe:
### Environment
**Vuetify Version:** 3.7.16
**Vue Version:** 3.5.13
**Browsers:** Google Chrome (134.0.6998.89)
**OS:** Windows 10

### Steps to reproduce
This bug occurs only in a really specific way and it's only visible on chrome (works fine on firefox at least)

- v-select must be attached to an element with a relative position
- v-select must be at the bottom of a scrollable container with no scrollbar yet (overflow-y: auto)

### Expected Behavior
The menu should open above the select, like it does on firefox

### Actual Behavior
The menu flickers above and under the select

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqdUctugzAQ/JUVPXApIa1UqaI0UpVPaG9JDgY2warxWmsDiaL8ex0MeR17QfYMs57ZWR0jy2X6ZcysazHKotxhY5RwuFhrgLxLhDHD0V8q2UGphLWf66gk7YTUyOtopAEUMTZ5wZBeIGls+wD9+6+fWlqwqLB00KBuwdbUqgq2Spa/yEAaypqpQSiYeoucTfppwpBAVt5+GJP07OPdZRgyB/YKeTuiQOV134GJbrlM+p2dd7KKl0LJLbGWIn6GeEk+hago3jwIzu4Tw2TOqiMI50RZZxA/3duK4XQrvG4iT32QqZXLOU+ntvL0pkV/tSVL4/zuXGsWeRquI+UOKnQ9u1QKxzC6l5Xztt7mc7P/CFCNcle7DF7fr1hBXPllw4vZgyUlK9gxHkaSOuStoj45ZCBaRwN8CpU8xJ2eNWSlk6QzYPQZZIeTyFsf7UanZ90qNXw2fwru30Y=)

### Other comments
I resolved this problem by setting overflow-y: scroll to the container for the moment.

TedVu:
Look like this specific condition causes https://github.com/vuetifyjs/vuetify/blob/7c339338ac174aed2e6f22286d06308990f2bf19/packages/vuetify/src/components/VOverlay/locationStrategies.ts#L202 to be constantly called.

jcjp:
[...]

PR #21143, opened 2025-03-23
fix(location-strategies): infinite loop on update location

jcjp:
## Description

On Google Chrome browser the repositioning of the `VMenu` on a `VSelect` flips infinitely. This changed will ensure that we only do a maximum of 2x to reposition. The first reposition will position the menu correctly on the second reposition it will resize the menu correctly. Caveat this somehow changes the animation and now it's a little bit awkward, if you guys have an idea how to fix the animation feel free to update or message me about it thanks!

- Fix: #21098

### DIFF
--- a/packages/vuetify/src/components/VOverlay/locationStrategies.ts
+++ b/packages/vuetify/src/components/VOverlay/locationStrategies.ts
@@ -309,7 +309,10 @@ function connectedLocationStrategy (data: LocationStrategyData, props: StrategyP
     let x = 0; let y = 0
     const flipped = { x: false, y: false }
     const available = { x: 0, y: 0 }
+
     // Make the retry max x2 to get the correct position
+    // Addition happens on resizeObserver see line 207
+    // eslint-disable-next-line no-unmodified-loop-condition
     while (retries < 2) {
       const { x: _x, y: _y, overflows } = checkOverflow(placement)
 
@@ -402,14 +405,14 @@ function connectedLocationStrategy (data: LocationStrategyData, props: StrategyP
       maxWidth: available.x > 0 &&
         convertToUnit(pixelCeil(clamp(
           available.x, minWidth.value === Infinity
-          ? 0
-          : minWidth.value, maxWidth.value))
+            ? 0
+            : minWidth.value, maxWidth.value))
         ),
       maxHeight: available.y > 0 &&
         convertToUnit(pixelCeil(clamp(
           available.y, minHeight.value === Infinity
-          ? 0
-          : minHeight.value, maxHeight.value))
+            ? 0
+            : minHeight.value, maxHeight.value))
         ),
     })
 
--- a/packages/vuetify/src/components/VOverlay/locationStrategies.ts
+++ b/packages/vuetify/src/components/VOverlay/locationStrategies.ts
@@ -6,7 +6,6 @@ import { computed, nextTick, onScopeDispose, ref, watch } from 'vue'
 import { anchorToPoint, getOffset } from './util/point'
 import {
   clamp,
-  consoleError,
   convertToUnit,
   destructComputed,
   flipAlign,
@@ -146,7 +145,11 @@ function getIntrinsicSize (el: HTMLElement, isRtl: boolean) {
   return contentBox
 }
 
+// Prevents infinite loop
 let retries = 0
+// Track the last position
+let previousX = 0
+let previousY = 0
 
 function connectedLocationStrategy(data: LocationStrategyData, props: StrategyProps, contentStyles: Ref<Record<string, string>>) {
   const activatorFixed = Array.isArray(data.target.value) || isFixedPosition(data.target.value)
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VOverlay/locationStrategies.ts
```diff
@@ -298,15 +307,13 @@ function connectedLocationStrategy (data: LocationStrategyData, props: StrategyP
...
-      if (resets++ > 10) {
-        consoleError('Infinite loop detected in connectedLocationStrategy')
-        break
-      }
+    const available = { x: 0, y: 0 }
 
+    // Make the retry max x2 to get the correct position
+    // Addition happens on resizeObserver see line 207
+    // eslint-disable-next-line no-unmodified-loop-condition
+    while (retries < 2) {
```
KaelWD, line 316: This loop has nothing to do with the linked issues, it doesn't mutate the dom and should run until the flip block is happy with the chosen placement. Linking the loop counter to something only modified outside it is also fragile and more likely to introduce an infinite loop if anything, what if the flip block never hits `break`?

packages/vuetify/src/components/VOverlay/locationStrategies.ts
```diff
@@ -199,21 +204,25 @@ function connectedLocationStrategy (data: LocationStrategyData, props: StrategyP
...
+    // if (oldContentEl) observer.unobserve(oldContentEl)
+    // if (newContentEl) observer.observe(newContentEl)
   }, {
     immediate: true,
   })
 
   onScopeDispose(() => {
     observer.disconnect()
+    // Reset retry on close component
+    retries = 0
```
KaelWD, line 225: It also needs to reposition on scroll or if the content size changes.

packages/vuetify/src/components/VOverlay/locationStrategies.ts
jcjp: As per observation of Ted Vu on #21098, the bug is experienced since it reposition infinitely and unable to settle to the correct position. There is another fix that I found which is to pass a new direction props to the `VSelect`  and pass it to the `VMenu` under the `VSelect` for example passing `top` direction props to `VMenu` fixes the issue let me know if this is a viable solution to the first issue.

packages/vuetify/src/components/VOverlay/locationStrategies.ts
jcjp: I tried testing it by changing resizing the browser size, it seems the content still is resizing and flips the `VMenu` but the last position is unpredictable.

packages/vuetify/src/components/VOverlay/locationStrategies.ts
KaelWD: The loop in #21098 is through the resize observer:

menu opens  
menu width > container width, scrollbar appears  
target width shrinks due to scrollbar  
updateLocation sets menu width < container width, scrollbar hides  
target width grows due to hidden scrollbar  
updateLocation sets menu width > container width, scrollbar appears  
target width shrinks due to scrollbar  
updateLocation sets menu width < container width, scrollbar hides  
etc...

KaelWD: Disabling the observer for a frame was supposed to prevent this, maybe it just needs to be longer? https://github.com/vuetifyjs/vuetify/blob/3024bb10411f78246d8caeba94b0893a1f148670/packages/vuetify/src/components/VOverlay/locationStrategies.ts#L221-L222

johnleider: Could we approach this in a different way maybe? Such as, we limit the width of the container to the screen width minus the pixel padding on each side? It would probably solve the infinite loop and also resolve issues like this: https://play.vuetifyjs.com/#eNpNjssKwjAQRX/lmk0V1OylCu78AxfWRUmnUpg8SNJuSv/dJFXoZpjD3Me8ZhG8knfnztNI4iLqSNpxG+nWGKCeTq1zZS2g28H8qHAgJhWRDrYjvjZCh08jIP8OubFkWMNquSlJGJQfXESgOK5dg3bWR8zw1GNB761Glf6rshxQ1oSIVIVrVuyrBzFbPK3nblcdcsEamcLEcjQjcxnvL9rXTpY=

johnleider: One final point, this tends to happen on smaller screens. Should we explore a different, maybe opt in behavior for menu that makes it behave similar to a dialog?

jcjp: Okay after checking if we could fix it by changing the width it seems that it still gets dynamically resize which makes it overflows the boundaries.

johnleider:
Using the supplied playground in the issue and latest changes on this P.R., I'm still seeing this;

chrome-issue

jcjp:
> Using the supplied playground in the issue and latest changes on this P.R., I'm still seeing this;
> 
> chrome-issue
> 

Oh I see it still happening, will furher investigate on this issue. Thanks for looking John.

jcjp:
Will close this PR for now needs more investigation and research.

PR #21154, opened 2025-03-25
feat(VConfirmEdit): add okButtonProps and cancelButtonProps

ikushum:
## Description
fixes https://github.com/vuetifyjs/vuetify/issues/20226

Implement customization option for actions of VConfirmEdit so that on VDateInput the cancel button can always be enabled.

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-date-input v-model="date" :hide-actions="false" label="Date" mobile />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const date = ref(null)
</script>


```


johnleider:
I mentioned wanting to take this on a similar path to that of VDatePickerControls:

image

Mainly because if the user wants to customize the buttons for either slot, it's just recommended that they use the slot and add their options there. We're mainly concerned with enabling or disabling the buttons. Also, I have an intent of adding a reset button and it would be nice to be able to easily control it's disabled state as well.

ikushum:
> I mentioned wanting to take this on a similar path to that of VDatePickerControls:

I don't think an approach similar to `VDatePickerControls` would work in case of `VConfirmEdit` unless we introduce breaking changes. 

In `VDatePickerControls` the disabled state of the buttons are controlled 100% from the props so simply passing array of disabled strings works fine. But in case of `VConfirmEdit` the disabled state of the buttons are controlled from inside the component and the problem we are trying to solve is concerned with actually enabling the button.

Please let me if I'm not understanding something

packages/docs/src/pages/en/components/confirm-edit.md
```diff
@@ -45,3 +45,9 @@ The `v-confirm-edit` component is an intuitive way to capture a model's changes
 It's easy to integrate pickers into the `v-confirm-edit` component. This allows you to provide a more user-friendly experience when selecting dates, times, or colors.
 
 <ExamplesExample file="v-confirm-edit/misc-date-picker" />
+
+### Disable actions
+
+You can control the disabled state of action buttons using `disabled` prop by either passing an array to disable targeted actions or a boolean value to disable all actions. If the `disabled` prop is not provided, the component will use internal logic to determine the disabled state.
```
johnleider, line 51:
```suggestion
You can control the disabled state of action buttons using **disabled** prop by either passing an array to disable targeted actions or a boolean value to disable all actions. If the **disabled** prop is not provided, the component will use internal logic to determine the disabled state.
```

packages/docs/src/examples/v-confirm-edit/misc-disable-actions.vue
```diff
@@ -0,0 +1,64 @@
...
+  </div>
+</template>
+
+<script setup>
+  import { ref } from 'vue'
+
+  const disabled = ref([])
+  const value = ref('Egg Plant')
+
+  const onClick = action => {
```
johnleider, line 53:
```suggestion
  function onClick (action) {
```

packages/docs/src/examples/v-confirm-edit/misc-disable-actions.vue
```diff
@@ -0,0 +1,64 @@
...
+            <v-spacer></v-spacer>
+            <component :is="actions"></component>
+          </template>
+        </v-card>
+      </template>
+    </v-confirm-edit>
+  </div>
+</template>
+
+<script setup>
```
johnleider, line 47: Needs a duplicated script in the options API. The user can choose on the docs.

packages/api-generator/src/locale/en/VConfirmEdit.json
```diff
@@ -1,8 +1,11 @@
 {
   "props": {
+    "modelValue": "Represents the commited v-model value",
+    "color": "The color of the component",
```
johnleider, line 4:
```suggestion
```

This already exists in generic

johnleider (changes requested):

johnleider (approved):

PR #21160, opened 2025-03-26
docs: updated missing descriptions on API

eesayas:
## Description
1. Filled out documentation on component json files. Lifting to generic base file when applicable.
2. Omit `HTMLInputElement` from `VTextField` type to fix missing descriptions on `VAutocomplete`, `VCombobox`, `VDateInput`, and `VSelect`. Workaround until doc generation recursion issue is fixed.
3. If accessing `VTextField` ref and requiring `HTMLInputElement` props, use `<VTextField & HTMLInputElement>` as Ref type
4. `VTimePickerControls` correct type for `period` prop.

## Markup:
N/A


packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -224,11 +224,59 @@ export const VAutocomplete = genericComponent<new <
...
+      const next = selectionIndex.value + 1
+
+      if (model.value[next]) {
+        selectionIndex.value = next
+      } else {
+        selectionIndex.value = -1
+        vTextFieldRef.value?.setSelectionRange(0, 0)
+      }
+    }
+
```
KaelWD, line 275: What does this have to do with documentation?

packages/api-generator/src/locale/en/VDataTableVirtual.json
```diff
@@ -25,6 +26,9 @@
     "item.data-table-expand": "Slot to replace the default `v-icon` used when expanding rows.",
     "item.data-table-select": "Slot to replace the default checkbox used when selecting rows.",
     "loading": "Defines content for when `loading` is true and no items are provided.",
+    "tbody": "Slot to replace the default table `<tbody>`.",
+    "thead": "Slot to replace the default table `<thead>`.",
+    "tfoot": "Slot to replace the default table `<tfoot>`.",
```
KaelWD, line 31: These slots have misleading names, tbody and tfoot are *after* the default tbody and thead is between the default thead and tbody.

packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
eesayas: - Automplete was not rendering its exposed prop because `vTextFieldRef` did not have type.
- After type was added. Cognitive complexity was triggered. Hence the need for the refactor.

KaelWD:
Screenshot_20250327_141851

`source: calendar` means it can be set in `calendar.json` and will apply to any other components that are using that composable instead of duplicating the description in each component

packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -292,4 +292,5 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
   },
 })
 
-export type VTextField = InstanceType<typeof VTextField>
+// Omit HTMLInputElement prop to prevent inference when generating docs
+export type VTextField = Omit<InstanceType<typeof VTextField>, keyof HTMLInputElement>
```
KaelWD, line 296: This is not a good solution, those properties are actually accessible on the instance and should be present in the public types. If something is causing incorrect inference it should be fixed in https://github.com/vuetifyjs/vuetify/blob/3fde52a7634f32918049c8f44acd019f0a536a55/packages/api-generator/templates/component.d.ts#L49-L54 https://github.com/vuetifyjs/vuetify/blob/42dba04743421d5f8032f598ccacb613b8f78ba3/packages/vuetify/src/composables/forwardRefs.ts#L29-L41 or https://github.com/vuetifyjs/vuetify/blob/44be6034569b938621c44dd877e76ea6314c2f78/packages/api-generator/src/types.ts#L412-L568

packages/api-generator/src/locale/en/VDataTableVirtual.json
eesayas: on second look, maybe these slots doc are not required for `VDataTableVirtual` API docs. I have to double check why it was shouting missing when I was generating the api.

Anyways, could your comment be applicable to the current doc for `VDataTable`? 
- https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VDataTable/VDataTable.tsx#L258-L271
- https://github.com/vuetifyjs/vuetify/blob/master/packages/api-generator/src/locale/en/VDataTable.json#L67-L69

KaelWD: Yeah that's wrong too, if it replaced the default it would be `slots.tbody?.() ?? (<tbody>`

packages/vuetify/src/components/VTextField/VTextField.tsx
eesayas: yeah I don't like it either. I added `<TextField & HTMLInputElement>` as type on all components that use the `TextField` instance as a workaround. But thanks I'll look into this, I think there's an underlying recursion error in generating the docs that must be its own PR.

packages/api-generator/src/locale/en/VDataTableVirtual.json
eesayas: Thanks, I'll update `VDataTable` current doc too. `Slot after the default...`

packages/vuetify/src/components/VTextField/VTextField.tsx
eesayas: Just saw the last line of your comment 😅 . Yes I was in this file trying to figure out how I can correctly infer `HTMLInputElement`. `VTextField` is inferring it fine anyway 🤷‍♂️ . But when `VTextField` is a ref then there's the issue. I'll defer the work to another PR if it's okay with you. Then merge it back here.

packages/api-generator/src/locale/en/VDataTableVirtual.json
MajesticPotatoe: these slots are meant to used with the `hide-default-(body|footer|header)` props to replace the slot, which should be worth mentioning in the descriptions.

KaelWD (approved):
You can open another PR if you have more changes

PR #21167, opened 2025-03-27
refactor: replace computed() with toRef()

KaelWD:
Computed refs that are infrequently accessed and don't do anything complex can be replaced with getter refs which saves the setup and caching overhead of computed.

time to render 1000x `<v-btn icon="$vuetify" variant="flat" />`:
dev: 182ms (189 180 162 176 202)
pr: 139ms (146 136 133 135 144)

Some of these changes might actually hurt performance but I'm not gonna microbenchmark every single one, overall it's a significant improvement.


ikushum (commented):
Looking at the number of files changed, is it safe to merge this into master directly?

johnleider:
Where is your sense of adventure?

ikushum:
> Where is your sense of adventure?

Just worried about an potential adventurous rollback later 😅

KaelWD:
Should be fine, `toRef()` has the same behaviour as `computed()` just without caching.

Issue #21153, opened 2025-03-25
[Feature Request][VDataTable] Make the last v-table-column fixed


Jayczee:
### Environment
**Vuetify Version:** 3.7.6
**Vue Version:** 3.4.21
**Browsers:** Chrome 134.0.0.0
**OS:** Windows 10

### Steps to reproduce
Just define a table and set the last column fixed

### Expected Behavior
last column fixed

### Actual Behavior
nothing happened

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrtVEtuwjAQvcrIXYRKEBogrWQBUk/RRWHhJkZYzcdyBgpC3L1jO4FEBfWjdtds8vImM29+9vOBVSYZPmodbjeScTZFmetMoJwvCoDpdiC0dtB9JGWBQhXS1JQjU4FigOIlkw0JwNdSpNJUswXzaMFaRkUi1uTebctK7WQ6qD1ObIX7TNLvudgN3lSKaw7juzu9O3ue0qGEmgLgxoYPX+U+bpldxi9YzEWCqiyA4HTomVaMYacLnmoX2nSEyG5LLOM71olBn1VilEaoJG58Q1WuS4NwACNXfUjKXG9QpnCElSlzCGgcgXUEMhUVgusVzOzfvWcvfwAqLuIQ9S0YcRg5MKb2ODDhMHEg5hDDsf/v9Ttey9v2YPy20mSaEfZ6tzCbw8ELGBq4KaAeGQme1wwVZpJDkJRZFNT52IfUiLV5tdmKtsXuHoeVyCp5sjSVXAs9uhi6w14LDZCr4smfuCCyR+7s9Zns+KJsh/2abPwt2clF2Q77F9XGF2U77E9k6xuRA5rNx4kv7etIyzgd+tuF7hV27LNxOAlHEbPgIbxny3c8tqHi)

jcjp:
Is this behavior what you are looking for?

<img src="https://github.com/user-attachments/assets/5fb8ae69-5170-4701-b362-0c249cd6b16e" width="200" />

Jayczee:
> Is this behavior what you are looking for?
> 
[...]

PR #21168, opened 2025-03-27
feat(VDataTable): add new lastFixed property (#21153)

jcjp:
## Description

Utilize props `lastFixed` and make column header fixed on the right.

- Fixes: #21153


eesayas:
Is it possible to not change the interface? Like it's still setting `fixed: true` in the last header not a prop.

johnleider (changes requested):
This needs language in VDataTableHeaders.json, and also newIn.json

jcjp:
> Is it possible to not change the interface? Like it's still setting `fixed: true` in the last header not a new prop.

I played around with it last time Isias but I could not get it work with the `fixed` props.

PR #21172, opened 2025-03-28
fix(VEmptyState): pass href/to props to VBtn

### LINKED ISSUE #21170: [Bug Report][3.7.19] v-empty-state to/href props don't work
### Environment
**Vuetify Version:** 3.7.19
**Vue Version:** 3.5.13
**Browsers:** Chrome 134.0.0.0
**OS:** Windows 10

### Steps to reproduce
Set either the `to` or `href` props with the action button included and the links don't apply

### Expected Behavior
action button inherits the `to`/`href` prop

### Actual Behavior
it doesn't

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqFkcFOxCAQhl9lwnlbPOxpUzcxvoCJJh6sB6TTFqUMgWndxvjuUtyNbWKyFwI//8cw8798iRi0vPO+nEYUB1ExDt4qxmPtAKqpUN7nbT5ocqyMw3CWspgAnovICbqoAD2qxibnbS2eeyIfd7C/2dfiz8GG7XL9oDoERwwtja7ZOPDEyfDUI/jFNNMInxgQLNGHcV0iAjSEMeN4MpHXuBkSlPie2ceDlKlBNu38Hsu3QjeudMiyIR1lNkZpqaMop9K7bv2M0mzIFefP3C8T0AyPo/cUNvX6gO2/5TQNm7bouutYyc1gLxEkdZvBovxGVMlVdOJ750Zr8/L6AwzcnN0=)

SonTT19:
## Description

fixes #21170 

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-empty-state
        action-text="Contact Support"
        headline="Whoops, 404"
        href="https://vuetifyjs.com"
        image="https://vuetifyjs.b-cdn.net/docs/images/logos/v.png"
        text="The page you were looking for does not exist"
        title="Page not found"
      />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VEmptyState/VEmptyState.tsx
+++ b/packages/vuetify/src/components/VEmptyState/VEmptyState.tsx
@@ -185,7 +185,9 @@ export const VEmptyState = genericComponent<VEmptyStateSlots>()({
                   VBtn: {
                     class: 'v-empty-state__action-btn',
                     color: props.color ?? 'surface-variant',
+                    href: props.href,
                     text: props.actionText,
+                    to: props.to,
                   },
                 }}
               >


jcjp:
I tested this and it fixes the bug reported, although there is another issue wherein if we pass both `to` and `href` it doesn't work properly if we pass both external url the scenario is not ideal but should we still fix it?

SonTT19:
> I tested this and it fixes the bug reported, although there is another issue wherein if we pass both `to` and `href` it doesn't work properly if we pass both external url the scenario is not ideal but should we still fix it?

Not sure about it. Isn't `href` meant to use with external url & `to` is for `vue-router`

jcjp:
> > I tested this and it fixes the bug reported, although there is another issue wherein if we pass both `to` and `href` it doesn't work properly if we pass both external url the scenario is not ideal but should we still fix it?
> 
> Not sure about it. Isn't `href` meant to use with external url & `to` is for `vue-router`?

Correct `href` for external `to` for `vue-route`, I just found a scenario if we pass both this what happens:

```ts
<v-empty-state
  headline="Whoops, 404"
  title="Page not found"
  text="The page you were looking for does not exist"
  image="https://vuetifyjs.b-cdn.net/docs/images/logos/v.png"
  action-text="Contact Support"
  to="https://vuetifyjs.com"
  href="https://vuetifyjs.com"
/>
```

The result of the link would be like this (Highly unlikely that a user will use both at the same time):



I realized that we don't need to address this scenario, since we could just point out to the user that they need to use either `href` or `to` but no both at the same time.

johnleider (approved):

Issue #21163, opened 2025-03-26
[Bug Report][3.7.19] change event of VFileInput doesn't respond to drag-and-dropped files


donalmurtagh:
### Environment
**Vuetify Version:** 3.7.19
**Vue Version:** 3.5.13
**Browsers:** Chrome 134.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
 - If a file is drag-dropped onto the `<VFileInput>`, the change handler is not called
 - If a file is selected via the file explorer dialog, the change handler is called

### Expected Behavior
The change handler should be called in both cases.

### Actual Behavior
The change handler is only called when a file is selected via the file explorer dialog

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNUk1rwzAM/Ssi7NBCkzJaGIy2bJdB79su66BuoiYG1za2nG6U/vcpcbYmh229GEtPek9fb6fEu3z6aG1WB0zukwXhwSpBuNpogEWdCmvbb2vkRpOQGl3nYmdQkCvh/XKTHEQ63yQ/EINK9iyA9R4E7KVCkB4KJ8q0cMZaLIB5DVDFgLaBYIfKHCetI6+ELhH4LRQ6TuwTakOQC6Ww6IlOB6p/1uBRYU6sX0vRqrUAflhlHLorCrhCfK3hNSDJ/SfMsrvsdhZpL3wdCbcOO0MVmx79P108V5wnvQ8IR8E/Tc4UIY8sA7k5sBC3JD01ccBaUKPz0uhBJ6tlF/9LM4tpUBejTptJpXFbSvC+eP8vPDZRwBMjfpPAQ5wcA7HXCPOBMFc///u82Dm8r8YTz28x7Z0lmz530hKvj0K8Tk701M20K2MJN1ijJlh2AoLHTaPtzSn6MxKuRMqaQnymUJdUndviR34MoSXBYjtu1KMe8yTniQ5Ktc/7F8k1AV8=)

### Other comments
This bug was introduced in v3.7.14. In v3.7.13 the handler was called in both cases. In all versions later than v3.7.13 the handler is only called when a file is selected via a file explorer

PR #21182, opened 2025-03-29
fix(VFileInput): emit change with event on drop

jcjp:
## Description

Fixes: https://github.com/vuetifyjs/vuetify/issues/21163

Props `@change` not triggered on file drop.

## Markup:

```vue
<template>
  <v-app>
    <v-file-input label="Upload Files" @change="handleUpload" />
  </v-app>
</template>

<script setup>
const handleUpload = $event => {
  if ($event.dataTransfer) {
    alert(`${$event.dataTransfer.files.length} File(s) uploaded`)
  } else {
    alert(`${event.target.files.length} File(s) uploaded`)
  }
}
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VFileInput/VFileInput.tsx
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.tsx
@@ -89,6 +89,7 @@ export const VFileInput = genericComponent<VFileInputSlots>()({
     'mousedown:control': (e: MouseEvent) => true,
     'update:focused': (focused: boolean) => true,
     'update:modelValue': (files: File | File[]) => true,
+    'change': (e: DragEvent) => true,
   },
 
   setup (props, { attrs, emit, slots }) {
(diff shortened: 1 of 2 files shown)


Arslan038 (commented):
Can you please add a spec test for this ? Other then this it is working fine for me as expected.

packages/vuetify/src/components/VFileInput/VFileInput.tsx
```diff
@@ -161,6 +161,11 @@ export const VFileInput = genericComponent<VFileInputSlots>()({
       if (!e.dataTransfer) return
 
       model.value = [...e.dataTransfer.files ?? []]
+
+      // INFO: If we add this to emits property we get an error on the test
```
eesayas, line 165: curious, can you paste the error here for posterity.

eesayas (approved):

johnleider (approved):

PR #21184, opened 2025-03-30
fix(VTooltip, VSnackbar): increasecontrast on dark theme

J-Sek:
I can barely see the text. It is fine with light theme.

Current APCA score is [41.8](https://colorcontrast.app/#fg=%23424242&bg=%23a3a3a3&level=aa&format=rgb&algo=APCA&filter=none) - not enough for any text

Dropped text color to pure black to make minimal adjustment to `surface-variant`. Otherwise it would need to change to `#e0e0e0` and it would be more drastic change for other places (e.g. VSlider, VCarousel)

Black on `#c8c8c8` gets [74.3](https://colorcontrast.app/#fg=%23000000&bg=%23c8c8c8&level=aa&format=rgb&algo=APCA&filter=none) which is still below the threshold, but good enough and I'd rather not reach some arbitrary number and have the text readable than wait for `inverse-surface` to be adopted.

---

Before:

image

image

After:

image

image


J-Sek:
Colors from MD3 spec get [82.4 for dark theme colors](https://colorcontrast.app/#fg=%23322F35&bg=%23E6E0E9&level=aa&format=rgb&algo=APCA&filter=none)

johnleider (approved):

PR #21198, opened 2025-04-02
feat(VTextField): inline prefix/suffix

KaelWD:
Screenshot_20250402_223716

maybe a bit too cursed, idk if there's a way to do this without contenteditable

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-number-input :precision="null" control-variant="stacked" suffix="%" />
      <v-number-input control-variant="stacked" prefix="$" reverse />
    </v-container>
  </v-app>
</template>
```


J-Sek:
I would have some use for it, but also decided long time ago to pull more complex suffixes consistently to the right.

image

PR #21199, opened 2025-04-02
fix(VDatePicker): use start of week to calculate week numbers

### LINKED ISSUE #20490: [Bug Report][3.6.7] Wrong calculation of the calendar week in date composable

### LINKED ISSUE #20724: [Bug Report][3.7.4] Wrong number week with locale-provider "fr"

J-Sek:
## Description

fixes #20724
fixes #20490

- ~~ISO-ish 4-day requirement for the first week~~ defaults to match `en-US`

## Markup:

```vue
<template>
  <v-app style="background-color: rgba(var(--v-theme-on-surface), 0.2)">
    <v-container>
      <div class="d-flex justify-center align-center ga-3">
        <span class="opacity-70">locale:</span>
        <v-chip-group v-model="currentLocale">
          <v-chip filter label text="en" value="en"></v-chip>
          <v-chip filter label text="pl" value="pl"></v-chip>
          <v-chip filter label text="fr" value="fr"></v-chip>
          <v-chip filter label text="pt" value="pt"></v-chip>
        </v-chip-group>
      </div>
      <div class="d-flex justify-center align-center ga-3">
        <span class="opacity-70">startOfWeek:</span>
        <v-chip-group v-model="startOfWeek">
          <v-chip
            v-for="o in startOfWeekOptions"
            v-bind="o"
            :key="o.value"
            filter
            label
          ></v-chip>
        </v-chip-group>
      </div>
      <div class="d-flex flex-wrap justify-center ga-6">
        <v-locale-provider :locale="currentLocale">
          <v-date-picker
            v-for="d in 6"
            :key="d"
            :elevation="5"
            :first-day-of-week="startOfWeek"
            :model-value="new Date(2023 + d, 0, 1)"
            class="mt-3"
            hide-header
            show-adjacent-months
            show-week
          ></v-date-picker>
        </v-locale-provider>
      </div>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const currentLocale = ref('en')
  const startOfWeek = ref(undefined)
  const startOfWeekOptions = 'Sun|Mon|Tue|Wed|Thu|Fri|Sat'
    .split('|')
    .map((x, i) => ({ text: x, value: i }))
</script>
```

### DIFF
--- a/packages/vuetify/src/composables/calendar.ts
+++ b/packages/vuetify/src/composables/calendar.ts
@@ -186,7 +186,7 @@ export function useCalendar (props: CalendarProps) {
 
   const weekNumbers = computed(() => {
     return weeksInMonth.value.map(week => {
-      return week.length ? getWeek(adapter, week[0]) : null
+      return week.length ? getWeek(adapter, week[0], { startOfWeek: Number(props.firstDayOfWeek ?? 0) }) : null
     })
   })
 
--- a/packages/vuetify/src/composables/calendar.ts
+++ b/packages/vuetify/src/composables/calendar.ts
@@ -186,7 +186,7 @@ export function useCalendar (props: CalendarProps) {
 
   const weekNumbers = computed(() => {
     return weeksInMonth.value.map(week => {
-      return week.length ? getWeek(adapter, week[0], { startOfWeek: Number(props.firstDayOfWeek ?? 0) }) : null
+      return week.length ? adapter.getWeek(week[0], props.firstDayOfWeek) : null
     })
   })
 
--- a/packages/vuetify/src/composables/calendar.ts
+++ b/packages/vuetify/src/composables/calendar.ts
@@ -1,5 +1,5 @@
 // Composables
-import { getWeek, useDate } from '@/composables/date/date'
+import { useDate } from '@/composables/date/date'
 import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utilities
--- a/packages/vuetify/src/composables/date/DateAdapter.ts
+++ b/packages/vuetify/src/composables/date/DateAdapter.ts
@@ -37,6 +37,7 @@ export interface DateAdapter<T = unknown> {
   getDiff (date: T, comparing: T | string, unit?: string): number
   getWeekArray (date: T, firstDayOfWeek?: number | string): T[][]
   getWeekdays (firstDayOfWeek?: number | string): string[]
+  getWeek (date: T, firstDayOfWeek?: number | string, firstWeekMinSize?: number): number
   getMonth (date: T): number
   setMonth (date: T, month: number): T
   getDate (date: T): number
--- a/packages/vuetify/src/composables/date/adapters/vuetify.ts
+++ b/packages/vuetify/src/composables/date/adapters/vuetify.ts
@@ -6,160 +6,41 @@ import type { DateAdapter } from '../DateAdapter'
 
 type CustomDateFormat = Intl.DateTimeFormatOptions | ((date: Date, formatString: string, locale: string) => string)
 
-// https://simplelocalize.io/data/locales/
-// then `new Intl.Locale(...).getWeekInfo()`
-const weekInfo: Record<string, { firstDay: number, firstWeekSize: number }> = {
-  '001': { firstDay: 1, firstWeekSize: 1 }, // not covered by Intl
-  AD: { firstDay: 1, firstWeekSize: 4 }, // ca-AD
-  AE: { firstDay: 6, firstWeekSize: 1 }, // ar-AE
-  AF: { firstDay: 6, firstWeekSize: 1 }, // uz-AF or ps-AF | tk-AF
-  AG: { firstDay: 0, firstWeekSize: 1 }, // en-AG
-  AI: { firstDay: 1, firstWeekSize: 1 }, // en-AI
-  AL: { firstDay: 1, firstWeekSize: 1 }, // sq-AL
-  AM: { firstDay: 1, firstWeekSize: 1 }, // hy-AM
-  AN: { firstDay: 1, firstWeekSize: 4 }, // nl-AN
-  AR: { firstDay: 1, firstWeekSize: 1 }, // es-AR
-  AS: { firstDay: 0, firstWeekSize: 1 }, // sm-AS or en-AS
-  AT: { firstDay: 1, firstWeekSize: 4 }, // de-AT
-  AU: { firstDay: 1, firstWeekSize: 1 }, // en-AU
-  AX: { firstDay: 1, firstWeekSize: 4 }, // sv-AX
-  AZ: { firstDay: 1, firstWeekSize: 1 }, // az-AZ
-  BA: { firstDay: 1, firstWeekSize: 1 }, // sr-BA or bs-BA | hr-BA
-  BD: { firstDay: 0, firstWeekSize: 1 }, // bn-BD
-  BE: { firstDay: 1, firstWeekSize: 4 }, // fr-BE or nl-BE
-  BG: { firstDay: 1, firstWeekSize: 4 }, // bg-BG
-  BH: { firstDay: 6, firstWeekSize: 1 }, // ar-BH
-  BM: { firstDay: 1, firstWeekSize: 1 }, // en-BM
-  BN: { firstDay: 1, firstWeekSize: 1 }, // ms-BN
-  BR: { firstDay: 0, firstWeekSize: 1 }, // pt-BR
-  BS: { firstDay: 0, firstWeekSize: 1 }, // en-BS
-  BT: { firstDay: 0, firstWeekSize: 1 }, // dz-BT
-  BW: { firstDay: 0, firstWeekSize: 1 }, // tn-BW or en-BW
-  BY: { firstDay: 1, firstWeekSize: 1 }, // be-BY
-  BZ: { firstDay: 0, firstWeekSize: 1 }, // en-BZ
-  CA: { firstDay: 0, firstWeekSize: 1 }, // en-CA or fr-CA
-  CH: { firstDay: 1, firstWeekSize: 4 }, // de-CH or fr-CH | it-CH | rm-CH
-  CL: { firstDay: 1, firstWeekSize: 1 }, // es-CL
-  CM: { firstDay: 1, firstWeekSize: 1 }, // fr-CM or en-CM
-  CN: { firstDay: 1, firstWeekSize: 1 }, // zh-CN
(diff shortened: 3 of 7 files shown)


packages/vuetify/src/composables/calendar.ts
```diff
@@ -186,7 +186,7 @@ export function useCalendar (props: CalendarProps) {
 
   const weekNumbers = computed(() => {
     return weeksInMonth.value.map(week => {
-      return week.length ? getWeek(adapter, week[0]) : null
+      return week.length ? getWeek(adapter, week[0], { startOfWeek: Number(props.firstDayOfWeek ?? 0) }) : null
```
johnleider, line 189: This will probably have to change: https://github.com/vuetifyjs/vuetify/pull/20794

johnleider (commented):
Please make sure to check https://github.com/vuetifyjs/vuetify/pull/20794 as I think some of the changes might effect your P.R.

J-Sek:
If we decide to trust `Intl.Locale` we could rewrite `firstDay = [ ... ]` into `weekInfo = [ ... ]`
It is just a matter of mapping all the codes

```js
[
    {code: 'US', intlCode: 'en-US'},
    {code: 'FR', intlCode: 'fr'},
    {code: 'PL', intlCode: 'pl'},
    // etc...
]
.map(({code, intlCode}) => {
    const {firstDay, minimalDays} = new Intl.Locale(intlCode).getWeekInfo()
    return `'${code}': { firstDay: ${firstDay}, minimalDays: ${minimalDays} },`
})
.join('\n')
```

J-Sek:
Ok, this PR was wrong in the state when `firstWeekSize` was forced to `4` (ISO) while it is `0` in United States.
I am trying to align with #20794 so we can merge it together.
When generating `weekInfo` dictionary in place of `firstDay` map, there are sometimes multiple matches and the old value do not align. It would be great if we could resolve this quickly.

J-Sek:
Glad I had this test case for Portugal. It was incorrectly assigned during refactoring.

johnleider:
> this PR can be merged instead of https://github.com/vuetifyjs/vuetify/pull/20794 once we resolve changes to firstDay

What changes are left for firstDay?

packages/vuetify/src/composables/date/DateAdapter.ts
```diff
@@ -37,6 +37,7 @@ export interface DateAdapter<T = unknown> {
   getDiff (date: T, comparing: T | string, unit?: string): number
   getWeekArray (date: T, firstDayOfWeek?: number | string): T[][]
   getWeekdays (firstDayOfWeek?: number | string): string[]
+  getWeek (date: T, firstDayOfWeek?: number | string, firstWeekMinSize?: number): number
```
johnleider, line 40: I know that I added new arguments to some of the typings but it was probably the wrong move. We need to figure out a way to get this information without needing to change the interface.

johnleider (approved):

packages/vuetify/src/composables/date/adapters/vuetify.ts
```diff
@@ -6,166 +6,46 @@ import type { DateAdapter } from '../DateAdapter'
...
-  XK: 1,
-  YE: 0,
-  ZA: 0,
-  ZW: 0,
+function weekInfo (locale: string): { firstDay: number, firstWeekSize: number } | null {
+  // https://simplelocalize.io/data/locales/
+  // then `new Intl.Locale(...).getWeekInfo()`
+  const code = locale.slice(-2).toUpperCase()
+  switch (true) {
+    case code === '001': {
```
KaelWD, line 14: `.slice(-2)` will never be 3 characters

packages/vuetify/src/composables/date/adapters/vuetify.ts
```diff
@@ -6,166 +6,46 @@ import type { DateAdapter } from '../DateAdapter'
...
+      return { firstDay: 1, firstWeekSize: 1 }
+    }
+    case locale === 'GB-alt-variant': {
+      return { firstDay: 0, firstWeekSize: 4 }
+    }
+    case `AG AI AL AM AR AS AU AZ BA BD BM BN BR BS BT BW BY BZ CA CL CM CN CO CR CY DM DO
```
KaelWD, line 20: Where is this data from? AU should be monday.

KaelWD:
Sounds like we should just be using ISO rules for week of year https://unicode-org.atlassian.net/browse/CLDR-18275

packages/vuetify/src/composables/date/adapters/vuetify.ts
J-Sek: I have regenerated the switch statement using these scripts:

- [compress with Intl](https://bin.vuetifyjs.com/bins/90jY6w)
- [compress from static info](https://bin.vuetifyjs.com/bins/xHGx8A)

Both yield us the same result. Second one is just to prove the commit prior to [6674fa23](https://github.com/vuetifyjs/vuetify/pull/21199/commits/6674fa23306c7b1c34eef86c2c89d3127e75227b) had correct data.

J-Sek:
> Sounds like we should just be using ISO rules for week of year [unicode-org.atlassian.net/browse/CLDR-18275](https://unicode-org.atlassian.net/browse/CLDR-18275)

We could, but I can only see this being rolled out in Vuetify if we also let users opt-out with `first-day-of-year` prop or some other non-hacky way.

J-Sek:
One last fix.. hopefully. I just realized when clicking on the playground - after choosing non-us locale, selecting `startOfWeek: Sun` was ignored.

ikushum (approved):

johnleider (approved):

PR #21204, opened 2025-04-03
docs(api): update firstDayOfWeek description

NoctalIvan:
Simple string for https://vuetifyjs.com/en/api/v-calendar/#props-first-day-of-week
Wording is ISO datePicker

PR #21207, opened 2025-04-03
feat: deprecate `firstDayOfWeek` prop

J-Sek:
Marks `first-day-of-week` as deprecated because it does not work when using date adapters. We want developers to use `<v-locale-provider>` instead and get both first day of the week as well as correct week number calculations.

closes #21204


J-Sek:
Looking a this after a while.. users sometimes get a dropdown to change week start - overriding their locale preferences. It is a valid scenario and I don't see any way to achieve this with VLocaleProvider.

KaelWD:
Reverted: #21454

Issue #21245, opened 2025-04-11
[Bug Report][3.8.1] VDateInput When multiple="range", pressing the Enter key in the input field will clear its content


Im-Victor:
### Environment
**Vuetify Version:** 3.8.1
**Last working version:** 3.8.1
**Vue Version:** 3.5.13
**Browsers:** Chrome 135.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Steps:
	1.	Click the input field
	2.	Select a date range
	3.	Click the input field again
	4.	Press the Enter key

### Expected Behavior
The Enter key has no effect or does not trigger any action

### Actual Behavior
The input field is cleared.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpVkMFugzAMhl/FyqmVCKhCmypEkfYMO44dMmLWTAYiEhgT4t1nSLfSU5zPv/3bfpuF66vkxdp4HFBkIvfYWFIei7IFyLUZoSLl3KUUWtaEE3wNzpv6R1bYeuxLsQlZOkrNZdK0dvABAYyy6TQSF29vKf4SpD42/IqElYdetZ94zzZqkt9G+ysr0ufzLjGQN5aQ+UNJkSd7+zB7wsNzlCe7lfjrqt7YIMHJdr0HjbXixjCHZtxHZXA4wqWAw42x9bpABu1AFAW2HG9Bg/7aaZfBvGxkWU3/bcQSiTR+ik+pWINzfBJRrchh9HD6wN5/ASWTflE=)

Issue #20964, opened 2025-02-06
[Bug Report][3.7.3] v-text-input not update modelvalue if date value is written with keyboard and then blur component without press enter key


ingfada:
### Environment
**Vuetify Version:** 3.7.3
**Vue Version:** 3.5.12
**Browsers:** Firefox 135.0
**OS:** Windows 10, Windows

### Steps to reproduce
Write date with keyboard and blur the component pointing mouse anywhere.

### Expected Behavior
Data on span text would change because of modelvalue of v-date-input

### Actual Behavior
Modelvalue of v-date-input not change 

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptUUtuwyAQvcqIjROprveRY7VSb9F0Qc1ERcKAYHBbWb57B5LINMoGwfswj8f7ImIYu1fvn+eE4iB6wskbSTicLEA/t9L7si2H0VmS2mK4QgVUrG619YluIMDcTk6hOZ4Es+zIpNjo0RkXmPRBTzL81tQsg5aWmExWYTA8TdW8kZ/l3jeeWuNfWmErR9LOxjtUIUcwkU0ykatNL8nn9IcStp2lScgqwh/K18MRHqQf+q5+89ZE9NIOy3Jzr2vfFejaHrv+15eRS7t9V7XOxzgG7QkiUrqUryfvAsECAc+wwjm4CRr+sCbLc5020paVc7NuZ/EbcpDdfr+JqrdlTdMwxznLQB4l1iebjCnLxx/RUq73)

PR #21249, opened 2025-04-11
fix(VDateInput): Add updateOn prop

ikushum:
fixes https://github.com/vuetifyjs/vuetify/issues/21245
fixes https://github.com/vuetifyjs/vuetify/issues/20964

## Summary of changes
1. Added a new prop `update-on`. This lets us define when the text input should update the model value. If empty, the text field will go into read-only state.
2. Model value is not set to null when an invalid text input is entered. (because most of the display values are invalid text inputs)

## Markup:

```vue
<template>
  <div class="d-flex ga-6">
    <v-date-input label="Single" />
    <v-date-input label="Multiple" multiple />
    <v-date-input label="Range" multiple="range" />
  </div>
</template>
```


packages/vuetify/src/labs/VDateInput/VDateInput.tsx
```diff
@@ -121,7 +125,14 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
     })
 
     const isInteractive = computed(() => !props.disabled && !props.readonly)
-    const isReadonly = computed(() => !(mobile.value && isEditingInput.value) && props.readonly)
+
+    const isReadonly = computed(() => {
+      const canUpdate = props.updateOn.length
+      if (!canUpdate) return true
```
johnleider, line 131:
```suggestion
      if (!props.updateOn.length) return true
```

packages/vuetify/src/labs/VDateInput/VDateInput.tsx
```diff
@@ -121,7 +125,14 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
...
+
+      const isEditingOnMobile = mobile.value && isEditingInput.value
+      return !isEditingOnMobile && props.readonly
```
johnleider, line 134:
```suggestion
      return !(mobile.value && isEditingInput.value) && props.readonly
```

johnleider (changes requested):
Few small tweaks, looks good.

johnleider (approved):

johnleider:
This looks great. Only feedback would be a unit test.

packages/vuetify/src/labs/VDateInput/__tests__/VDateInput.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { VDateInput } from '../VDateInput'
+
+// Utilities
+import { mount } from '@vue/test-utils'
+import { createVuetify } from '@/framework'
+
+global.ResizeObserver = require('resize-observer-polyfill')
```
johnleider, line 7: How do we do this in other places? Surely this isn't the first time we're unit testing a component that utilizes the resize observer.

ikushum: [This](https://github.com/vuetifyjs/vuetify/blob/04551e2c54454ed9ae0221b6e9fbb0789912342e/packages/docs/src/pages/en/getting-started/unit-testing.md?plain=1#L80) was the only place where `resize-observer-polyfill` was referenced in the entire codebase. So I just followed this

ikushum: I converted it to `VDateInput.spec.browser.tsx` and removed the necessity to mock ResizeObserver

johnleider (approved):

Madcheese101:
Can we backport this to 3.7.19? unfortunetly we cannot update to the latest version yet as the system still uses vue 3.3

Issue #20237, opened 2024-07-29
[Feature Request] Prevent VSelect automatic scrolling to the selected item


emreseyit:
### Problem to solve
VSelect opens without scrolling to the selected item and also shows it as an option.

### Proposed solution
Add a prop to disable auto scrolling. It can be named `disable-auto-scroll` and be `false` as default. [The code line](https://github.com/vuetifyjs/vuetify/blob/c0526963a937e7800d0168a930262312d984e0c6/packages/vuetify/src/components/VSelect/VSelect.tsx#L310) that prop going to added to.

I have a demo [here](https://play.vuetifyjs.com/#eNqVUMFqhDAU/JVpLrp0s3Z7FFfYPyj0WEsRfW4DMQlJFIr4701iwba3vSR5k3nzZt7bwpztiqsxp3kiVrLK02hk66luFFDNvDUmPVPRaeVbocj+QAl0JKnzmPmoe5KXhm0A9Q1DKYKeC1i6A1DsnQ+c48Vqox31cFpOXmgFzv8y7h1QtpPX3HVWy0gdWumoYUW9C1fF/yAR2XJWxa/8oQw6wng48tO2BjEabT0WWBqwYrB6RBZWl0U6EHRdpG/+cIm0PIvezk/ZYWckt+H7am37dYoq+QJJ6uY/SzyfsR6RfxwhDrjUSP0ZHiFOXr96K9QtPwSxqtjsBWNsPapJynS8fwPhyJhW). Code is like below:

```html
<template>
  <v-app>
    <v-container>
      <v-select v-model="selected" :items="items" />
      <!-- Proposed solution -->
      <!-- <v-select v-model="selected" :items="items" :auto-scroll="false"/> -->
    </v-container>
  </v-app>
</template>

<script setup>,
  import { ref } from 'vue'

  const selected = ref('item5')
  const items = Array.from({ length: 21 }, (_, i) => 'item' + i.toString())

</script>
```

KaelWD:
Why?

emreseyit:
It's affecting UX and this feature was not available in the migrated project from Vue 2 to Vue 3. Customers are annoyed by this feature and I had to write a patch to remove it.

dawid-polak:
I have a similar problem. With a long multiple list, after each selection of an item I get an automatic scroll to the first selected item. 
I added your parameter (auto-scroll), but this parameter did not change anything. 
Do you have any ideas?

canbax:
[...]

PR #21254, opened 2025-04-12
feat(VSelect): add new auto scroll props (#20237)

jcjp:
## Description

- Resolves: https://github.com/vuetifyjs/vuetify/issues/20237
- Add a new prop for `VSelect` to control auto scroll on selected item by default this is set to `true` so it would still be the same behavior as before

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-select v-model="selected" :items="items" :auto-scroll="false" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selected = ref('item10')
  const items = Array.from({ length: 21 }, (_, i) => 'item' + i.toString())
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -121,8 +121,7 @@ export const VSelect = genericComponent<new <
     returnObject?: ReturnObject
     multiple?: Multiple
     modelValue?: V | null
-    'onUpdate:modelValue'?: (value: V) => void,
-    autoScroll?: boolean,
+    'onUpdate:modelValue'?: (value: V) => void
   },
   slots: Omit<VInputSlots & VFieldSlots, 'default'> & {
     item: { item: ListItem<Item>, index: number, props: Record<string, unknown> }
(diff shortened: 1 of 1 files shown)


johnleider:
Typically we try to avoid Boolean props that default to true. I don't have a good suggestion of how to name it where the default value of false is the current behavior, and true would be opting in to the different behavior. If this doesn't make sense, lmk.

jcjp:
> Typically we try to avoid Boolean props that default to true. I don't have a good suggestion of how to name it where the default value of false is the current behavior, and true would be opting in to the different behavior. If this doesn't make sense, lmk.

How about `prevent-auto-scroll` or `no-auto-scroll`?

johnleider:
> > Typically we try to avoid Boolean props that default to true. I don't have a good suggestion of how to name it where the default value of false is the current behavior, and true would be opting in to the different behavior. If this doesn't make sense, lmk.
> 
> How about `prevent-auto-scroll` or `no-auto-scroll`?

I like `no-auto-scroll`

eesayas (dismissed):

johnleider (dismissed):
Need updates to VSelect.json in API with description. Also updates to newIn.json.

johnleider (dismissed):

johnleider (approved):

PR #21258, opened 2025-04-13
framework: create vuetify-scope package

johnleider:
POC for moving shared resources to a @vuetify/shared packages. This is a precursor to moving theme into its own package, @vuetify/theme. In order to do that, I need to pull the shared stuff out first.

This is a new thing for us, so I'm looking for any and all feedback on process.

TODO:

- [ ] Fix all broken imports in Vuetify
- [ ] Discussion about independent versioning for @vuetify packages
- [ ] Independent packages build process, types, changelogs
- [ ] How do we handle Globals?
- [ ] ...


KaelWD:
> moving theme into its own package

I thought you meant making vuetify work without theme, not making theme work without vuetify.

PR #21259, opened 2025-04-14
chore: use pnpm catalog to manage dependencies

johnleider:
Pending: https://github.com/lerna/lerna/issues/4055


KaelWD:
This also breaks `pnpm update`: https://pnpm.io/catalogs#caveats

userquin:
https://github.com/antfu-collective/taze

package.json
```diff
@@ -24,66 +24,66 @@
...
   },
   "packageManager": "pnpm@10.8.0",
   "pnpm": {
     "patchedDependencies": {
-      "@mdi/js@7.4.47": "patches/@mdi__js@7.4.47.patch",
-      "@testing-library/vue": "patches/@testing-library__vue.patch",
-      "vue-gtag-next": "patches/vue-gtag-next.patch"
+      "@mdi/js@7.4.47": "/home/john/sites/vuetify/patches/@mdi__js@7.4.47.patch",
+      "@testing-library/vue": "/home/john/sites/vuetify/patches/@testing-library__vue.patch",
+      "vue-gtag-next": "/home/john/sites/vuetify/patches/vue-gtag-next.patch"
```
userquin, line 86: Remove prefix `/home/john/sites/vuetify/`, should be relative.

We also need to update to pnpm 10.8.1, `pnpm patch` fixed in this release: https://github.com/pnpm/pnpm/releases

ghiscoding:
If you don't need or care about Nx, then you could also switch to use [Lerna-Lite](https://github.com/lerna-lite/lerna-lite) as an alternative (which I maintain and is closer to the original Lerna code, before Nx came in). Lerna-Lite fully supports Workspace and Catalog protocols since a couple months already for both lerna version & publish commands (see Lerna-Lite [`catalog:`](https://github.com/lerna-lite/lerna-lite/tree/main/packages/publish#catalog-protocol) docs). For more info about Lerna-Lite itself, then see this [section](https://github.com/lerna-lite/lerna-lite?tab=readme-ov-file#about-lerna-lite) on the main readme, the commands are pretty much the same and so nothing new to learn. Also both Workspace and Catalog protocols are purposely used within Lerna-Lite to make sure everything runs smoothly. 

Cheers

ghiscoding:
I see that @KaelWD migrated the project to Lerna-Lite in this commit 7ce2ce62ff11daf73cc336dabe20d60b1df53d01 to now only use and rely on `lerna version` to bump the versions. So with that in place, you guys can now officially switch to pnpm `catalog:` without any problems... you're welcome 😉 

I'm the maintainer of Lerna-Lite and if it's useful to you then upvoting it would be nice 😄 I might add your project to the list of big users (you would become the 2nd biggest user after Jest, so thank you). I also saw that I missed the `lerna clean` command, I might add it in the future, it seems simple enough (EDIT: I worked on the clean command, but then realized that the clean command is probably easier to deal with other simpler tools and so I won't add it to Lerna-Lite).

Also `pnpm update` is now supported in pnpm since June of this year in v10.12.x  (see this pnpm PR [feat: support `catalog:` protocol on `pnpm update`](https://github.com/pnpm/pnpm/pull/9517))

Cheers

ghiscoding:
I know you guys have your own way of publishing with OIDC but I'm happy to say that both Lerna and Lerna-Lite now support OIDC Trusted Publishing, see [Lerna-Lite@4.9.0](https://github.com/lerna-lite/lerna-lite/releases/tag/v4.9.0) with a working demo: https://github.com/lerna-lite-test/oidc

Combine that with Catalogs and life is good 😄

J-Sek (dismissed):

pnpm-workspace.yaml
```diff
@@ -3,6 +3,209 @@ packages:
 
 autoInstallPeers: false
 
+catalog:
+  # Vue ecosystem
```
KaelWD: Note that this grouping is not persistent, anything that rewrites this file such as `pnpm patch-commit` will rearrange everything into alphabetical order.

J-Sek: Good to know 👍🏼  I don't feel strongly about the order and groups, but will try to restore it if I catch it before PR is merged.

PR #21262, opened 2025-04-14
feat(VDatePicker): date picker keyboard controls

### LINKED ISSUE #5629: [Bug Report] add keyboard controls to v-datepicker
### Problem to solve
Currently user's need to use the mouse to select date. Which is not very quick and efficient.

### Proposed solution
Use the arrow keys to move around the calendar when it is open. Press enter to select the date the user has currently selected.

johnleider:
## Motivation and Context
Fixes #5629

TODO:
Why is this happening?
date

## Markup:
<details>

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker v-model="msg"></v-date-picker>
      <pre>{{ JSON.stringify(msg, null, 2) }}</pre>

      <input type="date" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref()
</script>

```
</details>


ikushum (dismissed):

J-Sek:
- (probably not intended) clears selection when using `multiple`
- does not respect min/max and `allowed-dates`

I tried to hook up `focusgroup` polyfill, but it has a different problem. Regardless, I have a feeling we should move focus rather than selection.

KaelWD:
- Throws `RangeError: date value is not finite in DateTimeFormat.format()` if there is no initial selection because `adapter.addDays(undefined, 1)` is `Invalid Date`
- Does not respect `:weekdays="[1, 2, 3, 4, 5]"`

packages/vuetify/src/components/VDatePicker/__tests__/VDatePicker.spec.browser.tsx
```diff
@@ -72,4 +72,36 @@ describe('VDatePicker', () => {
...
+      ))
+
+      await userEvent.click(await screen.findByText(15))
+      expect(model.value).toBeTruthy()
+
+      const initialDate = new Date(model.value)
+      const calendarContainer = screen.getByCSS('.v-date-picker-month__days')
+
+      const keyEvent = new KeyboardEvent('keydown', { key, bubbles: true })
+      calendarContainer.dispatchEvent(keyEvent)
```
KaelWD, line 117: Prefer `userEvent` whenever possible, `dispatchEvent` doesn't always behave like a real user.

KaelWD (changes requested):

J-Sek:
reopening as #22857 so there is no confusion about the author (to blame in case of bugs)

PR #21266, opened 2025-04-15
feat(validation): add optional support for global rules

johnleider:
## Motivation and Context
- add optional global rules support

## Markup:
<details>

```js
// vuetify/dev/index.js
import vuetify from './vuetify'
import App from './App.vue'

import { routes } from './router'
import viteSSR from 'vite-ssr/vue'

import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
import { library } from '@fortawesome/fontawesome-svg-core'
import { fas } from '@fortawesome/free-solid-svg-icons'
import { createRulesPlugin } from '../src/labs/rules/plugin'

library.add(fas)

export default viteSSR(App, { routes }, ({ app }) => {
  app.use(vuetify)
  app.component('FontAwesomeIcon', FontAwesomeIcon)
  app.use(createRulesPlugin({}, vuetify.locale))
})
```

```vue
<template>
    <!-- vuetify/dev/Playground.vue -->
    <v-container class="pa-md-12">
      <v-row>
        <v-col cols="12">
          <v-text-field :rules="['$required']" label="Text field" ></v-text-field>
        </v-col>

        <!-- <v-col cols="12">
          <v-textarea :error-messages="['Required']" label="Textarea"></v-textarea>
        </v-col>

        <v-col cols="12">
          <v-select :error-messages="['Required']" label="Select"></v-select>
        </v-col>

        <v-col cols="12">
          <v-combobox :error-messages="['Required']" label="Combobox"></v-combobox>
        </v-col>

        <v-col cols="12">
          <v-radio-group :error-messages="['Required']" label="Combobox">
            <v-radio label="Option 1" value="option1"></v-radio>
            <v-radio label="Option 2" value="option2"></v-radio>
            <v-radio label="Option 3" value="option3"></v-radio>
          </v-radio-group>
        </v-col>

        <v-col cols="12">
          <v-checkbox :error-messages="['Required']" label="Combobox"></v-checkbox>
        </v-col>

        <v-col cols="12">
          <v-switch :error-messages="['Required']" label="Combobox"></v-switch>
        </v-col>

        <v-col cols="12">
          <v-slider :error-messages="['Required']"></v-slider>
        </v-col>

        <v-col cols="12">
          <v-range-slider :error-messages="['Required']"></v-range-slider>
        </v-col> -->
      </v-row>
    </v-container>
</template>

<script setup>
  import { computed, ref } from 'vue'
  import { useRules } from '../src/labs/rules/rules'

  const rules = useRules()
[...]

PR #21267, opened 2025-04-15
feat(validation): resolve global rules

KaelWD:
#21266


kohenkatz:
Should there be a documentation update that goes with this change?

The example code for `useRules` in the documentation and the 3.8.0 release blog post no longer works after this commit.

The call to `const rules = useRules()` throws `error TS2554: Expected 1 arguments, but got 0` and the use of `rules.required()` throws `Error TS2339: Property 'required' does not exist on type 'any[] | readonly (string | boolean | PromiseLike<ValidationResult> | ((value: any) => ValidationResult) | ((value: any) => PromiseLike<ValidationResult>) | [...])[]'.`

laventnc:
I'm confused, the composable no longer works and I don't see any guidance on how to get it working

laventnc:
If anyone is here and curious about getting this working, check my comment on the linked issue report ^

PR #21221, opened 2025-04-06
feat(VDateInput): Add input-format prop

ikushum:
fixes #19803

## What problem does this solve?
DIfferent regions in the world use different formats for entering date. For example:
- US uses MM/DD/YYYY
- Europe/Asia uses DD/MM/YYYY
- Russia uses DD.MM.YYYY

Vuetify as well as other date adapters use `Date.parse` to parse date string to date object and `Date.parse` is not locale aware. For example if we enter `01/02/2025` in VDateInput then `Jan 2, 2025` is selected which is correct for US but in Europe/Asia  region it should be `Feb 1, 2025`.

## How is the problem solved?
A new prop `input-format` is added which can be used to specify the format in which users can enter a date in the VDateInput. Values can be a combination of `yyyy`, `mm` and `dd` with separators `.`, `-`, `/` (e.g. `yyyy-mm-dd`, `dd/mm/yyyy`, `dd.mm.yyyy`) or a custom function. 

If `input-format="dd/mm/yyyy"` is passed then the selected date will be `Feb 1, 2025` instead of current behaviour which will select `Jan 2, 2025`

## Future Improvements
Vuetify will eventually support [Masked Inputs](https://github.com/vuetifyjs/vuetify/pull/19207), so `input-format` can be used in the future to mask the input with mask placeholder as well.

## Markup:
For German locale
```vue
<template>
  <v-app>
    <v-container>
      <v-locale-provider locale="de-DE">
        <v-date-input
          v-model="pickedDate"
          input-format="dd.mm.yyyy"
          placeholder="dd.mm.yyyy"
        />
      </v-locale-provider>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const pickedDate = ref()
</script>
```

For UK locale
```vue
<template>
  <v-app>
    <v-container>
      <v-locale-provider locale="en-GB">
        <v-date-input
          v-model="pickedDate"
          input-format="dd/mm/yyyy"
          placeholder="dd/mm/yyyy"
        />
      </v-locale-provider>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const pickedDate = ref()
</script>
```

PR #21273, opened 2025-04-16
fix(VDateInput): accept value when picker is hidden

J-Sek:
## Description

`[Enter]` should not clear the input.

related to #21221

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <pre>value: {{ value }} | {{ typeof value }}</pre>
      <v-date-input v-model="value" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const value = ref()
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
+++ b/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
@@ -221,7 +221,6 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
 
       if (!menu.value || !isFocused.value) {
         menu.value = true
-        return
       }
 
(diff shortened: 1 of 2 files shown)


ikushum (approved):
Looks good

ikushum:
Looks like there are lint issues

johnleider (approved):

ikushum:
This bug is quite obvious and annoying. What do we need to merge this? @J-Sek  @johnleider

johnleider:
It was approved, author never merged.

J-Sek:
I don't like the fact the tests are a bit brittle... needed to move the new one as first, otherwise it fails. But getting the fix out is more important atm.

PR #21274, opened 2025-04-16
feat(VDateInput): add close-on-selection prop

### LINKED ISSUE #21234: [Feature Request] Provide option to close VDateInput when the range is selected
### Problem to solve
The "range" VDateInput is made to select a range of dates. As soon as the range is selected, the picker is no longer necessary, and so it should be closed (Which would be the same behavior than for simple VDateInput: When you select a date, it closes the picker)

### Proposed solution
- Solution 1: Automatically close the picker when the date range is fully selected (when the user selected the start date and end date). This would be a breaking change.
- Solution 2: Provide a boolean option to enable this behavior
- Solution 3: Provide a way to control the visibility of the picker, so this behavior can be implemented using custom code

GaetanWcz:
## Description

Resolves #21234 
Adding a `close-on-selection` to ensure the menu is closing once the range has been selected. 

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-input
        v-model="model"
        label="Select range"
        max-width="368"
        multiple="range"
        close-once-range-selected
      >
      </v-date-input>

      <v-date-input
        v-model="model2"
        label="Select range"
        max-width="368"
        multiple="range"
      >
      </v-date-input>

      <v-date-input
        v-model="model3"
        label="Select range"
        max-width="368"
        close-once-range-selected
      >
      </v-date-input>

      <v-date-input
        v-model="model4"
        label="Select range"
        max-width="368"
      >
      </v-date-input>
    </v-container>
  </v-app>
</template>

<script setup>

  const model = shallowRef(null)
  const model2 = shallowRef(null)
  const model3 = shallowRef(null)
  const model4 = shallowRef(null)
</script>

```


J-Sek:
IMO it should be `close-on-selection` (`true` by default) and apply to both single and range selection. We'd only ignore it for `multiple === true`.

GaetanWcz:
Switching to close-on-selection then ! :)

packages/vuetify/src/labs/VDateInput/VDateInput.tsx
```diff
@@ -232,8 +235,14 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
                           } else {
                             model.value = value
 
-                            if (!props.multiple) {
-                              menu.value = false
+                            const isRange = props.multiple === 'range'
+                            const isSingle = props.multiple !== true && !isRange
+
+                            if (isSingle) {
+                              menu.value = props.closeOnSelection
```
ikushum, line 242: Looks like it should be:

```suggestion
                              menu.value = !props.closeOnSelection
```
?

ikushum (approved):

ikushum:
After some discussion with the internal team, we have decided to fix this issue using a different approach - by exposing menu property to the parent component and providing full control over the picker's visibility. A [PR](https://github.com/vuetifyjs/vuetify/pull/21298) is already up for this

PR #21252, opened 2025-04-12
feat(date): export VuetifyDateAdapter

mueller-jens:
## Description

resolves #19904

exports the VuetifyDateAdapter so it is possible to extend it

implement functions-option 

## Markup:

```vue
<template>
<v-card>
<v-row>
<v-col>
{{" Previous Month: "+adapter.getPreviousMonth(new Date())}}

</v-col>
</v-row>
<v-row>
<v-col>
{{" MyDateFunction: "+adapter.getFirstDayOfYear()}}
</v-col>
</v-row>
</v-card>
</template>

<script setup lang="ts">
  import { DateOptions, VuetifyDateAdapter,DateAdapter } from '@/composables/date'
  const  adapter: MyDateAdapter = new MyDateAdapter ({ locale: 'de-DE' })


</script>
<script lang="ts">

  export class MyDateAdapter extends VuetifyDateAdapter  {
    getFirstDayOfYear() {
      return new Date(new Date().getFullYear()-1,0,1);
    }
  }
</script>
```

PR #21290, opened 2025-04-20
feat(VDatePicker): restore `weekday-format`

J-Sek:
## Description

`weekdayFormat` existed in v2, but was not yet re-implemented.

Note: It is going to be ignored by custom date adapters until those packages get updated by community.
- `date-fns` is already printing with `EEEEEE` which equals to Intl's `short`
- day.js has `dd` hardcoded - i.e. `narrow`
- luxon uses `Intl` with `short` hardcoded

## Markup:

```vue
<template>
  <v-container>
    <v-select
      label="Locale"
      variant="outlined"
      v-model="locale.current.value"
      :items="['en', 'de', 'pl', 'fr', 'ar']"
    />
    <v-date-picker
      :key="locale.current.value"
      class="mt-3"
      weekday-format="short"
    />
  </v-container>
</template>

<script setup>
  import { useLocale } from 'vuetify'
  const locale = useLocale()
</script>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VCalendarMonthly.json
+++ b/packages/api-generator/src/locale/en/VCalendarMonthly.json
@@ -13,7 +13,6 @@
     "showMonthOnFirst": "Whether the name of the month should be displayed on the first day of the month.",
     "showWeek": "Whether week numbers should be displayed when using the `month` view.",
     "start": "The starting date on the calendar (inclusive) in the format of `YYYY-MM-DD`. This may be ignored depending on the `type` of the calendar.",
-    "weekdayFormat": "Formats day of the week string that appears in the header to specified locale.",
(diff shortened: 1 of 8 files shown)


johnleider:
> Devs can extend adapters to force particular weekday format or make it support new arguments by copy/pasting from the date-io PR linked above.

Is this something worth putting into an alert in the docs?

J-Sek:
> Is this something worth putting into an alert in the docs?

Alert.. not really. More like a full blog post to cover differences & limitations. For now we have the disclaimer at the end of the prop description in API docs and I feel it is fine. People are asking for help with their specific adapters from time to time, so I guess we'll see their voices on Discord. You could touch on it in v3.8.6 announcement with something like "_week format supported by default date adapter, see [issueID] for more details_"

IMO, getting built-in adapter better (and extendable: [#21252](https://github.com/vuetifyjs/vuetify/pull/21252)) makes it easier for devs to achieve expected i18n goals without having to use 3rd party libraries.

johnleider (approved):

KaelWD:
Moved to dev

PR #21295, opened 2025-04-22
docs: app bar fully visible in mobile

eesayas:
## Description
In mobile screens, the login button is cut off. This change will update the login button to an icon button.

Before:
image

After:
Screen Shot 2025-04-21 at 11 44 07 PM

## Note
https://github.com/vuetifyjs/one/pull/23 needs to be merged & released first. Then this PR will be updated to the latest `@vuetify/one` release version

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -11,7 +11,7 @@
     "directory": "packages/docs"
   },
   "scripts": {
-    "dev": "vite",
+    "dev": "vite --force",
     "build": "vite build",
     "preview": "vite preview",
     "preview-https": "HTTPS=true vite preview",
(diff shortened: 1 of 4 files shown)


johnleider (approved):

packages/docs/package.json
```diff
@@ -11,7 +11,7 @@
     "directory": "packages/docs"
   },
   "scripts": {
-    "dev": "vite",
+    "dev": "vite --force",
```
KaelWD, line 14: ?

eesayas: https://vite.dev/guide/dep-pre-bundling.html#caching

This is for the sole purpose of linking. So it pre-bundles if there are changes to the externally linked package. Might as well just have that as default behavior in development.

KaelWD: I thought that would make startup really slow but it seems to be around 1200ms either way.

PR #21298, opened 2025-04-22
feat(VDateInput): expose menu as v-model

### LINKED ISSUE #21234: [Feature Request] Provide option to close VDateInput when the range is selected

ikushum:
fixes #21234

## Description
Lets us have full control over VDateInput's menu property via `v-model:menu`

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-date-input
        v-model="date"
        v-model:menu="menu"
        multiple="range"
        @update:model-value="onUpdateModelValue"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const date = ref()
  const menu = ref(false)

  const onUpdateModelValue = (value) => {
    if (!value) return

    const isFirstSelection = value.length === 1
    menu.value = isFirstSelection
  }
</script>


```

### DIFF
--- a/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
+++ b/packages/vuetify/src/labs/VDateInput/VDateInput.tsx
@@ -39,6 +39,7 @@ export const makeVDateInputProps = propsFactory({
     type: String as PropType<StrategyProps['location']>,
     default: 'bottom start',
   },
+  menu: Boolean,
 
   ...makeDisplayProps(),
   ...makeFocusProps(),
@@ -64,6 +65,7 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
     save: (value: string) => true,
     cancel: () => true,
     'update:modelValue': (val: string) => true,
+    'update:menu': (val: boolean) => true,
   },
 
   setup (props, { emit, slots }) {
@@ -79,7 +81,7 @@ export const VDateInput = genericComponent<VDateInputSlots>()({
       val => Array.isArray(val) ? val.map(item => adapter.date(item)) : val ? adapter.date(val) : val
     )
 
-    const menu = shallowRef(false)
+    const menu = useProxiedModel(props, 'menu')
     const isEditingInput = shallowRef(false)
     const vTextFieldRef = ref<VTextField>()
     const disabledActions = ref<typeof VConfirmEdit['props']['disabled']>(['save'])


ikushum:
Good catch. @J-Sek 

We have the same issue with VSelect:

```vue
<template>
  <v-app>
    <v-container>
      <v-select label="select menu"
        v-model:menu="menu"
        :items="['Item 1', 'Item 2', 'Item 3']"
        @update:model-value="selectkeepOpen" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const menu = ref(false)

  const keepOpen = (value) => {
    menu.value = true
  }
</script>
```

J-Sek:
All dropdown-based componets get closed when using [enter] instead of [space]. `close-on-selection` could help suppress this behavior and let us keep menu open when we drop [space] for selection with VAutocomplete and VCombobox (#21311). Is there any chance this PR can get us the same thing?

johnleider (approved):

J-Sek:
I would use it with a small delay to improve UX a bit: [playground](https://play.vuetifyjs.com/#eNplUstuwjAQ/JWVLwQpDxBFqlCgrfoDPfRWerDCBiw5tuVHUIX4965taKG9xPbM7GpmlI8Tc7ZrXoypx4BsxVqPg5Hc42arANqx4saka3p0WnkuFNoLlMAdqSuhTPBXEGCsBr1Dud6yyG7ZP2Y1oApEx+OWHoL0wkgkynK1vxt9DiZuW6UF1chliDKtXg/3yuZquPnrOCI5UNvcBKWn66wwHhz6kPOKwWjr4QQWezhDb/UAE+poEuUAtNd5iHZgHSXF9BeNmS5oz6VDoiLJ3ZfqoA+q80IruPouximcsuE4riXWUu8JzZjooRifaolq7w+wgfmPmjYeufCg8AhvZE84LCysNzHEuxhQB1/YEhaz2fSyi+ola3Vqjgwmc5k5x4M+bZOLoArYuWSLelnPH1i8PNZLVqaJ8u6XydjnNxsSuKY=)

PR #21311, opened 2025-04-25
fix(VAutocomplete, VCombobox): space key should not select

### LINKED ISSUE #20728: [Bug Report][3.7.4] VAutocomplete: Highlighted list item gets selected when trying to continue the search by pressing the space bar
### Environment
**Vuetify Version:** 3.7.4
**Vue Version:** 3.5.13
**Browsers:** Chrome 130.0.0.0
**OS:** Mac OS 10.15.7

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptkDFrwzAQhf/Kqxa14Na03YxT6Nahe4c4g3AvRXCyhHTyEvzfK9sxJCGL0D0+vXtP+5NKsa8/Q3gZM6lGtUIusBH66AagHZ9NCMt1GXo/iLEDxbO0Ell878srEtpkoLHFKe06tT9BrDA10N8+koMNKTtdYTSci/qKqcItY66Zt7sMX0PvmA6d2hLUW+r6NvasrK3a+qJtGVMfbRAkkryWti74KGVzpCMmHKN30OWj9IwDxTcJXPrDbiYe9Rcxe/z4yL8P+mlesFoWMzVVQ2ZejsM/bJ16UQ==)

### Steps to reproduce
Type "lorem", press `Arrow Down`, and then try to continue the search by pressing `Space Bar`

### Expected Behavior
Focus back to the search field to continue the search

### Actual Behavior
Selects the highlighted list item

### Other comments
[VListItem.tsx#L197](https://github.com/jay-zahiri/vuetify/blob/54e430ff1aba72cbcd72ea3683e745020100504e/packages/vuetify/src/components/VList/VListItem.tsx#L197) dispatches `click` event when `Space Bar` is pressed.
```ts
    function onKeyDown (e: KeyboardEvent) {
      if (e.key === 'Enter' || e.key === ' ') {
        e.preventDefault()
        e.target!.dispatchEvent(new MouseEvent('click', e))
      }
    }
```

Also this👇 ([link](https://github.com/vuetifyjs/vuetify/pull/21311#issuecomment-2832755429)):
> ...
> Now that I test some other use cases, I see that `Backspace` is also excluded. So if the user goes through the list with `ArrowDown`/`ArrowUp`, then there's no way to alter the search value again by pressing `Space Bar` or `Backspace` – without having to reach for the mouse. But also, typing anything other than `Space Bar` or `Backspace` does give the user ability to alter its search, as expected. So conceptually, I'd suggest `Space Bar` and `Backspace` to be considered in the typing context, I guess.
> 
> `Backspace` is probably even more important to be recognized as typing in a multi-selection context as well, as the user might want to continue searching and selecting – just not with what's currently typed in. Again, the user has to reach for the mouse, click the text input field, and delete the input value – instead of just pressing `Backspace` immediately to continue the search. I've updated [the repro link](https://play.vuetifyjs.com/#eNrNkkFOwzAQRa8yeGOQQqK2FKEoRWIHEnsWTRdWmCBLdmzZ42yq3B27oaIOPQA7z/fz92jm74/Mu656sbYcA7KaNYTaKkH43A4AzXgvrD0dT0VnBhJyQPcjzUQg05n4CgnPMkAto5PftWx/BJKksAb+bhxqkNYHzQsYhQpRXcFUwJIRObO+yqgc2sB0aNm5g+qftngJvaVbUAn9ZR6uMZQz28T8pYaceswGAqCDIhlnsBxRUy03m5R58U11EYhY+s5JS+CRwpwLqa1xFFtx2MMEvTMaeMwSTzhA9PUE2n/BLhG3/BWVMvBhnPq84Xfpg9kymrGpYJtyW642LB2eyjUreqE8FllIZ+3wDRGa3QY=) accordingly. The second input is marked as `multiple`, and you can try typing "lorem", press `ArrowDown` to focus on a `VListItem`, press `Space Bar` to select it, and then try to press `Backspace` to start searching for items starting with "ipsum" instead, for instance.
> 
> I'd like to suggest that pressing `Space Bar` when not marked as `multiple`, and `Backspace` – regardless of `multiple` or not, would perform like any other typing.

J-Sek:
## Description

- prevents selection with [space]
- suppresses ripple effect
- [ ] keep menu open when using [enter] to select items and parent component has `multiple`
  - ideally new prop `close-on-selection='false'` from #21274 could help suppress [VMenu.tsx:169](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VMenu/VMenu.tsx#L169)

resolves #20728

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete label="autocomplete" :items="items" />
      <pre>search: '{{ search }}'</pre>
      <pre>menu: '{{ menu }}'</pre>
      <v-autocomplete label="autocomplete + multiple" multiple :items="items" v-model:search="search" v-model:menu="menu" />
      <v-combobox label="combobox" :items="items" />
      <v-combobox label="combobox + multiple" multiple :items="items" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const search = ref()
  const menu = ref()
  const items = [{ title: 'Lorem ipsum', value: 1 }, { title: 'Lorem apsum', value: 2 }, { title: 'Lorem alpsum', value: 3 }]
</script>
```


ikushum:
Authoring Practices Guide (APG) from the W3C recommends using space key for toggling [checkbox](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) and [multi selection lists](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)

image
image

Does this PR mean we are taking away this feature?

I think the underlying [issue](https://github.com/vuetifyjs/vuetify/issues/20728) can be not considered a bug because:
1. When the user presses "Arrow Down", the input element loses focus state, so pressing "Space Bar" should not make any impact to the input.
2. When the user presses "Arrow Down", the focus state is rather shifted to the select option, so pressing "Space Bar" should toggle the selection as it currently does.

J-Sek:
I think the checkbox within selection dropdown is not a regular use case that falls under this spec. Some frameworks show checkmark without a box. We render `v-checkbox-btn` because it looks nice this way, but it cannot be selected for explicit interaction.

Moreover, user cannot easily exit the menu when they click "ArrowDown" - as "Shift" + "Tab" closes the menu and in case of VAutocomplete, the focus does not land back on the input. The only "trick" I see here is to type {anything except space}{backspace}{space} which is annoying highlights overall inconsistency.

jay-zahiri:
Hi, original poster here.

Granted, pressing `Space Bar` in a multi-selection context like checkbox or even radio button probably should perform a selection. But I like to suggest that autocompletes that don't have multi-selection, creates a different type of expectation and context, as you perform more of a search action. I'd say, pressing `ArrowDown` in that context to go through the suggestions (maybe it's a long list that overflows in it's dropdown menu or maybe you're just too lazy to type more), not finding what you were looking for, and so you type in a space (in this case) to continue your search – should not break that experience by selecting whatever random list item you were on because you were just scrolling through.

Now that I test some other use cases, I see that `Backspace` is also excluded. So if the user goes through the list with `ArrowDown`/`ArrowUp`, then there's no way to alter the search value again by pressing `Space Bar` or `Backspace` – without having to reach for the mouse. But also, typing anything other than `Space Bar` or `Backspace` does give the user ability to alter its search, as expected. So conceptually, I'd suggest `Space Bar` and `Backspace` to be considered in the typing context, I guess.

`Backspace` is probably even more important to be recognized as typing in a multi-selection context as well, as the user might want to continue searching and selecting – just not with what's currently typed in. Again, the user has to reach for the mouse, click the text input field, and delete the input value – instead of just pressing `Backspace` immediately to continue the search. I've updated [the repro link](https://play.vuetifyjs.com/#eNrNkkFOwzAQRa8yeGOQQqK2FKEoRWIHEnsWTRdWmCBLdmzZ42yq3B27oaIOPQA7z/fz92jm74/Mu656sbYcA7KaNYTaKkH43A4AzXgvrD0dT0VnBhJyQPcjzUQg05n4CgnPMkAto5PftWx/BJKksAb+bhxqkNYHzQsYhQpRXcFUwJIRObO+yqgc2sB0aNm5g+qftngJvaVbUAn9ZR6uMZQz28T8pYaceswGAqCDIhlnsBxRUy03m5R58U11EYhY+s5JS+CRwpwLqa1xFFtx2MMEvTMaeMwSTzhA9PUE2n/BLhG3/BWVMvBhnPq84Xfpg9kymrGpYJtyW642LB2eyjUreqE8FllIZ+3wDRGa3QY=) accordingly. The second input is marked as `multiple`, and you can try typing "lorem", press `ArrowDown` to focus on a `VListItem`, press `Space Bar` to select it, and then try to press `Backspace` to start searching for items starting with "ipsum" instead, for instance.

I'd like to suggest that pressing `Space Bar` when marked as `multiple` would focus back up to the text input field (with cursor at the end of the input value), `Backspace` would perform like any other typing – regardless of `multiple` or not, and `Space Bar` would perform like any other typing when not `multiple`.

jay-zahiri:
GitHub's or vscode's own command palette could perhaps be an example of what I would expect. Google's search both in Chrome and its start page considers `Space Bar` as typing in its own autocomplete when arrowing down to the suggestions – arguably this would be de facto digital UX of the majority of normies and superusers alike =P

jay-zahiri:
@J-Sek I noticed you removed `Space Bar` (`e.key !== ' '`) completely, shouldn't it be more like [if not `Space Bar` OR not `multiple`], because you are choosing to show check boxes?

J-Sek:
@jay-zahiri, `multiple` was discuss because of checkbox, but if we agree it is not a regular checkbox, but just a decoration for the list item, then whatever spec says about checkboxes can be ignored. Moreover @ikushum linked a "listbox" pattern that refers to a scrollable list detached from the input. Actual guidance is outlined as "Combobox pattern" ([here](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)), but I cannot really agree space should hide the menu... what it says seems a bit disconnected from reality and common sense.

Regardless, `multiple` changes nothing. The critical variable for the behavior of [space] and [backspace] is the existence of free-text input that helps user narrow down options displayed in the menu. It is already covered by `checkPrintable` and should not require any more code within `onListKeydown`.

johnleider:
What are our available options to choose from? I'll throw it up in a poll and see what people think. There is definitely a deeper discussion that needs to happen here.

J-Sek:
> What are our available options to choose from?

- leave it as-is
  - space is selecting, backspace is ignored
- follow majority (see [analysis](https://github.com/vuetifyjs/vuetify/issues/20728#issuecomment-2830359527))
  - space & backspace land in the primary input

GitHub is acutally so inconsistent (e.g. label selection between PRs and Issues) - it is a hot mess, out of the table.
PrimeVue is similar to the current behavior, but kinda worse. Out of the table as well.

jay-zahiri:
@johnleider agreed. To emphasize my previous two cents; I just think since the initial input is free text, there should be a consistent way of navigating back to that initial input via keyboard only. Typing any other letter performs the expected behavior and I'd consider `Space Bar` and `Backspace` to be associated with that expectation of typing in a single line free text input as well.

That's it =) I'll let whatever plan or process you have run its course and I'm not going to ramble more now.

PS. Solid job on the framework, followed since v1.4 👏

MatthewAry:
I think we should follow majority and to add to @J-Sek own investigation

- https://mui.com/base-ui/react-autocomplete/  [space] does not select (same goes for other mui autocomplete components)
- https://ant.design/components/auto-complete [space] does not select
- https://www.heroui.com/docs/components/autocomplete [space] does not select

In general, if any kind of input field is focused, then it should take priority. Any key bindings that might interfere with the use of the input field should be disabled.

eesayas:
I agree with going with the majority as well. Another example from [Material Angular ](https://material.angular.dev/components/autocomplete/examples).  

Moreover,  `space` & `backspace` can be relevant on suggestions filtering especially for items where the `space` is a character. For example: `San Francisco`, `San Diego`, `Santorini`.

ikushum:
If we go for the majority then it's also worth keeping the input in focused state (caret in input field) even after pressing `up` or `down` arrow.

J-Sek:
> If we go for the majority then it's also worth keeping the input in focused state (caret in input field) even after pressing `up` or `down` arrow.

It will be a welcome improvement on top of these changes as focus-based approach is borked when we use `#item` slot and include anything focusable within (I was sharing this [example](https://play.vuetifyjs.com/playgrounds/1dMGUw) many times before I realized that keyboard navigation is totally broken when it lacks `tabindex='-1'`). Out of scope for this PR though.. it's going to be a challenge and greater risk of regression.

packages/api-generator/src/locale/en/VList.json
```diff
@@ -3,6 +3,7 @@
     "itemType": "Designates the key on the supplied items that is used for determining the nodes type.",
     "activatable": "Designates whether the list items are activatable.",
     "disabled": "Puts all children inputs into a disabled state.",
+    "filterable": "Prevents list item selection using [space] key and pass it back to the text input. Used internally for VAutocomplete and VCombobox.",
```
johnleider, line 6:
```suggestion
    "filterable": "**FOR INTERNAL USE ONLY** Prevents list item selection using [space] key and pass it back to the text input. Used internally for VAutocomplete and VCombobox.",
```

packages/vuetify/src/directives/ripple/index.ts
```diff
@@ -283,6 +287,9 @@ function updateRipple (el: HTMLElement, binding: RippleDirectiveBinding, wasEnab
     el._ripple.class = value.class
   }
 
+  const allowedKeys = (typeof value === 'object' ? value.keys : null) ?? [keyCodes.enter, keyCodes.space]
```
johnleider, line 290: We're already calling **isObject(value)** above, can we just assign that to a const and reuse it as opposed to calling typeof value === 'object'?

johnleider (commented):
A few questions, overall looks good.

johnleider (approved):

Issue #21291, opened 2025-04-21
[Bug Report][3.8.2] Clicking label set cursor position to beginning of the input in VTextField


muzakon:
### Environment
**Vuetify Version:** 3.8.2
**Vue Version:** 3.5.13
**Browsers:** Chrome 135.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
1. Create a text field with label
2. Click the label to check cursor position when focusing input

### Expected Behavior
Clicking on the input's label should place the text cursor at the position corresponding to where the user clicked

### Actual Behavior
Clicking the input's label moves the text cursor to the beginning of the input field

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNplU01v2zAM/SucLm6BKK4bbBgCt9uw0267zzsoNt0I0xck2WlQ9L+Plp3Edg8GSL1HUnx++vPGgq/zH85t+w7ZnpURtVMi4nNlAMqet9brFKestiYKadBDqzrZTECCvD1d04mrgL7wVLHisWIQNEVfKjZjEc9BiGeFBGnhX6ThBxuj1XvYPbjXFRngp5L1P2leIB4RpHFdzAIocUAFd78yDWJM7kHbHkNiRXyNUHc+WA/RpqMD0iAztLHtrRO0ElWzoSREFM2AkRL1QJMRTkf01IzIXaD16+Ei2GwXu+RuuVvPh+E89V3u0XNtG1S0tRMhnKxvKrZk7IVzaBpOF0XPJQlP5HC0pwK+QaYbyfGMGeyvMbdtm33oEs9u0PZaOFwoVV3mfihJAlLJTM41xQg9NE2q8eKBF2vC9yTPYoXrJZ7gUwrWNb3wUphIvM406BW5bCXKc5nPFZ27jQCy2s2OlN/8OKKTcUdf08lk7DKfOZ7SUHvpIgSM3fg3pXbWR3gDjy28Q+uthoweSzbQgRxOdoHLasS5a4UKeH/DLkpPcPb7ojxxynycR5PY+4bttp+3xY4NwdftI9ukVpvFEx3P/v4H4zMg3Q==)

### Recommended Fix:

Avoid using padding on the input element.

- Use margin instead to position the input within the v-field.

- Set the height of the `.v-field__field` element using: `height: var(--v-input-control-height);`

- Remove any min-height property from the input element.

[...]

PR #21322, opened 2025-04-27
fix(VTextField): on click label moves cursor (#21291)

jcjp:
## Description

- Fixes: #21291 
- Maintain the looks of the `VTextfield` as it is and prevent cursor movement on click label
- Affected components `VAutocomplete` and `VCombobox` set the `margin-top` to `0` to prevent increase of size

## Markup:

```vue
<template>
  <v-form>
    <v-container fluid>
      <v-row>
        <v-col cols="12" sm="6">
          <p style="margin-bottom: 30px">
            Clicking the input's label (I'm a label) moves the text cursor to
            the beginning of the input field, instead of placing it where the
            user clicked.
          </p>
          <v-text-field
            v-model="password"
            :append-inner-icon="show1 ? 'mdi-eye' : 'mdi-eye-off'"
            :type="show1 ? 'text' : 'password'"
            label="I'm a label"
            name="input-10-1"
            @click:append-inner="show1 = !show1"
            variant="underlined"
          />
        </v-col>
      </v-row>
    </v-container>
  </v-form>
</template>

<script setup>
  import { ref } from 'vue'

  const show1 = ref(false)
  const password = ref('Password')
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.sass
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.sass
@@ -84,6 +84,7 @@
         right: 0
         width: 100%
         padding-inline: inherit
+        margin-top: 0
 
       .v-field--active
         input
--- a/packages/vuetify/src/components/VCombobox/VCombobox.sass
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.sass
@@ -84,6 +84,7 @@
         right: 0
         width: 100%
         padding-inline: inherit
+        margin-top: 0
 
       .v-field--active
         input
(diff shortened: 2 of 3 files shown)


johnleider (commented):
Are the padding and min-height declarations needed? I wasn't actually able to reproduce this locally, so I'm looking for more information.

jcjp:
> Are the padding and min-height declarations needed? I wasn't actually able to reproduce this locally, so I'm looking for more information.

John let me retest this on our latest version, and check if the issue still persist and close this one if the fix is not needed thanks for checking.

PR #21342, opened 2025-04-30
fix(VOtpInput): trim clipboard text and update focus on finish

Dyskal:
## Description
This PR implements improvements to the `v-otp-input` component, addressing paste trimming and enhancing the autofill focus behavior discussed in #21131.

Specifically, the changes in this PR include:

1.  **Improved Paste Handling:**
    *   Leading/trailing whitespace is now removed using `trim()`.

2.  **Enhanced `finish` event:**
    *   Within the watcher that detects when the model's length equals the target length (signaling a completed OTP, often after autofill), the component now explicitly sets the `focusIndex` to the last input field.
    *   This change ensures that after a successful autofill, the focus is directed to the final input element, which provides a more intuitive user experience compared to the focus remaining on the second element.

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input v-model="otp" />
    </v-container>
  </v-app>
</template>

<script>
  import { ref } from 'vue'

  export default {
    name: 'Playground',
    setup () {
      const otp = ref('')
      return {
        otp,
      }
    },
  }
</script>

### DIFF
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -166,7 +166,7 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
       e.preventDefault()
       e.stopPropagation()
 
-      const clipboardText = e?.clipboardData?.getData('Text').slice(0, length.value) ?? ''
+      const clipboardText = e?.clipboardData?.getData('Text').trim().slice(0, length.value) ?? ''
 
       if (isValidNumber(clipboardText)) return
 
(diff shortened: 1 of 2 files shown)


johnleider (approved):
Looks good, just needs a unit test.

Dyskal:
Does VOtpInput currently have unit tests?
It only have an old Cypress test that is not ran by Vitest

johnleider:
> Does VOtpInput currently have unit tests? It only have an old Cypress test that is not ran by Vitest

Guess those hadn't been converted yet. I went ahead and updated them. Should be easy to understand what's going on, lmk if you have questions though. https://github.com/vuetifyjs/vuetify/commit/e140a2e2fe289c3473743e9f8d590410fe819a97

johnleider (approved):

PR #21354, opened 2025-05-02
docs: save or favorite search results

eesayas:
## Description
Favorite search results in documentation, allowing users to quickly save and revisit searches.

## Markup:
N/A

Screen Shot 2025-05-02 at 11 48 43 AM


johnleider (approved):

packages/docs/src/components/app/search/SearchDialog.vue
```diff
@@ -152,20 +181,39 @@
...
-  function onClickResult (result: any) {
+  function deleteFavorite (index: number) {
+    deleteItem(favorites, index)
+  }
+  function selectResult (result: any) {
     const array = searches.value.slice(0, 6)
-
     array.unshift(result)
-
     searches.value = array
```
J-Sek, line 198: I see some duplicates on my "Recent" list. Maybe it's just a matter of including `= [...new Set(array)]` here.

J-Sek (changes requested):

packages/docs/src/components/app/search/SearchDialog.vue
eesayas: yes I was beginning to see that too. thank you for catching this! it also uncovered the edge case that iff `searches` has a defined non-array value then we'll run into errors.

J-Sek: I'd call it "deduplication" instead of "uniqueness" (nice to have)

J-Sek (approved):

packages/docs/src/components/app/search/SearchDialog.vue
eesayas: alrighty 😆

Issue #21275, opened 2025-04-16
[Bug Report][3.8.1] v-otp-input loses focus after token is pasted


Icebreaker83:
### Environment
**Vuetify Version:** 3.8.1
**Vue Version:** 3.5.13
**Browsers:** Chrome 135.0.0.0
**OS:** Windows 10

### Steps to reproduce
Copy/paste some token into otp input

### Expected Behavior
otp input doesn't lose focus and form can be submited

### Actual Behavior
otp input loses focus after token is pasted, and form can't be submitted.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdj80KwkAMhF8l5KJCfyhFEGkFn8N6KDWFYrsbdtNeSt/d/VFQb8kw801yW9GaLr8yZ8tMeMZKaOKxFbo0CqBa0pY5jGHptJJ2UGTeUhC1cDoongWWdNIPGusGRT9JNQj5J5v/h70S2VX+1elW25mBBSzJHKuHibURWMFQDxv0Rk+wc+fuvB3Aca1AaITae/YHD40YB8AtwTI7ZkWJfjhlBSZ9O1pKfn6P2v0FXHtZUQ==)

### Other comments
Behavior is caused by [line](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx#L175)

jcjp:
Maybe we could add an option to stop it from losing focus on copy paste, as per our documentation it is intended to lose focus after paste.

Icebreaker83:
@jcjp There is simmilar [issue](https://github.com/vuetifyjs/vuetify/issues/18016) but for manual entry, and blur was removed.  
Most common usecase is to have v-otp-input inside a form that is submited on Enter, where user copy/pastes the token.  

I personally would remove blur completely, and if blur is needed you can always handle that in [finish event](https://vuetifyjs.com/en/api/v-otp-input/#events-finish), although I can't imagine a single scenario where blur is needed.

My workaround is to refocus the component on finish event

jcjp:
> [@jcjp](https://github.com/jcjp) There is simmilar [issue](https://github.com/vuetifyjs/vuetify/issues/18016) but for manual entry, and blur was removed. Most common usecase is to have v-otp-input inside a form that is submited on Enter, where user copy/pastes the token.
> 
[...]

PR #21356, opened 2025-05-03
feat(VOtpInput): allow to keep focus on paste (#21275)

jcjp:
## Description

- Resolves: #21275
- Add new prop to keep focus on paste to `VOtpInput`, retains the default behavior of blur on paste

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input v-model="token" keep-focus  />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const token = ref()
</script>
```


johnleider (changes requested):
Need to add prop description to VOtpInput.json and newIn.json

KaelWD:
Does this even need a new prop? The current behaviour seems pretty useless.

jcjp:
> Does this even need a new prop? The current behaviour seems pretty useless.

Good question I should just delete the `onBlur` thanks Kael!

johnleider (approved):

PR #21359, opened 2025-05-04
feat(VTextField, VTextarea): suppress browser's autocomplete

### LINKED ISSUE #21353: [Feature Request] Autocomplete="off" on all input fields globally
### Problem to solve
It would solve the problem to manually put autocomplete="off" on all input fields on the systems. This could work on small systems, but on huge projects it becomes a very big effort.

### Proposed solution
Make a prop that could work globally on vuetify.js settings, that we could insert the value of autocomplete on all used inputs of vuetify.

J-Sek:
## Description

- expose `autocomplete` as prop so it can be used in `defaults: { global: { ... } }`
- detect custom value `suppress` to append unique suffix to field's `name` and help developers prevent field suggestions on browsers that ignore `autocomplete='off'`
- supports SSG by updating generated `name` on focus

resolves #21353

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-form>
        <v-text-field name="email" label="E-mail" autocomplete="suppress" />
        <v-textarea name="comment" label="Comment" autocomplete="suppress" />
        <v-btn type="submit">Submit</v-btn>
      </v-form>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.tsx
+++ b/packages/vuetify/src/components/VTextField/VTextField.tsx
@@ -7,6 +7,7 @@ import { makeVFieldProps, VField } from '@/components/VField/VField'
 import { makeVInputProps, VInput } from '@/components/VInput/VInput'
 
 // Composables
+import { makeAutocompleteProps, useAutocomplete } from '@/composables/autocomplete'
 import { useAutofocus } from '@/composables/autofocus'
 import { useFocus } from '@/composables/focus'
 import { forwardRefs } from '@/composables/forwardRefs'
@@ -43,6 +44,7 @@ export const makeVTextFieldProps = propsFactory({
   },
   modelModifiers: Object as PropType<Record<string, boolean>>,
 
+  ...makeAutocompleteProps(),
   ...makeVInputProps(),
   ...makeVFieldProps(),
 }, 'VTextField')
@@ -94,6 +96,7 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
     const vInputRef = ref<VInput>()
     const vFieldRef = ref<VField>()
     const inputRef = ref<HTMLInputElement>()
+    const autocomplete = useAutocomplete(props)
     const isActive = computed(() => (
       activeTypes.includes(props.type) ||
       props.persistentPlaceholder ||
@@ -101,6 +104,10 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
       props.active
     ))
     function onFocus () {
+      if (autocomplete.isSuppressing.value) {
+        autocomplete.update()
+      }
+
       if (!isFocused.value) focus()
 
       nextTick(() => {
@@ -217,7 +224,8 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
                         autofocus={ props.autofocus }
                         readonly={ isReadonly.value }
                         disabled={ isDisabled.value }
-                        name={ props.name }
+                        name={ autocomplete.fieldName.value }
+                        autocomplete={ autocomplete.fieldAutocomplete.value }
                         placeholder={ props.placeholder }
                         size={ 1 }
                         type={ props.type }
--- a/packages/vuetify/src/components/VTextarea/__tests__/VTextarea.spec.browser.tsx
+++ b/packages/vuetify/src/components/VTextarea/__tests__/VTextarea.spec.browser.tsx
@@ -19,7 +19,7 @@ describe('VTextarea', () => {
       </Application>
     ))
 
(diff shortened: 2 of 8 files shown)


J-Sek:
Browser extensions (password managers) will probably still attach to the field as they tend to look at the field's **label** as well. Depending on the use case it might be desired or not. The suppression on that level is very hard and developers might need to target specific extensions individually
- risk sacrificing some a11y by including zero-width space (e.g. `label="E-m&ZeroWidthSpace;ail"`) - the only way to stop NordPass
- include custom CSS (e.g. `[data-protonpass-role] { display: none }`) - the only way to stop ProtonPass
upon research with those 2 extensions mentioned above, I decided to avoid any more suppression logic as those can be easily supplemented on top of Vuetify.

packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -27,6 +28,7 @@ import type { VInputSlots } from '@/components/VInput/VInput'
 const activeTypes = ['color', 'file', 'time', 'date', 'datetime-local', 'week', 'month']
 
 export const makeVTextFieldProps = propsFactory({
+  autocomplete: String as PropType<'suppress' | string>,
```
johnleider, line 31: Move this to reusable props like other composabes.

johnleider (changes requested):
Looks good

packages/vuetify/src/composables/autocomplete.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface InputAutocompleteProps {
+  autocomplete?: 'suppress' | string
+  name?: string
+}
+
+// Composables
+export function useAutocomplete (props: InputAutocompleteProps) {
+  const reloadTrigger = ref(0)
+
+  const isSuppressing = computed(() => props.autocomplete === 'suppress')
```
KaelWD, line 14: These can all be toRef instead of computed.

packages/vuetify/src/composables/autocomplete.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const fieldName = computed(() => {
+    return isSuppressing.value
+      ? `${props.name}-${new Date().getTime()}-${reloadTrigger.value}`
```
KaelWD, line 18: This will cause SSR hydration mismatch

packages/vuetify/src/composables/autocomplete.ts
```diff
@@ -0,0 +1,34 @@
+// Utilities
+import { computed, ref } from 'vue'
+
+// Types
+export interface InputAutocompleteProps {
+  autocomplete?: 'suppress' | string
```
KaelWD, line 6: Use `| undefined` not `?:` or it doesn't actually enforce that the prop is declared.

KaelWD (dismissed):

packages/vuetify/src/composables/autocomplete.ts
```diff
@@ -0,0 +1,43 @@
...
+import type { PropType } from 'vue'
+
+// Types
+export interface InputAutocompleteProps {
+  autocomplete: 'suppress' | string | undefined
+  name?: string
+}
+
+// Composables
+export const makeAutocompleteProps = propsFactory({
```
J-Sek, line 15: single character difference with `makeVAutocompleteProps`... but maybe it does not matter

packages/vuetify/src/composables/autocomplete.ts
J-Sek: switched to `{name}-{uid}-{timestamp}`

packages/vuetify/src/components/VTextarea/__tests__/VTextarea.spec.browser.tsx
```diff
@@ -43,7 +43,7 @@ describe('VTextarea', () => {
       </Application>
     ))
 
-    const el = screen.getByCSS('#input-v-0')
+    const el = screen.getByCSS('textarea[rows]')
```
J-Sek, line 46: It fails after `useId()` being called and the `id` is effectively `input-v-1`

alternatively I could try `useToggleScope` to run the composable only for `suppress`

johnleider (dismissed):
Approved pending api description update.

johnleider (dismissed):

johnleider (dismissed):

johnleider (dismissed):
Looks good. If you think there is value for unit tests on autocomplete.ts, add them.

McCdama:
If the type of <VTextFiled .. /> is `password` it won't work!

PR #21375, opened 2025-05-05
fix(VDateInput): accept typing/changing more than one date

J-Sek:
## Description

- follow-up to [#21249](https://github.com/vuetifyjs/vuetify/pull/21249#review-thread-or-comment-id-1323997059)
- makes it possible to enter or edit dates by typing value directly in the text field

Note: considered a bug fix because single value works fine while "multiple" and "range" do not

## Markup:

```vue
<template>
  <div class="d-flex ga-6">
    <v-date-input label="Single" />
    <v-date-input label="Multiple" multiple />
    <v-date-input label="Range" multiple="range" />
  </div>
</template>
```


johnleider (approved):

J-Sek:
Included in #21409

PR #21263, opened 2025-04-14
feat(VNumberInput): extract number from pasted text

J-Sek:
## Description

Currently user action is completely ignored if even one character causes `onBeforeinput` to bail. A nicer UX would be to extract the digits to support copy/pasting from spreadsheets or PDFs.

I would prefer to merge it to `master` as it can be considered a bug by many developers at this point.

## Markup:

```vue
<template>
  <v-container>
    <v-row>
      <v-col cols="6" lg="3">
        <h5>(default precision="0" max="200")</h5>
        <v-number-input v-model="example1" clearable hide-details="auto" :max="200" />
        <code class="d-block pt-3">value: {{ example1 ?? 'null' }}</code>
        <input class="border-md pa-2 mt-2" type="number" v-model="example1" />
      </v-col>
      <v-col cols="6" lg="3">
        <h5>(precision="1")</h5>
        <v-number-input v-model="example2" :precision="1" hide-details="auto"></v-number-input>
        <code class="d-block pt-3">value: {{ example2 ?? 'null' }}</code>
        <input class="border-md pa-2 mt-2" type="number" v-model="example2" />
      </v-col>
      <v-col cols="6" lg="3">
        <h5>(precision="5")</h5>
        <v-number-input v-model="example3" :precision="5" hide-details="auto"></v-number-input>
        <code class="d-block pt-3">value: {{ example3 ?? 'null' }}</code>
        <input class="border-md pa-2 mt-2" type="number" v-model="example3" />
      </v-col>
      <v-col cols="6" lg="3">
        <h5>(precision="3" min='4' max='120') w/ initial null</h5>
        <v-number-input v-model="example4" :precision="3" hide-details="auto" :min='4' :max='120'></v-number-input>
        <code class="d-block pt-3">value: {{ example4 ?? 'null' }}</code>
        <input class="border-md pa-2 mt-2" type="number" v-model="example4" />
      </v-col>
    </v-row>
    <v-alert class="mt-12">
      Values to copy/paste:
      <ul>
        <li><v-kdb>421.00</v-kdb></li>
        <li><v-kdb>23 567.20</v-kdb></li>
        <li><v-kdb>8,999.99</v-kdb></li>
        <li><v-kdb>$ 120.95</v-kdb></li>
      </ul>
    </v-alert>
  </v-container>
</template>

<script setup>
  import { ref } from 'vue'
  const example1 = ref(150.5)
  const example2 = ref(25.5)
  const example3 = ref(0.052)
  const example4 = ref(null)
</script>
```

PR #21391, opened 2025-05-08
feat(VNumberInput): custom decimal separator

### LINKED ISSUE #20254: [Feature Request] v-number-input: Using the locale decimal separator
### Problem to solve
I want to let the users input a decimal value in a v-number-input.
It is possible to do so using the dot: 3.8
In many other languages, the comma is used as a decimal separator: 3,8
It is possible to type numbers this way in a native `<input type="number" />`, but not using this component.

### Proposed solution
None, but I wish the behavior of v-text-input were the same as the native input tag on this regard.

J-Sek:
## Description

- only decimal separator, no dependency on the locale, no grouping
- we would do this simple prop anyway
- needs to be coordinated with #21263, not sure which one will be merged first

resolves #20254

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="400">
      <v-number-input
        v-model="num"
        control-variant="stacked"
        decimal-separator=","
        variant="outlined"
        :precision="2"
        :step=".5"
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'

  const num = ref(32.5)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/components/VNumberInput/VNumberInput.tsx
@@ -16,7 +16,7 @@ import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utilities
 import { computed, nextTick, onMounted, ref, shallowRef, toRef, watch, watchEffect } from 'vue'
-import { clamp, extractNumber, genericComponent, omit, propsFactory, useRender } from '@/util'
+import { clamp, escapeForRegex, extractNumber, genericComponent, omit, propsFactory, useRender } from '@/util'
 
 // Types
 import type { PropType } from 'vue'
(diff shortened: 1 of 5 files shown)


packages/vuetify/src/components/VNumberInput/VNumberInput.tsx
```diff
@@ -196,21 +206,28 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
...
+
+      const sep = '\\^$*+?.()|{}[]'.includes(decimalSeparator.value)
+        ? `\\${decimalSeparator.value}`
+        : decimalSeparator.value
+
+      // Allow only numbers, "-" and {decimal separator}
+      // Allow "-" and {decimal separator} only once
+      // Allow "-" only at the start
+
+      console.log(sep, `^-?\\d*${sep}?\\d*$`)
```
johnleider, line 218:
```suggestion
```

johnleider (changes requested):

johnleider (approved):
Just needs a unit test.

J-Sek:
Rebased and reworked a bit to support parsing pasted text
\+ more unit tests

johnleider (approved):

PR #21400, opened 2025-05-10
fix(VNavigationDrawer): prevent list item jump when rail width is passed

eesayas:
## Description
List item prepend now uses the rail width to maintain center alignment and prevent layout shifting when the rail opens or closes.

## Markup
```vue
<template>
  <v-navigation-drawer rail rail-width="70">
    <v-list nav>
      <v-list-item></v-list-item>
    </v-list>

    <v-divider></v-divider>

    <v-list density="compact" nav>
      <v-list-item></v-list-item>
      <v-list-item></v-list-item>
      <v-list-item></v-list-item>
    </v-list>
  </v-navigation-drawer>
</template>

```


eesayas:
> Is there not a way to fix this without introducing a new property?

Yes. I made new commit to use provide/inject instead. Thanks for the nudge 😎

packages/docs/src/pages/en/components/navigation-drawers.md
```diff
@@ -87,6 +81,18 @@ When using the **rail** prop, the drawer will shrink (default 56px) and hide eve
...
 
+#### Expand on hover
+
+Places the component in **rail** mode and expands once hovered. This **does not** alter the content area of **v-main**. The width can be controlled with the **rail-width** property.
+
+<ExamplesExample file="v-navigation-drawer/prop-expand-on-hover" />
+
+#### Rail width
+
+The drawer will resize to the value of **rail-width** (or default to `56px` if undefined) and will hide all **v-list** items except the first. This must be used in **rail** mode.
```
johnleider, line 92: You don't have to change this, but putting default values in content is just one more place we have to remember to update if it ever changes.

packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -227,6 +237,7 @@ export const VNavigationDrawer = genericComponent<VNavigationDrawerSlots>()({
                 'v-navigation-drawer--floating': props.floating,
                 'v-navigation-drawer--is-hovering': isHovering.value,
                 'v-navigation-drawer--rail': props.rail,
+                'v-navigation-drawer--rail-width': props.railWidth !== Number(56),
```
johnleider, line 240: This assumes the default will always be 56.

packages/vuetify/src/components/VList/VListItem.tsx
```diff
@@ -180,6 +181,19 @@ export const VListItem = genericComponent<VListItemSlots>()({
     const { roundedClasses } = useRounded(roundedProps)
     const lineClasses = toRef(() => props.lines ? `v-list-item--${props.lines}-line` : undefined)
 
+    // Determine if VListItem is within a VNavigationDrawer
+    // prepend width is important for drawer rail mode
+    const navigationDrawer = inject(VNavigationDrawerSymbol, null)
+    const prependWidth = computed(() => {
+      if (navigationDrawer) {
+        const value = Number(navigationDrawer.prependWidth.value)
+        return value !== 56 ? convertToUnit(value) : 'auto'
```
johnleider, line 190: This assumes the value will always be 56

johnleider (changes requested):
If we truly need to make this many changes in order to fix this, the P.R. needs to rebase to `dev`.

PR #21401, opened 2025-05-10
feat(VDateInput): sync with placeholder, infer from locale

J-Sek:
## Description

- drops 'function' mode
- detects the separator and sign order
- infers format from locale
- supplements placeholder
- included #21375 to avoid breaking `multiple`

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-defaults-provider :defaults="{ VDateInput: { persistentPlaceholder: true, persistentHint: true } }">
      <v-container max-width="400" class="d-flex flex-column ga-6 py-6">

        <v-locale-provider locale="de-DE">
          <!-- without locale configured -->
          <v-date-input label="matching locale" input-format="dd.mm.yyyy" hint="locale: de-DE (dd.mm.yyyy)" />
        </v-locale-provider>

        <v-locale-provider locale="en-GB">
          <!-- without locale configured -->
          <v-date-input label="from locale" hint="locale: en-GB (dd/mm/yyyy)" />
        </v-locale-provider>

        <v-locale-provider locale="hr-HR">
          <!-- without locale configured -->
          <v-date-input label="from locale" hint="locale: hr-HR (dd. mm. yyyy.)" />
        </v-locale-provider>

        <!-- without locale configured -->
        <v-date-input label="just format" input-format="dd.mm.yyyy" hint="locale: en-US (mm/dd/yyyy)" />

        <v-locale-provider locale="az-AZ">
          <!-- different than locale -->
          <v-date-input label="override locale" input-format="dd-mm-yyyy" hint="locale: az-AZ (yyyy-mm-dd)" />
        </v-locale-provider>
      </v-container>
    </v-defaults-provider>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/composables/dateFormat.ts
+++ b/packages/vuetify/src/composables/dateFormat.ts
@@ -57,7 +57,7 @@ export function useDateFormat (props: DateFormatProps, locale: Ref<string>) {
 
   function inferFromLocale () {
     return Intl.DateTimeFormat(locale.value ?? 'en-US', { year: 'numeric', month: '2-digit', day: '2-digit' })
-      .format(new Date(1999, 11, 7))
+      .format(adapter.toJsDate(adapter.parseISO('1999-12-07')))
       .replace(/(07)|(٠٧)|(٢٩)|(۱۶)|(০৭)/, 'dd')
       .replace(/(12)|(١٢)|(٠٨)|(۰۹)|(১২)/, 'mm')
       .replace(/(1999)|(2542)|(١٩٩٩)|(١٤٢٠)|(۱۳۷۸)|(১৯৯৯)/, 'yyyy')
@@ -90,7 +90,7 @@ export function useDateFormat (props: DateFormatProps, locale: Ref<string>) {
     }
 
     function autoFixYear (year: number) {
-      const currentYear = new Date().getFullYear()
+      const currentYear = adapter.getYear(adapter.date())
       if (year > 100 || currentYear % 100 >= 50) {
         return year
       }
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/composables/dateFormat.ts
```diff
@@ -0,0 +1,131 @@
...
+    function validateDateParts (dateParts: Record<string, number>) {
+      const { y: year, m: month, d: day } = dateParts
+      if (!year || !month || !day) return null
+      if (month < 1 || month > 12) return null
+      if (day < 1 || day > 31) return null
+      return { year: autoFixYear(year), month, day }
+    }
+
+    function autoFixYear (year: number) {
+      const currentYear = new Date().getFullYear()
```
ikushum, line 93: how about using `adapter.getYear()`?

packages/vuetify/src/composables/dateFormat.ts
```diff
@@ -0,0 +1,131 @@
...
+    validator: (v: string) => !v || DateFormatSpec.canBeParsed(v),
+  },
+}, 'lazy')
+
+export function useDateFormat (props: DateFormatProps, locale: Ref<string>) {
+  const adapter = useDate()
+
+  function inferFromLocale () {
+    return Intl.DateTimeFormat(locale.value ?? 'en-US', { year: 'numeric', month: '2-digit', day: '2-digit' })
+      .format(new Date(1999, 11, 7))
```
ikushum, line 60: How about:
```js
.format(adapter.parseISO('1999-11-07'))
```

Want to avoid using JS Date as much as possible and reuse our adapter so that when we switch to Temporal, the transition would be smooth

J-Sek: You only get `Date` if you use default adapter. `Intl` only works with native `Date`.
It will need to be `adapter.toJsDate(adapter.parseISO('1999-11-07'))`.

Good catch! I get different results when simulating time zone from left hemisphere.

```js
.format(adapter.toJsDate(adapter.parseISO('1999-11-07'))) // --> 11/07/1999
.format(new Date('1999-11-07')) // --> 11/06/1999
```

PR #21423, opened 2025-05-16
fix(VSelect): remove redundant css selector

kingyue737:
## Description

```sass
.v-select
  .v-field
    &.v-field
```
will be compiled to 
```css
.v-select .v-field.v-field
```
which has a redundant `.v-field` and also won't affect selector specificity

## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.sass
+++ b/packages/vuetify/src/components/VSelect/VSelect.sass
@@ -10,7 +10,7 @@
       .v-text-field__prefix,
       .v-text-field__suffix,
       .v-field__input,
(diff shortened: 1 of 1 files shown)


KaelWD:
> won't affect selector specificity

It does, this is needed to override

https://github.com/vuetifyjs/vuetify/blob/e820ca7f01a7443ad8c5d586c2c9deee02af44f7/packages/vuetify/src/components/VTextField/VTextField.sass#L24-L25

kingyue737:
Oh sorry for bothering, cheated by gpt. I should have done more search and test

PR #21424, opened 2025-05-16
feat(VAutocomplete): add prop to allow sorting filter results

akpath:
## Description
This PR adds a prop that allows you to specify a sort order for the filter results.

It would be nice to be able to override the filter results sort order for the VAutocomplete. For instance, it can be desirable that items which _start with_ the search text appear before those where the item merely _contains_ the search text. See the Playground below: with the sort prop specified, I can search for "a" and "Alligator" will appear first even though "Bear" also contains "a" and appears first in the default items ordering. Without a sort prop, this behavior takes a lot more work to implement.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-autocomplete v-model:search="search" :filtered-items-sort="filteredItemsSort" :items="items" label="With Sorting" />
      <v-autocomplete :items="items" label="Without Sorting" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
    data () {
      return {
        items: ['Lion', 'Tiger', 'Bear', 'Alligator'],
        search: '',
      }
    },
    methods: {
      filteredItemsSort (a, b) {
        const searchLower = this.search.toLowerCase()
        const aLower = a.title.toLowerCase()
        const bLower = b.title.toLowerCase()

        return aLower.indexOf(searchLower) - bLower.indexOf(searchLower)
      },
    },
  }
</script>

```


J-Sek:
I don't think it makes much difference to a simple computed ([example](https://play.vuetifyjs.com/#eNp1VE1vm0AQ/SujvWBLGFy1UVWEHaU9VYrUSo3UQ8hhgcFeZWHR7uI4ivzfM8tiILF82vl8M+8N4vGNGV3Ed20bHTpkCUst1q3kFrdZA5AeVrxte7N3CtVYLhrUUPPj6kWUdr/J2M16nbGhiMpaKCQ3hhJ1vvo2ywA8vLbocJ7zcsvT2BsR/JW8EM0OMnYnpdhxq3TGIMdKaaTgT+TOr/kzGqgpNgEabAxG4+j4vKtfvbOqUEQH7azlsKpViTIxhFq49b2RsakkEaSCI2CUtlj+dt48L3mOktL/hd3DP6qh5ad8fG2HCbd/idIcSHV2whpBnEij6v4mFPFHSePZscg1hRat7Yvw2BIUlFjxTlp482AltxwWy7MLoNF2upl8gH6zBB6De6GaIITgQexQO8Ndwb3jiYKncH4IJ2ICQTAGT944DQGnQkdqJtO4mbzzrVxtYyy0e80NwgbsXpjIT7iNrLpXL6h/UWqxnFoGKr5nCgPc+vaeGDU7ibFcLHgI+RI2W5+thLSoh11cSZ8PB7jlbA59HTPAK1xrpHuWpOPI6WIAfJxwyZ73NIk9v0bZ1+Xnuvxz3YU4HjISTYnHP9XiPHo1YHxOjOQGVv4lqmk8fmrsFLKv0U305Ttzxo9ozcKKS4Phhx+Ljz29A7OyXBc=)), but has a potential to be be much more performant when instead of function prop, we could just focus on strictly supporting order by first match.

Let's say we have `filteredOrder: 'first-match' | 'default'` and internally we could run:

```ts
return search.value && props.filteredOrder === 'first-match'
  ? filteredItems.value.toSorted((a, b) => getMatches(a)!.title![0][0] - getMatches(b)!.title![0][0])
  : filteredItems.value
```

My assumption is that it would take less effort than lowercasing each title and finding a phrase inside, but it would need to be verified.

Issue #21045, opened 2025-03-01
[Bug Report][3.7.14] [VTabs]  Height of VTabsWindowItem worngly calculated


mueller-jens:
### Environment
**Vuetify Version:** 3.7.14
**Vue Version:** 3.5.13
**Browsers:** Edge 133.0.0.0
**OS:** Windows 10

### Steps to reproduce
Open the link, you will see that the end of the vertical scrollbar is not visible, because the height of the card, which contains the tabs and the tab window, has the same height 

### Expected Behavior
The end of the scrollbar should be visible. The height of the inner window should be card - height of the tab

### Actual Behavior
The end of the scrollbar is not visible

### Reproduction Link
[...]

PR #21435, opened 2025-05-19
fix(VCard): Move v-card diplsay to flex to adjust the hight of v-card-text

mueller-jens:
## Description

I have changed the display attribute of v-card to flex added flex-direction: column and min-height: 0 to v-card style.
to have the complete v-card-text visible inside the v-card.
Second change is to set the flex-shrink attribute to 0 so that the v-tabs will not shrink.

## Markup:

```vue
<template>
  <v-card style="height: 50%" id="card">
    <v-tabs v-model="tab" bg-color="primary" id="tabs">
      <v-tab value="one"> Item One</v-tab>
      <v-tab value="two">Item Two</v-tab>
      <v-tab value="three">Item Three</v-tab>
    </v-tabs>

    <v-card-text class="ma-0 pa-0" style="height: 100%">
      <v-tabs-window v-model="tab" id="windowItem" style="height: 100%">
        <v-tabs-window-item
          class="ma-0 pa-0"
          value="one"
          style="overflow-y: auto; background-color: green"
        >
          Height of the card: {{cardHeight}}<br />
          Height of the tabs: {{tabsHeight}}<br />
          Height of the window item:{{widnowItemHeight}}<br /><br />
          Height of the window item should be {{cardHeight - tabsHeight}}
          <br /><br /><br /><br /><br /><br />
          <br /><br /><br /><br /><br /><br /><br /><br /><br /><br /><br />
        </v-tabs-window-item>

        <v-tabs-window-item
          value="two"
          minHeight="100%"
          style="overflow-y: auto; background-color: blue"
        >
          Two
        </v-tabs-window-item>

        <v-tabs-window-item value="three"> Three </v-tabs-window-item>
      </v-tabs-window>
    </v-card-text>
  </v-card>
</template>
<style>
  html {
    overflow-y: hidden;
  }
  #app {
    height: 100vh;
  }
</style>

<script>
  export default {
    data: () => ({
      tab: null,
    }),
    computed: {
      cardHeight() {
        return document.getElementById('card').getBoundingClientRect().height
      },
      widnowItemHeight() {
        return document.getElementById('windowItem').getBoundingClientRect()
          .height
      },
      tabsHeight() {
        return document.getElementById('tabs').getBoundingClientRect().height
      },
    },
  }
</script>
```
Before the change, the scroll-down button is not visible andf the height of the v-card-text is not the expected number of pixel:

image

After the change the scroll-down butto is visible and the calculated height has the expected number of pixel (heigh of the card - height of the tabs):

image

### DIFF
--- a/packages/vuetify/src/components/VCard/VCard.sass
+++ b/packages/vuetify/src/components/VCard/VCard.sass
@@ -4,7 +4,8 @@
 
 @include tools.layer('components')
   .v-card
-    display: block
+    display: flex
+    flex-direction: column
     overflow: hidden
     overflow-wrap: $card-overflow-wrap
     position: relative
@@ -157,6 +158,7 @@
     font-size: $card-text-font-size
     font-weight: $card-text-font-weight
     letter-spacing: $card-text-letter-spacing
+    min-height: 0px
     opacity: $card-text-opacity
     padding: $card-text-padding
     text-transform: $card-text-text-transform
--- a/packages/vuetify/src/components/VTabs/VTabs.sass
+++ b/packages/vuetify/src/components/VTabs/VTabs.sass
@@ -7,6 +7,7 @@
 @include tools.layer('components')
   .v-tabs
     display: flex
+    flex-shrink: 0
     height: var(--v-tabs-height)
 
     @at-root


J-Sek:
I acknowledge that placing tabs in card might require a bit of finesse, but these changes are a bit too risky.

Please take a look at the [example](https://github.com/vuetifyjs/vuetify/issues/21045#issuecomment-3050263568) solution for the issue that works solely on built-in helper classes. You can experiment with the linked playground by removing secondary `d-flex flex-column` or replacing `pa-0` with `pa-3` on the `v-card-text`, to assess if it is works as you intended.

If you have any follow-up questions, please reach out to us in our [Discord community](https://community.vuetifyjs.com/).

PR #21438, opened 2025-05-20
docs: trim and dont search empty string

MajesticPotatoe:
## Description

Prevents empty search of space

### DIFF
--- a/packages/docs/src/components/app/search/SearchDialog.vue
+++ b/packages/docs/src/components/app/search/SearchDialog.vue
@@ -119,8 +119,8 @@
   const model = defineModel<boolean>()
   const searchString = defineModel('search', { type: String, default: '' })
 
-  // Algolia doesn't tokenize hyphens by default. Sanitize search string.
-  const searchTerm = computed(() => searchString.value.replace(/-/g, ' ').trim())
+  // trim and replace starting instance of -someText to prevent broad queries
+  const searchTerm = computed(() => searchString.value.replace(/^-\S+\s*/, '').trim())
 
   const list = ref<InstanceType<typeof SearchResults>>()
   const searchClient = algoliasearch(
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/app/search/SearchDialog.vue
```diff
@@ -151,7 +151,8 @@
   })
 
   function searchFunction (helper: AlgoliaSearchHelper) {
-    if (helper.state.query) helper.search()
+    console.log('moo')
+    if (helper.state.query && searchTerm.value) helper.search()
```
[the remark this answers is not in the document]
MajesticPotatoe: must have slipped into rogue branch when trying to figure out push changes.

KaelWD:
@eesayas "hyphens are not tokenized" explain this then?

Screenshot_20250520_150433
image

KaelWD:
We have this specifically configured: 
image

eesayas:
ok if hyphen is being configured as separators. then i think we just need to prevent search term starting with non-alphanumeric @MajesticPotatoe

packages/docs/src/components/app/search/SearchDialog.vue
```diff
@@ -120,7 +120,7 @@
   const searchString = defineModel('search', { type: String, default: '' })
 
   // Algolia doesn't tokenize hyphens by default. Sanitize search string.
-  const searchTerm = computed(() => searchString.value.replace(/-/g, ' '))
+  const searchTerm = computed(() => searchString.value.replace(/-/g, ' ').trim())
```
eesayas, line 123:
```suggestion
    const searchTerm = computed(() => searchString.value.replace(/^[^a-zA-Z0-9]/, ' ').trim())
```
replace only the first char if it's alpha numeric

KaelWD: This still breaks advanced query syntax: https://www.algolia.com/doc/api-reference/api-parameters/advancedSyntax/

eesayas: hmm, so should we just filter out the empty search results instead? keep the search term as it was before 🤔

PR #21454, opened 2025-05-21
chore: un-deprecate firstDayOfWeek

J-Sek:
Looking a this after a while.. users sometimes get a dropdown to change week start - overriding their locale preferences. It is a valid scenario and I don't see any way to achieve this with VLocaleProvider

### DIFF
--- a/packages/api-generator/src/locale/en/generic.json
+++ b/packages/api-generator/src/locale/en/generic.json
@@ -18,7 +18,7 @@
     "end": "Applies margin at the start of the component.",
     "falseIcon": "The icon used when inactive.",
     "falseValue": "Sets value for falsy state.",
-    "firstDayOfWeek": "Deprecated, Sets the first day of the week, starting with 0 for Sunday.",
+    "firstDayOfWeek": "Sets the first day of the week, starting with 0 for Sunday.",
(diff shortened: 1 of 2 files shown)


johnleider:
What does all of this look like in the precursor to Temporal taking over date stuff? Are we able to do this in a way that regardless of what adapter the developer uses, it will work?

J-Sek:
> What does all of this look like in the precursor to Temporal taking over date stuff? Are we able to do this in a way that regardless of what adapter the developer uses, it will work?

Option 1: Keep disclaimers about "no guarantee" to work with custom adapters and rely on developers to extend the adapter class (we have a minimal guide in the docs, but we could add more with copy-pastable snippets).

Option 2: Introduce a proxy around the adapter so we can detect non-empty `firstDayOfWeek` argument and redirect method calls to our internal implementation.

johnleider:
I like Option 2 as long as it can be a small code footprint and require **zero** changes from the user.

J-Sek:
> roduce a proxy around the adapter so we can detect non-empty `firstDayOfWeek` argument and redirect method calls to our internal implementation.

Do you need a PoC before we merge this PR?

johnleider (approved):

PR #21466, opened 2025-05-22
feat(VDatePicker): disable months and years if not allowed

### LINKED ISSUE #20465: [Feature Request] Date picker shown years which are not valid
### Problem to solve
if I use allowed-dates, in the year selection all years are visible/selectable, even if the selection of dates is not valid

https://play.vuetifyjs.com/#eNpVjsEKwjAMhl8l9KTg3H10BcG3sB5KG2FYs9K13UF8d2u04C4h+fIn/395iiXa/hTCsWQUg5AJH8GbhEoTgCydCYFbHuxMyUyE8YcYuqruwmTvGBsFGIz384qOt8uoRTE+I4wKuFFAuMK57nZ7LdqZkv3mXTOudOv8Id9gsv8LnMnhrYqcJvE6UPaey/UN+qVKwg==

### Proposed solution
Only years should be shown/active if at least one date is selectable

ikushum:
fixes #20465

## Description
Disable month and year buttons based on `allowed-dates` prop when:
- [x]  `allowed-dates` is array
- [x]  `allowed-dates` is function (open for suggestion/help)

## Markup:

```vue
<template>
  <v-app>
    <v-container>
        <h2>Date picker</h2>
        <v-date-picker
          mode="month"
          hide-header
          :allowed-dates="['2025-05-01', '2025-03-02', '2024-02-01']"
        ></v-date-picker>

        <h2>Date picker years</h2>
        <VDatePickerYears
          :allowed-years="[2025, 2026]"
          class="mb-10"
        />

        <h2>Date picker months</h2>
        <VDatePickerMonths
          :allowed-months="[2, 3]"
        />
    </v-container>
  </v-app>
</template>

```

## Screenshots
image
image

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
@@ -194,6 +194,20 @@ export const VDatePicker = genericComponent<new <
       return targets
     })
 
+    function allowedYears (year: number) {
+      if (!Array.isArray(props.allowedDates) || !props.allowedDates.length) return true
+      return props.allowedDates.map(date => adapter.getYear(adapter.date(date))).includes(year)
+    }
+
+    function allowedMonths (month: number) {
+      if (!Array.isArray(props.allowedDates) || !props.allowedDates.length) return true
+      if (!allowedYears(year.value)) return false
+
+      return props.allowedDates
+        .filter(date => adapter.getYear(adapter.date(date)) === year.value)
+        .map(date => adapter.getMonth(adapter.date(date))).includes(month)
+    }
+
     // function onClickAppend () {
     //   inputMode.value = inputMode.value === 'calendar' ? 'keyboard' : 'calendar'
     // }
@@ -340,19 +354,21 @@ export const VDatePicker = genericComponent<new <
                       key="date-picker-months"
                       { ...datePickerMonthsProps }
                       v-model={ month.value }
-                      onUpdate:modelValue={ onUpdateMonth }
                       min={ minDate.value }
                       max={ maxDate.value }
                       year={ year.value }
+                      allowed-months={ allowedMonths }
+                      onUpdate:modelValue={ onUpdateMonth }
                     />
                   ) : viewMode.value === 'year' ? (
                     <VDatePickerYears
                       key="date-picker-years"
                       { ...datePickerYearsProps }
                       v-model={ year.value }
-                      onUpdate:modelValue={ onUpdateYear }
                       min={ minDate.value }
                       max={ maxDate.value }
+                      allowed-years={ allowedYears }
+                      onUpdate:modelValue={ onUpdateYear }
                     />
                   ) : (
                     <VDatePickerMonth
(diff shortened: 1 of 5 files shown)


ikushum:
As mentioned in the PR description, this works only when `allowed-dates` is an array. 

I'm open for suggestion for how we can implement this (without heavy computation) for cases when `allowed-dates` is a function.

ikushum:
@J-Sek, For the computation heavy changes, I created a [new PR](https://github.com/vuetifyjs/vuetify/pull/21470) that targets this branch

packages/vuetify/src/components/VDatePicker/VDatePickerMonths.tsx
```diff
@@ -76,6 +78,18 @@ export const VDatePickerMonths = genericComponent<VDatePickerMonthsSlots>()({
...
+        return props.allowedMonths.includes(month)
+      }
+
+      if (typeof props.allowedMonths === 'function') {
+        return props.allowedMonths(month)
+      }
+
+      return true
+    }
+
```
johnleider, line 92: This is a good example of how the other ones should look.

johnleider (commented):
Looks good

packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
```diff
@@ -194,6 +194,44 @@ export const VDatePicker = genericComponent<new <
...
+      if (Array.isArray(props.allowedDates) && props.allowedDates.length) {
+        if (!allowedYears(year.value)) return false
+
+        return props.allowedDates
+          .filter(date => adapter.getYear(adapter.date(date)) === year.value)
+          .map(date => adapter.getMonth(adapter.date(date))).includes(month)
+      }
+
+      return true
+    }
```
johnleider, line 233: This looks great. Is it worth making some of the above reusable to avoid duplication?

ikushum: made validation logic for function `allowedDates` reusable

packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
```diff
@@ -340,19 +380,21 @@ export const VDatePicker = genericComponent<new <
                       key="date-picker-months"
                       { ...datePickerMonthsProps }
                       v-model={ month.value }
-                      onUpdate:modelValue={ onUpdateMonth }
                       min={ minDate.value }
                       max={ maxDate.value }
                       year={ year.value }
+                      allowed-months={ allowedMonths }
```
KaelWD, line 386: Always use camelCase in jsx

KaelWD:
> up to 31 executions per month in the grid

For some reason this is calling allowedDates 2061 times instead of just 365

> up to 366 executions per year in the grid

&times;152 years is over 55000 calls, should we only do this if min and max are explicitly set?

ikushum:
> > up to 31 executions per month in the grid
> 
> For some reason this is calling allowedDates 2061 times instead of just 365
> 
> > up to 366 executions per year in the grid
> 
> ×152 years is over 55000 calls, should we only do this if min and max are explicitly set?

@KaelWD 

I just tested for worst case senario and `allowedMonths` makes `353` calls as expected.

image.

Maybe you mixed it up with calls from `allowedYears` too?

KaelWD:
My last commit fixed it. 

> Maybe you mixed it up with calls from allowedYears too?

Well yeah I don't care how many times a specific function calls it directly, only how many times it is invoked overall.

ikushum:
> ×152 years is over 55000 calls, should we only do this if min and max are explicitly set?

Sounds fair, should we mention this somewhere in the documentation?

johnleider (approved):
Approved pending added unit test.

PR #21470, opened 2025-05-22
feat(VDatePicker): Calculate disabled year and month 

ikushum:
fixes https://github.com/vuetifyjs/vuetify/issues/20465

## Description
Disable month and year buttons based on `allowed-dates` prop when:
- [x]  `allowed-dates` is array (done in #21466)
- [x]  `allowed-dates` is function 

‼️  This is a follow up PR for https://github.com/vuetifyjs/vuetify/pull/21466 and completes the second checkbox from list above. The changes in this PR are slightly computation heavy so wanted to have some discussion and see if the code is worth it.

If the `allowed-dates` prop is function then:
- Code inside `allowedYears` will execute approx 366 * 150 = **54900 times**
- Code inside `allowedMonths` will execute approx 31 * 12 = **372 times**

## Markup:

```vue
<template>
  <v-app>
    <v-container>
        <h2>Date picker</h2>
        <v-date-picker
          mode="month"
          hide-header
          :allowed-dates="allowedDates"
        ></v-date-picker>

        <h2>Date picker years</h2>
        <VDatePickerYears
          :allowed-years="[2025, 2026]"
          class="mb-10"
        />

        <h2>Date picker months</h2>
        <VDatePickerMonths
          :allowed-months="[2, 3]"
        />
    </v-container>
  </v-app>
</template>

<script setup>
const allowedDates = value => value > new Date()
</script>


```

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePicker.tsx
@@ -18,7 +18,7 @@ import { useProxiedModel } from '@/composables/proxiedModel'
 
 // Utilities
 import { computed, ref, shallowRef, toRef, watch } from 'vue'
-import { genericComponent, omit, propsFactory, useRender, wrapInArray } from '@/util'
+import { createRange, genericComponent, omit, propsFactory, useRender, wrapInArray } from '@/util'
 
 // Types
 import type { VPickerSlots } from '@/labs/VPicker/VPicker'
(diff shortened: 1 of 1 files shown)


J-Sek:
Yup. Worst case scenario is ~55k roundtrips, but also it takes < 300ms (with console log commented out), so it basically depends on complexity of that function... in addition to the devs being careless by forgetting to put `min` and `max`

```vue
<template>
  <h2>Select winter holiday with the most epic party</h2>
  <v-date-picker :allowed-dates="allowedDates" mode="month" hide-header />
</template>

<script setup lang="ts">
  const allowedDates = (v: Date) => {
    // console.count('allowedDates') // over 54500
    return v.getMonth() === 11 && v.getDate() > 21
  }
</script>
```

J-Sek (approved):

johnleider (approved):

PR #21478, opened 2025-05-24
docs(VMenu): more examples and fixes for non responsive examples

eesayas:
## Description
Additional Menu examples:
1. Absolute positioning within the activator
2. Close on content click
3. Persistent
4. Misc: Usage in a Nav Menu (utilizing `activator` for single instance)

Fixed non-responsive examples:
1. Usage
2. Transitions

## Markup:
N/A


eesayas:
Will fix double quotes

packages/docs/src/examples/v-menu/misc-nav-menu.vue
```diff
@@ -0,0 +1,69 @@
...
+        <v-app-bar-nav-icon></v-app-bar-nav-icon>
+      </template>
+
+      <v-app-bar-title>My Website</v-app-bar-title>
+
+      <v-btn
+        v-for="page in pages"
+        :key="page.name"
+        append-icon="mdi-chevron-down"
+        @mouseenter="register($event, page.name)"
```
J-Sek, line 12: Interesting approach. I tried to make it keyboard-friendly, but `@focus="register($event, page.name)"` is not enough... Worst case, we will make it work with pointer and somehow gracefully fallback for keyboard.

packages/docs/src/examples/v-menu/prop-absolute-positioning.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div class="d-flex justify-center">
+    <v-menu target="cursor">
```
J-Sek, line 3: When I see it (and imagine myself as regular developer) I immediately think about applying it as context menu... which annoyingly is not that straightforward. We could drop 2 examples here, anticipating typical usage patterns.
- [playground](https://play.vuetifyjs.com/#eNqVVU2P2jAQ/Suj9AAr5YMElgW0W7WqetxL1UMr4OA1TvCSxJbtsIsQ/71jJ+SjYiv1ksQzb2bezBvD+uxpRaOvUobHinkr79GwQubEsM+bEuBxx49Ac6L108bbBWnO3uG10oanp4Cy0jAF8hTMISPBfOO5EAw6BgUrKzBEZcxgIK2UFqr1I+JaBD4RaviRGKEQdwaphNRw6UFdOkrUrjMA8IJkDAP2xki9iiK6Ky19S+tVh1QUkUPoyAbiU4gDL7NQltnG6+fZM57tLcNkMhl63vjO7NEx/dvxhQps+93YDkOp2BGngDhHPBTlt5zTQz+i30hUd9KNIRoMu2lW7xkz7dAlCaYbr+WTxMinn/OGQPYRvCkirSzJAO4KEJw3UXAMUjd1CrwEKnKhNBZaHdjJGu2nMzaHaNAIVu110XC+yo9nOx08tvvQF/A/xdsJetDhq+xJd1u2m5J9IBcKhckHYvXYNyINtvmaEE9ix3LM4axtoVW77db+0x06r2NRmuCqFMU7JIqgIVcXGPJwhXOue5twCmKrCselsWf3RoN9B/XFuYrUk6BTa7BteNRUcWlAM1NJyEmZYU6DCV0QL6RQBs6g9yTPxdsPlsIFUiUKGKFYI4tB9kjPXfWnHm6cklyzuyGiHsgQt574MNkC0bAuq+KFKR/q9xaDu3DXJ0au697OYLjJ2QpGz7ay9UI88uFI8gqtMVz8j4FJB0z+CZx2wGkD3PZJ1felYzVS2cs4fpj5ECdTH5aTu1GT3XkWDz7M5z7MFkN7EqMxXi7x8XA/dMXLBK0LnFE8nzWuhkJalfirKUrorfEYmT6LSrPvdsXv4Fyn4imM3e67ZlozOFFqIzbhFKs9F2D4PcTV4rXoNQtpzrHKLx+un7+R2o3ERlXXvPaFj8eo3jvcMu/ie9PwPoxnnv1YhHPPd0z8wX9Sbdv+Ab2aC9c=)

J-Sek: The description could be

> There are 2 ways to achieve the context menu. You can either set `target` to `"cursor"` and bind right click using the `activator` slot. Alternatively, manually pass `x` and `y` coordinates from the click event.

packages/docs/src/examples/v-menu/misc-nav-menu.vue
```diff
@@ -0,0 +1,69 @@
...
+      >
+        {{ page.name }}
```
johnleider, line 16: prefer to use the **text** prop for regular text.

packages/docs/src/examples/v-menu/misc-nav-menu.vue
```diff
@@ -0,0 +1,69 @@
...
+      </v-btn>
+    </v-app-bar>
+
+    <v-menu :activator="activator" location="bottom">
+      <v-list :items="subMenu"></v-list>
+    </v-menu>
+
+    <v-main></v-main>
```
johnleider, line 24: Nit pick: Could do something like:

```html
<v-main>
  <!-- Main Content -->
</v-main>
```

packages/docs/src/examples/v-menu/misc-transition.vue
```diff
@@ -1,68 +1,35 @@
 <template>
-  <div class="d-flex justify-space-around">
-    <v-menu
-      transition="scale-transition"
-    >
+  <div class="d-flex justify-space-around flex-wrap ga-1">
+    <v-menu transition="scale-transition">
       <template v-slot:activator="{ props }">
```
johnleider, line 4: If we're already here, should update props to look like this:

```html
```

Since that's how we recommend users do it.

johnleider (changes requested):
Looking good overall. All examples need to have a <script setup> and a regular <script> for compositional

packages/docs/src/examples/v-menu/prop-absolute-positioning.vue
J-Sek: Actually... `target="cursor"` is having some issues when used for context menu:
- secondary click on the card does not hide it
- quickly opening and closing (click outside the card) makes closing transition include unnecessary movement to the left bottom

Unless we can resolve those somehow, I would probably leave the `prop-absolute-positioning.vue` as-is while supplementing context menu example (with direct `x` & `y` manipulation) as a second piece.

Issue #20825, opened 2025-01-03
[Feature Request]  Request to improve the color transition when switching the variant property of the VInput component from solo-filled to solo


ahas:
### Problem to solve
[Example](https://play.vuetifyjs.com/#eNptkMFuwyAQRH9ly8WtVNt3y2nVS3+gx7oHBIuChAHB2moV5d+7Nk3tRLmg3WFmBO/zJHJS7VuMzTyh6ERPOEYnCV8GD9DPtYxxHddFBU/Sekx/0ioSflNtLDp9EQHmegwaXWeCmjLqwyBsfi/zIDZbN8tkpaf9PbxClYMLFXRl4G7nUFdXQcNv5NTD3Vr+gMJjcBoTez5QJnVsmmZvSWHymlP/gkafLf2wX4UxSkWbu70AaG8JLEoB1Lc7cLxmlWwkyEhT4WfHGBLBCRIaOINJYYSKmVeLHYB7M8EG4bD4Ho10GZ+W9tLHTeL87Cfn1uPrF15KjvM=)

In the example, when switching the variant property from solo-filled to solo, a black flickering issue occurs. This issue was not present in Vuetify 2.

### Proposed solution
[Vuetify 2 Implementation](https://codepen.io/ahas/pen/WbeXwGd)
I hope the black flickering issue can be resolved so that it behaves like it did in Vuetify 2, without any flickering.

J-Sek:
CSS transition timing is misaligned with `isFocused` propagation. It won't work even if we force `transition: all...` on all elements.  Just avoid this antipattern.

PR #21482, opened 2025-05-25
feat(VTextField): improve color transition on change variant #20825 [WIP]

jcjp:
## Description

- Resolves: https://github.com/vuetifyjs/vuetify/issues/20825

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        v-model:focused="isFocused"
        :variant="isFocused ? 'solo' : 'solo-filled'"
        :flat="!isFocused"
        placeholder="Search..."
        rounded
        density="compact"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const isFocused = ref(false)
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VField/VField.sass
+++ b/packages/vuetify/src/components/VField/VField.sass
@@ -493,7 +493,7 @@
     .v-field__overlay
       background-color: currentColor
       opacity: $field-overlay-filled-opacity
-      transition: opacity $field-subtle-transition-timing
+      transition: none
 
     @media (hover: hover)
       &:hover .v-field__overlay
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VField/VField.sass
```diff
@@ -493,7 +493,7 @@
     .v-field__overlay
       background-color: currentColor
       opacity: $field-overlay-filled-opacity
-      transition: opacity $field-subtle-transition-timing
+      transition: none
```
jcjp, line 496: Removing the transition fixes the background flicker on change variant, but is this the correct solution? What do you guys think?

jcjp: Yeah I agree, maybe he could just overwrite the transition on his side what do you think?

PR #21487, opened 2025-05-26
docs(v-data-table): prevent object reference mutation in add and save methods

mhdna:
## Description
Instead of assigning the new `record` value in the add and save methods to the `DEFAULT_RECORD` reference, it must be assigned to its value, otherwise you'll run into object reference mutation, so when you add another record you'll be mutating the old object reference.


MajesticPotatoe:
superseded by #21737

PR #21498, opened 2025-05-28
chore(api-generator): add public composables to web-types

johnleider:
## Description
I'm not sure if it makes sense to have composables in web-types.json, but it's what we are currently using for MCP data fetching. This would give access to things like useDisplay.


johnleider:
bugbot run

johnleider:
bugbot run

Issue #18846, opened 2023-12-10
[Bug Report][3.4.6] Disabling utility classes breaks some components blueprints and styling


mcfarljw:
### Environment
**Vuetify Version:** 3.4.6
**Vue Version:** 3.3.11
**Browsers:** Chrome 120.0.0.0
**OS:** Mac OS 10.15.7

![CleanShot 2023-12-10 at 15 17 05@2x](https://github.com/vuetifyjs/vuetify/assets/4167341/425a94a7-a55b-4f55-a7d5-794f2fbc9ce6)

### Steps to reproduce
1. Clone https://github.com/mcfarljw/vuetify-disabled-class-utilities
2. Run `npm i`
3. Run `npm run dev`

### Expected Behavior
Disabling utility classes doesn't break the default Vuetify components.

- `VBtn` should be rounded with a cursor on hover
- `VTab` should use the correct colors with a cursor when hovering and be rounded

### Actual Behavior
Disabling utility classes cases several components to not be styled correctly.

- `VBtn` is not rounded without a cursor on hover
- `VTab` is not the correct color when hovering and is not rounded

### Reproduction Link
[https://github.com/mcfarljw/vuetify-disabled-class-utilities](https://github.com/mcfarljw/vuetify-disabled-class-utilities)

### Other comments
In general I've been having a lot of trouble following the documentation for SASS Variables (https://vuetifyjs.com/en/features/sass-variables). I'm trying to disable the utility classes so that I can use TailwindCSS or UnoCSS without conflicts but in this example I'm just trying to disable the utility classes without breaking the default styling based on the selected blueprint.

mcfarljw:
@kingyue737 thanks for linking the discussion!

I realize it might not be easy/feasible to do but it would be nice if all components came bundled with required stylings based on the blueprint being used. As is mentioned in the linked thread TailwindCSS (or similar libraries like UnoCSS and PandaCSS) have become somewhat of a standard in terms of usage. They also come with VSCode extensions that provide a better developer experience. It seems like component libraries are embracing this like https://tailwind.primevue.org/ and which give developers the ability to mix and match without consequences.

Still love Vuetify more than other component libraries but at times it feels like utility classes are limited slightly different subset of more comprehensive existing styling solutions.

kingyue737:
I have the same feeling. Using CSS utility classes is the worst DX I've experienced in Vuetify. No intellisense. I need to check the document each time I want to add a class. Not util at all.

[...]

PR #21500, opened 2025-05-29
refactor: grid system overhaul 

### LINKED ISSUE #8611: [Bug Report] New Grid System Negative Margins
### Environment
**Vuetify Version:** 2.0.7
**Vue Version:** 2.6.10
**OS:** Windows 10

### Steps to reproduce
Using v-row creates a negative right margin that causes overflows / horizontal scrolling on mobile.

Is there any solution to get rid of this instead of adding ma-0 to every row? Is this the expected functionality?

Thank you!

### Expected Behavior
Should not cause any problems with overflow / horizontal scrolling.

### Actual Behavior
Causes issues with overflows / horizontal scrolling, especially on mobile.

### Reproduction Link
<a href="https://codepen.io/VeronicaGavrila/pen/mdbEQLM" target="_blank">https://codepen.io/VeronicaGavrila/pen/mdbEQLM</a>

AndreyYolkin:
## Description

Resolves #8611, related to #18846

This PR:
1. Swaps VCol/VRow spacing system to use gap instead of negative margin hack
2. Generates prefixed set of grid utilities to be independent from utility classes
3. Deprecates `dense` and `no-gutters` props in favour of widely adopted `density` prop

Docs are also updated (both examples and app itself)

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-container>
        <v-row density="compact">
          <v-col cols="12">
            <v-sheet class="pa-2">
              v-col-12
            </v-sheet>
          </v-col>
        </v-row>
        <v-row density="comfortable">
          <v-col cols="6">
            <v-sheet class="pa-2">
              v-col-6
            </v-sheet>
          </v-col>
        </v-row>
      </v-container>
    </v-main>
  </v-app>
</template>
```


KaelWD:
> Swaps VCol/VRow spacing system to use gap instead of negative margin hack

This will break a lot of layouts which is why I put the issue on a major milestone not 3.9.0. Still unsure what the impact of this will be, we need to check a bunch of real examples and see if it's possible to make the required changes automatically.

packages/vuetify/src/components/VGrid/VGrid.sass
```diff
@@ -19,28 +21,19 @@
...
+      @include make-row(settings.$form-grid-gutter)
 
-      > .v-col,
-      > [class*="v-col-"]
-        padding: settings.$form-grid-gutter * .5
-
-    // Remove the negative margin from default .v-row, then the horizontal padding
-    // from all immediate children columns (to prevent runaway style inheritance).
-    &.v-row--no-gutters
-      margin: 0
```
KaelWD, line 28: This used to override line 26 (`.v-row + .v-row`)

packages/vuetify/src/components/VGrid/VCol.ts
```diff
@@ -133,8 +133,8 @@ export const VCol = genericComponent()({
         'v-col': !hasColClasses || !props.cols,
         [`v-col-${props.cols}`]: props.cols,
         [`offset-${props.offset}`]: props.offset,
```
KaelWD: Still using the global offset util classes

AndreyYolkin: We don't have offset util classes, however it's good to rename to avoid further confusion

packages/docs/src/examples/grid/misc-equal-width-columns.vue
```diff
@@ -1,28 +1,28 @@
 <template>
   <v-container class="bg-surface-variant">
-    <v-row no-gutters>
+    <v-row cols="[16, 8]">
```
KaelWD: What's this?

J-Sek: typo, should be `:gap=`

J-Sek:
I am going to introduce `--v-col-size` and `--v-row-columns`. This should make it significantly easier to "opt-out" back to negative margins making it much easier for people to migrate gradually.

J-Sek:
- [x] fix for auto broke it `.v-row > .v-col` overrides `.v-row > [class*='v-col-']`, maybe we will need to restore `make-col`, as CSS ifs nor style queries are available yet
  - [x] once reverted, rework migration (CSS for restoring legacy grid)
- [x] migrate docs (ready to append on a separate branch, chilling until we fix ☝🏼 )

KaelWD:
I cleaned it up a bit then broke it more, `[class*='v-col-']` still matches `.v-col-md-6` so columns with no xs size are full-width instead of 0+grow

- [x] proper test suite is needed, checking breakpoints in playground is tedious

KaelWD:
```html
<template>
  <v-container>
    <v-row>
      <v-col md="4">xs: grow, md: 4</v-col>
      <v-col>grow</v-col>
    </v-row>
    <v-row>
      <v-col cols="4" md>xs: 4, md: grow</v-col>
      <v-col>grow</v-col>
    </v-row>
    <v-row>
      <v-col cols="auto" md="4">xs: auto, md: 4</v-col>
      <v-col>grow</v-col>
    </v-row>
    <v-row>
      <v-col cols="4" md="auto">xs: 4, md: auto</v-col>
      <v-col>grow</v-col>
    </v-row>
    <v-row>
      <v-col cols="4" md="12">xs: 4, md: 12</v-col>
    </v-row>
    <v-row>
      <v-col cols="4" offset-md="4">offset-md: 4</v-col>
      <v-col>grow</v-col>
    </v-row>
  </v-container>
</template>

<style>
.v-col {
  outline: 1px solid red;
}
</style>
```

J-Sek:
We can push it even more into parsing `cols="1 of 5"` or `cols="1/5"`.

Following example already works, but is too much noise.

```html
<v-row>
  <v-col cols="1" style="--v-row-columns: 5">1/5</v-col>
  <v-col cols="1" style="--v-row-columns: 5">1/5</v-col>
  <v-col cols="1" style="--v-row-columns: 2">1/2</v-col>
  <v-col cols="1" style="--v-row-columns: 10">1/10</v-col>
</v-row>
```

~~It can be a follow-up PR.~~ #22572

J-Sek:
The last thing that bothers me is that "auto" all over the docs examples instead of "grow". It should be fixed in `master`.

packages/docs/src/pages/en/getting-started/upgrade-guide.md
```diff
@@ -381,3 +381,100 @@ createVuetify({
...
+#### Prop changes on VRow
+
+| Previous                          | New                                              |
+|-----------------------------------|--------------------------------------------------|
+| `dense`                           | `density="comfortable"` or `gap="8"`             |
+| `no-gutters`                      | `density="compact"` or `gap="0"`                 |
+| `align` prop on VRow              | use utility class (e.g., `align-start`)          |
+| `justify` prop on VRow            | use utility class (e.g., `justify-center`)       |
+| `align-content` prop on VRow      | use utility class (e.g., `align-content-center`) |
+| `align-sm`, `justify-md`, etc.    | use responsive utility classes                   |
```
KaelWD: Why were these props removed?

J-Sek: I have hinted this idea to Andrey in December.

(from internal chat)
> Those convenience attributes were already a problem if you setup Vuetify with TailwindCSS (or UnoCSS). I don't consider attributify a valid solution here because it is global (cannot be scoped to specific components). Forcing migration in v4.x will also make it easier to opt-in to official UnoCSS support in v5. Migration is straightforward and can be handled with ESLint plugin (once we update it).

KaelWD: If they're being deprecated there's no reason to remove them too, dense and noGutters still work just with a warning.

> I don't consider attributify a valid solution here because it is global

unocss-preset-vuetify should implement it for those components: https://github.com/vuetifyjs/unocss-preset-vuetify/issues/4

J-Sek: I have some doubts if UnoCSS preset would be able to scope those to VRow/VCol which seems crucial for `align` prop especially.

Anyway, attributes are restored as deprecated

PR #21504, opened 2025-05-30
feat(VNavigationDrawer): add force-expand to override hover expand

### LINKED ISSUE #20591: [Feature Request] VNavigationDrawer force expand
### Problem to solve
A way to force the VNavigationDrawer to be expanded even when the user isn't hovering.
User case - allow user to edit/rename items in the VNavigationDrawer without shifting the main UI, disabling rail mode would visibly shift the main UI which we want to avoid.

### Proposed solution
add a forceExpand value that when set it overrides the isHovering value.
and only works when expandOnHover is enabled.

eesayas:
## Description

Resolves #20591 

Allows the drawer to remain visually expanded even when the user is not hovering, without disabling `rail` mode. 
Only applies when `expand-on-hover` is enabled. Useful for cases like renaming items where a stable expanded view is needed.

## Markup:
```vue
<template>
  <v-app>
    <v-main>
      <v-navigation-drawer expand-on-hover force-expand rail>
        <v-list density="compact" nav>
          <v-list-item></v-list-item>
          <v-list-item></v-list-item>
          <v-list-item></v-list-item>
        </v-list>
      </v-navigation-drawer>
    </v-main>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
+++ b/packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
@@ -52,6 +52,7 @@ export const makeVNavigationDrawerProps = propsFactory({
   disableRouteWatcher: Boolean,
   expandOnHover: Boolean,
   floating: Boolean,
+  forceExpand: Boolean,
   modelValue: {
     type: Boolean as PropType<boolean | null>,
     default: null,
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VNavigationDrawer/VNavigationDrawer.tsx
```diff
@@ -52,6 +52,10 @@ export const makeVNavigationDrawerProps = propsFactory({
   disableRouteWatcher: Boolean,
   expandOnHover: Boolean,
   floating: Boolean,
+  forceExpand: {
+    type: Boolean,
+    default: false,
+  },
```
johnleider, line 58: Booleans are already default false.

johnleider (changes requested):

Issue #20837, opened 2025-01-05
[Feature Request] Dynamically set size of badge


Tauromachian:
### Problem to solve
- Change size without recurring to SCSS or the `:deep` helper
- Dynamically set the size of the badge for complex scenarios.
- Set a default size and overwrite it on demand (again without having to use the :deep helper)

### Proposed solution
Make a `size` prop that accepts a number and (string with number). This number will then be applied as the height and width of the badge with the `px` unit.

jcjp:
Added a new props `width` and `height` to resize the `VBadge`.

PR #21509, opened 2025-05-31
feat(VBadge): add size width and height props to VBadge (#28037)

jcjp:
## Description

- Resolves: https://github.com/vuetifyjs/vuetify/issues/20837
- Adds new props `width` and `height` using our `DimensionProps`

## Markup:

```vue
<template>
  <v-container>
    <v-badge width="28px" height="30px" content="1" />
  </v-container>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VBadge/VBadge.tsx
+++ b/packages/vuetify/src/components/VBadge/VBadge.tsx
@@ -51,7 +51,7 @@ export const makeVBadgeProps = propsFactory({
   ...makeTagProps(),
   ...makeThemeProps(),
   ...makeTransitionProps({ transition: 'scale-rotate-transition' }),
-  ...(makeDimensionProps() as Omit<ReturnType<typeof makeDimensionProps>, 'maxHeight' | 'maxWidth' | 'minHeight' | 'minWidth'>),
+  ...makeDimensionProps(),
 }, 'VBadge')
 
 export const VBadge = genericComponent<VBadgeSlots>()({
@@ -80,10 +80,7 @@ export const VBadge = genericComponent<VBadgeSlots>()({
       )
     })
 
-    const { dimensionStyles } = useDimension({
-      height: props.height,
-      width: props.width,
-    })
+    const { dimensionStyles } = useDimension(props)
 
     useRender(() => {
       const value = Number(props.content)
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VBadge/VBadge.tsx
```diff
@@ -50,6 +51,7 @@ export const makeVBadgeProps = propsFactory({
   ...makeTagProps(),
   ...makeThemeProps(),
   ...makeTransitionProps({ transition: 'scale-rotate-transition' }),
+  ...(makeDimensionProps() as Omit<ReturnType<typeof makeDimensionProps>, 'maxHeight' | 'maxWidth' | 'minHeight' | 'minWidth'>),
```
J-Sek: `Pick<...., 'height' | 'width'>` and later `useDimension(pick(props, ['height', 'width'])` might be more future proof.
I would even simplify the composable call to `useDimension(props)`, otherwise it suggests VBadge has some `min*` or `max*` props that would be in conflict, which is not the case, so it's just cognitive load.

J-Sek (dismissed):

J-Sek:
Is it OK that these are not reactive? Ofc, we could patch it later after learning about some real-life use cases proving it matters.

<details>
<summary>Playground</summary>

```vue
<template>
  <v-container max-width="400">
    <v-slider
      v-model="height"
      label="height"
      max="100"
      min="0"
      step="1"
    />
    <v-slider
      v-model="width"
      label="width"
      max="100"
      min="0"
      step="1"
    />
  </v-container>
  <v-container max-width="400">
    <v-badge :key="height + width" :height="height" :width="width" content="123" />
  </v-container>
</template>

<script setup lang="ts">
  import { ref } from 'vue'
  const height = ref<number | null>(null)
  const width = ref<number | null>(null)
</script>
```

</details>

packages/vuetify/src/components/VBadge/VBadge.tsx
jcjp: Done thanks for the review, yeah I agree with your idea makes more sense to just call `useDimension(props)` and not omit any dimension props.

jcjp:
> Is it OK that these are not reactive? Ofc, we could patch it later after learning about some real-life use cases proving it matters.

Do you mean to say that the `width` and `height` are not reactive?

J-Sek:
> Do you mean to say that the `width` and `height` are not reactive?

The changes did not seem to be reflected unless I had `:key=` to reset the badge. I was using the playground from my previous comment.

Your last commit fixed it by passing `props` straight to the composable without extracting width and height. :)

Issue #10890, opened 2020-03-20
[Feature Request] Bring masks back


andreixk:
### Problem to solve
Vuetify 1.5 used to have this functionality implemented properly. Current proposition of using 3rd party plugins (especially vue-the-mask) is useless. Those plugins are clumsy, incomplete and generally require a lot of re-work to integrate. each v-text-field becomes a 10-15 line monstrosity, forms are a nightmare to create and even worse to debug.

### Proposed solution
Simple: bring back the masks that were in v1.5

Mikilll94:
Yes, It would be great to bring mask functionality again to Vuetify 2.x

About [vue-the-mask](https://github.com/vuejs-tips/vue-the-mask) I will add also that:
- This plugin is outdated. Latest version was released on 10 Oct 2017. There are many bugs in this plugin which were not fixed.
Just look here:
https://github.com/vuejs-tips/vue-the-mask/issues/137
https://github.com/vuejs-tips/vue-the-mask/issues/112

- This plugin has no MIT license, so theoretically you cannot use it.
https://github.com/vuejs-tips/vue-the-mask/issues/111

MajesticPotatoe:
We have plans to bring this back for v3. The likelihood of it making its way into v2 is undetermined yet slim.

rpaggi:
What is the way to apply masks in the current version?

andreixk:
> What is the way to apply masks in the current version?

make your own plugin or use another framework. there's no good solution for vuetify

RonaldJerez:
I am the author of [vue-input-facade](https://github.com/RonaldJerez/vue-input-facade).  It started as a fork of vue-the-mask to fix some of the outstanding issues, but it has evolved while still maintaining a similar API.   I'll be willing to help out if you guys want to include it in vuetify.  It is currently been used in a high visible project at my current employment (Fortune 500 company), so I'll be maintaining it for quite some time.

gogangipark:
> I am the author of [vue-input-facade](https://github.com/RonaldJerez/vue-input-facade). It started as a fork of vue-the-mask to fix some of the outstanding issues, but it has evolved while still maintaining a similar API. I'll be willing to help out if you guys want to include it in vuetify. It is currently been used in a high visible project at my current employment (Fortune 500 company), so I'll be maintaining it for quite some time.


It will be very helpful to use vuetify.

SkyaTura:
These is any perspective on this topic? Or a recommended workaround

[...]

Issue #9028, opened 2019-09-19
[Bug Report] text-fields mask


SP86:
### Environment
**Vuetify Version:** 1.5.18
**Vue Version:** 2.6.10
**Browsers:** Chrome 76.0.3809.132
**OS:** Windows 10

### Steps to reproduce
1. Set in data 
mask: '+38(###)-###-##-##',

2.Click to text field and and start entering number 097

### Expected Behavior
That the cursor snaps into place 3 characters

### Actual Behavior
The cursor is in the middle of the entered characters

### Reproduction Link
<a href="https://codepen.io/heraqles/pen/oMNZzX" target="_blank">https://codepen.io/heraqles/pen/oMNZzX</a>

### Other comments
This is a video link for a demonstration.
https://itpromotion.eu/ice_video_20190919-161919.webm

Teellcory:
Are you trying to modify a time example to do phone numbers? As you can see in the pattern code: 7 doesn't work because there's no such time as 9:70AM.

alexandre-le-borgne:
The bug is not due to the change of mask but to the mask in itself:

Works :'+++(###)-#####' (the cursor is placed after the first #)
Does not works :'+38(###)-#####' (the cursor is misplaced after the first '3' when typing the first character)

J-Sek:
Does not seem to occur with VMaskInput ([demo](https://play.vuetifyjs.com/#eNqFkWFrgzAQhv/KcX6wZWqoZdCVdLDfMfch2HOTGROS2C+l/30XnUzLoJCE3JM3d5c371f0rhZv1haXgfCIMpC2nQr0WvUA8pIra8ftGNSmD6rtyf2iEWrlv/O2t0OYIQBTc6buVKH2n7sK/06imvHT/rBJkmSb88KTx1LEHdT0ZbozuQda8X8n6wbKCue6SXKM1+8qzHTOJsX9WyOZrJBiYRGHvnatDeApDJNTrbbGBbiCowZu0DijIWV30ygH4Lw+QLQFTlGySdPtipdLLsWUnzPjLcN98VzsDhg3L0WJWaM6T9nqDyf28QPmWIuj))

Issue #4484, opened 2018-07-04
[Feature Request] Hex Mask


artroup:
### Problem to solve
Currently for a text field that are predefined masks, and the ability to create a custom mask. However the custom mask feature still does not allow for me to create a Hex based value. I would like to see a hex based mask included for the custom mask feature. 

This would allow me to create a mask for hex based fields such as Mac Address.

### Proposed solution
Add a hex based mask value, to the already custom entries, or create a predefined mask for a Mac Address.

jacekkarczmarczyk:
masks have been removed in 2.0

J-Sek:
`VMaskInput` with custom token matching `/[0-9a-f]/i` looks like a perfect solution:
- [playground](https://play.vuetifyjs.com/#eNpdksFugzAMhl/FyoVOKtBtmrQiWqnqpZcdd2p7yMCoqDSJEpML4t3ntGNAUSTsz47584djJ5wt0p0xiW9RZCInvJlGEm5PCiD3sTTmHt6TQiuStUL7hxgai1ufQdeBh77P05D/F318k+4a18q0NEAAprrEZnMS/iRG3MifO/za7WFXlhadm5azMIrL4TXlrLbAi25KtFxdrbL5mrbKlnSh+XxIyL26qmaD0LraESqKL7Wa6LVIrVX3o2AZe9m0OBTTwZv02ZxAHt7l6cRTTl1ha0PgeOrD2vpmtCXo+EMV9FBZfYOIryMK7QA81xG7uwn1RRS9THEQxZVuVBtIBtHhkM1XtBx7SF9ROb61EQFviZ4IeyKJ0KoM0uMqXsu4Oqf1ZE54WIdHS1lQuAWfkP42bOVeOlyw1LGxn2zsh5CDPH04wl6Ifinek4/k9VOEYJ28iWUlG4fL2W/6YOdfyD7Lhw==)

Issue #3618, opened 2018-03-19
[Bug Report] Cursor Placement when using masks


matthanson12:
### Versions and Environment
**Vuetify:** 1.0.8
**Vue:** 2.5.16
**Browsers:** Chrome 64.0.3282.186
**OS:** Windows 10

![image](https://user-images.githubusercontent.com/6919027/37614410-11e5285e-2b81-11e8-90d9-dd770adbaf9f.png)

### Steps to reproduce
Type one extra number in the codepen to finish the month. The flashing cursor appears on the left of the static character when the next input will appear on the right of the static. 

### Expected Behavior
The cursor should appear where the user is about to type

### Actual Behavior
The cursor appears in the previous 'section' 

### Reproduction Link
https://codepen.io/anon/pen/JLEYGL?&editors=101

SOFSPEEL:
Agree with expected behavior but would also add that the cursor not being in the correct position is even more disorienting when using type="'password'"

![image](https://user-images.githubusercontent.com/1612646/42003344-e6c377b4-7a1f-11e8-80ff-ce4c1e3c3fff.png)

Andrey7287:
I have same kind of bug using phone mask with preset digit "+7 (###) ###-##-##" .
[https://codepen.io/Andrey7287/pen/aRYjww?&editors=101](url)

If works greate when first input digit is 7, so if you type any other digit cursor is mooving before it !

jacekkarczmarczyk:
masks have been removed in 2.0

PR #21519, opened 2025-06-03
feat(VMaskInput): Create new component

ikushum:
fixes https://github.com/vuetifyjs/vuetify/issues/10890
fixes https://github.com/vuetifyjs/vuetify/issues/3618
fixes https://github.com/vuetifyjs/vuetify/issues/9028
fixes https://github.com/vuetifyjs/vuetify/issues/4484

## Description
Created a new component `v-mask-input`.
This PR has been continued from https://github.com/vuetifyjs/vuetify/pull/19207

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-card>
        <v-card-title>
          <h1>Mask</h1>
        </v-card-title>

        <v-card-text>
          <v-text-field v-model="mask" label="Mask"></v-text-field>
        </v-card-text>
      </v-card>

      <br />
      <br />

      <v-card>
        <v-card-title>
          <h1>prop: returnMaskedValue = false</h1>
          <h3>v-model: {{ value }}</h3>
        </v-card-title>

        <v-card-text>
          <VMaskInput
            v-model="value"
            :mask="mask"
            label="Value"
          />
        </v-card-text>
      </v-card>

      <br />
      <br />

      <v-card>
        <v-card-title>
        <h1>prop: returnMaskedValue = true</h1>
        <h3>v-model: {{ value1 }}</h3>
        </v-card-title>

        <v-card-text>
          <VMaskInput
            returnMaskedValue
            v-model="value1"
            :mask="mask"
            label="Value"
          />
        </v-card-text>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const mask = ref('(###) ### - ####')
  const value = ref('4443331111')
  const value1 = ref('4443331111')
</script>

```


packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+          { ...attrs }
+          v-model={ model.value }
+          ref={ vTextFieldRef }
+        >
+          {{ ...slots }}
+        </VTextField>
+      </div>
+    ))
+
+    return {}
```
J-Sek: `return forwardRefs({}, vTextFieldRef)` to expose `reset()`, `validate()`, etc.

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+    onMounted(() => {
+      if (props.returnMaskedValue) {
+        emit('update:modelValue', model.value)
+      }
+    })
+
+    useRender(() => (
+      <div>
+        <VTextField
+          { ...props }
```
J-Sek: `const textFieldProps = VTextField.filterProps(props)` like in VSelect, to avoid passing local props as attrs

packages/docs/src/examples/v-mask-input/usage.vue
```diff
@@ -0,0 +1,44 @@
...
+
+<script setup>
+  const name = 'v-mask-input'
+  const model = ref('default')
+  const options = ['outlined', 'underlined', 'solo', 'solo-filled', 'solo-inverted']
+  const clear = ref(false)
+  const counter = ref(false)
+  const disabled = ref(false)
+  const props = computed(() => {
+    return {
```
J-Sek, line 28: It lacks `mask` here, so acts more like typical text input. Probably oversight.

packages/docs/src/examples/v-mask-input/custom-token.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div>
+    <v-mask-input
+      v-model="model"
+      :mask="{
+        mask: '&&&&-&&&&',
+        tokens: {
+          '&': {
+            pattern: /[0-9]/,
```
J-Sek: Uppercase with digits makes no sense.
Maybe include `a-f` to make a Git hash input? (with uppercase, 7-character, no separators)

packages/docs/src/pages/en/components/mask-inputs.md
```diff
@@ -0,0 +1,95 @@
...
+export default createVuetify({
+  components: {
+    VMaskInput,
+  },
+})
+```
+
+## Usage
+
+At its core, the `v-mask-input` component is a basic container that extends [v-text-field](/components/text-fields).
```
J-Sek: - (suggestion) "_basic container that extends_" » "_wrapper around_"
- we could include few words about intended purpose. It should probably include "data collection" keyword

J-Sek:
It would be great if it could trim pasted values or extract characters ignoring everything else (like VNumberInput). Probably out of scope for this particular PR as it does not touch `useMask`.

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+    onMounted(() => {
+      if (props.returnMaskedValue) {
+        emit('update:modelValue', model.value)
+      }
+    })
+
+    useRender(() => (
+      <div>
```
J-Sek: Can we drop this `<div>`?

packages/docs/src/examples/v-mask-input/custom-token.vue
KaelWD: I've never seen anyone write git hashes in uppercase.

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
KaelWD: Doesn't need an extra div

packages/docs/src/examples/v-mask-input/custom-token.vue
ikushum: Added a better example with a License Plate

packages/docs/src/pages/en/components/mask-inputs.md
ikushum: Intended purpose is already described right at the top
image

J-Sek (dismissed):

J-Sek:
Nice to have (maybe follow-up, possible in user-land):

image

Quick draft: [8d46f5e](https://github.com/vuetifyjs/vuetify/commit/8d46f5ec4830f8d2074cf8200d518fcb7375546d)
Note: I think it would make sense to move it inside composable and replace all tokens (default and custom) with `_`

packages/vuetify/src/composables/mask.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  function convert (mask: string, char: string): string {
+    const item = tokens.value[mask]
+    return item.convert ? item.convert(char) : char
+  }
+
+  function maskText (text: string | null | undefined): string {
+    if (text == null) return ''
+
+    if (!mask.value.length || !text.length) return text
```
J-Sek: I would trim it before this line to have better support for copy pasting. Character extraction (like in VNumberInput) can be considered as a follow-up or we could wait for feedback from users.

J-Sek: I have reworked it a bit here so it can be greedy [3cc88e5](https://github.com/vuetifyjs/vuetify/commit/3cc88e5e1284311f74ed0e3cc378d63213b2c1a7). It does not need explicit `trim` in this version.

J-Sek (changes requested):

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+      }
+    })
+
+    useRender(() => {
+      const textFieldProps = VTextField.filterProps(props)
+
+      return (
+        <VTextField
+          { ...textFieldProps }
+          { ...attrs }
```
johnleider: Does this need to be done? It's already the root element.

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+          updateRange()
+          return returnMaskedValue.value ? maskText(val) : val
+        }
+        return val
+      },
+    )
+
+    onMounted(() => {
+      if (props.returnMaskedValue) {
+        emit('update:modelValue', model.value)
```
johnleider, line 65: Any reason for this to be in onMounted instead of onBeforeMounted? What is the exact purpose of it?

johnleider (dismissed):
This looks amazing. Great spread of information, unit tests, and clean code. I think the only area that could use some love would be more examples, like a lot. Maybe @MajesticPotatoe or @eesayas could chip in.

packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
ikushum: I've replaced by `onBeforeMount`.

The purpose of this is, if `returnMaskedValue` prop is passed with un-masked model-value then this will ensure, the model-value becomes masked at all times.

For example, in the following case, value of `phone` which is initially unmasked `4443331111` will be changed to masked `(444) 333 - 1111`  immediately. 
```vue
<template>
  <v-mask-input
    v-model="phone"
    return-masked-value
    mask="(###) ### - ####"
  />
</template>

<script setup>
  import { ref } from 'vue'
  const phone = ref('4443331111')
</script>
```

packages/vuetify/src/composables/mask.ts
ikushum: Implemented trim() but character extraction logic looks promising, we could add in a follow up PR

johnleider (approved):
Pending the unit tests being fixed (if any of them are even related to this P.R.)

PR #21450, opened 2025-05-21
fix(VDateInput): use common date range parser

ikushum:
Fixes #21448

## Description
Use common function to create date range.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-input label="Date input" multiple="range" v-model="val" />
    </v-container>
  </v-app>
</template>

<script>
  export default {
    data: () => ({
      val: [],
    }),
    watch: {
      val(newVal) {
        console.log('VALUE CHANGED', newVal)
      },
    },
  }
</script>

```

PR #21525, opened 2025-06-04
fix(date): expect adapter without internal extensions

J-Sek:
## Description

Recently extended `DateInstance` is affecting `VuetifyOptions` and now TS complains about missing `createDateRange` when using custom adapters.

- fixed by introducing `BaseDateInstance`

## Markup:

```vue
// vuetify.ts
import LuxonAdapter from '@date-io/luxon'

export const vuetify = createVuetify({
  date: {
    adapter: LuxonAdapter, // this line should be accepted without TS errors
  },
  theme: {
    defaultTheme: 'light',
  },
})
```

### DIFF
--- a/packages/vuetify/src/composables/date/date.ts
+++ b/packages/vuetify/src/composables/date/date.ts
@@ -13,8 +13,11 @@ import type { LocaleInstance } from '@/composables/locale'
 // Adapters
 import { VuetifyDateAdapter } from './adapters/vuetify'
 
-export interface DateInstance extends DateModule.InternalAdapter {
+export interface BaseDateInstance extends DateModule.InternalAdapter {
   locale?: any
+}
+
+export interface DateInstance extends BaseDateInstance {
   createDateRange: (start: unknown, stop?: unknown) => unknown[]
 }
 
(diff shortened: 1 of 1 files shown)


J-Sek:
Should we update any of those places as well?
image

J-Sek:
> Should we rather just make `createDateRange` optional?

Easy way out from TS error, yes. However the type will suggest that developer can define `createDateRange` on their end. We always override it, so there is no point.

ikushum:
@J-Sek I agree. 

In that case, I think we should remove `createDateRange` from the `DateInstance`. It was supposed to be an internal util and not necessary meant to be made available for developers.

Only reason it was kept inside DateInstance was [this](https://github.com/vuetifyjs/vuetify/pull/21450#issuecomment-2910319860)

How about making it a normal util function [like before](https://github.com/vuetifyjs/vuetify/blob/c59a661590d6b3f46bb3aa5f989343d086a8b0b4/packages/vuetify/src/composables/date/date.ts#L120)

J-Sek:
I would vote for making it an internal utility function. When it is exposed outside, we will soon be asked to fix all the edge cases that do not matter for internal uses (i.e. `createDateRange(start, start)` returing 2 items instead of 1).

@johnleider , any objections?

johnleider:
> I would vote for making it an internal utility function. When it is exposed outside, we will soon be asked to fix all the edge cases that do not matter for internal uses (i.e. `createDateRange(start, start)` returing 2 items instead of 1).
> 
> @johnleider , any objections?

None

ikushum:
Approved the alternative one - https://github.com/vuetifyjs/vuetify/pull/21531

PR #21530, opened 2025-06-06
feat: respect prefers-reduced-motion

### LINKED ISSUE #19622: [Bug Report][3.5.15] Transitions and animations are not disabled when the user disables them
### Environment
**Vuetify Version:** 3.5.15
**Vue Version:** 3.4.22
**Browsers:** Firefox 124.0
**OS:** Windows 10

### Steps to reproduce
1. Disable animations in your OS and/or browser
2. Cause any animation or transition to happen. I left the playground contents bog standard, because even something so simple as hovering over the text input causes a transition that ought to be disabled.

### Expected Behavior
No animations when the user chooses to disable them

### Actual Behavior
Animations

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqLrq7VySvNyQETsQApPwV/)

### Other comments
It bares noting that this is an *accessibility issue*. I do not consider this a feature request, for 2 reasons:
1. Users suffering from motion sensitivity (such as yours truly) do not consider having animation to be a missing feature. It causes nauscia and is therefor well and truly a bug.
2. This has never been taken care of, neither in Vetify 3, nor in 2, which frankly baffles me. I would expect a professional UI framework to have this nailed, and yet it seems to have been put to one side for all these years.

KaelWD:
## Description

closes #19622 

## Markup:

```vue

```


invotech-magnus:
This is not a good solution! A lot of users have animations disabled in the OS for performance reasons, but that doesn't mean they don't want them in apps and web pages where they often serve a good purpose. If you want to force this accessibility feature on your users/visitors, you're free to do so in your own project by overriding styles, but don't force it on all of us in the framework!

PR #21378, opened 2025-05-06
fix(date): week number for DST

J-Sek:
## Description

- [x] similar change in `createRange` from #21450

fixes #21366

## Markup:

```vue
<template>
  <v-app>
    <v-container max-width="1200">
      <div class="d-flex ga-6">
        <!-- open DevTools » Sensors -->
        <!-- change Location to Berlin -->
        <!-- reload the page -->
        <v-date-picker :month="3" show-week />
        <v-row>
          <v-col>
            <v-date-input v-model="rangeSelection" :month="2" :year="2025" multiple="range" show-week />
          </v-col>
          <v-col>
            <v-alert type="info" variant="outlined">
              <ul>
                <li>1. select 3/29 ~ 3/31.</li>
                <li>2. re-open the picker.</li>
              </ul>
            </v-alert>
            <pre>rangeSelection: {{ rangeSelection }}</pre>
          </v-col>
        </v-row>
      </div>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'

  const rangeSelection = ref([])
</script>
```

PR #21531, opened 2025-06-06
fix(date): keep `createDateRange` internal

J-Sek:
alternative to #21525


ikushum (approved):

J-Sek:
@johnleider, I'd like to merge it and adjust #21378 on top, so the specific lines get marked in blame. ~~Need your vote on this one to be sure we are on the same page.~~ [noted](https://github.com/vuetifyjs/vuetify/pull/21525#issuecomment-2951406452)

PR #21549, opened 2025-06-10
fix(VTimePicker): Value not changing on wheel

### LINKED ISSUE #21524: [Bug Report][3.8.8] Scrollable on VTimePicker doens't work
### Environment
**Vuetify Version:** 3.8.8
**Vue Version:** 3.3.12
**Browsers:** Chrome 134.0.0.0, Google Chrome
**OS:** Linux x86_64, Linux

### Steps to reproduce
1. pick any hours or minute
2. the data should be changed

### Expected Behavior
when stop scroll, the data changed

### Actual Behavior
scrollbale didn't nothing

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdUEuuwjAMvIrlDe9J/eh9QAgVJM5BWYTiioikiRzTTdW7k7QIATt7PJ4Z+zBg4Kbce1/0N8INVkLWGyW0qzuAqs+V91M5NY3rROmO+AFNoGhLudfNlRj63LozmW2NZyWqRggNO2PUyRC0jq2SOPr9v3AclU+RYUjscXz4lJ9GCZlzVOVLvthGee0FAsltjqmtdywwAFMLI7TsLCziaYtEB4i6QSC5wTZRvr6T5qwS93HM8K9YFj8rTMW6WGPWKhMoe3vTjB3vv35ncA==)

### Other comments
it's lab i know but it some tiny functionallity should work

Haviles04:
## Description
Added Debounce to `wheel()` which will `emit('change')` switching hours to minutes and updating value
fixes #21524 

## Markup:
```
<template>
  <v-app>
    <v-container>
      <v-time-picker v-model="data" format="24hr" scrollable />
      <p>Current Time: {{ data }}</p>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const data = ref()
</script>
```

### DIFF
--- a/packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
+++ b/packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
@@ -66,6 +66,7 @@ export const VTimePickerClock = genericComponent()({
     const isDragging = ref(false)
     const valueOnMouseDown = ref(null as number | null)
     const valueOnMouseUp = ref(null as number | null)
+    const wheelTimeout = ref<NodeJS.Timeout | undefined>(undefined)
 
     const { textColorClasses, textColorStyles } = useTextColor(() => props.color)
     const { backgroundColorClasses, backgroundColorStyles } = useBackgroundColor(() => props.color)
@@ -115,6 +116,8 @@ export const VTimePickerClock = genericComponent()({
       if (value !== props.displayedValue) {
         update(value)
       }
+      clearTimeout(wheelTimeout.value)
+      wheelTimeout.value = setTimeout(() => emit('change', value), 1000)
     }
 
     function isInner (value: number) {


packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
```diff
@@ -66,6 +66,7 @@ export const VTimePickerClock = genericComponent()({
     const isDragging = ref(false)
     const valueOnMouseDown = ref(null as number | null)
     const valueOnMouseUp = ref(null as number | null)
+    const wheelTimeout = ref<NodeJS.Timeout | undefined>(undefined)
```
johnleider:
```suggestion
    const wheelTimeout = shallowRef(-1)
```

johnleider (dismissed):

johnleider (approved):

packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
```diff
@@ -66,6 +66,7 @@ export const VTimePickerClock = genericComponent()({
     const isDragging = ref(false)
     const valueOnMouseDown = ref(null as number | null)
     const valueOnMouseUp = ref(null as number | null)
+    const wheelTimeout = shallowRef<NodeJS.Timeout>()
```
J-Sek: Is there a reason for it to be any kind of ref? We usually just have `let timeout = -1`.

J-Sek (changes requested):

packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
Haviles04: No real reason, just didn't know what was preferred, I'm okay with using let

packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
```diff
@@ -116,8 +116,8 @@ export const VTimePickerClock = genericComponent()({
       if (value !== props.displayedValue) {
         update(value)
       }
-      clearTimeout(wheelTimeout)
-      wheelTimeout = setTimeout(() => emit('change', value), 1000)
+
+      debounceWheel(value)
```
J-Sek: With `debounce` we moved the action outside and have a naming problem. When reading this part we should not force ourselves to lookup the implementation. `debounceWheel` does not ring any bells for me unfortunately and I have a feeling that it ties the thing to the usage too much. The straight-to-the-point (albeit verbose) name would be `emitChangeDebounced`. We could make it shorter by assuming 'change' is the only reasonable event type to emit here and land with `emitDebounced` or `debounceEmit`. What do you think?

J-Sek (changes requested):

packages/vuetify/src/labs/VTimePicker/VTimePickerClock.tsx
Haviles04: I think you are right and `emitChangeDebounced` is probably the best name, since you can't really reuse this function for anything else.

PR #21574, opened 2025-06-14
fix(VBtn): correct letter-spacing compensation for RTL

J-Sek:
`letter-spacing` requires `text-indent` to make the button content appear centered (especially with numbers in `.v-btn--icon` that appear in VDatePicker). RTL layout requires negative offset.

Before:

image

After:

image

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.sass
+++ b/packages/vuetify/src/components/VBtn/VBtn.sass
@@ -29,6 +29,9 @@
     vertical-align: $button-vertical-align
     flex-shrink: 0
 
+    .v-locale--is-rtl &
+      text-indent: -1 * $button-text-letter-spacing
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VBtn/VBtn.sass
```diff
@@ -29,6 +29,9 @@
     vertical-align: $button-vertical-align
     flex-shrink: 0
 
+    .v-locale--is-rtl &
+      text-indent: -1 * $button-text-letter-spacing
```
MatthewAry, line 33: This is not great. 
https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing#syntax
We're supposed to be allowed to use strings for `letter-spacing`

Issue #15398, opened 2022-07-07
[Bug Report][3.0.0] v-btn-toggle clipping on narrow screens


ThibaultNocchi:
### Environment
**Vuetify Version:** 3.0.0-beta.5
**Vue Version:** 3.2.37
**Browsers:** Firefox 102.0
**OS:** Linux x86_64

### Steps to reproduce
* use a somewhat long `v-btn-toggle` group
* try on a narrow screen and see the buttons go out of screen

### Expected Behavior
Buttons could go to the next line or they could be scrollable?

### Actual Behavior
They go out of screen and those are untogglable.

![Screenshot 2022-07-07 at 10-46-51 BarjvqK](https://user-images.githubusercontent.com/1619359/177732110-60e141b2-b995-4e11-9f3d-7a19cc2aa486.png)

### Reproduction Link
<a href="https://codepen.io/tiwenty/pen/BarjvqK" target="_blank">https://codepen.io/tiwenty/pen/BarjvqK</a>

johnleider:
Oddly enough this is an issue on v2 as well. @KaelWD should `v-btn-group` use a slide group or just enable overflow?

johnleider:
@KaelWD still need feedback on this

johnleider:
As a workaround for now, wrap the component in a div with **overflow-auto**.

```html
<div class="overflow-x-auto">
  <v-btn-group>
    ...
  </v-btn-group>
</div>
```

ThibaultNocchi:
Thank you, this does help. :)

pierrejoye:
[...]

PR #21577, opened 2025-06-15
fix(VBtnToggle): buttons should be accessible despite overflow (#15398)

jcjp:
## Description

- Fixes: https://github.com/vuetifyjs/vuetify/issues/15398

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-btn-toggle v-model="msg">
        <v-btn v-for="n in 6" :key="n" :value="n">{{ n }}</v-btn>
      </v-btn-toggle>
      <v-text-field v-model="msg" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref(0)

  const callNext = () => {
    msg.value++
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VBtnGroup/VBtnGroup.sass
+++ b/packages/vuetify/src/components/VBtnGroup/VBtnGroup.sass
@@ -11,7 +11,8 @@
     flex-wrap: nowrap
     max-width: 100%
     min-width: 0
-    overflow: hidden
+    overflow-y: hidden
+    overflow-x: auto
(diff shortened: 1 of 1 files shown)


jcjp:
> I would love to see a follow-up PR to `dev` for `&--vertical` variant with:
> 
> ```css
> max-height: 100%;
> overflow-x: hidden;
> overflow-y: auto;
> ```

Will work on this thanks!

PR #21587, opened 2025-06-17
feat(VWindow): vertical arrows

J-Sek:
## Description

On v2 `vertical-delimiters` on VCarousel caused the arrows to be hidden (to avoid obstructing the delimiters). I think we can have both coexisting.

Example in Docs is VCarousel scope because the feature is primarily intended for carousels.

image

## Markup:

```vue
<template>
  <v-app>
    <v-container max-width="700">
      <v-card elevation="24" rounded="xl">
        <v-carousel
          direction="vertical"
          height="400"
          vertical-arrows="left"
          vertical-delimiters="right"
        >
          <v-carousel-item v-for="i in 5" :key="i">
            <v-sheet height="100%">
              <div class="d-flex h-100 justify-center align-center">
                <div class="text-h2">Slide #{{ i }}</div>
              </div>
            </v-sheet>
          </v-carousel-item>
        </v-carousel>
      </v-card>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/docs/src/examples/v-carousel/misc-vertical.vue
+++ b/packages/docs/src/examples/v-carousel/misc-vertical.vue
@@ -19,7 +19,7 @@
 
         <v-overlay
           :scrim="false"
-          content-class="w-100 h-100 d-flex flex-column align-center justify-space-between click-through py-3"
+          content-class="w-100 h-100 d-flex flex-column align-center justify-space-between pointer-pass-through py-3"
           contained
           model-value
           no-click-animation
@@ -123,13 +123,3 @@
     },
   }
 </script>
-
-<style>
-.click-through {
-  pointer-events: none;
-}
-
-.click-through > * {
-  pointer-events: auto;
-}
-</style>
--- a/packages/docs/src/examples/v-carousel/misc-vertical.vue
+++ b/packages/docs/src/examples/v-carousel/misc-vertical.vue
@@ -39,7 +39,7 @@
             </v-sheet>
           </v-scroll-x-transition>
           <v-chip
-            :text="`{{ currentIndex + 1 }} / ${items.length }`"
+            :text="`${ currentIndex + 1 } / ${items.length }`"
             color="#eee"
             size="small"
             variant="flat"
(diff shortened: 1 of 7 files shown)


packages/docs/src/examples/v-carousel/misc-vertical.vue
```diff
@@ -0,0 +1,132 @@
...
+    computed: {
+      currentItem () {
+        return this.items[this.currentIndex]
+      },
+    },
+  }
+</script>
+
+<style scoped>
+.click-through {
```
J-Sek: I wish we had this as a utility class.

johnleider: Is there anything existing that it would fit under? Like `.cursor-click-through`.

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -244,7 +246,16 @@ export const VWindow = genericComponent<new <T>(
           { slots.default?.({ group }) }
 
           { props.showArrows !== false && (
-            <div class="v-window__controls">
+            <div
+              class="v-window__controls"
+              style={{
```
johnleider: This should probably be a handled in a class.

packages/docs/src/examples/v-carousel/misc-vertical.vue
J-Sek: Not a good fit for `cursor-*`, but can be a cohesive new group when we include typical TailwindCSS utilities for `auto` and `none` [[1]](https://tailwindcss.com/docs/pointer-events#ignoring-pointer-events)

```sass
.pointer-events-none
  pointer-events: none !important

.pointer-events-auto
  pointer-events: auto !important

.pointer-pass-through
  pointer-events: none !important
  > *
    pointer-events: auto !important
```

Note: we could move those simple ones to the utilities generator, but I don't like the idea of having them separated. Keeping it in one place is less mental overhead (my opinion). Let me know if you'd like to change it.

PR #21597, opened 2025-06-18
fix(VTimePicker): adapt width for seconds and AM/PM

J-Sek:
## Description

We can leave it for users to guess the optimal width around ~350px and use `height prop... or we can make the picker adapt.

We could target a combination of AM/PM and seconds, but the code did not look elegant...

```sass
&:has(.v-time-picker-controls__time--with-seconds__btn):has(.v-time-picker-controls__time--with-ampm__btn)
```

...let me know if you'd rather not extend the width when the picker has 24h format and seconds that can still fit with previous 328px width.

image

## Markup:

```vue
<template>
  <v-app>
    <v-container
      class="d-flex align-start ga-6 justify-center flex-wrap"
      width="1600"
    >
      <v-time-picker v-model="time1" elevation="24" rounded="lg" ampm-in-title />
      <v-time-picker v-model="time2" elevation="24" rounded="lg" ampm-in-title use-seconds />
      <v-time-picker v-model="time1" elevation="24" format="24hr" rounded="lg" />
      <v-time-picker v-model="time2" elevation="24" format="24hr" rounded="lg" use-seconds />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const time1 = ref()
  const time2 = ref()
</script>
```


KaelWD:
wdyt about #21613 instead?

J-Sek:
> wdyt about #21613 instead?

Much better, thanks

PR #21598, opened 2025-06-18
feat(VHotkey): add VHotkey component

MatthewAry:
- Introduced VHotkey component to render keyboard shortcuts
- Added localization entries for hotkey terms in multiple languages.
- Created JSON files for hotkey localization and updated auto-imports for new components.
- Included styles for the VHotkey component
- Enhanced documentation for the new component in the auto-imports and relevant markdown files.

## Markup:

```vue

```


packages/vuetify/src/labs/VHotkey/VHotkey.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+    // If we requested icon mode but no icon is available, fallback to text mode
+    if (mode === 'icon' && !keyConfig.icon) {
+      return ['text', value]
+    }
+
+    // If we requested symbol mode but no symbol is available, fallback to text mode
+    if (mode === 'symbol' && !keyConfig.symbol) {
+      return ['text', value]
+    }
```
Haviles04: Can these be combined into one? 


```suggestion
 
    if (!keyConfig[mode]) {
      return ['text', value]
    }
```

Haviles04 (commented):
Cool stuff, 1 small nitpick :)

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
MatthewAry: The current code structure is:

```typescript
type KeyConfig = {
  symbol?: string
  icon?: string
  text: string
}
```

Notice that:
1. `symbol` and `icon` are **optional** properties (marked with `?`)
2. `text` is **required** 
3. The `value` variable is set to `keyConfig[mode] ?? keyConfig.text`

Your suggested change would have these problems:

1. If `mode` is `'text'`, then `keyConfig[mode]` would be the text value (which is always present), so the condition would never be true for text mode.
2. The current code specifically handles the cases where:
   - They request `'icon'` mode but no icon is defined → fallback to text
   - They request `'symbol'` mode but no symbol is defined → fallback to text
   - For `'text'` mode, it always uses the text value (no fallback needed)
3. The current approach ensures that when an icon or symbol isn't available, it falls back to text mode.

The existing code is more explicit and handles the three different modes correctly. It could potentially be simplified to:

```typescript
if ((mode === 'icon' && !keyConfig.icon) || (mode === 'symbol' && !keyConfig.symbol)) {
  return ['text', value]
}
```

But I think the current version is easier to understand.

MatthewAry:
`packages/docs/src/examples/v-hotkey/prop-key-map.vue` has some pretty dumb lint errors. Not sure what to do about that. I fixed bugs with platform handling and added a new prop to VHotkey for DX, Testing, and Documentation purposes.

packages/vuetify/src/components/VKbd/VKbd.sass
```diff
@@ -3,11 +3,15 @@
...
+    align-items: center
+    align-self: stretch
     background: rgb(var(--v-theme-kbd))
     color: rgb(var(--v-theme-on-kbd))
     display: $kbd-display
     font-size: $kbd-font-size
     font-weight: $kbd-font-weight
+    justify-content: center
+    min-width: $kbd-min-width
     padding: $kbd-padding
```
J-Sek: `line-height: 1` could help avoid inconsistent height

image

packages/vuetify/src/labs/VHotkey/VHotkey.scss
```diff
@@ -0,0 +1,46 @@
...
+    &__divider {
+      opacity: 0.5;
+      font-size: 0.75rem;
+    }
+
+    &__combination {
+      display: inline-flex;
+      align-items: center;
+      gap: 2px;
+    }
```
J-Sek, line 238: `opacity`, `font-size` and `gap` should definitely get variables. Not sure about `padding`, but it would not hurt.

packages/vuetify/src/labs/VHotkey/VHotkey.scss
```diff
@@ -0,0 +1,46 @@
+@use '../../styles/tools';
+
+@include tools.layer('components') {
+  .v-hotkey {
+    display: inline-flex;
+    align-items: center;
+    gap: 4px;
+
+    &__key.v-kbd {
```
J-Sek: I would set `min-width: 1.5rem` (configurable with another Sass variable). It makes `display-mode="icon"` combinations look much better.

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
```diff
@@ -0,0 +1,430 @@
...
+  }),
+  arrowright: createKey({
+    default: { symbol: '→', icon: '$arrowright', text: '$vuetify.hotkey.rightArrow' },
+  }),
+  backspace: createKey({
+    default: { symbol: '⌫', icon: '$backspace', text: '$vuetify.hotkey.backspace' }, // Backspace symbol
+  }),
+  escape: createKey({
+    default: { text: '$vuetify.hotkey.escape' },
+  }),
```
J-Sek: I think we are missing `space` here. There is a dedicated icon `mdi-keyboard-space` ready to use.

packages/vuetify/src/components/VKbd/VKbd.sass
J-Sek: And additional `vertical-align: middle` could help within text

image

packages/vuetify/src/labs/VHotkey/VHotkey.scss
```diff
@@ -0,0 +1,106 @@
...
+    &__key-symbol.v-kbd {
+      line-height: normal;
+      font-size: 1em;
+    }
+
+    &--inline {
+      align-items: baseline;
+      max-height: 1lh;
+      vertical-align: baseline;
+      font-size: $hotkey-inline-font-size;
```
J-Sek: dividers within `--inline` ends up with 10.5px font size. Should they be set to 1em as well? Perhaps with a dedicated Sass variable?

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
MatthewAry: Yes, we could add space. But do we really need/want to? I personally hate the fact that Macs have symbols on the keyboard to begin with.

packages/vuetify/src/labs/VHotkey/VHotkey.scss
MatthewAry: Set to `1.5em`

packages/vuetify/src/components/VKbd/VKbd.sass
MatthewAry: I agree about need for improved height consistency. Specifically with symbol mode vs others. I would love to hear your thoughts after you review the latest changes.

MatthewAry:
I updated my playground!

MatthewAry:
image
Looks pretty good!

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
J-Sek: I would use it, but it surely depends on the target audience. Tech-savvy users will understand the symbol just fine, while others might be a bit confused at first. If you have any point of reference I would just follow.

m50S79sM6SRNp8Jn:
Excellent work, keep it up!

packages/vuetify/src/components/VKbd/_variables.scss
```diff
@@ -1,11 +1,17 @@
...
-$kbd-font-size: 0.75rem !default;
+// Kbd elements are used inline with text, so we should use the same font size as the text it's with.
+$kbd-font-size: 0.875em !default;
 $kbd-font-weight: normal !default;
-$kbd-padding: .2rem !default;
+$kbd-line-height: 1 !default;
+$kbd-padding: 3px 6px !default;;
+$kbd-min-height: 1em !default;
 $kbd-min-width: 20px !default;
+$kbd-margin-x: 1px !default;
```
J-Sek: we usually call it `-margin-inline`

packages/vuetify/src/iconsets/mdi-svg.ts
```diff
@@ -48,6 +48,16 @@ const aliases: IconAliases = {
...
   color: 'svg:M17.5 12a1.5 1.5 0 0 1-1.5-1.5A1.5 1.5 0 0 1 17.5 9a1.5 1.5 0 0 1 1.5 1.5a1.5 1.5 0 0 1-1.5 1.5m-3-4A1.5 1.5 0 0 1 13 6.5A1.5 1.5 0 0 1 14.5 5A1.5 1.5 0 0 1 16 6.5A1.5 1.5 0 0 1 14.5 8m-5 0A1.5 1.5 0 0 1 8 6.5A1.5 1.5 0 0 1 9.5 5A1.5 1.5 0 0 1 11 6.5A1.5 1.5 0 0 1 9.5 8m-3 4A1.5 1.5 0 0 1 5 10.5A1.5 1.5 0 0 1 6.5 9A1.5 1.5 0 0 1 8 10.5A1.5 1.5 0 0 1 6.5 12M12 3a9 9 0 0 0-9 9a9 9 0 0 0 9 9a1.5 1.5 0 0 0 1.5-1.5c0-.39-.15-.74-.39-1c-.23-.27-.38-.62-.38-1a1.5 1.5 0 0 1 1.5-1.5H16a5 5 0 0 0 5-5c0-4.42-4.03-8-9-8',
+  command: 'svg:M6 2a4 4 0 0 1 4 4v2h4V6a4 4 0 0 1 4-4a4 4 0 0 1 4 4a4 4 0 0 1-4 4h-2v4h2a4 4 0 0 1 4 4a4 4 0 0 1-4 4a4 4 0 0 1-4-4v-2h-4v2a4 4 0 0 1-4 4a4 4 0 0 1-4-4a4 4 0 0 1 4-4h2v-4H6a4 4 0 0 1-4-4a4 4 0 0 1 4-4m10 16a2 2 0 0 0 2 2a2 2 0 0 0 2-2a2 2 0 0 0-2-2h-2zm-2-8h-4v4h4zm-8 6a2 2 0 0 0-2 2a2 2 0 0 0 2 2a2 2 0 0 0 2-2v-2zM8 6a2 2 0 0 0-2-2a2 2 0 0 0-2 2a2 2 0 0 0 2 2h2zm10 2a2 2 0 0 0 2-2a2 2 0 0 0-2-2a2 2 0 0 0-2 2v2z',
+  ctrl: 'svg:m19.78 11.78l-1.42 1.41L12 6.83l-6.36 6.36l-1.42-1.41L12 4z',
```
J-Sek: Not sure if it matters, but I would copy those straight from `@mdi/js`

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
```diff
@@ -0,0 +1,551 @@
...
+  // How to display keys: 'symbol' uses special characters (⌘, ⌃), 'icon' uses SVG icons, 'text' uses words
+  displayMode: {
+    type: String as PropType<DisplayMode>,
+    default: 'icon',
+  },
+  keyMap: {
+    type: Object as PropType<KeyMap>,
+    default: keyMap,
+  },
+  overridePlatform: {
```
J-Sek: How about instead of `overridePlatform` make it into `platform: 'auto' | 'pc' | 'mac'` with default `'auto'`?

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
MatthewAry: I have already added a title property to icon/symbol representations such that they should be able to see the translation of the icon. I can add the space icon. I just think it's stupid. I wonder if I should change the way the keymap is exposed through the prop such that people would be able to exclude icons/symbols. Right now, they can only modify or add.

packages/vuetify/src/labs/VHotkey/VHotkey.scss
```diff
@@ -0,0 +1,232 @@
...
+      font-size: $hotkey-inline-divider-font-size;
+    }
+
+    // Variant-specific overrides
+    &--variant-elevated &__key.v-kbd {
+      @include tools.elevation($hotkey-elevation);
+    }
+
+    &--variant-flat &__key.v-kbd {
+      @include tools.elevation(0);
```
J-Sek: Those are not that unique in our codebase (found 3 other components calling that helper with `0`), but I am not sure if it brings any value over `box-shadow: none`.

packages/vuetify/src/iconsets/mdi-svg.ts
MatthewAry: They came from https://icones.js.org/ record of those icons. Will update from OG reference

packages/vuetify/src/labs/VHotkey/VHotkey.scss
MatthewAry: Good idea.

packages/vuetify/src/labs/VHotkey/VHotkey.tsx
MatthewAry: Great Idea. Will introduce that after this up-coming push.

packages/vuetify/src/labs/VHotkey/VHotkey.scss
MatthewAry: Will do

packages/vuetify/src/iconsets/mdi-svg.ts
J-Sek: I did the same mistake few weeks ago

packages/vuetify/src/util/hotkey-parsing.ts
```diff
@@ -0,0 +1,184 @@
...
+  plus: '+',
+  minus: '-',
+  underscore: '_',
+}
+
+/**
+ * Splits a single combination string into individual key parts.
+ *
+ * A combination is a set of keys that must be pressed simultaneously.
+ * e.g. `ctrl+k`, `shift--`, `alt++`
```
J-Sek: Even though it is just a comment, examples should probably include only valid combinations.
Later code suggests that `-` after modifier key is not valid.
`alt++` also makes me pause here as I can only really set `alt+=`, because `+` requires `shift` (on a typical keyboard)

packages/vuetify/src/util/hotkey-parsing.ts
```diff
@@ -0,0 +1,184 @@
...
+        buffer += char
+      }
+    } else {
+      buffer += char
+    }
+  }
+  flushBuffer()
+
+  // Combinations like `ctrl-b` are not valid, `-` is a sequence separator.
+  // `-` is only a valid key if it's on its own.
```
J-Sek: It looks like `-` is repeated in many places and it would be risky to change at this point, so I just drop a loose idea without an expectation to have any changes.

My first pick for sequence separator would be `,`. I can see VSCode uses ` `(space). In my opinion both have a slight advantage over `-` when it comes to readability. That said, it is only visible to developers, so it matters a bit less than UI representation. If you based your pick on some common app or system, then it is probably fine.

Edit: weird that some sources list `CTRL++` as an existing combination on a Windows (or MS Excel). It is possible I am missing something here.

packages/vuetify/src/util/hotkey-parsing.ts
MatthewAry: Great point! I was so focused on making the parser resilient against weird or complex combinations that I totally overlooked how `+` would typically be unreachable. I think some of the tests for the parser check for it too! Hahahah. 😅 I'll update the comment and review the tests.

packages/vuetify/src/util/hotkey-parsing.ts
MatthewAry: This comment is wrong! This should be a valid combo! I blame AI.

johnleider (approved):

PR #21609, opened 2025-06-19
fix(VInput): Remove redundant aria roles in v-input

yncat:
## Description

There was a critical issue using v-input with screen readers. Because it had role="alert" applied immediately upon rendering, screen readers would repeat “alert” as many times as there were v-input components on the screen after page load. For example, if I had 5 input boxes in a page, I had to hear "alert alert alert alert alert" as the page finished loading. This issue significantly interrupted screen reader users' browsing experience.

I also removed aria-live="polite" as it was functionally redundant with role="alert".

Although alert role is no longer present, error messages are still announced correctly when they appear because the messages area is connected via aria-describedby. So, this change simply eliminates the redundant “alert” announcements without compromising the accessibility of the component.

I’ve tested this locally and confirmed that it works as expected. I hope the dev team can integrate this fix into the next patch version.

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-container>
        <v-row>
          <v-col cols="12">
            <h1 class="text-h2 mb-4">Page title</h1>

            <v-card>
              <v-card-title>Form</v-card-title>
              <v-card-text>
                <v-form @submit.prevent="onSubmit">
                  <v-text-field
                    v-model="name"
                    :rules="[rules.required]"
                    label="Name"
                    prepend-icon="mdi-account"
                  />
                  <v-text-field
                    v-model="email"
                    :rules="[rules.required]"
                    label="Email"
                    prepend-icon="mdi-email"
                    type="email"
                  />
                  <v-text-field
                    v-model="password"
                    :rules="[rules.required]"
                    label="Password"
                    prepend-icon="mdi-lock"
                    type="password"
                  />
                  <v-btn color="primary" text="Submit" type="submit" />
                </v-form>
              </v-card-text>
            </v-card>
          </v-col>
        </v-row>
      </v-container>
    </v-main>
  </v-app>
</template>

<script>
  export default {
    name: 'App',
    data () {
      return {
        name: '',
        email: '',
        password: '',
        rules: {
          required: value => !!value || 'This field is required',
        },
      }
    },
    methods: {
      async onSubmit (e) {
        const { valid } = await e
        if (valid) {
          alert(`Hello, ${this.name}!`)
        }
      },
    },
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VInput/VInput.tsx
+++ b/packages/vuetify/src/components/VInput/VInput.tsx
@@ -232,8 +232,6 @@ export const VInput = genericComponent<new <T>(
             <div
               id={ messagesId.value }
               class="v-input__details"
-              role="alert"
-              aria-live="polite"
             >
               <VMessages
                 active={ hasMessages }


J-Sek:
Reworked the playground:
- removed non-english symbols that were distracting screen reader
- one more field to try to get more alert announcements
- actual validation on submit

When testing locally with Orca, the only minor annoyance I can hear is that after tabbing over a required field (e.g. first one), the "Email entry" is followed by "This field is required" regarding the Name field that is not in focus. If we assume that people use `required` attribute that is spoken on focus, then the alert is redundant. However if the field has custom validation, user won't be notified about the validation failure. One way to work around it would be to auto-focus the first field with a validation error upon clicking Submit (similar to an example on react-aria docs).

Can someone confirm the original issue with the updated playground on some popular Screen Reader?

yncat:
Sorry, I completely missed the comment. I'll first test the reworked implementation, and if it looks good, I'll re-apply the same changes into the latest codebase.

PR #21639, opened 2025-06-26
feat(Vselect): Add Bg color to item list

### LINKED ISSUE #21579: [Feature Request] v-select prop for dropdown menu background color
### Problem to solve
Setting the color of the dropdown menu just as you would set the bg-color for the v-select bar. It would be ideal to have them both respect the bg-color. Instead of having to change the background the bar with a prop and then the dropdown menu with css selectors that arn't documented/hard to find.

### Proposed solution
bg-color sets the background color of the entire v-select component instead of just the clickable "bar"

Haviles04:
## Description
Added prop and documentation for Item list BG color 
fixes #21579 

## Markup:
```
<template>
  <v-select
    v-model="selected"
    :items="['Apple', 'Orange', 'Banana', 'Pear']"
    bg-color="blue"
    item-bg="black"
    item-color="green"
    width="500"
  /></template>

<script setup>
  import { ref } from 'vue'

  const selected = ref('Apple')
</script>

```

### DIFF
--- a/packages/api-generator/src/locale/en/Select.json
+++ b/packages/api-generator/src/locale/en/Select.json
@@ -4,7 +4,7 @@
     "chips": "Changes display of selections to chips.",
     "closableChips": "Enables the [closable](/api/v-chip/#props-closable) prop on all [v-chip](/components/chips/) components.",
     "hideSelected": "Do not display in the select menu items that are already selected.",
-    "itemBg" : "Sets the background color of the item list container.",
+    "itemBg": "Sets the background color of the item list container.",
     "itemColor": "Sets color of selected items.",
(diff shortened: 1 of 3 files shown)


J-Sek:
`item-bg="black"` becomes a shortcut to `:list-props="{ bgColor: 'black' }"`. I think it would be more intuitive to name it `list-bg-color` but even after... does not bring much value. How about a new example in docs for `list-props` in general (#21658)?

Haviles04:
> `item-bg="black"` becomes a shortcut to `:list-props="{ bgColor: 'black' }"`. I think it would be more intuitive to name it `list-bg-color` but even after... does not bring much value. How about a new example in docs for `list-props` in general (#21658)?

I think this is likely the more reasonable approach

PR #21649, opened 2025-06-27
fix(VDatePicker): avoid infinite loop when first day is out of range

J-Sek:
## Description

Anticipating a buggy UI where a careless developer (or LLM) creates an input element that does not prevent `first-day-of-week` outside of range.

## Markup:

```vue
<template>
  <v-app>
    <v-container max-width="400">
      <v-slider v-model="value" label="First day of week" max="8" step="1" />
      <pre>value: {{ value }}</pre>
      <v-date-picker :first-day-of-week="value" />
    </v-container>
  </v-app>
</template>

<script setup>
  const value = shallowRef(0)
</script>
```

### DIFF
--- a/packages/vuetify/src/composables/date/adapters/vuetify.ts
+++ b/packages/vuetify/src/composables/date/adapters/vuetify.ts
@@ -1,5 +1,5 @@
 // Utilities
-import { createRange, padStart } from '@/util'
+import { consoleWarn, createRange, padStart } from '@/util'
 
 // Types
 import type { DateAdapter } from '../DateAdapter'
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/composables/date/adapters/vuetify.ts
```diff
@@ -85,7 +85,12 @@ function getWeekArray (date: Date, locale: string, firstDayOfWeek?: number) {
 }
 
 function startOfWeek (date: Date, locale: string, firstDayOfWeek?: number) {
-  const day = firstDayOfWeek ?? weekInfo(locale)?.firstDay ?? 0
+  let day = firstDayOfWeek ?? weekInfo(locale)?.firstDay ?? 0
```
KaelWD:
```suggestion
  const day = (firstDayOfWeek ?? weekInfo(locale)?.firstDay ?? 0) % 7
```

J-Sek: I'd rather keep `let` fix it when the value is not a discrete value or a negative number.

PR #21658, opened 2025-06-30
docs(VSelect): example for `list-props`

### LINKED ISSUE #21579: [Feature Request] v-select prop for dropdown menu background color

J-Sek:
## Description

resolves #21579

## Markup:

```vue
<template>
  <v-select
    v-model="selected"
    :items="['Apple', 'Orange', 'Banana', 'Pear']"
    :list-props="{ bgColor: 'purple' }"
    item-color="yellow"
    label="Label"
    multiple
  ></v-select>
</template>

<script setup>
  import { ref } from 'vue'

  const selected = ref(['Apple'])
</script>
```

### DIFF
--- a/packages/docs/src/pages/en/components/selects.md
+++ b/packages/docs/src/pages/en/components/selects.md
@@ -95,6 +95,12 @@ Custom props can be passed directly to `v-menu` using **menu-props** prop. In th
 
 <ExamplesExample file="v-select/prop-menu-props" />
 
+#### List props
+
+Custom props can be passed directly to `v-list` using **list-props** prop. It helps address common use case of customizing dropdown list background color.
+
+<ExamplesExample file="v-select/prop-list-props" />
+
 #### Custom item props
 
 `item-title` and `item-value` are provided for convenience, and additional props can be passed to list items either through the **item** slot (see below) or with the **itemProps** prop.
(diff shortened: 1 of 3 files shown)


packages/docs/src/pages/en/components/selects.md
```diff
@@ -95,6 +95,12 @@ Custom props can be passed directly to `v-menu` using **menu-props** prop. In th
 
 <ExamplesExample file="v-select/prop-menu-props" />
 
+#### List props
+
+Custom props can be passed directly to `v-list` using **list-props** prop. It helps address common use case of customizing dropdown list background color.
```
Haviles04: I kinda like how the Menu props section is worded above and think it would fit in nicely here as well

```suggestion
Custom props can be passed directly to `v-list` using **list-props** prop. In this example a background color is added to the list.
```

Haviles04 (dismissed):

PR #20592, opened 2024-10-18
fix(VVirtualScroll): correct keyboard navigation for selection controls

J-Sek:
## Description

TODO:
- [x] avoid scrolling for `next` and `prev`
- [x] apply to VCombobox and VAutocomplete

fixes #18383

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container>
      <h5>VSelect</h5>
      <v-select v-model="selection1" :items="options" clearable />
      <h5>VCombobox</h5>
      <v-combobox v-model="selection2" :items="options" clearable />
      <h5>VAutocomplete</h5>
      <v-autocomplete v-model="selection3" :items="options" clearable />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const selection1 = ref(null)
  const selection2 = ref(null)
  const selection3 = ref(null)
  const options = [
    { title: 'Alabama', value: 'AL' },
    { title: 'Alaska', value: 'AK' },
    { title: 'American Samoa', value: 'AS' },
    { title: 'Arizona', value: 'AZ' },
    { title: 'Arkansas', value: 'AR' },
    { title: 'California', value: 'CA' },
    { title: 'Colorado', value: 'CO' },
    { title: 'Connecticut', value: 'CT' },
    { title: 'Delaware', value: 'DE' },
    { title: 'District Of Columbia', value: 'DC' },
    { title: 'Federated States Of Micronesia', value: 'FM' },
    { title: 'Florida', value: 'FL' },
    { title: 'Georgia', value: 'GA' },
    { title: 'Guam', value: 'GU' },
    { title: 'Hawaii', value: 'HI' },
    { title: 'Idaho', value: 'ID' },
    { title: 'Illinois', value: 'IL' },
    { title: 'Indiana', value: 'IN' },
    { title: 'Iowa', value: 'IA' },
    { title: 'Kansas', value: 'KS' },
    { title: 'Kentucky', value: 'KY' },
    { title: 'Louisiana', value: 'LA' },
    { title: 'Maine', value: 'ME' },
    { title: 'Marshall Islands', value: 'MH' },
    { title: 'Maryland', value: 'MD' },
    { title: 'Massachusetts', value: 'MA' },
    { title: 'Michigan', value: 'MI' },
    { title: 'Minnesota', value: 'MN' },
    { title: 'Mississippi', value: 'MS' },
    { title: 'Missouri', value: 'MO' },
    { title: 'Montana', value: 'MT' },
[...]

PR #21677, opened 2025-07-02
fix(VAutocomplete): allow autocomplte on active element with keyboard only

### LINKED ISSUE #21404: [Bug Report][3.8.4] Autocomplete on active element using only keyboard doesn't work
### Environment
**Vuetify Version:** 3.8.4
**Last working version:** 2.3.21
**Vue Version:** 3.5.13
**Browsers:** Chrome 135.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
* Select an item in between (not at the end, not at the start)
* Then use arrow key directly (on your keyboard) to find another value (don't click, use only arrow key)
* You should see that you are at the start of the component and not at the position of the active element.

--
* If you click on the component before using arrow key, you are at the right place (try to use keyboard only)

### Expected Behavior
When hitting arrow key you should be at the position of the active element (no mouse usage and it's directly after selecting a value)

### Actual Behavior
When hitting arrow key you are at the first element (no mouse usage and it's directly after selecting a value)

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdUU1PwzAM/Ssml7ZiTTXWSdXUIk27cETiwIFyCJ07ItImStJK07T/Tj42GFxi5/nZ7zl5OxGju2KrFJ0nJBtSWxyUYBYf2xGgnnOmVEjDpZOjZXxEfYEiY7Kyk64LLcKcD3KPomlJiC2BDXcjjQNCdIBgH4HwLA23XI4OKq4SxX8Nj0QLdXFjzV1Np7myYNBO0SEflNQWTqCxhzP0Wg6QuK0STwdwc42FwRyg8Yw0eUIhJLxKLfZ3SXZD8cavJF/4LYUVLqXoeKs1O1KvlYY0XT6sIYeyyugXHk2aZXRgKp2ZmBCan2d7sZqPh9C3+2R65yTTsoJ7CMRgBiAEd9RF3NV1k/OCrOiaLlfEJxUtyaJnwuDizz9G7P0bK8OZQQ==)

### Other comments
My users found a kinda "frustrating" issue, when they missed an option in the autocomplete if they use the keyboard directly after the selection they're not at the active element, they get back to the first one...

It used to work on Vuetify 2 and I found no options to make it work as it was. 
So I guess it's an issue.

erenertas123:
## Description
Autocomplete on active element using only keyboard doesn't work
I added new check for keyboardEvents for value selection.

resolves #21404

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -226,6 +226,23 @@ export const VAutocomplete = genericComponent<new <
         menu.value = false
       }
 
+      if(['ArrowDown', 'ArrowUp'].includes(e.key)) {
+        if (model.value.length > 0) {
+            const lastSelectedItem = model.value[model.value.length - 1];
+            const displayIndex = displayItems.value.findIndex(item => 
+              (props.valueComparator || deepEqual)(item.value, lastSelectedItem.value)
+            );
+            
+          if (displayIndex >= 0) {
+            vVirtualScrollRef.value?.scrollToIndex(displayIndex)
+            nextTick(() => {
+              listRef.value?.focus(displayIndex)
+            })
+            return;
+          }
+        }
+      }
+
       if (
         highlightFirst.value &&
         ['Enter', 'Tab'].includes(e.key) &&


J-Sek:
In this state it only works for short lists. It is an improvement, but I don't think the team wants to ship unstable behavior. Also, VSelect and VCombobox have to be included. Please take a look at #20592 to get a sense of complexity behind this issue. The very first thing I did was to replace `scrollToIndex`, because when the list is long (or the screen size is smaller), the virtualization kicks in and this method does not work as expected.

PR #21708, opened 2025-07-07
docs(useHotkey): Fix Eratta

MatthewAry:
- Fixes an example that suggests usage of the windows key is supported
- Adds link in introduction/why-vuetify page to useHotkey
- Adds link to VHotkey in labs/introduction


MatthewAry:
Pretty sure these CI failures are upstream of this PR.

Issue #21714, opened 2025-07-08
[Feature Request] VMaskInput and :mask property


Nadoedalo:
### Problem to solve
The current masked input doesn't fit the general data flow and doesn't allow for any programmatic control hence it's inferior to other options available.

I'm looking for these features available in other mask inputs:
1. The `VMask` for the actual `:mask` property - something that would provide following capabilities:
  1.1. The ability to validate the string against the mask
  1.2. The ability to un-mask string
  1.3. Dynamic resolution of the mask aka having more than 1 mask for the input
  1.4. Be reactive and play nicely with other flows
2. VMaskInput makes :rules break due to the mask changes. It supposes to be display-only instead the masked value gets validated.
3. It would be nice to see ahead of the mask. Kind of like a placeholder for the characters missing with "non-required" characters already present and in full color while the actual mask keys in gray / opaque
4. The UX issue - you don't allow user input if it doesn't fit the mask. Instead allow to enter character and delete it on keyup (it may depend on the `validate-on` property? For which there is little-to-no documentation)

### Proposed solution
For `#1` it may look like this:
```js
import { maskRef, ref } from 'vue';

const mask = maskRef('(AA)-###-###-##-##');
// alternatively const { refValue, test, mask } = maskRef('(AA)-###-###-##-##');
const testString = ref('AB1234567890')
mask.value // (AA)-###-###-##-##
mask.test(testString.value) // true
mask.test('(AB)-123-456-78-90') // true
mask.mask(testString.value) // (AB)-123-456-78-90
mask.unmask('(AB)-123-456-78-90') // AB1234567890

// and dynamic resolution
const dynamicMask = maskRef(['(AA)-###', '(##)-AAA']); // valid masks (AB)-123 and (12)-ABC
```

[...]

PR #21736, opened 2025-07-10
feat(mask): create useMask composable

ikushum:
Related to #21714

## Description
Create a useMask composable for working with masks

## Markup:
```vue
<template>
  <v-container class="pa-md-12">

    <h3>Mask: {{ maskString }}</h3>
    <br>
    isValid: {{ mask.test(string) }}
    <br>
    isValid Exact: {{ mask.test(string, { isExact: true }) }}
    <br>
    maskText: {{ mask.apply(string) }}
    <br>
    unmaskText: {{ mask.unapply(string) }}
    <br>
    <br>
    <v-text-field v-model="string" label="Input" />
  </v-container>
</template>

<script setup>
  import { ref } from 'vue'
  import { useMask } from '@/composables/mask'

  const maskString = '####-####'

  const mask = useMask({
    mask: maskString,
  })

  const string = ref('123')
</script>
```


packages/vuetify/src/composables/mask/mask.ts
```diff
@@ -145,7 +145,7 @@ export function useMask (props: MaskProps, inputRef: Ref<HTMLInputElement | unde
     return newText
   }
 
-  function unmaskText (text: string | null): string | null {
```
J-Sek: `unapply` translates to something weird and could drag unnecessary attention (cognitive load).

I would personally name it `unmask` or `remove` as I don't feel like it is important to have both method names strictly aligned.

It's not a strong suggestion, though :)

johnleider: I'm in favor of `unmask`

ikushum: I renamed a bunch of things:

apply --> mask
unapply --> unmask
test --> isValid
test({isExact: true}) --> isComplete

Example usage:
```js
const {
  isValid,
  isComplete,
  mask,
  unmask
} = useMask()
```

J-Sek (approved):
It might be better to wait until `dev` pulls [64943b3](http://github.com/vuetifyjs/vuetify/commit/64943b30c4a84ddfaa565f9b2d62285450aa128b) from `master`. There is one place that does not get recognized as conflict and will be calling old `maskText(...)` after successful merge. This would result in breaking the build, so I just suggest we avoid making Kael's life harder.

packages/vuetify/src/composables/mask/mask.ts
johnleider: love it

PR #21753, opened 2025-07-14
fix(VDateInput): align modelValue prop and event types

### LINKED ISSUE #21751: [Bug Report][3.9.0] The `modelValue` type in VDateInput is inconsistent between the prop and the event
### Environment
**Vuetify Version:** 3.9.0
**Vue Version:** 3.5.17
**Browsers:** Edge 138.0.0.0
**OS:** Windows 10

### Steps to reproduce
The types of `:model-value` and `@update:model-value` are different.

### Expected Behavior
The types should be consistent.

### Actual Behavior
The prop is of type any, while the event emits a string type.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNplUstqwzAQ/JVFl8TgR0vooalsWuglpx5yrHtQHTkYZElIa0Mx/vfqkWC7ve3OzI5mF31OxJqmeNM6HwdOjoTaxnQaQTB5LWuCtiZgOQ66qiVA12tlECYwvIUZWqN62LnB3UstPd8oaRH6n3eGHEqvor6s9okTLPTJqjOaTl5vGhuaqKJFTODecw3yXgvv4MfpmDEdg4TG2SHrJDc3KIAXp846qQe8gwBj1qsLF26fGK0mCyfYd2BC5Di3Yl8H7Q2PYT4bmRi40+5dkUBZ/VnFoTmq0/kjIvtkcaposY62BNaGV9N0P9k808Ij/+nlna3G+27v4JF4JlqszkfmlBzyp/zxQHzxnD+QtGXC8nTzAyL29Qu0PKwX)

OlTrenin:
## Description
fixes #21751 

## Markup:
```javascript
emits: {
    save: (value: any) => true,
    cancel: () => true,
    'update:focused': (val: boolean) => true,
    'update:modelValue': (val: any) => true,
    'update:menu': (val: boolean) => true,
  },
  
  function onSave (value: any) {
      emit('save', value)
      menu.value = false
    }
    
    function onUpdateModel (value: any) {
                          if (!props.hideActions) {
                            proxyModel.value = value
                          } else {
                            model.value = value

                            if (!props.multiple) {
                              menu.value = false
                            }
                          }

                          emit('save', value)

                          disabledActions.value = []
                        }

```


ikushum:
Instead of `any`, the correct type we should use here is `typeof VDatePicker['props']['modelValue']`

ikushum (changes requested):
Instead of `any`, the correct type we should use here is `typeof VDatePicker['props']['modelValue']`

OlTrenin:
> Instead of `any`, the correct type we should use here is `typeof VDatePicker['props']['modelValue']`

I'm change, thanks you

ikushum (approved):
Looks good

ikushum (approved):

J-Sek (changes requested):
It does not really work

Image

Image

OlTrenin:
do you mean need to replace VDateInputModelValue with unknown and refactor VDateInput?
Or have you already done it?

J-Sek:
Feel free to adjust your PR if you think you can get it to work. I am no TS wizard to tell if my way is the only solution. I have only copied code from VDatePicker and it feels like it satisfies the requirements - the event type matches `modelValue`.

OlTrenin:
I checked it, and it seems that my code also solves the problem

J-Sek:
Can we exchange screenshots on Discord? I mean.. I just shared mine, but we can discuss details, like IDE, TS version, anything that would explain why I cannot see it working and you can.

OlTrenin:
ok, bro
In runtime the type is really lost

KaelWD:
`typeof VDatePicker['props']['modelValue']` is still `any` because `VDatePicker['props']` is generic.

PR #21759, opened 2025-07-14
fix(useMask): handle number input in maskText function

Joao-Camilo-Mallmann:
**_fix(VMask): Ensure numeric values work correctly with v-model_**

Description:

This PR addresses an issue in the v-mask-input component that occurs when using a v-model bound to a number type.

Currently, if the initial value provided to the mask is a number, the maskText function fails when attempting to execute string operations, such as .trim(), resulting in an error and preventing the mask from being applied correctly.

The proposed solution modifies the maskText function to also accept the number type. A check has been added at the beginning of the function to convert any numeric value to a string before proceeding with the mask processing. This ensures compatibility and the expected behavior of the component without breaking existing functionality for string values.

How to Reproduce the Issue:

The error can be replicated by using a v-mask-input with the v-model initialized as a number, as shown in the example below:

Pull request

Before:
Pull request

### DIFF
--- a/packages/vuetify/src/composables/mask.ts
+++ b/packages/vuetify/src/composables/mask.ts
@@ -106,7 +106,9 @@ export function useMask (props: MaskProps, inputRef: Ref<HTMLInputElement | unde
     return item.convert ? item.convert(char) : char
   }
 
-  function maskText (text: string | null | undefined): string {
+  function maskText (text: string | number | null | undefined): string {
+    if (typeof text === "number") text = String(text);
+
     const trimmedText = text?.trim().replace(/\s+/g, ' ')
 
     if (trimmedText == null) return ''
--- a/packages/vuetify/src/composables/mask.ts
+++ b/packages/vuetify/src/composables/mask.ts
@@ -107,7 +107,7 @@ export function useMask (props: MaskProps, inputRef: Ref<HTMLInputElement | unde
   }
 
   function maskText (text: string | number | null | undefined): string {
-    if (typeof text === "number") text = String(text);
+    if (typeof text === 'number') text = String(text)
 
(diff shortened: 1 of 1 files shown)


J-Sek:
Actual markup would be useful to understand the use case. Please take a look at other open PRs to see and adapt to the standards.

I am guessing this is the use case:

```html
<v-mask-input mask="####" v-model.number="year" />
```

`year` type becomes `string` when the value is cleared out (when I delete all the digits).

<details>
<summary>Full playground</summary>

```vue
<template>
  <v-app>
    <v-container>
      <v-mask-input mask="####" v-model.number="year" />
      <pre>year: {{ year ?? 'null' }}</pre>
      <pre>type: {{ typeof year }}</pre>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref } from 'vue'
  const year = ref<number | null>(2020)
</script>
```

</details>

TBH, the casting is simple, but I have a feeling, it does not fit there. It makes more sense for this component to be strictly string-based to keep it simple. Having 2-way transformation with computed is both easy to understand and good for debugging issues.

<details>
<summary>Example target solution</summary>

```vue
<template>
  <v-app>
    <v-container>
      <v-mask-input mask="####" v-model="yearText" />
      <pre>year: {{ year ?? 'null' }}</pre>
      <pre>type: {{ typeof year }}</pre>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref, computed } from 'vue'

  const year = ref<number | null>(2020)
  const yearText = computed({
    get: () => year.value ? String(year.value) : null,
    set: (v: string) => year.value = v ? Number(v) : null,
  })
</script>

```

</details>

Also.. is there a probem with `v-number-input`? Year looks like a good fit for that component rather than a masked input.

PR #21770, opened 2025-07-16
chore(VPagination): replace cypress with vitest

J-Sek:
- migrates tests for VPagination (kept as-is, no meaningful changes)


packages/vuetify/src/components/VPagination/__tests__/VPagination.spec.browser.tsx
```diff
@@ -0,0 +1,219 @@
...
+    expect(screen.getAllByCSS('.v-pagination__item')).toHaveLength(6)
+  })
+
+  it('should render in RTL mode', () => {
+    render(() => (
+      <VLocaleProvider rtl>
+        <VPagination length="5" />
+      </VLocaleProvider>
+    ))
+  })
```
KaelWD: This test doesn't actually do anything

KaelWD (approved):

PR #21779, opened 2025-07-17
chore(VSlider): replace cypress with vitest

J-Sek:
- migrates tests for VSlider
- ~~`swipe(element, from, to)` inspired by [example](https://testing-library.com/docs/example-drag/) from the `testing-library`~~


KaelWD (commented):
- fireEvent doesn't use real events, only userEvent does
- there's already a `drag` command but it's relative to the viewport not a specific element: https://github.com/vuetifyjs/vuetify/blob/e5e85ee2faeffe7b39a3fbaaedbcb43f205943b3/packages/vuetify/src/directives/touch/__tests__/touch.spec.browser.tsx#L39 https://github.com/vuetifyjs/vuetify/blob/efaaa5d4f02e886b3e0c46e1f9dd9eed1cdf2e88/packages/vuetify/test/setup/browser-commands.ts#L18-L29

KaelWD (approved):

PR #21781, opened 2025-07-18
fix(VOtpInput): focus jumps to the end when correcting filled input

### LINKED ISSUE #21680: [Bug Report][3.8.9] v-otp-input jumps focus to the end when correcting input
### Environment
**Vuetify Version:** 3.8.12
**Vue Version:** 3.5.17
**Browsers:** Chrome 137.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
1. Enter a complete OTP (e.g., `563412`)
2. Try to correct the first digit by replacing it with another digit
3. Notice that focus jumps to the last digit

### Expected Behavior
After editing a digit, the focus should stay on the next one in line — not jump to the end unless I reach the end naturally. For example, if I fix the first digit, focus should move to the second, not to the end.

### Actual Behavior
When editing a filled OTP input like `5 6 3 4 1 2`, if I try to change the first digit (e.g., replace `5` with `6`), the component updates to `6 6 3 4 1 2`, but focus jumps to the last input. This makes it frustrating to correct codes manually — I have to keep navigating back after each correction.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdkMFqwzAQRH9lq4tbsGycNG0JTqG3/kEPdQ7CXgeBZC3S2gRC/j2SnYOT2+7wmJnd/4sIvi1/iIppRLEXNaMloxi/mwGgnqQimsd5ad3ASg/o79IsMp5Z9hpNB5O0rkNzaIQNp0ZAueYck9QDjQz7mZKTMiNGttps33cfK7wun6OSsjSpy1XDuIbWa2IIyONSVFtynuECHnu4Qu+dhSwelyUcIPoGhlgPDol4zX7RGAd/zpvuJXtLAYtlNBPXXGyLXVF9ijR8FdVG5L0yAfOHty3a8Qatpmpt)

partment:
fixes #21680 

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-otp-input :model-value="123456" />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -190,7 +190,7 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
 
       model.value = clipboardText.split('')
 
-      inputRef.value?.[finalIndex].focus()
+      focusIndex.value = finalIndex
     }
 
     function reset () {
@@ -225,10 +225,7 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
     }, { scoped: true })
 
     watch(model, val => {
-      if (val.length === length.value) {
-        focusIndex.value = length.value - 1
-        emit('finish', val.join(''))
-      }
+      if (val.length === length.value) emit('finish', val.join(''))
     }, { deep: true })
 
     watch(focusIndex, val => {
--- a/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
+++ b/packages/vuetify/src/components/VOtpInput/VOtpInput.tsx
@@ -225,7 +225,9 @@ export const VOtpInput = genericComponent<VOtpInputSlots>()({
     }, { scoped: true })
 
     watch(model, val => {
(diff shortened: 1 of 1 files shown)


partment:
Would it be possible to get an update on when this might be merged?

Dyskal:
@partment the problem is that the mobile autofill doesn't move the focus input, this was the reason of my fix
Screenshot_20250903_180802_Chrome
Screenshot_20250903_180843_Chrome

J-Sek:
@Dyskal, it seems to work when pasting values in desktop browser.
Do you experience regression on mobile?

Dyskal:
@J-Sek yeah it's only on mobile, the autofill from the keyboard (Gboard here, also on others) is handled strangely on web, pasting is correct tho

partment:
@Dyskal Pasting through Gboard will not fire onPaste event and backspacing when the character in the input is selected will not fire keydown event with keyCode "Backspace" but instead "Unidentified".
These two problem are probably chromium's bug I think, needs some workaround.

PR #21785, opened 2025-07-18
chore(VProgressLinear): replace cypress with vitest

J-Sek:
- migrates tests for VProgressLinear
- `indeterminate` is included among Percy snapshot.. I am not 100% sure it should though


KaelWD:
Can't have anything animated in percy snapshots, the timing isn't consistent so they'll be flaky

PR #21795, opened 2025-07-22
chore(VLayout): replace cypress with vitest

KaelWD:


packages/vuetify/src/composables/layout.ts
```diff
@@ -286,13 +286,16 @@ export function createLayout (props: { overlaps?: string[], fullHeight?: boolean
         const isHorizontal = position.value === 'left' || position.value === 'right'
         const isOppositeHorizontal = position.value === 'right'
         const isOppositeVertical = position.value === 'bottom'
-        const size = elementSize.value ?? layoutSize.value
+        const size = Number(elementSize.value ?? layoutSize.value)
+        const transformFunction = `translate${isHorizontal ? 'X' : 'Y'}`
+        const transformValue = active.value ? 0
+          : (size === 0 ? 100 : size + 1) * (isOppositeHorizontal || isOppositeVertical ? 1 : -1)
```
KaelWD, line 292: I added `size + 1` here, there was actually still 1px visible on screen before

packages/vuetify/src/components/VLayout/__tests__/VLayout.spec.browser.tsx
```diff
@@ -25,50 +24,54 @@ describe('VLayout', () => {
...
           </div>
         </VMain>
       </VLayout>
     ))
 
     await scroll({ top: 1000 })
-
-    // TODO: figure out how to make it work
-    expect(screen.getByCSS('nav')).toBeVisible()
+    await expect(screen.getByTestId('sticky')).toBeOnScreen()
```
J-Sek: So I see the problem I had previously was a duplicate `nav` in the DOM.

Other changes are welcome and make it easier to copy-paste code to VPlay for quick verification. But I don't trust the `.toBeOnScreen()`. I have removed `top: ...` from that `nav` and it still passes - so it does not detect VAppBar obstructing inner sticky element.

J-Sek: Same with the test case below. It only fails when the `.v-main__scroller` is missing, but `.toBeOnScreen()` passes ignoring the fact nav item can be effectively invisible.

PR #21857, opened 2025-07-30
docs(VAutocomplete): add filter-keys prop section and example

### LINKED ISSUE #21799: [Bug Report][3.9.2] VAutocomplete filter not work as expected when add filter-keys prop
### Environment
**Vuetify Version:** 3.9.2
**Vue Version:** 3.5.17
**Browsers:** Chrome 138.0.0.0
**OS:** Windows 10

### Steps to reproduce
set filter-keys props with key string lead to stop filtering

### Expected Behavior
when set filter-keys prop the filter should work  without add raw. string

### Actual Behavior
filter not work when add filter keys

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrdVMGOmzAQ/ZURF7cSBCVpD0XsSpX6F0kOXhg21hrbsg00ivLvHWO2MVG7t156ANvz3oxn3jAcrpmzTfndmM04YFZltcfeSO7x+agA6rFouG3nfTxp5blQaKGTg3gHZsjq6fdx4Uqgxz0ds+3umCUgwQYayV3APP70hcX2gQHgkNvmDEp7mLR9g+mMioweOiE92uINLw4mf9aDB2PRoGrB8ml1TWnW144FH7xuNNWIHtf3VUlcyuvARMtypniPtPC2tegcOx2ztZcgwQovvETyCeQ/EkYuh0AQVOYaHotetygJmznuEa+Cf8hnXh9RyV9m3x/Y8UH6Nfxcl+t6SYuPmvBqEdXf2vBhC/6N/hRsM/cgbJY+hO3/0ou7SvOw3IeJzvdpiugydnEqgyWOZV0m80pH11hhfGjREJUXvdHWwxUsdnCDzuoeGI06i58CBXZ+lsXBExyC6SraapsH9SrGpWD5InfF8PViPLvl76zdwvL0tgmv4Yq3PCHuF6Lj8pLw5OVFpLQvC03pM09oIbpoAvGUJh1bRFlTZZ+2n6NYKbRLsLqMypAm2S3P9puvm+0+C5tvm12Wd1w6zFe/wmg7/QLFSYmt)

### Other comments
:filter-keys="['key1',,'key2']" not work
:filter-keys="['raw.key1',,'raw.key2']" work

tommasogangemi:
Closes #21799

## Description

Added section to the documentation page explaining the behavior of the `filter-keys` prop with related example

### DIFF
--- a/packages/docs/src/examples/v-autocomplete/prop-filter-keys.vue
+++ b/packages/docs/src/examples/v-autocomplete/prop-filter-keys.vue
@@ -1,0 +1,34 @@
+<template>
+  <v-container fluid>
+    <v-autocomplete
+      :filter-keys="['title', 'raw.abbr']"
+      :items="states"
+      item-title="name"
+      label="State"
+    ></v-autocomplete>
+  </v-container>
+</template>
+
+<script setup>
+  const states = [
+    { name: 'Florida', abbr: 'FL' },
+    { name: 'Georgia', abbr: 'GA' },
+    { name: 'Nebraska', abbr: 'NE' },
+    { name: 'California', abbr: 'CA' },
+    { name: 'New York', abbr: 'NY' },
+  ]
+</script>
+
+<script>
+  export default {
+    data: () => ({
+      states: [
+        { name: 'Florida', abbr: 'FL' },
+        { name: 'Georgia', abbr: 'GA' },
+        { name: 'Nebraska', abbr: 'NE' },
+        { name: 'California', abbr: 'CA' },
+        { name: 'New York', abbr: 'NY' },
+      ],
+    }),
+  }
+</script>
--- a/packages/docs/src/pages/en/components/autocompletes.md
+++ b/packages/docs/src/pages/en/components/autocompletes.md
@@ -68,7 +68,7 @@ The `custom-filter` prop can be used to filter each individual item with custom 
 
 #### Filter keys
 
-The `filter-keys` prop can be used to filter items by properties other than the default `title`. Properties of the original items passed need to be accessed via the `raw.*` path, as `filter-keys` index the root level of `InternalItem`.
+When user is typing in the field to narrow the list of options, the input text is matched against the `title`. With `filter-keys` you can specify which properties should be used instead. Properties of original objects passed to `items` need to be accessed via the `raw.*` path, as `filter-keys` index the root level of `InternalItem`.
 
 <ExamplesExample file="v-autocomplete/prop-filter-keys" />
 
--- a/packages/docs/src/pages/en/components/autocompletes.md
+++ b/packages/docs/src/pages/en/components/autocompletes.md
@@ -66,6 +66,12 @@ The `custom-filter` prop can be used to filter each individual item with custom 
 
 <ExamplesExample file="v-autocomplete/prop-filter" />
 
+#### Filter keys
+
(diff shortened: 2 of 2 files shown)


packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
```diff
@@ -144,7 +144,12 @@ export const VAutocomplete = genericComponent<new <
         : model.value.length
     })
     const form = useForm(props)
-    const { filteredItems, getMatches } = useFilter(props, items, () => isPristine.value ? '' : search.value)
+    const { filteredItems, getMatches } = useFilter(
+      props,
+      items,
+      () => isPristine.value ? '' : search.value,
+      { transform: item => item.raw }
```
tommasogangemi: Tests are failing because when items are primitive this is not working. A solution might be something like:

```ts
{ transform: item => typeof item.raw === 'object' ? item.raw : item }
```

But there might be some better solution, I'll think of it and try to propose something else if comes to mind

Edit: this doesn't even work, because the raw object might not contain the title property, which is the default filterKey

tommasogangemi:
Update: I changed approach and ended up with a solution that looks to be working in all cases, other than the minor misbehavior documented [here](https://github.com/vuetifyjs/vuetify/pull/21857/files#diff-3df71dcef30db080ec919186fd82f35bba0ccc01932ada675a01ad776ace1296R75).

There might be a cleaner solution and I am open to suggestions. Refactors should be quite safe given the couple of tests I added!

J-Sek:
> _minor misbehavior_

I might have a `title` prop like `..., title: 'Mr', ...` or a professional title in a medical app - it will just break the search. I mean.. `raw` cannot just override the `internalItem` here.

Besides, it looks like we agreed with a couple of people to have `*.raw.*` thing in #17487. It would be a plain knife-in-the-back breaking change for them.

I believe we should rather improve the description for the prop and create an additional example.

tommasogangemi:
Hey thanks for the review! I feared that the proposed change wouldn't be acceptable and it makes total sense. 

I think the biggest confusion comes from the fact that `InternalItem` is, as the name says "internal", and invisible to the end user of the component Therefore the `filterKeys` kind of only makes sense (from a user perspective) if used to filter on the raw item since it wouldn't make much sense to filter based on keys of `ListItem` or `InternalItem`.

A solution might be to not add the filterKeys with a default on the `title`, but change it as an opt-in feature, that when passed directly and explicitly works on the raw item, while when left undefined it works with the `InternalItem` title, but this might be a breaking change as well in some cases, and there might be some intricacies that I am not seeing. Of course the highlight should be also adjusted (removed?) in this case.
 
What do you think about that? Should I just scrap this PR for now and update the doc about `filterKeys` so that it's clearer for now? Thanks!

J-Sek:
> _What do you think about that?_

We would not score points for breaking simple use cases to help with advanced ones.

> _Should I just scrap this PR for now and update the doc about `filterKeys` so that it's clearer for now?_

I feel it would be beneficial to the users. A single example for VAutocomplete should be fine. Can you make it so, the use case feels rational? I mean, it searching in invisible fields feels pointless.. maybe it should be combined with slots (e.g. list of tasks and a task status as VChip in the append slot). Utilizing country code in country selector would also count - e.g. typing US to find 'United States' on the list.

tommasogangemi:
Updated according to suggestions, let me know if it still needs something!

PR #21893, opened 2025-08-07
chore: replace tsc with tsgo

KaelWD:
Would be nice to have this for the docs too but https://github.com/vuejs/language-tools/issues/5381

Blocked by https://github.com/microsoft/typescript-go/issues/1034


KaelWD:
I can't get it to work at all in vscode, webstorm and build scripts are great though.

Issue #21900, opened 2025-08-08
[Bug Report][3.9.4] Setting the initial search value via `searchInput` on `VCombobox` does not work


jhadenfeldt:
### Environment
**Vuetify Version:** 3.9.4
**Vue Version:** 3.5.18
**Browsers:** Chrome 139.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Set the `search-input` prop on `VCombobox`.

### Expected Behavior
Have the initial value of the combobox be set and the results filtered by that value.

### Actual Behavior
The initial value of the combobox is unset and the results aren't filtered.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdkMFqwzAQRH9l0UUtODalLbTBKZRA+wOFHuIcFGftCiytWMkhJeTfK1kJTXKbfRpGO7s6CM9t9e5cuRtRzEUd0LhBBXxrLEC9mynnJjkNLdmgtEU+oRM0G9rQ/owA5h4Vtz8zbd0YFo0wvm/ExbOOn/jIV3KpBt0RW61kAXJJA7HaUtIfUerthD+RuM+OL9wrn8T3Lxlte7n+D67Oe1a3iyaSe9TVRb84+pa1C+AxjLmmNo44wAEYOzhCx2RAxtPIZAeIuT5A7AOL5LibCsj7FJyjYog4FuKxfC4fXkQSr+WTKDo1eCyubp3Z+g+e+XmD)

### Other comments
I believe this is supposed to work - I'm still seeing [a test](https://github.com/vuetifyjs/vuetify/blob/1a0a9e6f19d776fe104caaae4c659a3f6d3d3e55/packages/vuetify/src/components/VCombobox/__tests__/VCombobox.spec.ts#L72) that uses it, but am not able to set the initial value of the `VCombobox`.

jcjp:
The props `search-input` is not part of the `v-combobox` you should use `v-model` instead to set the initial value of the component.

<img width="510" height="657" alt="Image" src="https://github.com/user-attachments/assets/2de0fafd-8cfc-40c7-9cad-fe9ff8f27ad6" />

But even if there is initial value it doesn't filter based on initial value.

KaelWD:
Reverted due to #22077, follow #22060

PR #21901, opened 2025-08-09
fix(VCombobox): check if search value is not empty, if yes filter immediately (#21900)

jcjp:
## Description

Fixes: #21900 

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-combobox
        v-model="msg"
        :items="['Colorado', 'Florida', 'Georgia', 'Texas', 'Wyoming', 'California']"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref('Cali')
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -188,7 +188,7 @@ export const VCombobox = genericComponent<new <
         : (props.multiple ? model.value.length : search.value.length)
     })
 
-    const { filteredItems, getMatches } = useFilter(props, items, () => isPristine.value ? '' : search.value)
+    const { filteredItems, getMatches } = useFilter(props, items, () => isPristine.value && !_search.value ? '' : search.value)
 
     const displayItems = computed(() => {
       if (props.hideSelected) {
@@ -586,7 +586,7 @@ export const VCombobox = genericComponent<new <
                                 </>
                               ),
                               title: () => {
-                                return isPristine.value
+                                return isPristine.value && !search.value
                                   ? item.title
                                   : highlightResult('v-combobox', item.title, getMatches(item)?.title)
                               },


J-Sek (changes requested):
When placed side-by-side with VSelect and VAutocomplete, new behavior feels a bit off. - before: I can type `44` (no match) but still re-open full list
- after: list does not appear and it might feel unexpected/broken

[demo](https://play.vuetifyjs.com/#eNrNU8tuwjAQ/JWVL7SSk5DyUItCpd77B8DBSTYiqhNbtkOREP/ejRF5od45xZ4Zj2ej8e7CrMmiL63DU4NswxKHlZbC4ee+BkhOgdAa3BEr3O5ZLszPnnnGc5mqnShrNFCJc/Bb5u5IqvV83oluMmHybj85WMimHJKetigxc0MQQIoUJbl3LBvzm5KSWxLYozLuu7QPilNQqdx7+O+UziQKI1KJY1ijsWSGtQu8YshG0+CZqlKVqvN/0Tv+GcOLxikKqCW6iU0/wEjzZEMk0aBYff9atC/go6hFqOS0TqJB+WlrM1NqBxZd09IAZaVpMriAwQKuUBhVwYzezawlydWS+D47bGE35xBzeOOw4LDksOKwJoSgmLCYwJjQeHVob7sb+D9Dh+mKl+VrG+oWgwKwK2eLcBXG76xdfIRLxgshLfLRI75hhz8anxj9)

The original problem is the difference between typing the value and setting it programmatically. It would be best to target this specifically and leave the rest as-is.

J-Sek:
Alternative fix:

```diff
watch(model, value => {
  if (!props.multiple && !hasSelectionSlot.value) {
-    +search.value = value[0]?.title ?? ''
+    search.value = value[0]?.title ?? ''
  }
-})
+}, { immediate: true })
```

Breaks some tests... I don't have time to dig it this week. @jcjp, can you evaluate it?

jcjp:
> Alternative fix:
> 
> ```diff
> watch(model, value => {
>   if (!props.multiple && !hasSelectionSlot.value) {
> -    +search.value = value[0]?.title ?? ''
> +    search.value = value[0]?.title ?? ''
>   }
> -})
> +}, { immediate: true })
> ```
> 
> Breaks some tests... I don't have time to dig it this week. @jcjp, can you evaluate it?

I updated the solution, we are running into test issues whenever we try to update the `search.value` directly on the watch so instead I just emit an `update:search` for the new value do you think about this, is it okay? I updated my sample code above to use your example on your demo link.

No errors on the test as well, here is the result of the unit test.

#### VCombobox `run vitest VCombobox`
image

#### Full run `run vitest`
image

J-Sek:
`emit(..` does not break the tests, but also does not seem to have the desired effect - i.e. does not filter the list.

Anyway, my suggestion has the same problem as the previous attempt. Menu does not appear when there is no match (e.g. for `msg = ref("xx")`) which is inconsistent with typing those letters manually. We need to dig deeper.

jcjp:
> `emit(..` does not break the tests, but also does not seem to have the desired effect - i.e. does not filter the list.
> 
> Anyway, my suggestion has the same problem as the previous attempt. Menu does not appear when there is no match (e.g. for `msg = ref("xx")`) which is inconsistent with typing those letters manually. We need to dig deeper.

~~I recorded my test, I want to clarify if is this not the desired behavior?~~ EDIT: You are right as per my checking will dig deeper thanks Jacek!

J-Sek:
@jcjp, I have added a test against regression (re-opening menu after typing `cal` shows all options) and a second one for the new behavior we'd like to observe. Currently, both fail because re-opening does not preserve narrowed options.

Initially, I wrote those tests to be based on keyboard interactions (Tab » type » Escape » ArrowDown), but it looks like the options are only preserved if the user clicks outside... Clicking on the menu icon, Tab, Escape - all other methods of closing - seem to clear the "search". With your changes, click-outside + reopen aligns with all of the other methods, but I don't think it is a desired behavior... It is a tough battle. If you can make these new tests pass, we could merge it and handle the re-opening inconsistency later. Ideally, it should not matter if the user opens the menu for the first or n-th time with any of the methods.

jcjp:
> @jcjp, I have added a test against regression (re-opening menu after typing `cal` shows all options) and a second one for the new behavior we'd like to observe. Currently, both fail because re-opening does not preserve narrowed options.
> 
> Initially, I wrote those tests to be based on keyboard interactions (Tab » type » Escape » ArrowDown), but it looks like the options are only preserved if the user clicks outside... Clicking on the menu icon, Tab, Escape - all other methods of closing - seem to clear the "search". With your changes, click-outside + reopen aligns with all of the other methods, but I don't think it is a desired behavior... It is a tough battle. If you can make these new tests pass, we could merge it and handle the re-opening inconsistency later. Ideally, it should not matter if the user opens the menu for the first or n-th time with any of the methods.

I updated the solution and looks the issue is fixed, thanks for reviewing and adding the test.

packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -222,12 +225,28 @@ export const VCombobox = genericComponent<new <
...
+          isPristine.value = hadNoMatchOnLastOpen.value ? true : !search.value
+        }
+        hasEverOpened.value = true
+      }
+    })
+
     watch(model, value => {
       if (!props.multiple && !hasSelectionSlot.value) {
         _search.value = value[0]?.title ?? ''
+        emit('update:search', value[0]?.title ?? '')
```
J-Sek: Removing this line seems to keep the behavior we need. Could you check if it is still necessary? Looks like a duplicate emit.

J-Sek:
Overall, it looks great. Just anticipating a bug report below.

<details>
<summary>Recording</summary>

[Screencast from 2025-08-23 19-58-20.webm](https://github.com/user-attachments/assets/53dff6ae-f956-4e72-bea0-858c9e9b3c6a)

</details>

johnleider:
> Overall, it looks great. Just anticipating a bug report below.
> 
> Recording
>  Screencast.from.2025-08-23.19-58-20.webm

Recording won't load for me.

J-Sek:
@johnleider, when the text does not match any of the list options, every 2nd time I open the menu it either shows all options or a "No data available". So I am clicking menu icon to toggle it on/off and it consistently cycles between both of these results.

<details>
<summary>Screenshots</summary>

image

image

</details>

It would be great if we could avoid it.

johnleider:
The video is loading for me now, thank you.

packages/vuetify/src/components/VCombobox/VCombobox.tsx
jcjp: Remove this duplicate emit thanks!

jcjp:
I fixed the bug but there is a flicker animation happening when we blur the input field, looks annoying let me know if you guys are okay with this.

J-Sek:
- [x] opening first time with non-matching value should show all options (consistently, not only on subsequent attempts)
- [x] when I defer loading options (e.g. 2s delay), open, close, wait 2s, open - shows all items, not the matching one

I will attempt a cleanup. There are too many boolean flags scattered all over the place and it is making the whole thing very brittle.

J-Sek:
@jcjp, the last commit has changed quite a lot.
Could you please double-check for any regression?

J-Sek (approved):

jcjp:
> @jcjp, the last commit has changed quite a lot. Could you please double-check for any regression?

I did the test manually looks good thank you for your help on the flicker issue!

PR #21943, opened 2025-08-22
feat(VCardActions): add new "tag prop to improve semantics on cards

edfrias:
## Markup:

```vue

```

### DIFF
--- a/packages/vuetify/src/components/VCard/VCardActions.tsx
+++ b/packages/vuetify/src/components/VCard/VCardActions.tsx
@@ -1,14 +1,20 @@
 // Composables
 import { makeComponentProps } from '@/composables/component'
 import { provideDefaults } from '@/composables/defaults'
+import { makeTagProps } from '@/composables/tag'
 
 // Utilities
-import { genericComponent, useRender } from '@/util'
+import { genericComponent, propsFactory, useRender } from '@/util'
+
+export const makeVCardActionsProps = propsFactory({
+  ...makeComponentProps(),
+  ...makeTagProps(),
+}, 'VCardActions')
 
 export const VCardActions = genericComponent()({
   name: 'VCardActions',
 
-  props: makeComponentProps(),
+  props: makeVCardActionsProps(),
 
   setup (props, { slots }) {
     provideDefaults({
@@ -19,15 +25,14 @@ export const VCardActions = genericComponent()({
     })
 
     useRender(() => (
-      <div
+      <props.tag
         class={[
           'v-card-actions',
           props.class,
         ]}
         style={ props.style }
-      >
-        { slots.default?.() }
-      </div>
+        v-slots={ slots }
+      />
     ))
 
     return {}


J-Sek:
Please think about a bigger picture
- do we need the same treatment for VCardItem so it can have `tag="header"`?
- are there any other components beside VCard (VConfirmEdit, VPicker, come to mind... anything else?)
  - it does not mean they should be covered, but I'd like to understand next steps

Markup section is for a playground snippet including `<v-card>` with actions that utilize `tag`, so we can grasp what actually changes from the developer perspective. You can lookup recently merged PRs to understand the communication standards.

edfrias:
Hey @J-Sek !

Thank you so much for your feedback. I understand all your points and I think they are spot on. Here are my thoughts:

If I would like to have a header, I think I didn't use VCardItem as we already have VCardTitle that can change its tag for a proper "header" tag. 

But if you want a footer section into a Card, you wouldn't use a VCardItem right? Instead you should go for a VCardActions component as it is the proper component to do the job. 

So right now you don't have the chance to add semantics unless you properly enclose the component like so:

```
<footer>
  <v-card-actions>
    <v-btn>My Button</v-btn>
  </v-card-actions>
</footer>
```
While VCardItem and VCardTitle give you some flexibility, I feel VCardAction don't.

Sorry about the Markup section, I will update it right now!

J-Sek:
> If I would like to have a header, I think I didn't use VCardItem as we already have VCardTitle that can change its tag for a proper "header" tag.

While it works for your use case and specific design utilizing only the title, I wish the PR to consider wider audience. `VCardItem` is a bit unfortunate naming. I think it should have "*Header" suffix instead of "*Item". It is meant to contain title, subtitle, icon, avatar and top right actions. I am no expert on HTML semantics, but all of these elements seem to fit inside the `<header>`, not just the title.

BTW. I see `<main>` around the `<v-card-text>` in the updated markup section... Why not `tag="main"`?

edfrias:
Hey @J-Sek ,

Indeed, I've updated the markup section to better address the use case.

I think maybe all this stems from a vague use of the `VCardTitle` component.
 
While the `VCardItem` component should be doing the job, it doesn't allow you to change the HTML node that wraps it, while VCardTitle does, which prevents a title from being improperly wrapped inside a `<div>` and ensures proper semantics.

Should `VCardItem` allow you to improve its semantics by having a "tag" prop? I think so. This should let you to use it as a header of a card or as a container for a card's content outside of a header and inside a card's "body". 

If you ask me: `VCardItem` should be renamed into `VCardHeader`? I don't think so. `VCardHeader` seems to me like a more restricted component, while `VCardItem` is a more generic and reusable container that can be placed in different areas of the card.

Regarding your concerns, would you like me to update my pull request to cover the case where the `VCardItem` component had a prop "tag· that allowed it to change its semantics?

Please let me know what you think.

J-Sek:
> Regarding your concerns, would you like me to update my pull request to cover the case where the `VCardItem` component had a prop "tag· that allowed it to change its semantics?

Yes. I imagine using `tag` prop could be then specified in global configuration:

```ts
defaults: {
  VCardItem: { tag: 'header' },
  VCardText: { tag: 'main' },
  VCardActions: { tag: 'footer' },
},
```

and it will just work with:

```html
<v-card title="My Card" subtitle="Something" text="Main content">
  <template #actions>...buttons...</template>
</v-card>
```

resulting in:

image

edfrias:
Hello @J-Sek 

I just updated the PR to include the `VCardItem` component as we discussed. This way, we would now have a `VCardItem` component available to allow for more semantic headers while still being backwards compatible with the current codebase.

In its current state, `VCard` would already be much more semantic than what we have now without requiring changes outside of the framework.

On the other hand, what would you think if we left these changes here and proposed improving the `VCard` component in another PR? I feel we're backwards compatible so far, but if we make the changes you proposed in your last message, it would in some ways be a break from what was before.

Please let me know what you think.

J-Sek:
> if we make the changes you proposed in your last message

It's all fine. `defaults: { ... }` (_aka_ global configuration) is something meant to be controlled by the developers for every project. It was not my intention to replace current `div`-based defaults. We just made it possible.

edfrias:
Thanks for your help @J-Sek

PR #21946, opened 2025-08-22
feat(VAlert): High Contrast Support (Forced-colors mode)

06b:
## Description

This improves the appearance of the VAlert component when in forced-colors mode (For example, High Contrast Mode in Windows):
Screenshot 2025-08-22 102723

Since an alert is meant to get the user's attention, this will add a border (when forced-colors mode is active) to all alerts with the width reflecting based on the visual hierarchy of the variants. Since the elevated and flat variants appear to have the strongest visual hierarchy, they both have been grouped with a thick border. As text & plain have the weakest visual hierarchy they have been grouped with a thin border with the plain variant having a dashed border instead of a solid border since it stands out the least. 

_VAlert when not in forced-colors mode (High Contrast Mode in Windows):_
Screenshot 2025-08-22 102123

_VAlert Component when in forced-colors mode (High Contrast Mode in Windows) **before the changes in this pull request**:_
Screenshot 2025-08-22 102636

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-row dense>
        <v-col v-for="(variant, i) in variants" :key="i" cols="12">
          <v-alert
            :title="`Alert in ${variant}`"
            :variant="variant"
            text="Lorem ipsum dolor sit amet consectetur adipisicing elit. Commodi, ratione debitis quis est labore voluptatibus! Eaque cupiditate minima, at placeat totam, magni doloremque veniam neque porro libero rerum unde voluptatem!"
            type="info"
          />
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    data: () => ({
      variants: ['elevated', 'flat', 'tonal', 'outlined', 'text', 'plain'],
    }),
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VAlert/VAlert.sass
+++ b/packages/vuetify/src/components/VAlert/VAlert.sass
@@ -147,14 +147,11 @@
 
 @media (forced-colors: active)
   .v-alert
-    border-style: solid
-
-    &--variant-plain
-      border-style: dashed
-
-    &--variant-text,
-    &--variant-plain
-      border-width: thin
+    &:not(
+      &--variant-text,
+      &--variant-plain
+    )
+      border-style: solid
 
     &--variant-outlined,
     &--variant-tonal
--- a/packages/vuetify/src/components/VAlert/VAlert.sass
+++ b/packages/vuetify/src/components/VAlert/VAlert.sass
@@ -144,3 +144,22 @@
     text-transform: $alert-title-text-transform
     word-break: $alert-title-word-break
     word-wrap: $alert-title-word-wrap
+
+@media (forced-colors: active)
+  .v-alert
+    border-style: solid
+
+    &--variant-plain
+      border-style: dashed
+
+    &--variant-text,
+    &--variant-plain
+      border-width: thin
+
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/components/VAlert/VAlert.sass
```diff
@@ -144,3 +144,22 @@
...
+
+@media (forced-colors: active)
+  .v-alert
+    border-style: solid
+
+    &--variant-plain
+      border-style: dashed
+
+    &--variant-text,
+    &--variant-plain
```
J-Sek: Apart from being duplicated here, `*-plain` primary purpose is to be unstyled. Maybe we should leave it (and `*-text`) without the border (similar to VCard)?

J-Sek (changes requested):

packages/vuetify/src/components/VAlert/VAlert.sass
06b: Removed border from text & plain variants. 
image

PR #21949, opened 2025-08-22
fix(VProgressLinear): ensure visibility when using custom colors

J-Sek:
## Description

before:

image

after:

image

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-row dense>
        <v-col>
          <h1 class="text-subtitle-1">Native Progress Indicator Element</h1>
          <progress id="file" max="100" value="70" />
          <br>
          <progress id="progress-bar" aria-label="Content loading…" />
        </v-col>

        <v-col>
          <v-defaults-provider :defaults="{ VProgressLinear: { height: 12 } }">
            <h1 class="text-subtitle-1">Vuetify Progress Indicator</h1>
            <v-progress-linear color="orange" model-value="20" />
            <br>
            <v-progress-linear color="green" indeterminate />
            <br>
            <v-progress-linear model-value="20" rounded />
            <br>
            <v-progress-linear indeterminate rounded />
            <br>
            <v-progress-linear
              v-model="value"
              :buffer-value="bufferValue"
              buffer-color="yellow"
              buffer-opacity=".9"
              color="purple"
            />
            <br>
            <v-progress-linear
              buffer-value="60"
              color="red-lighten-2"
              model-value="40"
              stream
            />
          </v-defaults-provider>
        </v-col>

      </v-row>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    data () {
      return {
        value: 10,
        bufferValue: 20,
        interval: 0,
      }
    },

    watch: {
      value (val) {
        if (val < 100) return

        this.value = 0
        this.bufferValue = 10
        this.startBuffer()
      },
    },

    mounted () {
      this.startBuffer()
    },

    beforeUnmount () {
      clearInterval(this.interval)
    },

    methods: {
      startBuffer () {
        clearInterval(this.interval)

        this.interval = setInterval(() => {
          this.value += Math.random() * (15 - 5) + 5
          this.bufferValue += Math.random() * (15 - 5) + 6
        }, 2000)
      },
    },
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VProgressLinear/VProgressLinear.sass
+++ b/packages/vuetify/src/components/VProgressLinear/VProgressLinear.sass
@@ -31,8 +31,8 @@
 
   @media (forced-colors: active)
     .v-progress-linear__buffer
-      background-color: highlight
-      opacity: $progress-linear-stream-opacity
+      background-color: highlight !important
+      opacity: .5 !important
 
   .v-progress-linear__content
     align-items: center
@@ -53,7 +53,7 @@
     background: $progress-linear-background
 
     @media (forced-colors: active)
-      background-color: highlight
+      background-color: highlight !important
 
   .v-progress-linear__determinate
     height: inherit
--- a/packages/vuetify/src/components/VProgressLinear/VProgressLinear.tsx
+++ b/packages/vuetify/src/components/VProgressLinear/VProgressLinear.tsx
@@ -14,7 +14,7 @@ import { makeThemeProps, provideTheme } from '@/composables/theme'
 
 // Utilities
 import { computed, Transition } from 'vue'
-import { clamp, convertToUnit, genericComponent, IN_BROWSER, propsFactory, useRender } from '@/util'
+import { clamp, convertToUnit, genericComponent, propsFactory, useRender } from '@/util'
 
 type VProgressLinearSlots = {
   default: { value: number, buffer: number }
@@ -98,7 +98,6 @@ export const VProgressLinear = genericComponent<VProgressLinearSlots>()({
     const normalizedValue = computed(() => clamp(parseFloat(progress.value) / max.value * 100, 0, 100))
     const isReversed = computed(() => isRtl.value !== props.reverse)
     const transition = computed(() => props.indeterminate ? 'fade-transition' : 'slide-x-transition')
-    const isForcedColorsModeActive = IN_BROWSER && window.matchMedia?.('(forced-colors: active)').matches
 
     function handleClick (e: MouseEvent) {
       if (!intersectionRef.value) return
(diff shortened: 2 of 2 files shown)


J-Sek:
@06b, what's your opinion about this?

06b:
@J-Sek Setting `!important` on the `background-color` would allow for you to remove the isForcedColorsModeActive check in VProgressLinear.tsx — one benefit to this is that by removing the isForcedColorsModeActive check it would allow the visibility of the component in forced-colors mode without having to refresh the page if a user happens to already be on a page with the component before enabling forced-colors (Contrast themes within Windows Settings > Accessibility), this scenario could exist when using a public computer such as in a library. 

As for changing the `opacity`, from `$progress-linear-stream-opacity` to `.5 !important` — perhaps set it to `opacity: $progress-linear-stream-opacity !important` instead?

J-Sek:
> perhaps set it to `opacity: $progress-linear-stream-opacity !important` instead?

The problem is that the value is `.3` by default - this is not enough. I need it to be visible without forcing developers to change a value for the whole app and all regular users. If you think `.5` needs to be configurable, we could add a new Sass variable with `-forced-colors` suffix, yes... However, I wonder if by saying "_it should exist because someone would not be happy with `.5`_" we actually claim it is still too low and should be increased.. which we can just do right away and avoid unnecessary variable.

06b:
@J-Sek I understand, in my opinion I don't think a new Sass variable would be needed.

Since contrast perception varies between people, we could make use of the `prefers-contrast` media feature to adjust dynamically instead. 

Below shows your suggested `opacity: 0.5`, if a user prefers a higher contrast it uses `opacity: 0.6` instead, if `.5` is too high for a user and they would prefer a lower contrast then it uses `opacity: 0.4`
Contrast-State

https://play.vuetifyjs.com/#eNrFVO1ugjAUfZWmf9wSAfErGWNme465mAIXISstaSuJMb77bmGoxI85Y7Y/pL09557Tk0vfN1Sr2HsrS7daAQ1oaKAoOTMwmwtCwiSv6gUuM38Wevj53lZOqeRSgdYOzwUw1dQJiVZpCsqpGF/By5wO5rQ9iSWXCksKEuQsMwPCGe6PtVHAinYXZGAhCPd3LdDAkWzrJ1LEu4M3A4zvq4VMgO/gwwP835udnHAbr5n4T0/TE56kYmIJ5zIc3yfD0GsmM/QO5hW32qx5M7ruEXGxaNyTTdNclizOzTogA3fybEtb24KQ1wKSnJGHUgGitRNLYRTTJsCrKHhs6T8LdCSmtUQtcoUSx663Ko1PKGV+EESQov+WkuQac0NCxGX8ed39hUQjtggi3gdx3NsOg8C/Gym9Dqeu2169X6XRyf2ynIXeqNLJ/LKKhZ5VCb12Cum2T0fuxPVH1C6e3Cntp4xr6Hde3ab28QWeIb+3

J-Sek:
Yes, this means Sass variable (build time knob) has no utility. Devs looking to adapt to contrast preference will write additional CSS anyway. v4 might open the door for more centralized theme/colors management and `prefers-contrast` media queries could land somewhere there. For now, I will stick to a single value for simplicity. Thanks for the input 👍🏼 .

PR #21999, opened 2025-09-03
fix(VTextarea): hide autoGrow scrollbar until maxRows is reached

### LINKED ISSUE #21982: [Bug Report][3.9.6] VTextarea auto-grow grows too much when text width is similar to the textarea width
### Environment
**Vuetify Version:** 3.9.6
**Vue Version:** 3.5.20
**OS:** Windows

### Steps to reproduce
1. Have auto-grow v-textarea and type text that is about the same width as the textarea

### Expected Behavior
v-textarea height matches the text

### Actual Behavior
The height is twice the text

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNp1kMFOwzAQRH9ltSeQkjhQgSBKK/EdhMOSbCGSG1v2Ji1C/DvrlCjl0FN2Zl/GY79+YwytefG+mEbGCmvhg7ckvGsGgHrKgztClC/L2waPfSefFWzK0p8anImZaZ39E7MUPgkFpsUCqA6uY5tPZMeUk4AG1/VEoadBdLPvreXuckejuPxDW6yWiqjs3YrtarMeu/RSaymWZv1L59pcXFBlbEPvBSLL6Ge0dUMUSFmwBS3knHmnYK58lSgCeya5ebpN6ec8TcKfDDfFQ3FfYhqei0fM9mQjZ/9e/Oy9/QKUOnqR)

KaelWD:
## Description

fixes #21982

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <div style="width: 278px">
        <v-textarea
          :model-value="text"
          max-rows="9"
          rows="1"
          variant="filled"
          auto-grow
        />
      </div>
    </v-container>
  </v-app>
</template>

<script setup>
  const text = 'foo/bar/foo/bar/foo/bar/foo/bar '.repeat(8)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VTextarea/VTextarea.sass
+++ b/packages/vuetify/src/components/VTextarea/VTextarea.sass
@@ -33,6 +33,12 @@
     &--no-resize
       .v-field__input
         resize: none
+        scrollbar-width: none
+
+      &.v-textarea--height-capped
+        .v-field__input
+          scrollbar-gutter: stable
+          scrollbar-width: thin
 
     .v-field--no-label,
     .v-field--active
--- a/packages/vuetify/src/components/VTextarea/VTextarea.tsx
+++ b/packages/vuetify/src/components/VTextarea/VTextarea.tsx
@@ -98,6 +98,7 @@ export const VTextarea = genericComponent<VTextareaSlots>()({
     const vInputRef = ref<VInput>()
     const vFieldRef = ref<VInput>()
     const controlHeight = shallowRef('')
+    const heightCapped = shallowRef(false)
     const textareaRef = ref<HTMLInputElement>()
     const isActive = computed(() => (
       props.persistentPlaceholder ||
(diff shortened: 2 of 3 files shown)


packages/vuetify/src/components/VTextarea/VTextarea.sass
```diff
@@ -33,6 +33,12 @@
     &--no-resize
       .v-field__input
         resize: none
+        scrollbar-width: none
+
+      &.v-textarea--height-capped
+        .v-field__input
+          scrollbar-gutter: stable
+          scrollbar-width: thin
```
J-Sek, line 41: it makes it slightly inconsistent when textarea with `auto-grow` appears next to the regular one (without this prop `heightCapped` is always `false` and the `*--height-capped` is never present even though it has scrollbar, see [playground](https://play.vuetifyjs.com/#eNpVj+EKgkAQhF/l2N96omZEXELP0fZjsQ2Ccz30lCB69/LsSP/NfuwwM5cXDH2TnZ3T08hwBOO5dZY81yhKmSkl54IMR9OJp4dw/0MBen566pkiUqrtbmzTiezIJ4QcpUApUXYoFcoeIX7WJvvbY8oXbWNmsrQw2aodvBModaXzEmZx0AUkd7IDJ5tJC7t+AFe4SC0=)).

J-Sek (changes requested):

packages/vuetify/src/components/VTextarea/VTextarea.sass
J-Sek: I would drop `thin` and change `none` to:
```sass
&:not(.v-textarea--height-capped)
  .v-field__input
    scrollbar-width: none
```

PR #22005, opened 2025-09-03
fix(VChip): fix plain opacity

06b:
## Description

This pull request fixes the opacity for `v-chip` when set to the `plain` variant, all other components have their plain-opacity set to `.62`, however the `v-chip` had a default of `0.26` which that value seems to be used for `disabled-opacity` in other components.

## Markup:

```vue

```


J-Sek (approved):
The bug is so obvious.. it might be a sign that nobody uses this variant on chips.

Good catch anyway 👍🏼

PR #22020, opened 2025-09-05
docs: add v-model prop using OTP Input Divider

### LINKED ISSUE #18979: [Documentation] OTP Input Divider Example has no v-model
### Environment

**OS:** Windows 10

### Steps to reproduce
1. Enter any OTP you want.
2. Click on send new code.

### Expected Behavior
The OTP input field should clear itself once the button is pressed.

### Actual Behavior
The OTP input does not clear itself.

### Reproduction Link
[https://vuetifyjs.com/en/components/otp-input/#divider](https://vuetifyjs.com/en/components/otp-input/#divider)

### Other comments
The solution is to add a v-model to the v-otp-input. I could make a PR to resolve this!

antonhansson0624:
Add v-model prop using OTP Input Divider to be cleared when resetting or sending a new code.
Resolves #18979

### DIFF
--- a/packages/docs/src/pages/en/components/otp-input.md
+++ b/packages/docs/src/pages/en/components/otp-input.md
@@ -117,5 +117,7 @@ The following example is a detailed example of a `v-otp-input` component used to
 ### Divider
 
 The following example is a detailed example of a `v-otp-input` component used with a divider.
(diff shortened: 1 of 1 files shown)


J-Sek:
I don't think we need to be so explicit. It is very basic stuff. The original issue should be just closed after updating example year ago.

PR #22022, opened 2025-09-06
fix(VCombobox): placeholder visibility for empty string value

### LINKED ISSUE #21972: [Bug Report][3.9.6] v-combobox placeholder not showing initially if the model value is empty string
### Environment
**Vuetify Version:** 3.9.6
**Vue Version:** 3.5.20
**OS:** Windows 10 (current)

### Steps to reproduce
1. See the link. The placeholder should be visible on page load but it is not.

### Expected Behavior
The placeholder should be visible after the page is initially opened (on focus or always, if persistent-placeholder is set)

### Actual Behavior
The placeholder is missing when the page opens. If I enter a value and then clear it, the placeholder now shows as it should.

If instead of empty string, the initial v-model value (the ref) is set to null, the placeholder shows properly.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdUO1qwzAMfBWhP94gH2Njg5V0sOdo+sNxFGaIY2OrYVD67pObdk37QyCdTseddkdM0dTfIVTzgXCDDZMLo2b6aieAZi51COf2PBg/sbYTxQt0AV3nO/97hQDm0vmexm2LfYs3WHQN/fixpyir1XRHophsYpq4XDP+9xsrDpPc75RWheqkjNrfFOqr2/rRbkaWNE29SiljMtEGhkR8WMJaF3xkOEKkAU4wRO9AyYNUpgOIbmLoYZv3T0o9Z8lFRM7xVOBb9V69vmBuPqsPLAY9Jirufr1g+z/csHg5)

jcjp:
## Description

- Check the model value if it's an empty string and treat it as a not dirty value
- In this way we could show the placeholder from props

fixes #21972

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-combobox
        v-model="d"
        placeholder="placeholder"
        persistent-placeholder
        :items="['a','b','c']"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const d = ref('')
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -139,7 +139,7 @@ export const VCombobox = genericComponent<new <
       props,
       'modelValue',
       [],
-      v => transformIn(v === null ? [null] : wrapInArray(v, emptyValues.value)),
+      v => transformIn(wrapInArray(v, emptyValues)),
       v => {
         const transformed = transformOut(v)
         return props.multiple ? transformed : (transformed[0] ?? null)
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -478,8 +478,7 @@ export const VCombobox = genericComponent<new <
         slots['append-item'] ||
         slots['no-data']
       )
-      const isEmptyString = model.value.length === 1 && model.value[0].value === ''
-      const isDirty = model.value.length > 0 && !isEmptyString
+      const isDirty = model.value.length > 0
       const textFieldProps = VTextField.filterProps(props)
 
       return (
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -133,13 +133,13 @@ export const VCombobox = genericComponent<new <
     const vVirtualScrollRef = ref<VVirtualScroll>()
     const selectionIndex = shallowRef(-1)
     let cleared = false
-    const { items, transformIn, transformOut } = useItems(props)
+    const { items, transformIn, transformOut, emptyValues } = useItems(props)
     const { textColorClasses, textColorStyles } = useTextColor(() => vTextFieldRef.value?.color)
     const model = useProxiedModel(
       props,
       'modelValue',
       [],
-      v => transformIn(wrapInArray(v)),
+      v => transformIn(v === null ? [null] : wrapInArray(v, emptyValues.value)),
       v => {
         const transformed = transformOut(v)
         return props.multiple ? transformed : (transformed[0] ?? null)
@@ -478,7 +478,8 @@ export const VCombobox = genericComponent<new <
         slots['append-item'] ||
         slots['no-data']
       )
-      const isDirty = model.value.length > 0
+      const isEmptyString = model.value.length === 1 && model.value[0].value === ''
+      const isDirty = model.value.length > 0 && !isEmptyString
       const textFieldProps = VTextField.filterProps(props)
 
       return (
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -477,7 +477,8 @@ export const VCombobox = genericComponent<new <
         slots['append-item'] ||
         slots['no-data']
       )
-      const isDirty = model.value.length > 0
+      const isEmptyString = model.value.length === 1 && model.value[0].value === ''
```
J-Sek: It should rather be named `isArrayWithEmptyString`. Regardless... I don't think a combobox with `multiple` and `model-value="['']"` can be considered empty. It behaves differently when selecting more values. Showing placeholder in this case could be misleading.

J-Sek (dismissed):

packages/vuetify/src/components/VCombobox/VCombobox.tsx
jcjp: ~~Oh okay gotcha, so this should be regarded as intended.~~ I am a bit confused since the `model.value` is coming out as array of objects.

J-Sek: I'd just put `!props.multiple` there

J-Sek: Upon reading [comment](https://github.com/vuetifyjs/vuetify/issues/21972#issuecomment-3284313165), it might make sense to resolve `''` into empty array. I had an attempt on it in the past, but the scope might had been too broad (so #20742 got stuck).

J-Sek (dismissed):
Green light from me, but I wish others could have a day or two to voice their opinions. @vuetifyjs/core-team

johnleider:
Is there anywhere else in the framework where we parse or otherwise try to do something special when dealing with "empty" values? Are there any performance implication for building 2 more computed arrays?

packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -133,13 +133,13 @@ export const VCombobox = genericComponent<new <
...
     const selectionIndex = shallowRef(-1)
     let cleared = false
-    const { items, transformIn, transformOut } = useItems(props)
+    const { items, transformIn, transformOut, emptyValues } = useItems(props)
     const { textColorClasses, textColorStyles } = useTextColor(() => vTextFieldRef.value?.color)
     const model = useProxiedModel(
       props,
       'modelValue',
       [],
-      v => transformIn(wrapInArray(v)),
```
KaelWD, line 164: Is this change necessary? Combobox only works with returnObject so modelValue=null will never match any item unless they're doing something weird with the slots.

KaelWD:
> closes https://github.com/vuetifyjs/vuetify/pull/20742

this doesn't affect any other components, is that still a problem elsewhere?

J-Sek:
> > _closes #20742_
> _this doesn't affect any other components, is that still a problem elsewhere?_

They are overlaping, but could go together. The old #20742 is stuck a bit because, VSelect and VAutocomplete require more testing and I have dropped the ball. My concern is about `transformIn` not being called (?) after `items` are changed to include a value with `''`, so model/selection could be missing an item.

VCombobox is simplere and could be fixed separately/first with this PR - it would be a step forward. #20742 can stay open.

KaelWD:
> transformIn not being called (?) after items are changed

It should be, it's called in the computed getter in proxiedModel

J-Sek:
> It should be, it's called in the computed getter in proxiedModel

👍🏼  confirmed as working, shows Empty when the item is pushed with delay

```ts
setTimeout(() => {
  options1.value.push({ title: 'Empty', value: '' })
}, 2000)
```

That PR still needs some good testing. Assuming it gets done, we can consolidate both PRs. ~~@jcjp , is there any major difference between both implementation changes in `list-items.ts`?~~ Nevermind, it's Kael's performance improvement.

packages/vuetify/src/components/VCombobox/VCombobox.tsx
jcjp: You are right, after careful investigation, it seems to be not needed. During my investigation it seems that the `emptyValues` are not properly set on final render thus the placeholder not being shown properly will further investigate this thanks! EDIT: I updated the code and the fix seems to be working fine.

Issue #22017, opened 2025-09-05
[Bug Report][3.9.5] Accessibility: VTextField input element have role but not required aria attributes (following #21903)


Dyskal:
### Environment
**Vuetify Version:** 3.9.5
**Vue Version:** 3.5.21

### Expected Behavior
The input element inside the `<v-select>` should have role="combobox" and aria-expanded & aria-controls attributes.
https://dequeuniversity.com/rules/axe/4.10/aria-required-attr?application=playwright

### Actual Behavior
The input element have the role but not the required combobox attributes.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpdj0EKwjAQRa8SZpNNTGiLiCUKnsFl00UMIxRiDE3sRry7SWrBdvf+Y/gz070hjEZcvOfTC6EFGfHhrY54Vo4QOe209wVLME8X9eBw/KkiA1o0kVh9Q3tScC1RAWmH1BWS6WimijJSoF6goX0aE0u92PZnM6+X4u8s+DBo+J7XFWQ48gOwu7YB2eqX2fVfoMZGNQ==)

### Other comments
There was no problem before #21903. Those aria attributes must have been forgotten.

PR #22025, opened 2025-09-07
fix(VSelect, VAutocomplete, VCombobox): missing accessibility attributes (#22017)

jcjp:
## Description

- Fixes: [22017](https://github.com/vuetifyjs/vuetify/issues/22017)
- Includes additional components like `VSelect`, `VAutocomplete` and `VCombobox`

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-select label="Select" :items="['item1', 'item2', 'item3']"  />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -22,11 +22,12 @@ import { useForm } from '@/composables/form'
 import { forwardRefs } from '@/composables/forwardRefs'
 import { useItems } from '@/composables/list-items'
 import { useLocale } from '@/composables/locale'
+import { useMenuActivator } from '@/composables/menuActivator'
 import { useProxiedModel } from '@/composables/proxiedModel'
 import { makeTransitionProps } from '@/composables/transition'
 
 // Utilities
-import { computed, mergeProps, nextTick, ref, shallowRef, toRef, useId, watch } from 'vue'
+import { computed, mergeProps, nextTick, ref, shallowRef, watch } from 'vue'
 import {
   checkPrintable,
   deepEqual,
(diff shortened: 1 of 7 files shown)


packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -88,7 +88,12 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
 
       return props.counter
     })
-
+    const isCombobox = computed(() => props.role === 'combobox')
```
J-Sek: Could be simplified

```ts
const isCombobox = toRef(() => props.role === 'combobox')
const isAriaExpanded = toRef(() => isCombobox.value && attrs['aria-label'] === 'Open')
```

jcjp: Hi Jacek updated this part of the code thanks!

packages/vuetify/src/components/VTextField/VTextField.tsx
```diff
@@ -88,7 +88,8 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
 
       return props.counter
     })
-
+    const isCombobox = toRef(() => props.role === 'combobox')
+    const isAriaExpanded = toRef(() => isCombobox.value && attrs['aria-label'] === 'Open')
```
J-Sek: if an app adapts label for the user locale it won't work, right?

VSelect.tsx:420
```
aria-label={ t(label.value) }
```

Can we move it one level up (i.e. to VSelect, VAutocomplete and VCombobox)?

jcjp: All done let me know if there is still a need for a change Jacek thanks!

J-Sek:
- [x] [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls) needs to match with menu `id`
- [x] `attrs['aria-label'] === 'Open')` can be just `menu.value`
- [x] covers all components with dropdown lists, not just VSelect
- [x] `isCombobox` » `hasRoleCombobox` (feels redundant... maybe just inline the condition, code will be self-explanatory)
- [x] add unit test

PR #22029, opened 2025-09-07
docs: add item name slot with index prop documentation and example

### LINKED ISSUE #15466: [Documentation] VDataTable: `item.<name>` props have `index` prop, but haven't in document
### Problem to solve
After my test, I found that the `item.<name>` slot already has the prop of index, but it is not explained in the document. Would you like to add this to the document?

经过我的测试，发现`item.<name>`插槽已经带有index这个prop，但是在文档上没有说明。请问是否将这个补充到文档上吗？

image

image

### Proposed solution
Update document.

antonhansson0624:
Add item name slot with index prop documentation and example.
Resolves #15466

### DIFF
--- a/packages/docs/src/examples/v-data-table/slot-item-name.vue
+++ b/packages/docs/src/examples/v-data-table/slot-item-name.vue
@@ -1,0 +1,29 @@
+<template>
+  <v-data-table
+    :headers="headers"
+    :items="employees"
+    class="elevation-1"
+  >
+    <template v-slot:item.index="{ index }">
+      {{ index + 1 }}
+    </template>
+  </v-data-table>
+</template>
+
+<script setup>
+  const headers = [
+    { title: '#', key: 'index' },
+    { title: 'Name', key: 'name' },
+  ]
+
+  const employees = [
+    { name: 'Alice Johnson' },
+    { name: 'Bob Carter' },
+    { name: 'Clara Diaz' },
+    { name: 'David Evans' },
+    { name: 'Eva Flores' },
+    { name: 'Frank Green' },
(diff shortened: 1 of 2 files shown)


packages/docs/src/examples/v-data-table/slot-item-name.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <v-data-table
+    :headers="headers"
+    :items="employees"
+    class="elevation-1"
+  >
+    <template v-slot:item.index="{ index }">
```
J-Sek: It is confusing. Can we change the example to show index for the "name" column? We can also add an "age" field to make the content more realistic.

J-Sek (changes requested):

packages/docs/src/examples/v-data-table/slot-item-name.vue
```diff
@@ -0,0 +1,68 @@
...
+  <v-data-table
+    :headers="headers"
+    :items="employees"
+    :items-per-page="5"
+    :items-per-page-options="[5, -1]"
+    class="elevation-1"
+  >
+    <!-- Show index and alias -->
+    <template v-slot:item.name="{ item, index }">
+      {{ index + 1 }}. {{ item.name }}
```
J-Sek, line 11: It feels weird that sorting does not affect these `index` numbers - QA would definitely raise a ticket, as it might mislead or confuse users. Can you suggest a better use case for this value?

J-Sek (changes requested):

J-Sek:
the section already exists

Issue #22016, opened 2025-09-05
[Bug Report][3.9.7] Events are shown incorrectly when they span across multiple days


mathiasrando:
### Environment
**Vuetify Version:** 3.9.7
**Vue Version:** 3.5.21
**OS:** macOS 10.15.7 (current)

### Steps to reproduce
1. Add a VCalendar
2. Pass in an event spanning across multiple days

### Expected Behavior
Events spanning across multiple days should be shown on all the days

### Actual Behavior
Events spanning across multiple days only shows on the first and the last days

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqFkcFuwjAMhl/F8gUqQQtsbKMqkyZx2APsRjhkxZWqpU2UuK2mqu++lHRo7MLtzyf/9m/n2KOzefJmTNw2hClmTJVRkulV1ABZu5TGXOTlkeuaZVmTnVCAUlF9lhZSaqlmtxcYhEBoS+qWlT6Thx3Rl0fJb7vkf7+RhHFZ8ieGf7rclobBETchTVkZbRl6sFTAAIXVFcz8BrOxHMD3dQwhBezHovmxD2OlUgf5nUIhlaNFYLlW2qYg8FM1JHCifqkUaurg4HPMr0LgZrXZwmoHqyeBUexDvevGuvn6JYomq2Np+Y758da8vZq5ZEVjmg9yPKUZTtF4lXAHfwEcFvgQb+PNGkexi59xETa6+c/ATj+A3piZ)

### Images
![https://cdn.cosmicjs.com/3df86a40-8a5c-11f0-b401-3914bc05435b-Screenshot-2025-09-05-at-15-28-20.png](https://cdn.cosmicjs.com/3df86a40-8a5c-11f0-b401-3914bc05435b-Screenshot-2025-09-05-at-15-28-20.png)

PR #22030, opened 2025-09-08
fix(VCalendar): ensure events spanning multiple days display correctly

babu-ch:
fix(VCalendar): ensure events spanning multiple days display correctly

## Description

fix https://github.com/vuetifyjs/vuetify/issues/22016

Added a between condition when filtering.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-calendar :events="events" view-mode="week" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const events = ref([{
    allDay: false,
    color: "blue",
    end: new Date(new Date("2025 09 06").setHours(18)),
    start: new Date(new Date("2025 09 04").setHours(15)),
    title: "Test",
  }])
</script>

```


KaelWD:
Thanks but this is already being worked on in #21910

PR #22049, opened 2025-09-11
docs(VChipGroup): add example with center-active prop

### LINKED ISSUE #22046: [Bug Report][3.10.0] centerActive props does not work in ChipGroup
### Environment
**Vuetify Version:** 3.10.0
**Vue Version:** 3.5.21
**OS:** Windows 10 (current)

### Steps to reproduce
1. 

### Expected Behavior
Works normally like a slide.

### Actual Behavior
Move towards the center only when clicking on the edge

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNplkstugzAQRX9l5E02mLyUTURf6id0WbpwzZBaMcayDWqE+PeaR8BOV5g7M/aZufPZEWv49k3rtG2QnEnmsNKSOXzOFUDWUqb1eBx/eK0cEwrNLE3ij9D0YupGQymkQwMclf9Qxp1oESxK5A4LyiWz9iknDn8dNVjkZLlmuQhaWtbGJykQCg6nnMD5irdB8OnvQ0bXgYK+z7ZTxYoyCxOLlwNGK0WBM2REFxMEaVT4SaxB+A8WBlfGuMTK2nm9A2E/5jEk4OrLRSL0YXLAMZJ8uwV/foHXcnx/vQleYKONqJi5beAMjSqw9Ob4ucal97lXjB4eY75VX1bE4iuXgl8Hp0bSuCYiBQgtiXrwbvgmwvl65XHAkXlB8L5yg6XRzg3KtJLZNlhV0ifkmJ7S/ZEMh/0u3ZGkZNJiEi34pH39Ad3O3cg=)

antonhansson0624:
Add documentation for center-active prop usage on chip-groups.
Resolves #22046

### DIFF
--- a/packages/docs/src/examples/v-chip-group/prop-centeractive.vue
+++ b/packages/docs/src/examples/v-chip-group/prop-centeractive.vue
@@ -1,18 +1,16 @@
 <template>
-  <v-app>
-    <v-container>
-      <v-chip-group
-        v-model="selected"
-        selected-class="text-red"
-        center-active
-        filter
-      >
-        <v-slide-group-item v-for="n in 25" :key="n">
-          <v-chip :value="n">Chip {{ n }}</v-chip>
-        </v-slide-group-item>
-      </v-chip-group>
-    </v-container>
-  </v-app>
+  <v-sheet class="mx-auto" max-width="600">
+    <v-chip-group
+      v-model="selected"
+      selected-class="text-red"
+      center-active
+      filter
+    >
+      <v-slide-group-item v-for="n in 25" :key="n">
+        <v-chip :value="n">Chip {{ n }}</v-chip>
+      </v-slide-group-item>
+    </v-chip-group>
+  </v-sheet>
 </template>
 
 <script setup>


packages/docs/src/examples/v-chip-group/prop-centeractive.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <v-app>
```
J-Sek: We should not use `<v-app>` in examples. `<v-container>`, rarely.
I would suggest adapting to similar examples around, `<v-sheet>` with `max-width=600` and `mx-auto` class

antonhansson0624: I updated

J-Sek:
We might need to scrap it. It only works here because VChip `value` got the index value. It starts with `1`, so it is misaligned (not centered properly), but even if we make it center the items it is hard to call it "working".

antonhansson0624:
I will close this PR and delete the branch.

PR #22093, opened 2025-09-20
feat(VCombobox): always show only matching suggestions

### LINKED ISSUE #22060: [Bug Report] VCombobox - only show matching suggestions
### Steps to reproduce
1. type something that does not match
2. click away / blur
3. click to focus and type something more

### Expected Behavior
The field should act like more like an enhanced text field. Suggestions are not relevant.

### Actual Behavior
Shows the full list, distracting me from editing the field.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNplUk1PwzAM/StWLgWpHwMEiGlDAq6MExKHdYe0dWlE2kT5GEPT/jtOx7aunBL7vWc7z1lumTVl9qR1uvbIpmzmsNWSO3zMO4DZOuFag2uwxXnOKm6+ctYjPVaqznHRoYGWb5JvUbmGWHeTyZG0p3FTHeORsJZeDME/uC1UoTbDNMA6aVWFkjr0Z87O4amg0S2htlHGvQrrxgzJi17+wg0seOdrXjpv0Ix5pURueCHxPE22lNgoWaGhIph+pvAmrOXdWN9i5xNBbyTaGOPeKXqeluiCo6quxwyNxtLw2LmkH2SIZkMbs4GPJ7tD9uT3f1LI0E7pPssGu6bQlkZoBxadDzCAaDU5CVswWMMOaqNaiOibRAGkqpbIB69hDsvoXf0ox6MYooVw1hfCNiJEz4uPaBV6HGT9AklChS8uwyT73tSV7WJ2k96m11csXB7SexbXXFqMzz7qPrf6BTG40mQ=)

jcjp:
## Description

closes #22060
Relates to: https://github.com/vuetifyjs/vuetify/pull/21901

####
- Add new props `alwaysFilter` defaults to `false` to allow the previous behavior consistent with components like `VSelect` and `VAutocomplete`
- ~~If set to `true` it will prevent to set the `isPristine` to `false` on initial load~~
- ~~Instead of passing the `search` immediately check `isPristine` and determine the value to pass~~

## Markup:
```vue
// This example will show the previous behavior
// On menu show it should show all the items
// Pass always-filter to see it filter on load
<template>
  <v-app>
    <v-container>
      <v-combobox
        v-model="msg"
        :items="['Colorado', 'Florida', 'Georgia', 'Texas', 'Wyoming', 'California']"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const msg = ref('Cali')
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -190,7 +190,7 @@ export const VCombobox = genericComponent<new <
         : (props.multiple ? model.value.length : search.value.length)
     })
 
-    const { filteredItems, getMatches } = useFilter(props, items, () => isPristine.value ? '' : search.value)
+    const { filteredItems, getMatches } = useFilter(props, items, () => props.alwaysFilter ? search.value : isPristine.value ? '' : search.value)
 
     const displayItems = computed(() => {
       if (props.hideSelected) {
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -190,7 +190,11 @@ export const VCombobox = genericComponent<new <
         : (props.multiple ? model.value.length : search.value.length)
     })
 
-    const { filteredItems, getMatches } = useFilter(props, items, () => props.alwaysFilter ? search.value : isPristine.value ? '' : search.value)
+    const { filteredItems, getMatches } = useFilter(
+      props,
+      items,
+      () => props.alwaysFilter || !isPristine.value ? search.value : ''
+    )
 
     const displayItems = computed(() => {
       if (props.hideSelected) {
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -63,6 +63,10 @@ type Value <T, ReturnObject extends boolean, Multiple extends boolean> =
     : Val<T, ReturnObject> | null
 
 export const makeVComboboxProps = propsFactory({
+  alwaysFilter: {
+    type: Boolean,
+    default: false,
+  },
   autoSelectFirst: {
     type: [Boolean, String] as PropType<boolean | 'exact'>,
   },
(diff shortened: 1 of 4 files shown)


packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -63,6 +63,10 @@ type Value <T, ReturnObject extends boolean, Multiple extends boolean> =
     : Val<T, ReturnObject> | null
 
 export const makeVComboboxProps = propsFactory({
+  alwaysFilter: {
+    type: Boolean,
+    default: false,
+  },
```
J-Sek:
```suggestion
  alwaysFilter: Boolean,
```

J-Sek (changes requested):
Menu gets unstable (opens, closes, opens, etc.) again after typing few more characters with no match.

I deliberately removed `isPristine` from `useFilter(...)` to have `filterdItems` predictable and depending only on the `search` and `_searchLock` (to be introduced in one of the outstanding PRs). I don't mind having it back if we can avoid regression... just feeling like we made a wrong turn at some point and state/behavior is very hard to track and impossible to keep the bugs out.

jcjp:
> Menu gets unstable (opens, closes, opens, etc.) again after typing few more characters with no match.
> 
> I deliberately removed `isPristine` from `useFilter(...)` to have `filterdItems` predictable and depending only on the `search` and `_searchLock` (to be introduced in one of the outstanding PRs). I don't mind having it back if we can avoid regression... just feeling like we made a wrong turn at some point and state/behavior is very hard to track and impossible to keep the bugs out.

If I get this right we should skip this feature correct or we should continue adding this new property? Thanks Jacek.

J-Sek (changes requested):
Looks good 👍🏼 
We need an entry in `new-in.json` and a description for API docs that would include "matching", "suggestions" keywords and a hint about `:menu-icon="false"` and `hide-selected` for optimal UX.

J-Sek (approved):
In my opinion it is more of a fix than a feature. We just deliver it using a separate prop to avoid unexpected behavior change in all the use cases where it is expected to act like an enhanced VAutocomplete.

PR #22101, opened 2025-09-22
fix(VDialog): fix focus trap when tabbing forward

### LINKED ISSUE #21945: [Bug Report][3.9.5] vDialog focus trap is working only for Shift-Tab
### Environment
**Vuetify Version:** 3.9.5
**Vue Version:** 3.5.19
**Browsers:** Firefox 141.0, Google Chrome, Mozilla Firefox
**OS:** Windows, Linux

### Steps to reproduce
1. Open the dialog
2. Start pressing TAB

### Expected Behavior
After pressing TAB when the focus is on the last focusable element in the dialog (Agree) it should switch the focus back to the first focusable element (Disageee).

### Actual Behavior
Focus leaves the dialog.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVU01v2zAM/SuELtmA2Mmw7rAg7VagwC4DtsN2WnZgJSYRqi9IjNshyH8fbSe2U+yygw2KeO/xiaR+HVXJenGfUt0cSK3Umsknh0x3mwCwNrYB7bCU241ieuFKU2DKkLC62agOI6imMhZd3EFT+WjICbhPbBR4fKmereG9JG+Wy4EktEspoRUXeYWabYMcs0CPkHJMZQVD8nt7htNEoKv8yEH4jzYYYV2DBQrfEgV46N2tFx18NLCYXHZINpXGbMYaKZNomMrqGKSEN7bymOTLT5Q3agS2/RHAV2L4EuPOEezJJcAktg1J17wNBC5qZBtDDT/2toAnDAWKFLBhBxhi+OPjoQwwMMgIHM+Sc6BGLvS8l1+IvTZmgnwIQQTqKz+WHYmhn4XO7NlEt1BurKZPI2Pa13+NJoYygXSdKgk15bu2sedwbORkPp+1s/pp2Aq4hS26Qu18HmzBXSaaDOf/BO5fsQfmdLiXVD/by9rKsdfrd30hyy7Rq61YF51tYukXH1IHtD7FzHCETFs4wTZHDzN5PbPeuqxJYRiMCuhNZ/ZtK92LiYw6zdX7+kP97qNqg2W9VPMONr96kH3u91/WACyd)

### Other comments
Shift-Tab traps the focus as expected.

memorandumtk:
## Description
- Registers a keydown event listener on VDialog to focus on the first focusable element when the Tab key is pressed on the last focusable element in the dialog.
- Added two tests to test focus behavior when focus moves from the first and last buttons in a VDialog.

fixes #21945 

## Markup:
```vue
<template>
  <div class="text-center pa-4">
    <v-dialog v-model="dialog" max-width="400">
        <v-btn v-bind="activatorProps"> Open Dialog </v-btn>
      </template>

      <v-btn @click="dialog = false"> Disagree </v-btn>
      <v-btn @click="dialog = false"> Agree </v-btn>
    </v-dialog>
  </div>
</template>

<script setup>
  import { ref } from 'vue'

  const dialog = ref(false)

  // for debugging
  const showPressedKey = (e) => {
    const now = new Date()
    const msg = `[${now.toISOString()}] pressed: ${e.key}${e.shiftKey ? ' with shift key' : ''}`
    console.log(msg)
  }

  window.addEventListener('keydown', showPressedKey)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VDialog/VDialog.tsx
+++ b/packages/vuetify/src/components/VDialog/VDialog.tsx
@@ -14,7 +14,7 @@ import { useScopeId } from '@/composables/scopeId'
 
 // Utilities
 import { mergeProps, nextTick, onBeforeUnmount, ref, watch } from 'vue'
-import { focusableChildren, genericComponent, IN_BROWSER, propsFactory, useRender } from '@/util'
+import { focusableChildren, focusChild, genericComponent, IN_BROWSER, propsFactory, useRender } from '@/util'
 
 // Types
 import type { OverlaySlots } from '@/components/VOverlay/VOverlay'
@@ -81,22 +81,17 @@ export const VDialog = genericComponent<OverlaySlots>()({
     }
 
     function onKeydown (e: KeyboardEvent) {
-      if (e.key !== 'Tab' || !overlay.value?.contentEl) return
+      if (e.key !== 'Tab' ||
+        !overlay.value?.contentEl ||
+        // We're the topmost dialog
+        !overlay.value?.globalTop
+      ) return
 
       const focusable = focusableChildren(overlay.value.contentEl)
       if (!focusable.length) return
 
-      const firstElement = focusable[0]
-      const lastElement = focusable[focusable.length - 1]
-      const active = document.activeElement as HTMLElement | null
-
-      if (e.shiftKey && active === firstElement) {
-        e.preventDefault()
-        lastElement.focus()
-      } else if (!e.shiftKey && active === lastElement) {
-        e.preventDefault()
-        firstElement.focus()
-      }
+      e.preventDefault()
+      focusChild(overlay.value.contentEl, e.shiftKey ? 'prev' : 'next')
     }
 
     onBeforeUnmount(() => {
--- a/packages/vuetify/src/components/VDialog/VDialog.tsx
+++ b/packages/vuetify/src/components/VDialog/VDialog.tsx
@@ -66,17 +66,7 @@ export const VDialog = genericComponent<OverlaySlots>()({
         !overlay.value.contentEl.contains(after)
       ) {
         const focusable = focusableChildren(overlay.value.contentEl)
-
-        if (!focusable.length) return
-
-        const firstElement = focusable[0]
-        const lastElement = focusable[focusable.length - 1]
-
-        if (before === firstElement) {
-          lastElement.focus()
-        } else {
-          firstElement.focus()
-        }
+        focusable[0]?.focus()
       }
     }
 
@@ -107,7 +97,7 @@ export const VDialog = genericComponent<OverlaySlots>()({
     if (IN_BROWSER) {
       watch(() => isActive.value && props.retainFocus, val => {
         if (val) {
-          document.addEventListener('focusin', onFocusin)
+          document.addEventListener('focusin', onFocusin, { once: true })
           document.addEventListener('keydown', onKeydown)
         } else {
           document.removeEventListener('focusin', onFocusin)
--- a/packages/vuetify/src/components/VDialog/VDialog.tsx
+++ b/packages/vuetify/src/components/VDialog/VDialog.tsx
@@ -51,34 +51,6 @@ export const VDialog = genericComponent<OverlaySlots>()({
     const { scopeId } = useScopeId()
 
     const overlay = ref<VOverlay>()
-    function onFocusin (e: FocusEvent) {
-      const before = e.relatedTarget as HTMLElement | null
-      const after = e.target as HTMLElement | null
-
-      if (
-        before !== after &&
-        overlay.value?.contentEl &&
-        // We're the topmost dialog
-        overlay.value?.globalTop &&
-        // It isn't the document or the dialog body
-        ![document, overlay.value.contentEl].includes(after!) &&
-        // It isn't inside the dialog body
-        !overlay.value.contentEl.contains(after)
-      ) {
-        const focusable = focusableChildren(overlay.value.contentEl)
-
-        if (!focusable.length) return
-
-        const firstElement = focusable[0]
-        const lastElement = focusable[focusable.length - 1]
-
-        if (before === firstElement) {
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VDialog/VDialog.tsx
```diff
@@ -80,15 +80,39 @@ export const VDialog = genericComponent<OverlaySlots>()({
...
+      const lastElement = focusable[focusable.length - 1]
+      const active = document.activeElement as HTMLElement | null
+
+      if (e.shiftKey && active === firstElement) {
+        e.preventDefault()
+        lastElement.focus()
+      } else if (!e.shiftKey && active === lastElement) {
+        e.preventDefault()
+        firstElement.focus()
+      }
```
J-Sek, line 92: Looks like equivalent of:

```ts
e.preventDefault()
focusChild(overlay.value.contentEl, e.shiftKey ? 'prev' : 'next')
```

packages/vuetify/src/components/VDialog/VDialog.tsx
```diff
@@ -80,15 +80,39 @@ export const VDialog = genericComponent<OverlaySlots>()({
...
+    }
+
     onBeforeUnmount(() => {
       document.removeEventListener('focusin', onFocusin)
+      document.removeEventListener('keydown', onKeydown)
```
J-Sek, line 97: ~~new handler replaces `onFocusin`. Is there any reason left to keep both?~~ Yes, to capture the first focus.

Lines 69-79 could be replaced with `focusable[0]?.focus()` like in VMenu.

packages/vuetify/src/components/VDialog/VDialog.tsx
J-Sek: Well... nope, I withdraw my suggestion as it would interfere with inner VList.

packages/vuetify/src/components/VDialog/VDialog.tsx
memorandumtk: Oh, I see.
I edited it according to the comment before editing, so I'll correct it again.

memorandumtk: I have two questions.

1. Did you mean that the first focus was on the dialog itself (v-overlay__content)? I assumed it was on the first focusable element in the dialog (the "Disagree" button in the video below), so I just wanted to confirm.

2. I was checking the focus behavior of VMenu and noticed that when you press tab on Vmenu, the behavior is correct the first time you press tab, but the second time, Vmenu seems to close. Should I fix this?


[demo in play.vuetify](https://play.vuetifyjs.com/#eNqtVUtvm0AQ/isjLhAJsPPoBdlWo6S99HnpKc5hDWNnFdhFuwuxhfjvnV1sDG7rpFVPMDPf983T5qHxtEont2UZ1xV6iTczWJQ5M7hYCoBZxmtIc6b1fOkZ3JooRWFQQcmim6XnMISqo4yzXG6gjgqZYU7gzrH0oGDb6IVn5omcN9NpTyLaIRXRdC5NwlLDa2akImgDpZKlTqB3frc2tAMBl3llBPFXXGTEGoMJCt9KFHDfVTebOPixgMmg2d7ZSb5Pc54+943AHNYs12gl77lmG4X4q96r1Nt/5H1Fbp5o7GOmtTpst6wJbWvfyZsXV6Co/mojv99AKnMHKhUvmNrR3vudOJZr/iAGVvv8NnrxnGszTmc9EScspVi7pIG1QqB0uL2gB1ibUkLyjDuKu4A1a5ZXeHQMdMfKkeEmx0XTOKXYWdC2tuJTzKC0YXTQ1agHa/YjPyzs5A5nOlW8NKDRVKUD8qKUykADCtchSPFDFLKidWbQwlrJAnz68frd5lMptIH+hIgRuDO6GIVxVW0+yrTSBAnwAuYLaLoKLUDmGBM9aAxTGzQJYNy9hWBv6WjH1ryTVIowLWUAuo7TNJ9wl8kXcT4RLcqq0mMkk1GJNC4Tsyz7UNPLZxolClSBv7bVc+GHg2Yc9QyH5G0pB86+sv1kBmMNgmGhvaDCQtb41jpeI/6xGOp+tCx3zTS+h061AXd5Cfh39v8CvqAPbfi/Y3B1iD7aA+1Oko7Ra0PvOn4XX1169uVyGl97oTuwcPQl6XyPPwGnARXF)



[screen-capture (24).webm](https://github.com/user-attachments/assets/c45a752b-3836-468c-89ce-e736c603f316)

J-Sek: 1. VMenu and VDialog, both try to "capture" the focus when an interaction occurs and the focus is set outside. I don't know the origin of this implementation, but it feels slightly better than auto-focus within (done by native `<dialog>`) as it would be reasonable to load the dialog content progressively, show spinner, load stuff.

That said, the code between those is not the same because VDialog applied (bugged) focus trap. My suggestion would be to have `keydown` handler focus on the trap, while `focusin` handler can do the same thing as VMenu, so as a result, the code should look the same. No need to refactor/abstract a new composable. It is going to happen eventually (#22106). For now, we should just avoid 2 pieces of code having the same responsibility.

2. Better to keep the PR focused on one thing. It might feel a bit annoying that user tab-out of the menu and is supposed to use up/down arrows for a VList within.. Nobody rised this issue until now, so I guess it is not a bug. It might be equally annoying to be stuck within a menu/popover because developer forgot to put `:retain-focus="false"`, right? Either way, 3.11.0 will hopefully give us an option to set focus trap VMenu and VNavigationDrawer.

memorandumtk: Thank you for your explanation and sharing the current refactor PR 22106!
I've pushed a fix that brings the code as close as possible to Vmenu `onFocus`. I'd appreciate it if you could review it.

J-Sek (approved):

PR #22106, opened 2025-09-22
feat(VNagivationDrawer, VOverlay): add `retain-focus` prop

### LINKED ISSUE #16140: [Bug Report][3.0.1] V-Navigation-Drawer - Temporary V-Navigation-Drawer does not have focus trap
### Environment
**Vuetify Version:** 3.0.1
**Vue Version:** 3.2.45
**OS:** Mac OS 10.15.7

### Steps to reproduce
Just open the drawer and note that the focus is not trapped in the overlay element. I can see the same behaviour happening for the v-menu, the overlay is created but not focus trap is added.

The docs for v-menu state that (no reference of this for v-navigation-drawer)
By default, v-menu components are detached and moved to the root of your application. In order to properly support inserting dynamic content into the DOM, you must use the attach prop. This will ensure that focus transfers from the activator to the content when pressing the tab key.

### Expected Behavior
Once drawer is open, focus should be trapped within new element. As per Material UI drawers
https://mui.com/material-ui/react-drawer/

### Actual Behavior
Drawer and V-Menu is opened and focus is not set to overlay items.

### Reproduction Link
[https://codepen.io/rafaeldesousa/pen/abKqLgL](https://codepen.io/rafaeldesousa/pen/abKqLgL)

J-Sek:
- introduces `retain-focus` for VOverlay (VMenu) and VNavigationDrawer
  - consolidated with the same prop on VDialog
- supports multiple coexisting traps (only keeps focus within the closest trap)
- respects VList navigation with arrows only

closes #16140

> Note: considering `activate`, `deactivate` (like in [VueUse](https://vueuse.org/integrations/useFocusTrap/))

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-app-bar>
      <v-app-bar-nav-icon @click="sidebar = !sidebar" />
    </v-app-bar>
    <v-navigation-drawer v-model="sidebar" capture-focus retain-focus temporary>
      <div class="d-flex flex-column ga-3">
        <v-btn text="Button 1" />
        <v-btn text="Button 2" />
        <v-btn>
          Menu with list
          <v-menu activator="parent" location="right" retain-focus>
            <v-list>
              <v-list-item title="foo" @click="" />
              <v-list-item title="bar" @click="" />
              <v-list-item title="baz" @click="" />
            </v-list>
          </v-menu>
        </v-btn>
        <v-btn>
          Menu with buttons
          <v-menu activator="parent" location="right" retain-focus>
            <v-sheet>
              <v-btn-group direction="vertical">
                <v-btn text="foo" @click="" />
                <v-btn text="bar" @click="" />
                <v-btn text="baz" @click="" />
              </v-btn-group>
            </v-sheet>
          </v-menu>
        </v-btn>
      </div>
    </v-navigation-drawer>
    <v-main>
      <v-container>
        <v-row class="gc-6">
          <v-menu :close-on-content-click="false" width="500" retain-focus>
              <v-btn v-bind="props" text="Open Menu" />
            </template>

            <template #default="{ isActive }">
              <v-card title="Menu">
                <v-card-text>
                  <v-text-field />
                </v-card-text>

                <v-card-actions>
                  <v-btn text="Close" @click="isActive.value = false" />
                  <v-btn text="Save" @click="isActive.value = false" />
                </v-card-actions>
              </v-card>
            </template>
          </v-menu>
          <v-btn prepend-icon="mdi-earth">
            Dialog with radio group
            <v-dialog
              v-slot="{ isActive }"
              activator="parent"
              height="300"
              width="auto"
              capture-focus
              retain-focus
              scrollable
            >
              <v-card title="Select an option">
                <template #subtitle>
                  <small> would be better to let <v-kbd>Tab</v-kbd> escape to the actions, as the list might be very long </small>
                </template>
                <v-divider />
                <v-card-text class="px-4" style="height: 300px;">
                  <v-radio-group hide-details>
                    <v-radio v-for="i in 10" :key="i" :label="`Option #${i}`" :value="i" />
                  </v-radio-group>
                </v-card-text>
                <v-divider />
                <v-card-actions>
                  <v-btn text="Close" @click="isActive.value = false" />
                  <v-btn color="surface-variant" text="Save" variant="flat" @click="isActive.value = false" />
                </v-card-actions>
              </v-card>
            </v-dialog>
          </v-btn>
          <v-btn text="Open sidebar" @click="sidebar = true" />
          <!-- <v-btn text="Dummy" /> -->
        </v-row>
      </v-container>
    </v-main>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const sidebar = ref(false)
</script>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VDialog.json
+++ b/packages/api-generator/src/locale/en/VDialog.json
@@ -9,7 +9,6 @@
     "noClickAnimation": "Disables the bounce effect when clicking outside of a `v-dialog`'s content when using the **persistent** prop.",
     "openOnHover": "Designates whether component should activate when its activator is hovered.",
     "persistent": "Clicking outside of the element or pressing **esc** key will not deactivate it.",
-    "retainFocus": "Tab focus will return to the first child of the dialog by default. Disable this when using external tools that require focus such as TinyMCE or vue-clipboard.",
     "scrollable": "When set to true, expects a `v-card` and a `v-card-text` component with a designated height. For more information, check out the [scrollable example](/components/dialogs#scrollable)."
   },
   "events": {
--- a/packages/api-generator/src/locale/en/VMenu.json
+++ b/packages/api-generator/src/locale/en/VMenu.json
@@ -5,7 +5,7 @@
     "closeOnClick": "Designates if menu should close on outside-activator click.",
     "closeOnContentClick": "Designates if menu should close when its content is clicked.",
     "closeDelay": "Milliseconds to wait before closing component. Only works with the **open-on-hover** prop.",
-    "disableInitialFocus": "Prevents automatic redirect of first `focusin` event. Intended to use on permanently open menus or VSpeedDial.",
+    "disableInitialFocus": "Deprecated, use `capture-focus` instead. Prevents automatic redirect of first `focusin` event. Intended to use on permanently open menus or VSpeedDial.",
     "disableKeys": "Removes all keyboard interaction.",
     "internalActivator": "Detaches the menu content inside of the component as opposed to the document.",
     "minWidth": "Sets the minimum width for the component. Use `auto` to use the activator width.",
--- a/packages/api-generator/src/locale/en/generic.json
+++ b/packages/api-generator/src/locale/en/generic.json
@@ -20,6 +20,7 @@
     "falseValue": "Sets value for falsy state.",
     "firstDayOfWeek": "Sets the first day of the week, starting with 0 for Sunday. (Note: not guaranteed to work when using custom date adapters.)",
     "firstDayOfYear": "Sets the day that determines the first week of the year, starting with 0 for Sunday. For ISO 8601 this should be 4. (Note: not guaranteed to work when using custom date adapters.)",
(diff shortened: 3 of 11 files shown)


KaelWD:
Why is this separate from retainFocus? Seems kinda confusing to have two props that do almost the same thing but in slightly different ways.

J-Sek:
TODO:
- [x] ensure `capture-focus` works when activator is the last element in the DOM (like in the example in PR description)
  - maybe replace with `keydown`?
- [x] let **tab** escape VList and VRadioGroup

fanvyr:
Hm, can someone add a description to the [docs for that prop](https://v3.vuetifyjs.com/en/api/v-navigation-drawer/#props-capture-focus) please? not really sure what it does, but hoping it solves my current challenge :D

J-Sek:
@fanvyr, it is described in v4 docs. I have filled many missing props on the `master` branch, but did not have enough determination to port them to `v3-stable`.

fanvyr:
@J-Sek kk, easy. thanks for your reply

PR #22082, opened 2025-09-18
fix(V-Btn): prevent disabled links from opening on focus + enter key

Haviles04:
## Description
Fixes #22061 

## Markup:

```
<template>
  <v-app>
    <v-container class="d-flex flex-column">
      <v-text-field v-model="msg" />
      <v-btn @click="onClick">Active Button</v-btn>
      <v-btn color="secondary" :disabled="true" @click="onClick">I'm real button!</v-btn>
      <v-btn color="primary" href="http://www.google.com" target="_blank" :disabled="true" @click="onClick">I'm a link!</v-btn>
    </v-container>
  </v-app>
</template>

<script setup>
import { ref } from 'vue'

const msg = ref('Hello World!')
const onClick = () => {
  console.log('hi');
}
</script>

```

Issue #22071, opened 2025-09-17
[Feature Request] Add aria-disabled attribute to disabled VBtn components for improved accessibility


josemasf:
### Problem to solve
Currently, when a v-btn component is disabled using the disabled prop, it only applies visual styling (CSS classes) and prevents interactions, but lacks proper ARIA attributes for accessibility. This creates several issues:

Screen readers cannot properly announce the disabled state to users with visual impairments
Automated testing becomes difficult when trying to target disabled buttons
Does not follow WCAG 2.1 accessibility guidelines for disabled interactive elements
Inconsistent with modern web accessibility standards that recommend using aria-disabled="true" instead of or alongside the disabled attribute

```
<!-- Current output for disabled button -->
<button class="v-btn v-btn--disabled" disabled>
  Click me
</button>
```

```
<!-- Expected output with proper ARIA support -->
<button class="v-btn v-btn--disabled" disabled aria-disabled="true">
  Click me
</button>
```

This issue is part of a broader accessibility improvement effort. Similar issues may exist in other components like:

v-text-field (disabled inputs)
v-select (disabled selectors)
v-checkbox (disabled checkboxes)
Time picker clock items (already reported separately)

Visual error in storybook testing
<img width="1186" height="606" alt="Image" src="https://github.com/user-attachments/assets/6dc44897-248f-4f9c-8579-0a819c1284ba" />

### Proposed solution
Add the aria-disabled="true" attribute to VBtn components when the disabled prop is set to true. This should be implemented in the VBtn component to:

Automatically add aria-disabled="true" when disabled prop is true
Ensure the attribute is not present when the button is enabled
Maintain backward compatibility with existing implementations
Follow the same pattern used in other Vuetify components that support disabled states

johnleider:
Related #22061

Haviles04:
button fix : #22082 

still need to check 
v-text-field (disabled inputs)
v-select (disabled selectors)
v-checkbox (disabled checkboxes)

josemasf:
@Haviles04 yesterday i created https://github.com/vuetifyjs/vuetify/pull/22075 disabled and aria-disabled can be in the same button.

Haviles04:
> @Haviles04 yesterday i created https://github.com/vuetifyjs/vuetify/pull/22075 disabled and aria-disabled can be in the same button. 
> 

Wouldn't it be redundant with disabled on a regular button?

josemasf:
[...]

PR #22115, opened 2025-09-23
fix: (VListItem) fix incorrect tab index

Haviles04:
Fixes regression introduced by #22082 issue #22071 

## Description
Changed order so tabindex will overwrite the tabindex inside `linkProps`

## Markup:

```
<template>
  <v-app>
    <v-container>
      <v-select
        label="Select"
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">

</script>

```

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -283,13 +283,13 @@ export const VListItem = genericComponent<VListItemSlots>()({
             dimensionStyles.value,
             props.style,
           ]}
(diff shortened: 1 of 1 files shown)


J-Sek:
I see the same order (`tabindex={ ... }` before `{ ...link.linkProps }`) in VCard, VBtn and VChip. We might need to update those places - either by removing tabindex that's going to be overwritten by link props, or repeating the fix for VListItem.

TBH, it looks a bit brittle. Could we make it a parameter of `useLink(...)` instead?

PR #22117, opened 2025-09-23
feat(icon-set): add `mdi-unocss` icon set

userquin:
## Description

This PR adds a new icon set for MDI using UnoCSS.

### DIFF
--- a/packages/vuetify/src/composables/__tests__/icons.spec.ts
+++ b/packages/vuetify/src/composables/__tests__/icons.spec.ts
@@ -59,6 +59,7 @@ describe('icons.tsx', () => {
       md: await import('@/iconsets/md').then(m => m.aliases),
       mdi: await import('@/iconsets/mdi').then(m => m.aliases),
       mdiSvg: await import('@/iconsets/mdi-svg').then(m => m.aliases),
+      mdiUnoCss: await import('@/iconsets/mdi-unocss').then(m => m.aliases),
     }
     const mdi = unfill(sets.mdi)
 


packages/docs/src/pages/en/features/icon-fonts.md
```diff
@@ -202,6 +202,51 @@ Use this tool to search for any Material Design Icons and copy them to your clip
...
+```
+
+```bash [bun]
+bun add unocss @iconify-json/mdi -D
+```
+
+:::
+
+then, configure UnoCSS in your project adding the preset (read the [UnoCSS integration section](https://unocss.dev/integrations/) for further details).
+
```
userquin, line 237: Add a note about icon prefix at UnoCSS icons preset, should be the default `i-`, otherwise the icon set won't work => cannot make it dynamic here, we need some runtime (virtual vite/nuxt module) to allow UnoCSS pipeline to process the virtual module.

We can add some helpers at `@unvuetify/shared` to allow build custom Vuetify icon sets based on UnoCSS.

The next Nuxt module version will have this built-in feature: the icon set will be built with the prepare command, if added at `@unvuetify/shared` we can reuse it in the Nuxt module.

packages/docs/src/pages/en/features/icon-fonts.md
```diff
@@ -202,6 +202,51 @@ Use this tool to search for any Material Design Icons and copy them to your clip
...
+npm install unocss @iconify-json/mdi -D
+```
+
+```bash [bun]
+bun add unocss @iconify-json/mdi -D
+```
+
+:::
+
+then, configure UnoCSS in your project adding the preset (read the [UnoCSS integration section](https://unocss.dev/integrations/) for further details).
```
J-Sek, line 236: If those further details (at minimum) are just to add `uno.config.ts`

```ts
import { presetIcons, defineConfig } from 'unocss'

export default defineConfig({
  presets: [
    presetIcons(),
  ],
})
```

We could include it here and only send user out if he/she needs to learn more.

packages/vuetify/src/iconsets/mdi-unocss.ts
```diff
@@ -0,0 +1,83 @@
+// @unocss-include
```
J-Sek, line 1: I had exported a VPlay demo (to have a quickest setup) and applied those changes to verify if this file would be scanned by UnoCSS when it is burried in `node_modules`.

It seems to only scan the project files - which seems like a reasonable thing to do. Do you know how to force the scan? Does it require additional entry in the config file?

userquin: Uhmm, we also need to configure the pipeline, `[jt]s` files excluded by default, it works only with vite/webpack:

https://unocss.dev/config/#pipeline

So, VPlay cannot use this icon set.

J-Sek: I don't mind lack of support on VPlay, we can add it later (does not feel necessary though).

I only mentioned VPlay, because it exports `*.zip` file with a very basic project based on Vite. It was faster than `npm create vuetify`, lol :)

J-Sek: OK. It does work. Not sure why it did not pick up the file the first time.

J-Sek (approved):

PR #22120, opened 2025-09-25
docs(VDataTableVirtual): add example with itemRef slot prop

antonhansson0624:
Add example using itemRef slot with itemRef slot prop in virtual data table.

### DIFF
--- a/packages/docs/src/examples/v-data-table/virtual-custom.vue
+++ b/packages/docs/src/examples/v-data-table/virtual-custom.vue
@@ -1,0 +1,73 @@
+<template>
+  <v-data-table-virtual
+    :headers="headers"
+    :items="items"
+    item-key="id"
+    fixed-header
+  >
+    <template v-slot:item="{ columns, internalItem, props, itemRef }">
+      <tr v-bind="props" :ref="itemRef">
+        <td v-for="column in columns" :key="column.key">
+          {{ internalItem.raw[column.key] }}
+        </td>
+      </tr>
+    </template>
+  </v-data-table-virtual>
+</template>
+
+<script setup>
+  import { computed } from 'vue'
+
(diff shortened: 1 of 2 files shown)


packages/docs/src/examples/v-data-table/virtual-custom.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <v-data-table-virtual
+    :headers="headers"
+    :items="items"
+    item-key="id"
+    fixed-header
```
J-Sek: `fixed-header` only makes sense with limit on height (e.g. `height="400"`)

packages/docs/src/examples/v-data-table/virtual-custom.vue
```diff
@@ -0,0 +1,73 @@
...
+  const headers = [
+    { title: 'ID', value: 'id', sortable: true },
+    { title: 'Title', value: 'title', sortable: true },
+    { title: 'Name', value: 'name', sortable: true },
+    { title: 'Email', value: 'email', sortable: true },
+    { title: 'Role', value: 'role', sortable: true },
+  ]
+
+  const items = computed(() => [
+    { id: 1, title: 'Dr.', name: 'Sarah Johnson', email: 'sarah.johnson@company.com', role: 'Owner' },
```
J-Sek: Demo does not showcase the described functionality with a static items array with just a dozen items. Let's change it to `Array.from({ ... }).map(...)`.

I imagine items could be `computed` and some kind of chip-group above could let us toggle the overall size of the generated collection between 100/400/800. This way user could observe the scrollbar height changes - seeing a proof that it works as expected.

J-Sek (changes requested):

PR #22121, opened 2025-09-25
docs(transitions): add guidance on using padding with expand transitions

antonhansson0624:
Add guidance on using padding with expand transitions

### DIFF
--- a/packages/docs/src/examples/transitions/misc-expand-x-padding.vue
+++ b/packages/docs/src/examples/transitions/misc-expand-x-padding.vue
@@ -5,11 +5,15 @@
         Expand Transition
       </v-btn>
 
-      <v-expand-transition>
-        <div v-show="expand">
-          <v-alert text="Smooth transition here" type="info"></v-alert>
-        </div>
-      </v-expand-transition>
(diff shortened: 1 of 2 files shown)


packages/docs/src/examples/transitions/misc-expand-x-padding.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <v-app>
+    <v-container>
+      <v-btn class="ma-2" color="primary" @click="expand = !expand">
+        Expand Transition
+      </v-btn>
+
+      <v-expand-transition>
```
J-Sek: That `*-x-*` transition works just fine when alert is meant to extend to 100%, but requires more CSS skills to get something that fits the content. It might be helpful for the devs to see an example right here and learn it is possible with this little trick:

```html
<v-expand-x-transition>
  <v-responsive v-show="expand" width="fit-content">
    <v-alert
      text="Smooth transition here"
      type="info"
      class="text-no-wrap"
    ></v-alert>
  </v-responsive>
</v-expand-x-transition>
```

J-Sek (approved):

PR #22146, opened 2025-10-02
fix(validation): use correct translation placeholders

RobertMe:
## Description

When using validation in combination with vue-i18n (/`createVueI18nAdapter`) translation of messages containing placeholders are incorrect. This is due to the placeholders always being passed as an array (`[len]` in all cases). This leads to vue-i18n `JSON.stringify`ing the placeholder and thus leading to for example "You must enter a maximum of [ 1 ] characters" with the brackets surrounding the number. This seems to "accidentally" work with the vuetify adapter as this always `String(...)`s the value: https://github.com/vuetifyjs/vuetify/blob/80e154b8c295458b9ced262b3e20403fe07376c3/packages/vuetify/src/locale/adapters/vuetify.ts#L19

Checking other usages of `t()` in the Vuetify codebase also shows that parameters are passed as is, optionally adding more arguments like here: https://github.com/vuetifyjs/vuetify/blob/80e154b8c295458b9ced262b3e20403fe07376c3/packages/vuetify/src/components/VCarousel/VCarousel.tsx#L156

So using `[len]` instead of `len` seems wrong to me. And it just accidentally happens to seem to work fine for arrays with just one element due to how the string conversion works as the vuetify adapter applies.

## Markup:

This is hard to reproduce with just the playground, as the rules plugin isn't enabled (by default) and to actually reproduce the issue when using vue-i18n obviously also requires that to be installed and configured. But at least a starting point for the template is:
```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        :rules="[rules.maxLength(1)]"
      />
    </v-container>
  </v-app>
</template>

<script>
  import { useRules } from '@/labs/rules'

  export default {
    name: 'Playground',
    setup () {
      return {
        rules: useRules(),
      }
    },
  }
</script>
```

Which in addition requires modifying `index.js` as well:
```js
import { createRulesPlugin } from '@/labs/rules/index.js'
```
Plus in `export default viteSSR()`:
```js
app.use(createRulesPlugin({}, vuetify.locale))
```

My actual test case that it also works using vue-i18n was to just link the Vuetify clone (with fix) to my actual project.

### DIFF
--- a/packages/vuetify/src/labs/rules/rules.ts
+++ b/packages/vuetify/src/labs/rules/rules.ts
@@ -62,13 +62,13 @@ export function createRules (options: RulesOptions | undefined, locale: LocaleIn
       return (v: string) => (/^[A-Z]*$/.test(v)) || t(err || '$vuetify.rules.capital')
     },
     maxLength: (len: number, err?: string) => {
-      return (v: any) => (!v || v.length <= len) || t(err || '$vuetify.rules.maxLength', [len])
+      return (v: any) => (!v || v.length <= len) || t(err || '$vuetify.rules.maxLength', len)
     },
     minLength: (len: number, err?: string) => {
-      return (v: any) => (!v || v.length >= len) || t(err || '$vuetify.rules.minLength', [len])
+      return (v: any) => (!v || v.length >= len) || t(err || '$vuetify.rules.minLength', len)
     },
     strictLength: (len: number, err?: string) => {
-      return (v: any) => (!v || v.length === len) || t(err || '$vuetify.rules.strictLength', [len])
+      return (v: any) => (!v || v.length === len) || t(err || '$vuetify.rules.strictLength', len)
     },
     exclude: (forbiddenCharacters: string[], err?: string) => {
       return (v: string) => {


RobertMe:
Forgot to mention. I did check if there are tests targeting the validation rules code. But that doesn't seem to be the case so I couldn't add (regression) tests either. If there are tests which must be updated then please let me know (and point me to where they are :)) and I'll add them.

J-Sek:
I have only used `vue-i18n` years ago in a small project, but will try to help. Please bare with me :)

Either adapter is missing something or I can't figure out how to use it. In this [demo](https://play.vuetifyjs.com/#eNqlVl1r2zAU/SvCG9gBR1raDUZISvu2wR7GGHuJ86DYN6lXWTKSnKWE/PddSU7q2unXRilYV/eec3SuPrLYR0bnrOKlpL9NNI3Kqlbakj3JNXALN3VNDmStVUXibQNxJnsJvxqw5fq+m+TGw8QfjQDzXTSbUvaTmeArw7RLOEvwdfJZ3hS8tqCHpSrnAhgP04ZhfFxi/hDIoaSkMf6jizPIhzMKPQ0mndKcMyGHMvymj9zBmAHb1Oipr8mVNJY4HjLvyEn2mSREwIbn91Oy5sJA6iOebUpikLEPVGAM34CZtuoweBg9ALcqT9htUwK8vYUKsfakgDVvhP0ZxnHB9V3soR4IfQEhrZvTsy1I9n4hHStHHqMvSokCBbXSaGCgNpP9CCYld3CfkpprXpkRmV8FHRod1NLhuHmqoRY8hyR+f0SIUxLHo5RQSttaVPAggGOHjo5ghxL8P8lTyCzUJnEroRuhVlxQ24H2m5FWfPcN5MbeItPicjnC8meKn6zJJEqhaJevwIAbItC63FDQWukvXBYCN/ecgFv9kcPPJdAWuPpWHkYI8YHB2Ur2h7RnupOA6ZVqJKp8h1gxRqLUn/xjqj/8jKGV2P2CoCD0v1JbIOtSALGKwK5WBsjxqgjl7c7H2pmFChtk4cqxzbZj5PGffoBLslgIug35oIWdHa9LwI0y9ebNs2ixiN93LZwsl1lE2ItlPfOTyahbN2N9CS4SFM5YRzkOTa7L2hJ/fn3q6V5Aw73Tz99fxPfPEj/Glh6rEjR9xgI64qKBAXhccXdPKIkm+n2ftRMmi04nMouQy42z6Nba2kwZywvXhwJEudVUgmWyrtg1pqESacsKxoWqri/pJ3pxwYrS2G6cgqnGK63+GNCIkkX+CLc8bk2v4GozkWLygX4MFG1s7BxxHGexO369nUaUq049K2UBuydY/O7/R4bwrLyIPnx4/otvgDbgfgOHT72eTCj+tTvgBPtc+/Hp0lvQYw24eg36tduuV/Zo6/XmBvz+AcHbOzqkUajD8xH8idLwMD76oRJiy79plP3p) I had to patch `t()` from either `i18n.global.*` or `vuetify.locale.*`

```ts
const old = vuetify.locale.t
vuetify.locale.t = (key, params) => {
  return old(key.replace('$vuetify.', ''), ...params)
}
```

Without it, the messages are not resolved and `t('$vuetify.rules.maxLength')` returns "$vuetify.rules.maxLength".

The bug you experience occurs only if I prefix params with "..." - `(key, ...params) =>`, otherwise it works fine.

RobertMe:
Hi @J-Sek, I don't have my own code at hand right now. But this isn't a vue-i18n (adapter) issue. It's an issue with how `t()` is called in the validators. Using `[len]`, instead of `len`. This obviously makes it an array which is incorrect.

See for example the other `t()` call in Vuetify I linked earlier: https://github.com/vuetifyjs/vuetify/blob/80e154b8c295458b9ced262b3e20403fe07376c3/packages/vuetify/src/components/VCarousel/VCarousel.tsx#L156
here three arguments are given, the translation key, and two placeholders. And the language line also contains two placeholders `{0}` and `{1}`:https://github.com/vuetifyjs/vuetify/blob/80e154b8c295458b9ced262b3e20403fe07376c3/packages/vuetify/src/locale/en.ts#L62

So I don't see any reason why in the validation rules `[len]` must be used and not `len`.

And for the vuetify adapter it just accidentally works:
https://github.com/vuetifyjs/vuetify/blob/80e154b8c295458b9ced262b3e20403fe07376c3/packages/vuetify/src/locale/adapters/vuetify.ts#L19

In my example `params[Number(index)]` would be `[1]` (instead of `1`). And calling `String` constructor using an array just happens to concat/join all elements in the array (with `,` as separator). So `String([1])` equals `"1"` (/string `1`) which doesn't expose the issue.
But when using the vue-i18n adapter the logic when an array is used as placeholder differs (and it calls `JSON.stringify`). But this difference in translator implementation is irrelevant in this case. It's just `t()` getting called incorrectly.

And your proposed fix (spreading the array) would be wrong. Because in all other cases (once again, see the VCarousel example I linked) `params` would just be a single placeholder, and using the spread operator would thus not work (in the example `index + 1` is the first placeholder, so a `Number`, and you can't spread a `Number`).

J-Sek (approved):

PR #22152, opened 2025-10-03
feat(vuetify): export internal types

userquin:
This PR adds a new `internal-types` subpackage export to allow make configuration statically analisable at the new Vuetify Nuxt module when merging layers.

We should add some jsdoc to the `internal` types, the current version (3.10.4) with internal types there, just not exported from `framework.d.ts`.

Right now I'm getting some errors when building the configuration (`VuetifyOptions`) `d.ts` via `tsdown`:

tsdown build error

With this PR:

tsdown build with this PR


userquin:
I need to review labs, maybe we also need there expose some internals (Vuetify Rules seems to be fine)

userquin:
About fast-glob and node:fs:
- https://nodejs.org/api/fs.html#fsglobsyncpattern-options
- https://nodejs.org/api/fs.html#fspromisesglobpattern-options

J-Sek:
Is this PR still relevant? I have a working project with Nuxt layers monorepo rocking Vuetify v4 just fine with `vuetify-nuxt-module@1.0.0-beta.2`.

userquin:
We need the types for the next version (v2) of vuetify-nuxt-module where we're going to create the vuetify options at .nuxt/vuetify folder, nuxt will fail since not all internal types used at vuetify options are exported, check the screenshots in the pr description (cannot access to VariationOptions type error).

J-Sek:
Maybe we need to break some config types into smaller types. I'd like to play with this next version. Is what I see on these screenshots on a branch in your repo?

userquin:
> Maybe we need to break some config types into smaller types. I'd like to play with this next version. Is what I see on these screenshots on a branch in your repo?

Yes, but only on my local, I can share with you the changes via ZIP or just send a few screenshots or modules (I will need to resolve conflicts with huge changes added by Andrei for v1 alpha/beta versions). I will try to share the info in a few hours (DM at discord) or maybe tomorrow.

**NOTE**: maybe we don't need the types, I was checking creating the dts file using tsdown but cannot be used, since nuxt/nuxi will run creating files at .nuxt folder outside the lifecycle of nuxt (it is similar on how vite copies the public folder to the dist folder)

Issue #22174, opened 2025-10-09
[Bug Report][3.10.0] VMaskInput - clear does not work and throws error


gduliscouet-ubitransport:
### Environment
**Vuetify Version:** 3.10.5
**Vue Version:** 3.5.21
**OS:** macOS 10.15.7 (current)

### Steps to reproduce
1. Fill the input with "1" (for example)
2. Click on the clear icon

### Expected Behavior
It clears the input

### Actual Behavior
It doesn't clear the input and a JS error is thrown `Cannot read properties of null (reading 'split')`

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpVj0EOgjAQRa8ymTUU0bAxlcQDeALrYihjQiy1KYWN8e621AXsXt5k5v+5f3Dyuro6J5aZ8Ywy8OgMBW6VlUtJzkUAgMj6bQMNln02qxtpepWDdXMAbZg8dYYhyYtC7bkfQqnJ9wrBUMcm2lscwrqhsJXV9sQ/KbpdVBK5h6w27fBb4Ek04lhjgvogGiyeZCYudj9l9/gBK4ZNFg==)

### Other comments
According to the stacktrace, removeMaskDelimiters (VMaskInput.tsx) is called with null
FYI @ikushum 

### Images
![https://cdn.cosmicjs.com/2e149710-a50b-11f0-936e-dbe343b25d95-Screenshot-2025-10-09-at-14-26-07.png](https://cdn.cosmicjs.com/2e149710-a50b-11f0-936e-dbe343b25d95-Screenshot-2025-10-09-at-14-26-07.png)

PR #22175, opened 2025-10-09
fix(VMaskInput): handle null value when clearing input

babu-ch:
## Description

fixes https://github.com/vuetifyjs/vuetify/issues/22174

## Markup:

```vue
<template>
<v-app>
   <v-container>
     <v-mask-input clearable mask="credit-card" label="Mask input"></v-mask-input>
   </v-container>
 </v-app>
</template>

```

### DIFF
--- a/packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
+++ b/packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
@@ -47,7 +47,7 @@ export const VMaskInput = genericComponent<VMaskInputSlots>()({
       val => props.mask ? mask.mask(mask.unmask(val)) : val,
       val => {
         if (props.mask) {
-          const valueWithoutDelimiters = removeMaskDelimiters(val === null ? '' : val)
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -47,7 +47,7 @@ export const VMaskInput = genericComponent<VMaskInputSlots>()({
       val => props.mask ? mask.mask(mask.unmask(val)) : val,
       val => {
         if (props.mask) {
-          const valueWithoutDelimiters = removeMaskDelimiters(val)
+          const valueWithoutDelimiters = removeMaskDelimiters(val === null ? '' : val)
```
J-Sek:
```suggestion
          const valueWithoutDelimiters = val ? removeMaskDelimiters(val) : ''
```

babu-ch:
@J-Sek Thank you for the suggestion! Your code is better!

PR #22176, opened 2025-10-09
fix(VMaskInput): Fixes error when mask input is cleared

### LINKED ISSUE #22174: [Bug Report][3.10.0] VMaskInput - clear does not work and throws error

ikushum:
fixes #22174

## Description
Fixes error when mask input is cleared

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-mask-input label="Mask input" mask="credit-card" clearable persistent-clear />
    </v-container>
  </v-app>
</template>

```

### DIFF
--- a/packages/vuetify/src/labs/VMaskInput/__tests__/VMaskInput.spec.browser.tsx
+++ b/packages/vuetify/src/labs/VMaskInput/__tests__/VMaskInput.spec.browser.tsx
@@ -32,6 +32,21 @@ describe('VMaskInput', () => {
     expect(inputValue.value).toBe('(456) 7')
   })
 
+  it('should clear the input when clear icon is clicked', async () => {
+    const inputValue = ref('(456) 7')
+
+    render(() => (
+      <VMaskInput
+        v-model={ inputValue.value }
+        mask="(###) #"
+        clearable
+      />
+    ))
+
+    await userEvent.click(screen.getByLabelText('Clear'))
+    expect(inputValue.value).toBe('')
+  })
+
   describe('Caret Position and Formatting', () => {
     const renderComponent = ({
       defaultModel = '(AS)-123-XYZ-45-67',


packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx
```diff
@@ -72,6 +72,8 @@ export const VMaskInput = genericComponent<VMaskInputSlots>()({
     const validationValue = toRef(() => returnMaskedValue.value ? model.value : mask.unmask(model.value))
 
     function removeMaskDelimiters (val: string): string {
+      if (!val) return ''
```
J-Sek: Slightly buried... and looks suspicious without `val: string | null` above. @babu-ch already merged a fix one level above. His change could be further shortened to `removeMaskDelimiters(val ?? '')`, but it feels reasonable to skip the function, actually.

J-Sek:
We could benefit from making `model` type more precise with additional ` as string | null` at the end of line 47

https://github.com/vuetifyjs/vuetify/blob/46e3a84b582aa0c8f7e8aba4e00be3b1a6f7bddd/packages/vuetify/src/labs/VMaskInput/VMaskInput.tsx#L47

KaelWD:
Replaced with #22175

Issue #22052, opened 2025-09-12
[Bug Report][3.10.0] Checkboxes can become out of sync with VSelect model


RylanBueckert-Broadsign:
### Environment
**Vuetify Version:** 3.10.0
**Vue Version:** 3.5.21
**OS:** Windows 10 (current)

### Steps to reproduce
1. Open reproduction link
2. Select 6 days of the week from the select
3. Also select the last remaining weekday by clicking directly on the checkbox
4. See that all the days have become unselected except the checkbox for that last day still appears checked

### Expected Behavior
Each `VSelect` item's checkbox is only checked when its value is in the modelValue array.

### Actual Behavior
checkbox can get into a state where the item is not selected, but the checkbox is checked

### Reproduction Link
[...]

PR #22181, opened 2025-10-10
fix(VSelect): item checkbox gets out of sync

jcjp:
## Description

Fixes: #22052

####

- Prevent the checkbox to be out of sync in certain scenario, where we programmatically untick the checkboxes and disable them as in the case of the bug reported here. When we try to click all checkboxes from Sunday - Saturday, Everyday will be selected and the others will be unselected but in our previous behavior the last checkbox is left selected making it out of sync.

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-select
        v-model="selectModel"
        :items="dayOptions"
        :item-props="(item) => selectModel.includes('everyDay') && item.value !== 'everyDay' ? { disabled: true } : {}"
        :menu-props="{ maxHeight: undefined }"
        closable-chips
        multiple
        chips
      />
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref, computed, readonly } from 'vue'

  const weekDayTitles = readonly([
    'Sunday',
    'Monday',
    'Tuesday',
    'Wednesday',
    'Thursday',
    'Friday',
    'Saturday',
  ]);

  type OptionValue = number | 'everyDay';

  const dayOptions = computed<{ title: string; value: OptionValue }[]>(() => {
    return [
      {
        title: "Every Day",
        value: 'everyDay',
      },
      ...Array.from({ length: 7 }, (_, i) => ({
        title: weekDayTitles[i],
        value: i,
      })),
    ];
  });

  const model = defineModel<number[]>({ default: []});

  const selectModel = computed<OptionValue[]>({
    get() {
      if (model.value.length === 7) {
        return ['everyDay'];
      }

      return model.value;
    },
    set(value) {
      if (value.includes('everyDay')) {
        model.value = [0, 1, 2, 3, 4, 5, 6];
      } else {
        model.value = value as number[];
      }
    },
  });
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -495,7 +495,6 @@ export const VSelect = genericComponent<new <
                                         modelValue={ isSelected }
                                         ripple={ false }
                                         tabindex="-1"
-                                        onPointerdown={ (event: PointerEvent) => event.preventDefault() }
                                         onClick={ (event: MouseEvent) => event.preventDefault() }
                                       />
                                     ) : undefined }


packages/vuetify/src/components/VSelect/VSelect.tsx
```diff
@@ -495,6 +496,12 @@ export const VSelect = genericComponent<new <
                                         modelValue={ isSelected }
                                         ripple={ false }
                                         tabindex="-1"
+                                        onPointerdown={ (event: PointerEvent) => event.preventDefault() }
```
J-Sek: Do we actually need anything except

```jsx
onClick={ (event: MouseEvent) => event.preventDefault() }
```

?

Issue #18447, opened 2023-10-11
[Bug Report][3.3.21] v-tabs toggle auto-scroll to top


Sweet520:
### Environment
**Vuetify Version:** 3.3.21
**Vue Version:** 3.3.4
**Browsers:** Chrome 117.0.0.0
**OS:** Windows 10

### Steps to reproduce
Scroll down to see the tab content, switch tabs to the right, and the page will automatically scroll to the top

### Expected Behavior
By default, when switching tabs, the page will not scroll

### Actual Behavior
Switch the tab to the right and the page will automatically scroll to the top

### Reproduction Link
[...]

PR #22191, opened 2025-10-13
fix(VWindow): Maintain scroll position on window change

MatthewAry:
fix(VWindow): Maintain scroll position on window change

Fixes https://github.com/vuetifyjs/vuetify/issues/18447

It's possible this issue is being caused by something upstream. If that proves to be the case, then this fix is a hack and should be closed in favor of a solution that will resolve the problem upstream.

### DIFF
--- a/packages/vuetify/src/components/VWindow/VWindow.tsx
+++ b/packages/vuetify/src/components/VWindow/VWindow.tsx
@@ -17,6 +17,7 @@ import vTouch from '@/directives/touch'
 // Utilities
 import { computed, nextTick, provide, ref, shallowRef, toRef, watch } from 'vue'
 import { convertToUnit, genericComponent, IN_BROWSER, PREFERS_REDUCED_MOTION, propsFactory, useRender } from '@/util'
+import { getScrollParent } from '@/util/getScrollParent'
 
 // Types
 import type { ComputedRef, InjectionKey, PropType, Ref } from 'vue'
@@ -145,20 +146,7 @@ export const VWindow = genericComponent<new <T>(
       const savedScrollPosition = { x: 0, y: 0 }
 
       if (IN_BROWSER) {
-        // Find the first scrollable parent
-        let parent = rootRef.value?.parentElement
-        while (parent) {
-          const { overflowY, overflowX } = window.getComputedStyle(parent)
-          const isScrollable = /(auto|scroll)/.test(overflowY + overflowX)
-          if (isScrollable && (parent.scrollHeight > parent.clientHeight || parent.scrollWidth > parent.clientWidth)) {
-            scrollableParent = parent
-            break
-          }
-          parent = parent.parentElement
-        }
-
-        // Default to document root if no scrollable parent found
-        if (!scrollableParent) scrollableParent = document.documentElement
+        scrollableParent = getScrollParent(rootRef.value)
 
         // Save current scroll position
         savedScrollPosition.x = scrollableParent.scrollLeft


packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -139,7 +139,14 @@ export const VWindow = genericComponent<new <T>(
       return group.items.value.findIndex(item => group.selected.value.includes(item.id))
     })
 
+    // Fix for #18447
+    const savedScrollPosition = ref<{ x: number, y: number } | null>(null)
+
     watch(activeIndex, (newVal, oldVal) => {
+      const scrollX = typeof window !== 'undefined' ? (window as any).scrollX || 0 : 0
```
johnleider: Could probably use global constants to check window stuff

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -139,7 +139,14 @@ export const VWindow = genericComponent<new <T>(
       return group.items.value.findIndex(item => group.selected.value.includes(item.id))
     })
 
+    // Fix for #18447
```
johnleider:
```suggestion
    // Fix for https://github.com/vuetifyjs/vuetify/issues/18447
```

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -152,7 +159,23 @@ export const VWindow = genericComponent<new <T>(
       } else {
         isReversed.value = newVal < oldVal
       }
-    })
+
+      nextTick(() => {
+        const currentScrollY = typeof window !== 'undefined' ? (window as any).scrollY || 0 : 0
```
johnleider: Same here, use global constants for checking window.

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -152,7 +159,25 @@ export const VWindow = genericComponent<new <T>(
...
-    })
+
+      nextTick(() => {
+        if (!IN_BROWSER) return
+
+        const currentScrollY = window.scrollY
+
+        if (savedScrollPosition.value && currentScrollY !== savedScrollPosition.value.y) {
+          window.scrollTo(savedScrollPosition.value.x, savedScrollPosition.value.y)
+        }
```
johnleider: If this condition passes should it return early as opposed to proceeding to requestAnimationFrame?

MatthewAry: If you return early after the first restoration, you might miss a second scroll reset that happens during the browser's paint phase when the transition CSS (`position: absolute; top: 0`) is fully applied.

When investigating this issue, my debug logging showed that the scroll was being reset after the Vue component updates, which is exactly when requestAnimationFrame fires. This is why both callbacks are necessary - they're covering different phases of the browser's render pipeline.

packages/vuetify/src/components/VWindow/VWindowItem.tsx
```diff
@@ -81,8 +81,8 @@ export const VWindowItem = genericComponent()({
       isTransitioning.value = true
 
       if (window.transitionCount.value === 0) {
-        // Set initial height for height transition.
-        window.transitionHeight.value = convertToUnit(window.rootRef.value?.clientHeight)
```
J-Sek, line 85: Better to leave this file unchanged, unless these changes affect the behavior

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -139,7 +139,14 @@ export const VWindow = genericComponent<new <T>(
       return group.items.value.findIndex(item => group.selected.value.includes(item.id))
     })
 
+    // Fix for https://github.com/vuetifyjs/vuetify/issues/18447
+    const savedScrollPosition = ref<{ x: number, y: number } | null>(null)
+
     watch(activeIndex, (newVal, oldVal) => {
+      const scrollX = IN_BROWSER ? window.scrollX : 0
```
J-Sek: Should we rather traverse up to find the first scrollable parent?

I have changed to `<v-main scrollable>` and put the image spacer within. The unexpected jump up is back..

J-Sek: Quick draft that covers preventing only vertical scroll

```ts
const savedScrollPosition = { x: 0, y: 0 }

watch(activeIndex, (newVal, oldVal) => {
  let parent: HTMLElement | null = null
  if (IN_BROWSER) {
    parent = rootRef.value?.parentNode
    while (parent && !parent.scrollTop) parent = parent.parentNode as HTMLElement

    savedScrollPosition.x = parent?.scrollLeft ?? 0
    savedScrollPosition.y = parent?.scrollTop ?? 0
  }

  const itemsLength = group.items.value.length
  const lastIndex = itemsLength - 1

  if (itemsLength <= 2) {
    isReversed.value = newVal < oldVal
  } else if (newVal === lastIndex && oldVal === 0) {
    isReversed.value = true
  } else if (newVal === 0 && oldVal === lastIndex) {
    isReversed.value = false
  } else {
    isReversed.value = newVal < oldVal
  }

  nextTick(() => {
    if (!IN_BROWSER || !parent) return

    const currentScrollY = parent.scrollTop

    if (savedScrollPosition && currentScrollY !== savedScrollPosition.y) {
      parent.scrollTo({ left: savedScrollPosition.x, top: savedScrollPosition.y, behavior: 'instant' })
    }

    requestAnimationFrame(() => {
      const rafScrollY = parent.scrollTop

      if (savedScrollPosition && rafScrollY !== savedScrollPosition.y) {
        parent.scrollTo({ left: savedScrollPosition.x, top: savedScrollPosition.y, behavior: 'instant' })
      }
    })
  })
}, { flush: 'sync' }) // Run synchronously before DOM updates
```

`savedScrollPosition` does not seem to benefit from `ref` here

Edit: it does not need to fallback to `window`

MatthewAry: I have made changes according to your feedback. Please review.

J-Sek: In this scenario `window` is effectively the same as `document.body.parentNode` (that gets `<html>`), but the latter has the API of HTMLElement. It could help simplify the code.

We need `instant` back (probably accidentally overriden during merge conflict resolution).

```ts
parent.scrollTo({ left: savedScrollPosition.x, top: savedScrollPosition.y, behavior: 'instant' })
```

MatthewAry: Updated.

packages/vuetify/src/components/VWindow/VWindow.tsx
```diff
@@ -139,7 +139,32 @@ export const VWindow = genericComponent<new <T>(
...
 
+    // Fix for https://github.com/vuetifyjs/vuetify/issues/18447
     watch(activeIndex, (newVal, oldVal) => {
+      let scrollableParent: HTMLElement | undefined
+      const savedScrollPosition = { x: 0, y: 0 }
+
+      if (IN_BROWSER) {
+        // Find the first scrollable parent
+        let parent = rootRef.value?.parentElement
+        while (parent) {
```
J-Sek: Could be replaced with `getScrollParent` from `'@/util'`. Sorry for not suggesting it earlier.

MatthewAry: Tried it and it broke. Not sure why. I don't think adding the util will add any significant value anyways.

J-Sek: Weird. It works for me. Could you share a playground or a hint about the problematic scenario?

```ts
scrollableParent = getScrollParent(rootRef.value) ?? document.documentElement
savedScrollPosition.x = scrollableParent.scrollLeft
savedScrollPosition.y = scrollableParent.scrollTop
```

MatthewAry: Did I do it wrong? [`797b231` (#22191)](https://github.com/vuetifyjs/vuetify/pull/22191/commits/797b231ffb0e7e76eb98abe773193fa570a6d4ae)

MatthewAry: Seems I did. Will push again.

J-Sek (approved):

PR #22196, opened 2025-10-13
fix(VDateInput): apply min/max to text input

### LINKED ISSUE #22179: [Bug Report][3.10.5] VDateInput – min/max should apply to input validation
### Environment
**Vuetify Version:** 3.10.5
**Vue Version:** 3.5.22

### Steps to reproduce
1. Use the v-date-input component with a v-model bound to a reactive variable.
2. Set the min prop, e.g., min="2025-10-01".
3. Type a date before 2025-10-01 directly into the text field (e.g., 2025-09-15).
4. The value is accepted without validation error - modelValue updates to the invalid date.
5. Change the min prop dynamically (e.g., from 2025-10-01 to 2025-11-01) and observe that existing model values are not revalidated.

### Expected Behavior
When the min prop is set (e.g., min="2025-10-01"), the component should validate the current model value against this restriction.
If the user types a date earlier than the min value in the text field, the input should either:

- show a validation error, or

- prevent the model from being updated with the invalid value.

When the min prop changes dynamically, the component should revalidate the model and reset or mark it invalid if it no longer meets the new minimum date.

### Actual Behavior

- The min prop has no effect when the user types a date directly into the text field.

- The model (v-model / modelValue) updates successfully even if the new value is less than the min date.

- Changing the min prop after the component has been rendered does not trigger validation or update the model.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNpVUM1ugzAMfhUrl24SAUrFNE200l5hO44dsmDUTCZExNBOVd99IagqPdn+fvz3dRF+0Nm7c+k0ongTFWPnSDEeagtQNWYCTcr7fS0a2RKe4Xf0bNo/qdEyDrWIwiCdZBNs0lg38gIBTLLrG6RgjrEWN4LUT4Q/kVAzKJi9d7pTZ3kyDR+DZPfyuiKMDVCRF6Xc5jIvb8yhytbzl+WzsH3Iqmx1Uyi9Hoxj8Miji0LTuX5guIA/KqL+9IEtXKEd+g424Sub2QWge+sZ4h2wX0mf7Ej0PI9ZGoeW4pqIXVqmRSHmZJunpUhaRR6Th3cv2Pc/5td8kw==)

emmettxt:
## Description

 fixes #22179

```vue
<template>
  <v-app>
    <v-container>
      <v-card title="single input">
        <v-card-text>
          <VDateInput label="2026 Only" max="12/31/2026" min="1/1/2026" />
          <VDateInput label="Before 2027" max="12/31/2026" />
          <VDateInput label="After 2025" min="1/1/2026" />
          <VDateInput label="Any" />
        </v-card-text>
      </v-card>
      <v-card title="multiple">
        <v-card-text>
          <VDateInput
            label="2026 Only"
            max="12/31/2026"
            min="1/1/2026"
            multiple
          />
          <VDateInput label="Before 2027" max="12/31/2026" multiple />
          <VDateInput label="After 2025" min="1/1/2026" multiple />
          <VDateInput label="Any" multiple />
        </v-card-text>
      </v-card>

      <v-card title="Range">
        <v-card-text>
          <VDateInput
            label="2026 Only"
            max="12/31/2026"
            min="1/1/2026"
            multiple="range"
          />
          <VDateInput label="Before 2027" max="12/31/2026" multiple="range" />
          <VDateInput label="After 2025" min="1/1/2026" multiple="range" />
          <VDateInput label="Any" multiple="range" />
        </v-card-text>
      </v-card>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VNumberInput/VNumberInput.sass
+++ b/packages/vuetify/src/components/VNumberInput/VNumberInput.sass
@@ -54,7 +54,7 @@
           flex: 1
 
       .v-field--variant-underlined
-        > .v-field__prepend-inner:has(.v-number-input__control)
+        > .v-field__prepend-inner:has(.v-number-input__control),
         > .v-field__append-inner:has(.v-number-input__control)
           // drop input padding
           padding-top: var(--v-field-padding-top)
--- a/packages/vuetify/src/components/VNumberInput/VNumberInput.tsx
+++ b/packages/vuetify/src/components/VNumberInput/VNumberInput.tsx
@@ -135,7 +135,7 @@ export const VNumberInput = genericComponent<VNumberInputSlots>()({
       if (
         isFocused.value &&
           !controlsDisabled.value &&
-          Number(_inputText.value) === model.value
+          Number(_inputText.value?.replace(decimalSeparator.value, '.')) === model.value
       ) {
         // ignore external changes while typing
         // e.g. 5.01{backspace}2 » should result in 5.02


J-Sek:
I am a bit concerned about auto-clamp. We have something similar in VNumberInput but I believe we did a wrong turn there.

Let's say the preloaded value is out of range
- when the field has `readonly` user should not be able to change the value just by focusing and blurring
- in regular scenario (editable) it is also weird if the user opens/closes the menu/picker and gets surprised by "autofix"

Can we somehow apply the clamp to cover the actual gap - typing/pasting?

emmettxt:
Thanks for the feedback. I did take the clamp approach from the VNumberInput, for consistency.

>In regular scenario (editable) it is also weird if the user opens/closes the menu/picker and gets surprised by "autofix"

I think this behavior aligns with the default value of the updateOn prop ['blur', 'enter']. We can move the clamping into onUserInput method rather than useProxiedModel for semantics. However will still behave the same with the default updateOn. 


> when the field has readonly user should not be able to change the value just by focusing and blurring

With respect to this, adding a reaonly check to the onUserInput should be sufficient

J-Sek:
> _We can move the clamping into onUserInput_

It should be a step forward. Let's do it.

> _...readonly check to the onUserInput should be sufficient_

Sounds reasonable. Although, I would double-check if we even enter this method if `readonly` is present.

PR #22213, opened 2025-10-16
feat(VBreadcrumbs): pass props and width to items

### LINKED ISSUE #12249: [Feature Request] Breadcrumbs - add truncatable option
### Problem to solve
For `v-breadcrumbs` it would be nice to have a `truncatable` option. Generally it means that when a breadcrumbs content does not fit into container, it is cutted from the right and an ellipsis is shown.

Here is an example:
- `truncatable: false`
Screenshot 2020-09-18 at 22 52 14

- `truncatable: true`
Screenshot 2020-09-18 at 22 53 07

### Proposed solution
Add `truncatable` option to breadcrumb.

J-Sek:
resolves #12249

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container>
      <v-breadcrumbs
        :items="[
          'Foo',
          'Bar',
          {
            title: 'Baaaaaaaaaaaaaaaaaaaaaaaaaaaz',
            class: 'd-inline-block text-truncate',
            maxWidth: 120,
          }]"
      />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/docs/src/data/new-in.json
+++ b/packages/docs/src/data/new-in.json
@@ -31,6 +31,17 @@
       "bgColor": "3.4.0"
     }
   },
+  "VBreadcrumbs": {
+    "props": {
+      "itemProps": "3.11.0"
+    }
+  },
+  "VBreadcrumbsItem": {
+    "props": {
+      "width": "3.11.0",
+      "maxWidth": "3.11.0"
+    }
+  },
   "VBtn": {
     "props": {
       "text": "3.2.0",
--- a/packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
+++ b/packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
@@ -18,7 +18,7 @@ import { makeTagProps } from '@/composables/tag'
 
 // Utilities
 import { computed, toRef } from 'vue'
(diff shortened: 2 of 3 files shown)


ericcai0814:
Hey, thanks for this PR! I have a few questions about the item-props prop.(Note: not about the width/maxWidth support added here)

1. Current behavior seems redundant for object items
Looking at the implementation:
```
{ ...(typeof item === 'string' ? { title: item } : item) }
{ ...(props.itemProps && isObject(raw) ? raw : {}) }
```

The first spread already passes all object properties through, making the itemProps-controlled second spread redundant.
Test case: This works without item-props:
```
<v-breadcrumbs
  :items="[{ title: 'Test', class: 'text-truncate', style: 'max-width: 120px' }]"
/>
<!-- max-width works without item-props -->
```

2. Example in this PR might be misleading
The example shows item-props being used with max-width, which might lead users to think it's required — but it actually works without it.

3. Missing documentation
I noticed item-props is new in v3.11.0 but currently has no description in the docs ("MISSING DESCRIPTION").

Could you help clarify the intended use case for item-props? I might be missing something, but I'm having trouble understanding what it's supposed to do that isn't already happening by default.
Thanks! 🙏

J-Sek:
Good point. We will drop the redundant `item-props` in the next release.

PR #22215, opened 2025-10-17
feat(theme): add layers option

Niki2k1:
## Description
I was trying to get Tailwind working with Vuetify and enabled layers. I had some issues with the theme overwriting my tailwind colors and such. So I created an option which lets me enable a layer for the generated theme.

Usage:
```ts
theme: {
  unimportant: true,
  layer: true, // or specify a custom name: 'my-custom-layer'
},
```

which then generates:
```css
@layer vuetify-theme { 
 :root {
      --v-theme-background: 255,255,255;
      --v-theme-background-overlay-multiplier: 1;
      --v-theme-surface: 255,255,255;
      --v-theme-surface-overlay-multiplier: 1;
      /* ... */
```

### DIFF
--- a/packages/vuetify/src/composables/__tests__/theme.spec.ts
+++ b/packages/vuetify/src/composables/__tests__/theme.spec.ts
@@ -326,8 +326,8 @@ describe('createTheme', () => {
     expect(css).not.toContain('.border-primary')
   })
 
-  it('should generate layer with default name if layer option is provided', async () => {
-    const theme = createTheme({ layer: true })
+  it('should generate layers', async () => {
+    const theme = createTheme({ layers: true })
 
     theme.install(app)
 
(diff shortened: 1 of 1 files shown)


KaelWD:
This should use the same layer names as v4 for forwards compatibility: https://github.com/vuetifyjs/vuetify/blob/a9a1aa2d1215a63305933ea93b4cee0ba4aa3819/packages/vuetify/src/composables/theme.ts#L422-L432

Niki2k1:
So I would remove the option for the custom name and set the layer name to `vuetify.theme`?
I believe foreground and background layer will be a v4 Feature right?

Should it still be opt-in via the `layer: true` option?

J-Sek:
> _I believe foreground and background layer will be a v4 Feature right?_

**Edit**: sorry, linked `master` previously, this is `next`:

https://github.com/vuetifyjs/vuetify/blob/a9a1aa2d1215a63305933ea93b4cee0ba4aa3819/packages/vuetify/src/styles/generic/_layers.scss#L11

Niki2k1:
I updated the pr to align with the v4 version. If the layer option is set to true (is now a boolean only) the layer will get generated (with background and foreground like in v4) as well as the `vuetify.theme`

So this is basically a backport now.

Issue #21515, opened 2025-06-03
[Feature Request] High contrast support


Jesse205:
### Problem to solve
The high contrast ratio can make it clearer for the user.

### Proposed solution
When High Contrast is on:
All buttons and list items in the elevated, flat, and tonal variants are automatically changed to outlined.
If activated, the color will be changed to system colors automatically.
The card component uses the outlined variant anyway.

### Links: 
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors
https://developer.mozilla.org/en-US/docs/Web/CSS/system-color
https://support.microsoft.com/en-us/microsoft-edge/accessibility-features-in-microsoft-edge-4c696192-338e-9465-b2cd-bd9b698ad19a

06b:
I've fixed a few high contrast issues in the past, so if you want to assign this to me, I wouldn't mind tackling this when things are slow for me.

J-Sek:
- [x] VProgressLinear
- [x] VSlider
- [x] VSwitch
- [x] VCard
- [x] VBtn
- [x] VAlert
- [x] VList
- [x] VChip
- [x] VIconBtn
- [x] VBadge
- [x] VSnackbar
- [x] VTimeline
- [x] VTabs
- [x] VBtnToggle
- [x] VSkeletonLoader

06b:
- [x] VIconBtn
- [x] VSkeletonLoader
- [x] VTabs (#22271)
- [x] VBtnToggle (#22279)

J-Sek:
@06b that's a mountain of work done and a great job overall.
Of course, we probably still have can improve things here and there, but do you think the scope satisfies the claim "_Vuetify supports high-contrast mode_" or do we miss something?

06b:
I agree that there are some things that could be improved here and there but it looks like we did miss something.

- [x] VTimePicker

J-Sek:
VTimePicker will be fixed in #21601

06b:
Once VColorPicker (#22317) is pulled in I think it will be safe to be considered done.

@J-Sek it's been fun
[...]

PR #22216, opened 2025-10-17
fix(VSkeletonLoader): render background in forced-colors mode

06b:
## Description

This addresses the request for High Contrast Support https://github.com/vuetifyjs/vuetify/issues/21515 for the VSkeletonLoader component when in forced-colors mode (e.g. Windows Accessibility > Contrast themes) by rendering the canvastext background on the skeleton-loader when system colors is active as shown below.

_Vuetify-skeleton

**Note:** the cursor will change to progress to give feedback that something is still loading should they hover over the skeleton-loader bone while in forced-colors mode.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-row>
        <v-col cols="12" md="6">
          <v-skeleton-loader
            class="mx-auto"
            max-width="300"
            type="card-avatar, actions"
          />
        </v-col>

        <v-col cols="12" md="6">
          <v-skeleton-loader
            class="mx-auto"
            max-width="300"
            type="image, article"
          />
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VSkeletonLoader/VSkeletonLoader.sass
+++ b/packages/vuetify/src/components/VSkeletonLoader/VSkeletonLoader.sass
@@ -60,6 +60,7 @@
       @media (forced-colors: active)
         @media (pointer: fine)
           cursor: progress
+
     &__avatar
       border-radius: 50%
       flex: 0 1 auto


packages/vuetify/src/components/VSkeletonLoader/VSkeletonLoader.sass
```diff
@@ -54,6 +54,12 @@
         z-index: 1
         @include tools.absolute(true)
 
+      @media (forced-colors: active)
+        border: thin solid
```
J-Sek: Border affects size in some examples.
What do you think about the `outline` instead?

```css
outline-offset: -4px;
outline: thin solid;
```

06b: Good catch, after some thought - I decided it would probably make more sense to render the background instead.

PR #22225, opened 2025-10-21
docs: create new homepage

ikushum:
## Description
The new homepage focuses more on Vuetify ecosystem and tooling

## Todo:
- [x] Interactive section for components and themes 
- [x] Delete unused components from old homepage
- [x] Add missing links where necessary
- [x] Update copy where necessary

### DIFF
--- a/packages/docs/src/components/app/bar/Bar.vue
+++ b/packages/docs/src/components/app/bar/Bar.vue
@@ -18,7 +18,7 @@
     </template>
 
     <template #append>
-      <div v-if="mdAndUp" class="d-flex ga-1 pe-2">
+      <div v-if="showExtraLinks" class="d-flex ga-1 pe-2">
         <AppBarBlogLink />
 
         <AppBarLearnMenu />
@@ -27,12 +27,12 @@
 
         <AppBarEcosystemMenu />
 
-        <AppBarPlaygroundLink v-if="lgAndUp" />
+        <AppBarPlaygroundLink v-if="showPlaygroundLink" />
 
         <AppBarOneLink class="ml-4" />
       </div>
 
-      <AppVerticalDivider v-if="mdAndUp" />
+      <AppVerticalDivider v-if="showExtraLinks" />
 
       <div class="d-flex ga-1">
         <AppBarStoreLink v-if="smAndUp" />
@@ -49,8 +49,9 @@
 
 <script setup>
   const app = useAppStore()
-  const { smAndUp, lgAndUp, mdAndDown, width } = useDisplay()
+  const { smAndUp, mdAndDown, width } = useDisplay()
   const route = useRoute()
 
-  const mdAndUp = computed(() => width.value >= 1077)
+  const showExtraLinks = computed(() => width.value >= 1052)
+  const showPlaygroundLink = computed(() => width.value >= 1220)
 </script>


johnleider:
This looks amazing. Here are a few things that stand out to me:

- [x] All <script> tags should use lang="ts"
- [x] Update this to allow the user to click and select the package manager like we have on https://vuetify.new
image

- [x] The special sponsor should be between the first and section section
- [x]  Component gallery should have some interactive examples that show what they look like as opposed to just their code
- [x] Need to have an override for the "Open in Vuetify Bin" so we can instead say "Open in Playground"
image

- [x] The Vuetify tooling section should be broken up into alternating information sections like you see on https://developer.mozilla.org/en-US/plus
- [x] Vuetify One section should have a CTA button that says "Experience Vuetify One" that opens up the menu on click. In the future, we want a dedicated page for this feature that we can send the user to
- [x] Discord section looks great, I'd remove Discord icon from the button and move it above the header and add an "open-in-new" icon at the end
image

- [x]  I like the look of the blog section, considering if it would look better to change the layout and have the latest bigger than the rest like in this https://snips.vuetifyjs.com/marketing/page-sections/blog-sections#snips-blog-section-05
- [x] Need to add this section back in some capacity and expand it to talk about all of the stuff we do, also that we are partnered with EpicMax for x/y/z and we are partnered with HeroDevs for x/y/z
image

packages/docs/src/components/app/Markup.vue
```diff
@@ -53,18 +53,18 @@
...
               v-bind="activatorProps"
               variant="text"
-              @click="bin"
+              @click="openCode"
             />
           </v-fade-transition>
         </template>
 
-        <span>{{ t('open-in-vuetify-bin') }}</span>
+        <span>{{ openInPlayground ? t('open-in-playground') : t('open-in-vuetify-bin') }}</span>
```
Haviles04: small nitpick and preference 
```suggestion
        <span>{{ t(openInPlayground ? 'open-in-playground' : 'open-in-vuetify-bin') }}</span>
```

packages/docs/src/components/home/ComponentGallery.vue
```diff
@@ -0,0 +1,126 @@
...
+      </h4>
+
+      <v-responsive class="mx-auto mb-5" max-width="700">
+        <h6 class="text-h6 font-weight-regular text-medium-emphasis">
+          Vuetify provides a comprehensive collection of components that can be used to build your application.
+        </h6>
+      </v-responsive>
+
+      <div class="d-flex justify-center mb-10">
+        <v-chip-group
```
Haviles04: I think we should add the `mandatory` prop here so something is always selected, IMO it looks funny when nothing is selected

packages/docs/src/components/home/Store.vue
```diff
@@ -0,0 +1,105 @@
...
+              </a>
+            </v-col>
+          </v-row>
+        </template>
+      </v-data-iterator>
+
+      <v-btn
+        append-icon="mdi-open-in-new"
+        aria-label="See More Templates"
+        class="text-none"
```
Haviles04, line 75: I think we maybe want a little margin to space the button out from the store items?

MajesticPotatoe (commented):
Just a few comments:
- [ ] Default install command is for bun, should probably default it to npm
- [ ] Buttons "Get Started" have no functional link (assuming they are placeholders, but should be updated)
- [ ] Vuetify Tooling section feels a little long winded (little scroll fatigue). ESLint Plugin + ESLint Config could be merged into one (their titles are even the same aside of the "config/plugin"
- [ ] Vuetify Tooling - Would probably reorder these so Vuetify Create is first then ESLint / Figma following
- [ ] Vuetify One - Buttons to individual products don't work (should add links to them to open in new tabs)
- [ ] Support Services links not working

Other than that it looks nice and feels good!

ikushum:
> Just a few comments:
> 
> * [ ]  Default install command is for bun, should probably default it to npm
> * [ ]  Buttons "Get Started" have no functional link (assuming they are placeholders, but should be updated)
> * [ ]  Vuetify Tooling section feels a little long winded (little scroll fatigue). ESLint Plugin + ESLint Config could be merged into one (their titles are even the same aside of the "config/plugin"
> * [ ]  Vuetify Tooling - Would probably reorder these so Vuetify Create is first then ESLint / Figma following
> * [ ]  Vuetify One - Buttons to individual products don't work (should add links to them to open in new tabs)
> * [ ]  Support Services links not working
> 
> Other than that it looks nice and feels good!

@MajesticPotatoe 
> * [ ]  Default install command is for bun, should probably default it to npm

This is actually random, it can be any package managers


> * [ ]  Buttons "Get Started" have no functional link (assuming they are placeholders, but should be updated)

Yes, links and copy are just placeholders, will go through them again.

> * [ ]  Vuetify Tooling section feels a little long winded (little scroll fatigue). 

I agree, made some improvements. Would appreciate your feedback again

KaelWD:
All these strings should be localisable 
image

ikushum:
- [x] use proper image path where necessary (for example images in `Vuetify Support Services` section)

johnleider (approved):

packages/docs/src/components/doc/IconList.vue
```diff
@@ -23,14 +23,14 @@
     </template>
 
     <template #item="{ props, item }">
-      <v-list-item v-bind="props" :prepend-icon="'svg:' + item.path">
+      <v-list-item v-bind="props" :prepend-icon="'svg:' + item.raw.path">
```
KaelWD: @johnleider https://next.vuetifyjs.com/en/getting-started/upgrade-guide/#vselect-vcombobox-vautocomplete

You need to build before running checks.

PR #22257, opened 2025-10-27
fix(VRangeSlider): prevent thumb movement when disabled mid-interaction

### LINKED ISSUE #22248: [Bug Report][3.9.2] VRangeSlider thumb still movable if 'disabled' becomes 'true' before thumb is released
### Environment
**Vuetify Version:** 3.9.2
**Vue Version:** 3.5.13

### Steps to Reproduce
1. Start dragging one of the thumbs.  
2. Keep holding it (don’t release) and wait for the debounced fetch (and `:disabled="true"`) to trigger.  
3. While the fetch is in progress (`:disabled="true"`), observe that the thumb remains movable.

### Expected Behavior
The thumb should freeze when `disabled` becomes `true`.  
Not sure whether it should just unfreeze once `disabled` becomes `false` again (allowing the user to continue dragging), or if we should immediately finish the action on `:disabled="false"` (emit an `@end` event?) and require the user to click the thumb again to resume.

### Actual Behavior
The thumb remains movable if `disabled` becomes `true` before the thumb is released.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqFVMFunDAQ/ZUJh4RILEu2qlptd6P20lMr9dBbNgcHho1VYyPbkKxW/HsH2xi2qZST8fPze8/jMQ/nxOhy/a1t877DZJvsLDatYBbvDxJg169Y27pPNymVtIxL1AECOJ9BM3lEGIYJIl6r1VGjMauS67ITTEO/4vX+kHDzQ7GKy+MhAS4rtKgbLskO1lGS9jvJlRG8Qj3BQBqNqlCQjFs/JPPStmGvhN8VxSXKJaGXmLHYjtQLsOKGPQmsLiPOhK9dW1HKrQuw6pnokKgVPqlOllj9VOWf72jL53nPdJ7d+t+yjYiv6m69qDZNTal5a8Gg7XzRedMqbYFqjDUMUGvVwA1d1M1IB6jJ3HIlYQqSjkgGL4zbjDY3WHESv4Wzz9LTRVjeoOrsF49ostIyCqWR6sljcHy1sAf7zE0GTB8NTWjoGpTWBBWAUiDTv710Gixu4yqvIY1p4PoariaKc86pGuKUBjPvMm8OVLKlukwW/wu8pMpOiKjgA1wt6vG+6+CrOAGDG12LL4ou2MkdNoYI5ZT4Ar/orrjBNKV3oERPnP398gQBzlzoW+/j9Jk5yXJ2aabWmo8aOzR3jTjeju4wJGVj7JBtUxTFdIK3m2omjN9FvuNApTA2POf92HLpw6bIPhePXsMvR51AcSpLwttHQczYoPE4GUzpdmvf99TxyZAlH/KP+WaTjB93Rf4pyZxDdvGf8tjjX+8Mhis=)

hdJerry:
## Description
Fixes an issue where the `VRangeSlider` thumb could still move if `disabled` became `true` while the user was dragging it.

### Reproduction
1. Start dragging a thumb.
2. Programmatically set `disabled` to `true` before release.
3. Thumb continues to move despite being disabled.

### Fix
Added a check to stop tracking thumb movement when `disabled` state changes during interaction.

Closes #22248

```html
<template>
  <v-app>
    <v-container>
      {{ range }}
      <v-progress-circular v-if="isLoading" indeterminate />
      <v-range-slider
        v-model="range"
        :disabled="isLoading"
        :max="100"
        :min="0"
        :step="1"
        @end="log('end')"
        @start="log('start')"
        @update:model-value="debouncedMockFetch"
      />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  function log (v) {
    console.log(v)
  }

  function debounce (func, wait, immediate) {
    let timeout
    return function () {
      const context = this; const args = arguments
      clearTimeout(timeout)
      if (immediate && !timeout) func.apply(context, args)
      timeout = setTimeout(function () {
        timeout = null
        if (!immediate) func.apply(context, args)
      }, wait)
    }
  }
  function delay (time) {
    return new Promise(resolve => setTimeout(resolve, time))
  }
  async function mockFetch () {
    isLoading.value = true
    await delay(2000)
    isLoading.value = false
  }

  const range = ref([20, 80])
  const isLoading = ref(false)
  const debouncedMockFetch = debounce(mockFetch, 2000)
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VRangeSlider/VRangeSlider.tsx
+++ b/packages/vuetify/src/components/VRangeSlider/VRangeSlider.tsx
@@ -87,29 +87,30 @@ export const VRangeSlider = genericComponent<VSliderSlots>()({
       onSliderTouchstart,
       position,
       trackContainerRef,
+      disabled,
       readonly,
     } = useSlider({
       props,
       steps,
       onSliderStart: () => {
-        if (props.disabled || props.readonly) {
+        if (disabled.value || readonly.value) {
           activeThumbRef.value?.blur()
           return
         }
         emit('start', model.value)
       },
       onSliderEnd: ({ value }) => {
-        if (props.disabled || props.readonly) {
+        if (disabled.value || readonly.value) {
           activeThumbRef.value?.blur()
-          return
-        }
+        } else {
+          const newValue: [number, number] =
+            activeThumbRef.value === startThumbRef.value?.$el
+              ? [value, model.value[1]]
+              : [model.value[0], value]
 
-        const newValue: [number, number] = activeThumbRef.value === startThumbRef.value?.$el
-          ? [value, model.value[1]]
-          : [model.value[0], value]
-
-        if (!props.strict && newValue[0] < newValue[1]) {
-          model.value = newValue
+          if (!props.strict && newValue[0] < newValue[1]) {
+            model.value = newValue
+          }
         }
 
         emit('end', model.value)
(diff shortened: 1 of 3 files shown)


hdJerry:
> * This should apply to VSlider too
> * `end` is not emitted if the mouse button is released while disabled
> * Keyboard focus is retained while disabled and arrow keys still work
> * Use `disabled.value`/`readonly.value` instead of props so it inherits from VForm

Thank you @KaelWD

KaelWD (approved):

PR #22303, opened 2025-11-07
fix: prevent TimeoutNaNWarning caused by invalid timeout in defer()

### LINKED ISSUE #22302: [Bug Report][3.10.9] Vuetify triggers `TimeoutNaNWarning` in newer Node versions due to `NaN` timeout in `defer()`
### Environment
**Vuetify Version:** 3.10.9
**Vue Version:** 3.5.24

### Steps to reproduce
1. `git clone git@github.com:mdoesburg/vuetify-test-app.git`
2. `cd vuetify-test-app`
3. `pnpm install`
4. `pnpm test`

### Expected Behavior
`defer()` should execute the provided callback immediately (or after the specified delay) without producing any warnings in Node.js or the browser. Invalid timeout values (such as `NaN`) and zero delays (`0`) should both be treated as immediate execution, consistent with browser behavior.

### Actual Behavior
When running Vuetify tests or applications on newer Node.js versions (e.g. 24.11.0), any component that internally calls `defer()` (e.g. tooltips, menus, hover transitions) can trigger the following warning:

```
(node:58301) TimeoutNaNWarning: NaN is not a number.
Timeout duration was set to 1.
(Use `node --trace-warnings ...` to show where the warning was created)
```

### Reproduction Link
[https://github.com/mdoesburg/vuetify-test-app](https://github.com/mdoesburg/vuetify-test-app)

### Images

Image

mdoesburg:
fixes #22302

## Description

This PR fixes a `TimeoutNaNWarning` that occurs in newer Node versions when `defer()` is called with an invalid timeout value (`NaN`). Newer Node versions emit this warning when `setTimeout()` is invoked with `NaN`, while browsers silently coerce it to `0`.

### DIFF
--- a/packages/vuetify/src/util/__tests__/helpers.spec.ts
+++ b/packages/vuetify/src/util/__tests__/helpers.spec.ts
@@ -259,6 +259,13 @@ describe('helpers', () => {
       expect(mockCallback).toHaveBeenCalledWith()
     })
 
+    it('executes callback immediately if timeout is NaN', () => {
+      const mockCallback = vi.fn()
+      defer(NaN, mockCallback)()
+
+      expect(mockCallback).toHaveBeenCalledWith()
+    })
+
     it('executes callback after specified timeout', () => {
       const mockCallback = vi.fn()
       defer(1000, mockCallback)


J-Sek:
Is it fixing the problem or just hiding it?
Is there a good reason to leave `defer` executions with `NaN`?

mdoesburg:
> Is it fixing the problem or just hiding it? Is there a good reason to leave `defer` executions with `NaN`?

It’s not hiding the issue per se. In browsers `setTimeout(NaN)` runs immediately with no warning, while newer Node versions log a warning but behave the same. This change makes `defer()` match browser behavior and keeps test output and CI logs clean. The root cause is in `useDelay`, but fixing it here keeps behavior consistent across environments, and guards against future use.

Would you prefer addressing it only in `useDelay`, or in both places for extra safety?

J-Sek:
I get the argument. Thanks for the additional explanation. Seems like the cross-environment consistency calls for more changes, but I don't think we are into it to the extent that would justify a setTimeout wrapper. If we had it, seems more appropriate to throw an error instead. Browsers (and probably execution environments in general) have an incentive to silently accept NaN. This warning is probably a small step to eventually stop doing it.

I will keep the idea shelved for now. Thank you for the input and interest in improving Vuetify 👍🏼

PR #22304, opened 2025-11-08
feat(VCombobox): create new items when pasting with line breaks

### LINKED ISSUE #10813: [Feature Request] VCombobox - accept new line as values separator when pasting
### Problem to solve
Currently items are added into comboboxes with enter/tab. And also with the `delimiters` prop.

The issue is, due to the limitation of `input`, when pasting content with newlines (like a column from excel or an html table), they are converted to spaces.

With this feature, pasting column data would automatically create items for each row.

### Proposed solution
Intercept the paste event.

```
<input type="combo" @paste="handlePaste" />
```

```
class ComboBox extends ... {
  handlePaste(event: ClipboardEvent) {
    if (event.clipboardData?.getData("text")?.includes?("\n")) {
      this.addItems(...event.clipboardData.getData("text").trim().split("\n").map(x => x.trim()));
      event.preventDefault();
    }
  }
}
```

ikushum:
fixes #10813

## Description
When pasting content with newlines (like a column from excel or an html table), they are converted to spaces.
With this feature, pasting column data would automatically create items for each row.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      {{ comboboxModel }}
      <v-combobox
        v-model="comboboxModel"
        :delimiters="[',']"
        :items="items"
        chips
        multiple
      />
    </v-container>
  </v-app>
</template>
<script setup>
  import { ref } from 'vue'

  const comboboxModel = ref([])
  const items = [
    'Item 1',
    'Item 2',
    'Item 3',
  ]

  /*
  try pasting
    Item 1,
    Item 2,
    Item 3,
  or
    Item 4, Item 5,
    Item 6, Item 7
  */
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VCombobox/VCombobox.tsx
+++ b/packages/vuetify/src/components/VCombobox/VCombobox.tsx
@@ -165,8 +165,7 @@ export const VCombobox = genericComponent<new <
         }
 
         if (val && props.multiple && props.delimiters?.length) {
-          const signsToMatch = props.delimiters.map(escapeForRegex).join('|')
-          const values = val.split(new RegExp(`(?:${signsToMatch})+`))
+          const values = splitByDelimiters(val)
           if (values.length > 1) {
             selectMultiple(values)
             _search.value = ''
@@ -360,7 +359,7 @@ export const VCombobox = genericComponent<new <
     }
     function onPaste (e: ClipboardEvent) {
       const clipboardText = e?.clipboardData?.getData('Text') ?? ''
-      const values = clipboardText.split('\n').filter(item => item.trim() !== '')
+      const values = splitByDelimiters(clipboardText)
 
       if (values.length > 1 && props.multiple) {
         e.preventDefault()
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VCombobox/VCombobox.tsx
```diff
@@ -168,13 +168,7 @@ export const VCombobox = genericComponent<new <
...
-              v = v.trim()
-              if (v) {
-                select(transformItem(props, v))
-                await nextTick()
-              }
-            }
+            selectMultiple(values)
             _search.value = ''
           }
         }
```
J-Sek, line 173: Currently when I paste

```
Item 1,
Item 2,
Item 3
```

...duplicated commas appear - despite `delimiters` prop explicitly set to split by comma.

We can fix it (and also drop `onPaste`), if we inject `'\n'` into delimiters in the setter.

```suggestion
        const effectiveDelimiters = ['\n', ...props.delimiters ?? []]
        if (val && props.multiple && effectiveDelimiters.length) {
          const signsToMatch = effectiveDelimiters.map(escapeForRegex).join('|')

          const values = val
            .split(new RegExp(`(?:${signsToMatch})+`))
            .filter(item => item.trim())

          if (values.length > 0) {
            selectMultiple(values)
            _search.value = ''
          }
        }
```

J-Sek (changes requested):

packages/vuetify/src/components/VCombobox/VCombobox.tsx
ikushum: @J-Sek The above would fail if we paste:
```
Item 1
Item 2
Item 3
```
This is because the search.value setter would never have `\n`. This is why onPaste handler is necessary. I've updated the onPaste handler to also cater for delimiters.

J-Sek (approved):

PR #22313, opened 2025-11-10
feat(VDefaultProvider): Provides Defaults to component slots

blalan05:
```vue
<template>
  <v-container>
    <v-defaults-provider :defaults="blueprint">
      <v-alert text="Lorem ipsum dolor sit amet consectetur adipisicing elit.">
        <template #close>
          <v-icon-btn icon="$close" />
        </template>
      </v-alert>
    </v-defaults-provider>
  </v-container>
</template>
<script setup lang="ts">
  const blueprint = {
    VAlert: {
      class: 'pa-0',
      '#text': {
        class: 'pl-4 py-4',
      },
      '#close': {
        VIconBtn: {
          rounded: 'lg',
          class: 'mx-0 my-2',
        },
      },
    },
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VAlert/VAlert.tsx
+++ b/packages/vuetify/src/components/VAlert/VAlert.tsx
@@ -10,7 +10,7 @@ import { VIcon } from '@/components/VIcon'
 // Composables
 import { useTextColor } from '@/composables/color'
 import { makeComponentProps } from '@/composables/component'
-import { useDefaults, useSlotDefaults } from '@/composables/defaults'
+import { useSlotDefaults } from '@/composables/defaults'
 import { makeDensityProps, useDensity } from '@/composables/density'
 import { makeDimensionProps, useDimension } from '@/composables/dimensions'
 import { makeElevationProps, useElevation } from '@/composables/elevation'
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VAlert/VAlert.tsx
```diff
@@ -105,7 +106,9 @@ export const VAlert = genericComponent<VAlertSlots>()({
     'update:modelValue': (value: boolean) => true,
   },
 
-  setup (props, { emit, slots }) {
+  setup (_props, { emit, slots }) {
+    const props = useDefaults(_props)
```
KaelWD: `props` is already the result from `internalUseDefaults` here: https://github.com/vuetifyjs/vuetify/blob/23dfbb43323bc2766f3f7c0e48f0e63010abd85a/packages/vuetify/src/util/defineComponent.tsx#L119-L132

packages/vuetify/src/components/VAlert/VAlert.tsx
```diff
@@ -138,6 +141,29 @@ export const VAlert = genericComponent<VAlertSlots>()({
       },
     }))
 
+    // Helper function to wrap slot content with defaults
+    function wrapSlot (slotName: string, slotFn: (() => any) | undefined, fallbackContent?: any) {
```
KaelWD: The basic version of this already exists https://github.com/vuetifyjs/vuetify/blob/a507171d25b0d41a4be0d74095ac69afbc413922/packages/vuetify/src/util/helpers.ts#L719-L724

packages/vuetify/src/components/VAlert/VAlert.tsx
```diff
@@ -138,6 +141,29 @@ export const VAlert = genericComponent<VAlertSlots>()({
...
+
+      if (!slotDefaultsInfo) {
+        return slotFn?.() ?? fallbackContent
+      }
+
+      const { componentDefaults, directProps } = slotDefaultsInfo
+
+      return (
+        <VDefaultsProvider defaults={ componentDefaults as any }>
+          <div { ...directProps }>
```
KaelWD: Not every slot should create its own element, this probably breaks a lot of styles. 

```
'#text': {
  class: 'pl-4 py-4',
},
```

should either only work only for specific slots or not be supported at all.

packages/vuetify/src/composables/defaults.ts
```diff
@@ -8,9 +8,17 @@ import { injectSelf } from '@/util/injectSelf'
 import type { ComputedRef, InjectionKey, Ref, VNode } from 'vue'
 import type { MaybeRef } from '@/util'
 
+export type SlotDefaults = {
+  [slotName: string]: Record<string, unknown>
+}
+
+export type ComponentDefaults = Record<string, unknown> & {
+  [slotKey: `#${string}`]: Record<string, unknown>
+}
```
KaelWD: `string` includes `` `#${string}` `` and `unknown` includes `Record<string, unknown>` so this is the same as `ComponentDefaults = Record<string, unknown>`

packages/vuetify/src/components/VAlert/VAlert.tsx
blalan05: I understand the issue and agree, can we find an alternative ergonomic method for blueprints?
image
VS. 
image

PR #22320, opened 2025-11-12
docs: simplify nuxt installation

hymair:
## Description
Simplify nuxt installation docs.

### DIFF
--- a/packages/docs/src/pages/en/features/treeshaking.md
+++ b/packages/docs/src/pages/en/features/treeshaking.md
@@ -63,19 +63,18 @@ module.exports = {
 ```
 
 ::: tab Nuxt
-<p class="ma-4">Nuxt also uses the vite plugin but needs some extra configuration to load it in the correct order:</p>
+<p class="ma-4">Nuxt also uses the vite plugin:</p>
 
 ```js { resource="nuxt.config.js" }
 import vuetify from 'vite-plugin-vuetify'
 
 export default defineNuxtConfig({
-  modules: [
-    async (options, nuxt) => {
-      nuxt.hooks.hook('vite:extendConfig', config => config.plugins.push(
-        vuetify()
-      ))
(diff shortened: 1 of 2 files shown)


J-Sek:
👍🏼 confirmed as working
should we cover `treeshaking.md` as well?

hymair:
> 👍🏼 confirmed as working should we cover `treeshaking.md` as well?

@J-Sek Docs on treeshaking mention that the vuetify plugin is intentionally pushed at the end because it needs correct order. Is this still valid?

J-Sek:
I was referring to the fragment about Nuxt and snippent with `nuxt.hooks.hook('vite:extendConfig',`.

Could you verify(with `nuxt analyze`) if removing the hook in favor of `vite»plugins` affects the bundle composition or not?

hymair:
Yes, exactly that fragment, I was referring to the part above the code snippet.

What exactly should I check after running `nuxt analyze`? I tested on our fairly big Nuxt 4 app and everything seems to be working fine / the same with either implementation.

Maybe @KaelWD recalls why [he added it 3 years ago](https://github.com/vuetifyjs/vuetify/commit/77f345364e26976d8ff33c6aae5f97f28b401d9d#diff-bca8233d9f71271fe0a002ccb926b7f78cd8b4f6ddb582ccd786281c34615feaR81)

userquin:
What about using https://github.com/userquin/unvuetify-monorepo ? We've a lot of utils there for Vite and Nuxt (styles, auto-import, i18n).

KaelWD:
I think at the time `options.vite.plugins` was merged *before* the default nuxt config, so extendConfig was the only way to get it to run after vite-plugin-vue. If it doesn't throw an error on startup then that's no longer needed.

hymair:
Great, should be ready to merge then.

PR #22330, opened 2025-11-15
fix(VSelect): Use hidden select for native form submissions

### LINKED ISSUE #21097: [Bug Report][3.7.16] v-autocomplete, v-select accessibility problem,
### Environment
**Vuetify Version:** 3.7.16
**Vue Version:** 3.5.13
**Browsers:** Mozilla Firefox, Google Chrome, Microsoft Edge
**OS:** Windows

### Steps to reproduce
In our projects, we use the Laravel framework on the backend, while Vue and Vuetify handle component management on the frontend. On the site, users can select different values from a dropdown field. However, after selection, the field only presents the associated code (key) to screen readers, rather than the full textual description.

This poses an accessibility issue because the visually displayed value is not available to users relying on screen readers. As a result, users cannot verify which value they have actually selected.

The issue contradicts multiple principles of the WCAG (Web Content Accessibility Guidelines) 2.1 standard, particularly:
•	1.3.1 Info and Relationships – Information and relationships must be programmatically determinable.
•	4.1.2 Name, Role, Value – User interface components must have programmatically determinable names and states so that assistive technologies can interpret them correctly.

As a solution, a modification is needed to ensure that the full textual value is accessible to screen readers, rather than just the key or identifier.

### Expected Behavior
If you navigate on to the component, any try to read up the selected value of the component with screanreader, it should read up the selected text instead of the value.

### Actual Behavior
If you navigate on to the component, any try to read up the selected value of the component with screanreader, its read only the value.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqVUk1vAiEQ/SuTvdgmrt43atKY1kvjpSejHlgZlYpAgLU1m/3vhWXZj6Q26QWYN28ebwa2ZWL0Yfqi1ORWYJIlM4tXxYnFxU4AzG6pQY4H6wOAW3qVFPl8lzRoEvCMuSrj4HqPqA9SyyxHX2Cd5iBzI7zwGZLnOiY4yWv5j4G8Qm2YsShsemai8aLRFlqkMv9s7RkmThxTzgQGgGhGUormoFmONL+3xpGmPUN1q67Z2Lrr03Bps8BlUri6sjYN1S5p2I5vFBHA6C+qizLwJ3X7UFWzqSfHi6bDIU/jlF04yLnQe1c+AYDfSmoLFI+k4BbKIHZFe5bUZDF2CFFuYhkoLZWB+aLLuFF6bHLBO8ybc+2wI4SxhlxEq3E4xZ0SS56eO9mmpndNaMd5gnogGYzeuNSMktEY/Ht74H3UCsY/4brYdgj8p7xPXqHUJ9Yjr14ek9eYa2IuPfb69TF7STg7Si366ss/1b9gI/Wlp74ZsvftuepP2QXuz8THT6qxKDivl/0PXXsj6g==)

ikushum:
fixes #21097

## Description
This PR ensures
1. Screen readers read out selected item title instead of value
2. When performing native form submissions, the form data uses item value instead of title.

This is achieved by used a hidden `<select` and visible `<input`. The user and screen-readers interact with the visible `<input`. But when making native form submissions, the form data uses the hidden `<select` because it will contain the `name ` attribute

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-form ref="form">
        <v-select :items="selectItems" autocomplete="suppress" name="select" />

        <v-btn class="my-5" @click="refreshFormData()">Refresh form data</v-btn>
      </v-form>
      <pre>{{ formData }}</pre>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const form = ref()
  const formData = ref('')

  const selectItems = [
    { title: 'Item 1', value: 1 },
    { title: 'Item 2', value: 2 },
    { title: 'Item 3', value: 3 },
  ]

  function refreshFormData () {
    const fd = new FormData(form.value.$el)
    formData.value = JSON.stringify(fd.entries().toArray())
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VSelect/VSelect.tsx
+++ b/packages/vuetify/src/components/VSelect/VSelect.tsx
@@ -163,7 +163,6 @@ export const VSelect = genericComponent<new <
     const form = useForm(props)
     const autocomplete = useAutocomplete(props)
     const selectedValues = computed(() => model.value.map(selection => selection.value))
-    const displayValue = computed(() => model.value.map(v => v.props.title).join(', '))
     const isFocused = shallowRef(false)
 
     let keyboardLookupPrefix = ''
@@ -390,8 +389,8 @@ export const VSelect = genericComponent<new <
         <VTextField
           ref={ vTextFieldRef }
           { ...textFieldProps }
+          modelValue={ model.value.map(v => v.props.title).join(', ') }
           name={ undefined }
-          modelValue={ displayValue.value }
           onUpdate:modelValue={ onModelUpdate }
           v-model:focused={ isFocused.value }
           validationValue={ model.externalValue }


ikushum:
Related PR that targets this - https://github.com/vuetifyjs/vuetify/pull/22331

PR #22331, opened 2025-11-15
fix(select): fix screen reader reading out selected value twice

ikushum:
related to #21097

## Description
Screen readers were reading out the selected value in VSelect, VAutocomplete and VCombobox twice. This will ensure they are read only once

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-form ref="form">
        <v-select v-model="selected" :items="selectItems" name="select" />

        <v-autocomplete v-model="selected" :items="selectItems" name="autocomplete" />

        <v-combobox v-model="selected" :items="selectItems" name="combobox" />
      </v-form>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selectItems = [
    { title: 'Item 1', value: 1 },
    { title: 'Item 2', value: 2 },
    { title: 'Item 3', value: 3 },
  ]
  const selected = ref(selectItems[0])
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/VAutocomplete.tsx
@@ -603,6 +603,7 @@ export const VAutocomplete = genericComponent<new <
                   return (
                     <div
                       key={ item.value }
+                      aria-hidden="true"
                       class={[
                         'v-autocomplete__selection',
                         index === selectionIndex.value && [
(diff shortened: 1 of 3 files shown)


J-Sek:
I have some problems testing it. Orca does not seem to care, Windows VM has some bug with audio and I hoped to verify on a different machine, but headphones just died. I would appreciate a comparison recording.

ikushum:
Not able to reproduce this anymore so closing

PR #22336, opened 2025-11-16
fix(VCalendar): day button font size and no active state

### LINKED ISSUE #22335: [Bug Report][3.11.0-beta.0] v-calendar (month view) incorrectly adds visual selections
### Environment
**Vuetify Version:** 3.11.0-beta.0
**Vue Version:** 3.5.24
**OS:** Windows 10 (current)

### Steps to reproduce
1. Initialize a v-calendar component with type="month"
2. Bind v-model to a ref containing a single date string (e.g., '2025-11-16')
3. Click on multiple different dates sequentially (e.g., click Nov 24, then Nov 25, then Nov 26).

### Expected Behavior
The component should respect the single-string v-model binding. When a new day is clicked, the visual selection should move to that new day, and the previously selected day should be deselected.

This is the standard behavior for single selection, and it works correctly in type="week" view. The bug only appears in type="month" view.

### Actual Behavior
When a new day is clicked in type="month" view, the component adds the new day to the visual selection. The previously selected day is not deselected.
This leads to a visual bug where multiple days are highlighted simultaneously.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNptkU1ugzAQha8y8oZW4jdNuohIpB6gJyhduDAUJLAteyCKoty9tiGBRtnNvPfZnjf+ujCjy+RDqXgckO1ZTtirjhMeCwGQj5GWJ1/6ppTd3PjWNIgEDba/DR0Ktk3Tgt396QDvUFRcLyKAxtrCd4etvX0vK+yikXcDWohkxc8PhJCn544dTmprKd32XD+YdFbuwl4KatbOMU+WKZdsVvThbtEdNGd39bSUPFkty7am1K0iMEiD8mjbK6kJLiDFpxwEYRW69HCFWsseArvywHGlFIbgNgUcHPTyujg+7CwHm3Szi7Isyt4Di+TJ9Kp9j11D9hbv4s2WuSLL4jT6QeJxysKadwbDf389ad9/t5yaug==)

J-Sek:
aligned with v2.x
fixes #22335

**Note**: `v-model` has no visual representation

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-row>
      <v-col>
        <v-sheet height="400">
          <v-calendar
            v-model="selection"
            color="primary"
            type="month"
          />
        </v-sheet>
      </v-col>
    </v-row>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const selection = ref('2025-11-12')
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VCalendar/VCalendarWeekly.sass
+++ b/packages/vuetify/src/components/VCalendar/VCalendarWeekly.sass
@@ -109,9 +109,6 @@
     .v-icon-btn
       font-size: $calendar-weekly-day-label-font-size
 
-      &:focus > .v-icon-btn__overlay
-        opacity: calc(.25 * var(--v-theme-overlay-multiplier))
-
   .v-calendar-weekly__day-month
     position: absolute
     text-decoration: none


packages/vuetify/src/components/VCalendar/VCalendarWeekly.sass
```diff
@@ -106,9 +106,11 @@
     text-align: center
     margin: $calendar-weekly-day-label-margin
 
-    .v-btn
+    .v-icon-btn
       font-size: $calendar-weekly-day-label-font-size
-      text-transform: none
+
+      &:focus > .v-icon-btn__overlay
+        opacity: calc(.25 * var(--v-theme-overlay-multiplier))
```
KaelWD: I don't think this was an intentional design in v2, :focus-visible just didn't exist yet.

KaelWD:
> v-model has no visual representation

Yeah it doesn't really have selection like date pickers, the highlighted day is `now` and v-model just sets the visible range.

packages/vuetify/src/components/VCalendar/VCalendarWeekly.sass
J-Sek: Good point
leaving only font size

J-Sek:
> Yeah it doesn't really have selection like date pickers, the highlighted day is `now` and v-model just sets the visible range.

What would you say to change `now` to be outlined rather than filled? This would bring it closer to VDatePicker. We can later add `selectable` and it would not conflict with current day style.

KaelWD (approved):

PR #22347, opened 2025-11-19
fix(v-tooltip): prevent showing empty tooltip

### LINKED ISSUE #22345: [Bug Report][3.10.11] VTooltip with null or undefined value still show text on v-button (and empty tooltip with false value)
### Environment
**Vuetify Version:** 3.10.11
**Vue Version:**  3.5.24
**OS:** Windows 10 (current)

### Steps to reproduce
1. Create a v-btn with v-tooltip directive set to "undefined", "null" or "false"

### Expected Behavior
When using null or undefined value on v-tooltip directive on a v-btn component:

```
<v-btn 
  v-tooltip="null"
  text="Button with null tooltip"
/>
<v-btn 
  v-tooltip="undefined"
  text="Button with undefined tooltip"
/>
```

The tooltip is visible, similarly to setting v-tooltip to `true`.

Additionally, using false still display an empty tooltip:

```
<v-btn
  text="button with false tooltip"
  v-tooltip="false" 
/>
```

### Actual Behavior
We should expect the tooltip to not be displayed, similarly to:
```
<v-btn 
  text="Button without tooltip"
/>
```

And display no tooltip at all.

### Reproduction Link
[https://play.vuetifyjs.com/#eNqdkUsOgjAQhq/SzBqo+NiYSuI5xAWPEklq27RTNDHe3VJQ0YQNu/8x/dLMnB5gTUWPWied47AHhvyqRYE8yyUhrIsLrYMMplISi1ZyM0YhLFES5Hc85FA6RCXJrcULcbLmjZ+tCSolsNU5kC4etZ/99D6nX15ppm6WLp0QM+C+WsRsCmH5DDR0S6jK4QT5fsDo/zL7ZNg1o5MbeGsr02oklqPTGaOD9RU8I9gku2S9hV6kqyRNIQofjX6uOmTnF2CZnxo=](https://play.vuetifyjs.com/#eNqdkUsOgjAQhq/SzBqo+NiYSuI5xAWPEklq27RTNDHe3VJQ0YQNu/8x/dLMnB5gTUWPWied47AHhvyqRYE8yyUhrIsLrYMMplISi1ZyM0YhLFES5Hc85FA6RCXJrcULcbLmjZ+tCSolsNU5kC4etZ/99D6nX15ppm6WLp0QM+C+WsRsCmH5DDR0S6jK4QT5fsDo/zL7ZNg1o5MbeGsr02oklqPTGaOD9RU8I9gku2S9hV6kqyRNIQofjX6uOmTnF2CZnxo=)

Haviles04:
## Description
Fixes #22345 

Disables activator when `value === null` or `value === false`, undefined is intentional, but at least this will give users a way to not show the tooltip conditionally

```vue
<template>
  <v-app>
    <v-container>
      <v-btn text="button with undefined tooltip" v-tooltip />
      <br>
      <v-btn text="button with null tooltip" v-tooltip="nullTooltip" />
      <br>
      <v-btn text="button with false tooltip" v-tooltip="false" @click="handleClick" />
      <br>
      <v-btn text="button without tooltip" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'
  const nullTooltip = ref(null)

  function handleClick () {
    nullTooltip.value = 'Changed'
  }
</script>

```

### DIFF
--- a/packages/vuetify/src/directives/tooltip/index.ts
+++ b/packages/vuetify/src/directives/tooltip/index.ts
@@ -4,6 +4,9 @@ import { VTooltip } from '@/components/VTooltip'
 // Composables
 import { useDirectiveComponent } from '@/composables/directiveComponent'
 
+// Utilities
+import { isObject } from '@/util'
+
 // Types
 import type { DirectiveBinding } from 'vue'
 import type { Anchor } from '@/util'


packages/vuetify/src/directives/tooltip/index.ts
```diff
@@ -15,7 +15,7 @@ export interface TooltipDirectiveBinding extends Omit<DirectiveBinding<string>,
 
 export const Tooltip = useDirectiveComponent<TooltipDirectiveBinding>(VTooltip, binding => {
   return {
-    activator: 'parent',
+    activator: (binding.value === null || binding.value === false) ? null : 'parent',
```
userquin: cannot be simplified to this? `binding.value ? 'parent' : null`

Haviles04: This was intentional, the default behaviors is that if you don't pass a value the tooltip will use whatever value the component has so if you have a button that says 'hello' the tooltip will also say 'hello' 

I wanted to leave a way for that behavior to still work. But if you intentionally did not want to show it you could pass null or false

userquin: add a comment or maybe change it to (anyone just checking the code can change it):
- `(binding.value === undefined || binding.value === true) ? 'parent': null` or 
- `(typeof binding.value === 'undefined' || binding.value === true) ? 'parent': null`

userquin: arggg: add a comment ;)

J-Sek: We can get `undefined`, `null`, `boolean`, `string` or `object`
I think we don't want to show the tooltip for `false`, `null` and the empty string.

J-Sek: Anyone against?

```ts
const textOrValue = isObject(binding.value) ? binding.value.text : binding.value
const disabled = ['', false, null].includes(textOrValue) // intentionally excluding undefined
return {
  activator: disabled ? null : 'parent',
  //...
}
```

J-Sek: Updated to avoid showing `[Object object]` when there is no `text` field (e.g. `v-tooltip='{}'`).

```ts
const disabled = isObject(binding.value)
  ? !binding.value.text
  : ['', false, null].includes(binding.value) // undefined means true
return {
  activator: disabled ? null : 'parent',
  //...
}
```

flaviusbadescu:
Hi @Haviles04 @J-Sek,

This behavior needs reconsideration. While I understand the original intent from issue #22345, the current implementation creates inconsistencies in real-world usage. For instance, when applying the tooltip directive to v-tab components where only some tabs have tooltips, the directive incorrectly falls back to displaying the tab's text content for tabs without explicit tooltip values. This creates an inconsistent user experience where some tabs show meaningful tooltips while others display redundant text that's already visible in the tab itself. The undefined value should be treated as 'no tooltip specified' and behave identically to omitting the directive entirely, rather than triggering fallback behavior.

```
<template>
  <v-tabs>
    <v-tab 
      v-for="tab in tabs" 
      :key="tab.id"
      v-tooltip:top="tab.tooltip"
    >
      {{ tab.name }}
    </v-tab>
  </v-tabs>
</template>

<script setup>
const tabs = [
  { id: 1, name: 'Dashboard', tooltip: 'View your main dashboard' },
  { id: 2, name: 'Reports', tooltip: undefined }, // No tooltip intended
  { id: 3, name: 'Analytics', tooltip: 'View detailed analytics' },
  { id: 4, name: 'Settings', tooltip: undefined }, // No tooltip intended
]
</script>
```

J-Sek:
Can you switch to `false` or empty string instead of `undefined` there?

PR #22358, opened 2025-11-21
feat(VBreadcrumbs): improve accessibility and add menu when collapsed

jdflaugergues:
## Description
This PR improves the `VBreadcrumbs` component with several accessibility-focused and Material Design–aligned enhancements.
- semantic elements (nav, ol, li) for improved screen-reader support and better compliance with accessibility standards
- automatically wrap to a new line when the breadcrumb trail becomes too long, preventing overflow
- new `total-visible` prop limits how many items can be displayed before collapsing items into an ellipsis
- optional `collapse-in-menu` for displaying the hidden items inside an accessible dropdown menu
- `menu-props`, `list-props`, and `#list-item` slot - for deep customization

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <h5>show only 2 links</h5>
      <v-breadcrumbs :items="items" :total-visible="2" />
      <h5>responsive</h5>
      <v-breadcrumbs :items="items" :total-visible="$vuetify.display.mdAndDown ? 2 : 7" collapse-from="1" />
      <h5>menu (default)</h5>
      <v-breadcrumbs :items="items" :total-visible="3" collapse-in-menu />
      <h5>menu (customized)</h5>
      <v-breadcrumbs
        :items="items"
        :list-props="{ class: 'pa-1', nav: true, rounded: 'lg', bgColor: 'primary' }"
        :menu-props="{ location: 'left top' }"
        :total-visible="5"
        collapse-in-menu
      >
        <template #list-item="{ item }">
          <v-list-item
            v-bind="item"
            :prepend-icon="item.disabled ? 'mdi-lock' : 'mdi-link'"
            class="mt-0"
            prepend-gap="12"
          />
        </template>
      </v-breadcrumbs>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    data: () => ({
      items: Array.from({ length: 9 }, (k, v) => ({
        title: `Link ${v + 1}`,
        href: `breadcrumbs_link_${v + 1}`,
        disabled: v === 2,
        prependIcon: v === 2 ? 'mdi-lock' : 'mdi-link',
      })),
    }),
  }
</script>
```


packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -48,11 +52,15 @@ export const makeVBreadcrumbsProps = propsFactory({
     default: () => ([]),
   },
   itemProps: Boolean,
+  maxItems: {
+    type: Number,
+    default: 8,
+  },
```
J-Sek:
```suggestion
  maxItems: Number,
```

`8` feels too arbitrary. Better to have it unrestricted than surprise a user.

packages/api-generator/src/locale/en/VBreadcrumbs.json
```diff
@@ -1,10 +1,12 @@
 {
   "props": {
+    "collapseInMenu": "When true, overflowing breadcrumb items are collapsed into a contextual menu rather than a static ellipsis.",
     "divider": "Specifies the dividing character between items.",
     "icons": "Specifies that the dividers between items are [v-icon](/components/icons)s.",
     "justifyCenter": "Align the breadcrumbs center.",
     "justifyEnd": "Align the breadcrumbs at the end.",
-    "large": "Increase the font-size of the breadcrumb item text to 16px (14px default)."
+    "large": "Increase the font-size of the breadcrumb item text to 16px (14px default).",
+    "maxItems": "Determines how many breadcrumb items can be shown before middle items are collapsed."
```
J-Sek: `totalVisible` might fit the framework better. We have it in VPagination and it is planned for VSnackbarQueue as well.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -157,7 +174,80 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
...
+              <VBreadcrumbsItem
+                disabled={ false }
+                onClick={ onClickEllipsis }
+              >
+                { props.collapseInMenu ? (
+                <VMenu>
+                  {{
+                    activator: ({ props: activatorProps }) => (
+                      <VBtn
+                        icon="mdi-dots-horizontal"
```
J-Sek: VPagination has `ellipsis` prop that defaults to the string value `"..."`. It might be easier to align and  style consistently.

An idea:
```tsx
<VBreadcrumbsItem
  tabindex="0"
  onClick={ props.collapseInMenu ? noop : onClickEllipsis }
>
    { props.ellipsis }
    { props.collapseInMenu ? (
			<VMenu activator="parent">
         // ...      
      <VMenu>
  )}
</VBreadcrumbsItem>
```

This piece could be extracted into separate render function to keep the main render function lean (similar to VNumberInput inner controls).

Not sure if `onClick` is enough... The item is not a button, so it might require additional `onKeydown` to handle Space and Enter keys.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -91,24 +99,33 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
     const items = computed(() => props.items.map(item => {
       return typeof item === 'string' ? { item: { title: item }, raw: item } : { item, raw: item }
     }))
+    const showEllipsis = computed(() => items.value.length >= props.maxItems)
```
J-Sek:
```suggestion
    const showEllipsis = toRef(() => items.value.length >= props.maxItems)
```

We would prefer `toRef` over `computed` here.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -91,24 +99,33 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
...
-          ]}
-          style={[
-            backgroundColorStyles.value,
-            props.style,
-          ]}
+          aria-label="breadcrumbs"
         >
+          <ol
+            class={[
+              'v-breadcrumbs',
```
J-Sek: Is there a benefit to nesting `v-breadcrumbs` + all classes and styles and leaving `<nav>` almost empty? I don't think we have a precedence of this kind. Usually all those things go into the root element.

J-Sek (changes requested):

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -91,24 +99,33 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
     const items = computed(() => props.items.map(item => {
       return typeof item === 'string' ? { item: { title: item }, raw: item } : { item, raw: item }
     }))
+    const showEllipsis = computed(() => items.value.length >= props.maxItems)
+    const enableEllipsis = ref(showEllipsis.value)
```
J-Sek: This prevents the limit from being responsive. Let's say I resize the window and the limit changes, it won't be reflected until the component or page is fully refreshed.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
J-Sek: BTW. it should probably be `>` not `>=`, unless I misunderstand something

packages/api-generator/src/locale/en/VBreadcrumbs.json
jdflaugergues: I used this property name because it's the one used in MUI.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: I agree. I had set it to 8 because it's the arbitrary number used by Material UI, but I'm fine removing it.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: agree

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: agree. I added a watcher on showEllipsis to handle the responsive behavior.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: agree

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: ok. I added ellipsis props like VPagination
onClick is enough. enter and space keys works fine

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
J-Sek: If we leave it like this, it will only support on/off change while someone might want to change the limit based on the available width (e.g. `:total-visible='$vuetify.display.smAndUp ? 4 : 2').

BTW. it might be a personal preference, but I feel like those variables names should swap:
- `showEllipsis` » `ellipsisEnabled`
- `enableEllipsis` » `showEllipsis` or `hasEllipsis`

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -157,7 +177,68 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
...
+                class="v-breadcrumbs-item--ellipsis"
+              >
+                { props.ellipsis }
+                { props.collapseInMenu ? (
+                  <VMenu activator="parent">
+                    {{
+                      default: () => (
+                        <VList>
+                          { items.value.slice(1, items.value.length - 1).map(({ item }, index) => (
+                            <VListItem key={ index } value={ index } component="a" href={ 'href' in item ? item.href : undefined }>
```
J-Sek: Not sure if `component="a"` does anything here as `href` should be enough to make it into `<a>`.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -157,7 +177,68 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
...
+                tabindex="0"
+                onClick={ props.collapseInMenu ? noop : onClickEllipsis }
+                class="v-breadcrumbs-item--ellipsis"
+              >
+                { props.ellipsis }
+                { props.collapseInMenu ? (
+                  <VMenu activator="parent">
+                    {{
+                      default: () => (
+                        <VList>
```
J-Sek: - Would we benefit from `menu-props` and `list-props` pass-through (like in VSelect)?
- `#list-item` slot could be a reasonable addition

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: Yes, I'll renamed it

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: Right, `component="a"` is useless. I remove it

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: good idea. I add it

J-Sek:
It is coming up nicely.
- would it make sense to collapse only items that exceed the limit? Similar to VPagination. User could have more control and even set the value to `1` and have only the last item visible. I know you attempt to follow MUI, but I can almost foresee we get a feature request for this kind of enhancement in a few months.
- API should probably expose a simple method to collapse expanded items (both inline and menu) programatically
- it would be great to include some ARIA attributes and support screenreaders from day one, if possible

jdflaugergues:
> It is coming up nicely.
> 
> * would it make sense to collapse only items that exceed the limit? Similar to VPagination. User could have more control and even set the value to `1` and have only the last item visible. I know you attempt to follow MUI, but I can almost foresee we get a feature request for this kind of enhancement in a few months.
> * API should probably expose a simple method to collapse expanded items (both inline and menu) programatically
> * it would be great to include some ARIA attributes and support screenreaders from day one, if possible

@J-Sek 
* What would be the rule for choosing which breadcrumb items to collapse?
  - The first items? (e.g., for totalVisible = 3 and **Link1/Link2/Link3/Link4/Link5/Link6** → **…/Link4/Link5/Link6**)
  - The first items starting from the second one? (e.g., for totalVisible = 3 and **Link1/Link2/Link3/Link4/Link5/Link6** → **Link1/…/Link5/Link6**)
  - The last items? (e.g., for totalVisible = 3 and **Link1/Link2/Link3/Link4/Link5/Link6** → **Link1/Link2/…/Link6**)
*  Which ARIA attributes would you like to add to the component? We already have `aria-label="breadcrumbs"` on the `<nav>`, `aria-current="page"` on the active item, and `aria-hidden` on the dividers. Maybe add an `aria-label` on each breadcrumb item?

J-Sek:
> What would be the rule for choosing which breadcrumb items to collapse?

Good question. We could have `:collapse-from="{index}"` to support all of the scenarios.

> Which ARIA attributes would you like to add to the component?

Just something to support the new "ellipsis". I did not check it.. if it already says something reasonable upon keyboard focus, then we should be fine, otherwise I hope it only needs one `aria-label`.

packages/vuetify/src/components/VBreadcrumbs/__tests__/VBreadcrumbs.spec.browser.tsx
```diff
@@ -42,7 +42,7 @@ describe('VBreadcrumbs', () => {
 
   it('should use color', () => {
     render(() => (
-      <VBreadcrumbs items={['hello', 'world']} color="primary"></VBreadcrumbs>
+      <VBreadcrumbs items={['hello', 'world']} color="primary" activeColor="primary"></VBreadcrumbs>
```
J-Sek, line 45: Is this change related to anything?

packages/vuetify/playgrounds/Playground.breadcrumbs.vue
```diff
@@ -0,0 +1,26 @@
+<template>
```
J-Sek: We can drop this playground file from PR. Public examples and the snippet in PR description are enough.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
```diff
@@ -109,6 +158,7 @@ export const VBreadcrumbs = genericComponent<new <T extends BreadcrumbItem>(
             props.style,
           ]}
         >
+          <ol role="list">
```
J-Sek: I am pretty sure it is redundant. Did you observe a difference anywhere?

packages/vuetify/src/components/VBreadcrumbs/__tests__/VBreadcrumbs.spec.browser.tsx
jdflaugergues: Yes, because I fixed a bug where the last item of the breadcrumbs wasn’t being marked as active.
Now that it is `active`, the color used for the last active item is not `color `but `activeColor`.

packages/vuetify/src/components/VBreadcrumbs/VBreadcrumbs.tsx
jdflaugergues: you're right. I remove it

packages/vuetify/playgrounds/Playground.breadcrumbs.vue
jdflaugergues: 👍

PR #22379, opened 2025-11-29
docs: add Figma link support to examples

juanjcardona13:
## Description

This PR adds support for displaying Figma design links in example components. Examples can now include an `example-meta` block with a Figma URL that will be displayed as an action button in the example toolbar.

The `exampleMeta` computed property was designed with extensibility in mind, allowing for future integrations with other tools such as Storybook.

## Changes

- **Example.vue**: Added `exampleMeta` computed property to parse meta data from examples and `figmaLink` computed property to extract Figma URLs
- **vuetify.ts**: Added Figma icon alias to the Vuetify plugin configuration
- **en.json**: Added `view-in-figma` i18n translation key
- **vite.config.mts**: Added support for `example-meta` block processing in the Vite build configuration
- **misc-readonly.vue**: Added example-meta block with Figma link as a demonstration

## Benefits

- Provides easy access to Figma designs directly from documentation examples
- Maintains consistency with existing example action buttons (Playground, GitHub)
- Follows the same pattern as other example metadata (playground-resources, playground-setup)

## Testing

- Verified that the Figma button appears when an example includes an `example-meta` block with a `figma` property
- Confirmed the button is hidden when no Figma link is provided
- Tested the button opens the Figma link in a new tab
- Verified i18n translation works correctly


packages/docs/src/components/examples/Example.vue
```diff
@@ -209,6 +222,10 @@
     )
   })
 
+  const figmaLink = computed(() => {
+    return exampleMeta.value.figma ?? null
```
johnleider: Does this specifically need to be null?

packages/docs/src/components/examples/Example.vue
johnleider: This should probably have the type <HTMLElement | undefined>

packages/docs/src/components/examples/Example.vue
```diff
@@ -196,6 +196,19 @@
     set: val => _theme.value = val,
   })
 
+  const exampleMeta = computed<Record<string, any>>(() => {
+    const meta = (component.value as any)?.exampleMeta
```
johnleider: This shouldn't need any

packages/docs/src/components/examples/Example.vue
```diff
@@ -107,6 +107,7 @@
 
   // Utilities
   import { getExample } from 'virtual:examples'
+  import type { Component } from 'vue'
```
johnleider, line 112:
```suggestion

  // Types
  import type { Component } from 'vue'
```

johnleider (dismissed):

packages/docs/src/components/examples/Example.vue
```diff
@@ -196,8 +205,21 @@
     set: val => _theme.value = val,
   })
 
+  const exampleMeta = computed<Record<string, any>>(() => {
```
johnleider, line 208: Does this need explicit types?

johnleider (changes requested):

PR #22390, opened 2025-12-02
feat(VDatePicker): add readonly prop

### LINKED ISSUE #22282: [Feature Request] VDatePicker Readonly Property
### Problem to solve
Currently VDatePicker is missing a readonly property. Adding one would allow for the DatePicker to still be navigable, while not allowing for inputs to be changed. The only option currently available is the disabled property, but that prevents changing the year or months.

### Proposed solution
Add a readonly property, similar to VTimePicker. This change should allow for calendar navigation, while not allowing the value(s) to change.

See: https://play.vuetifyjs.com/#eNrdk81OwzAQhF/F8qWXOH9VJahCERJ3JK5tD9t4aQyOY9lOEEK8O3ab1ongjsTN8+144x1lt5/Umjp70DodeqRrWjlstQSHm50ipBoYaH06nkTdKQdCoRmRh9aZTh03zwicPCn5cV9lI7paBsZ9Q6ZF/YbmQglZtx1HyQaQPd7t6HZR5uWKFQUrbxcJuaplHlXJisV+R2OPtpdOaBnuG1BHnNbACGDGv6vzz/IGZ/pZfW2b7p0Bf4UalWOtn62xv/guLSJpBEfWeDwdByUO4ESnfIt8ev/kdsJJjKyWYMO3NLBojYlls8giPxiS/Yj+UVg4SOT/Lng+DvbXwQc6//MDOS9GlU0WxktbG6Edsej6UA6GM/KCfiV0ma7SsqThUOTpDU1eQFpMZlt4Zvtvb/oNOw==

Haviles04:
## Description
fixes #22282

```vue
<template>
  <v-app>
    <v-container>
      <strong>Read Only?</strong>
      <v-date-picker
        :show-adjacent-months="true"
        aria-readonly="true"
        class="pa-"
        elevation="0"
        multiple="range"
        hide-header
        hide-title
        readonly
      />
      <br>
      <strong>Disabled</strong>
      <v-date-picker
        :model-value="['2025-11-29', '2025-11-30', '2025-12-1']"
        :show-adjacent-months="true"
        aria-readonly="true"
        class="pa-"
        elevation="0"
        multiple="range"
        disabled
        hide-header
        hide-title
      />
    </v-container>
  </v-app>
</template>

<script setup>

</script>

```

### DIFF
--- a/packages/vuetify/src/components/VDatePicker/VDatePickerControls.tsx
+++ b/packages/vuetify/src/components/VDatePicker/VDatePickerControls.tsx
@@ -52,10 +52,6 @@ export const makeVDatePickerControlsProps = propsFactory({
     type: [Boolean, String, Array] as PropType<boolean | string | string[] | null>,
     default: null,
   },
-  readonly: {
-    type: Boolean,
-    defauly: null,
-  },
   nextIcon: {
     type: IconValue,
     default: '$next',


J-Sek:
Should probably target `dev` branch

packages/vuetify/src/composables/calendar.ts
```diff
@@ -218,7 +223,7 @@ export function useCalendar (props: CalendarProps) {
   const { minDate, maxDate } = useCalendarRange(props)
 
   function isDisabled (value: unknown) {
-    if (props.disabled) return true
+    if (props.disabled || props.readonly) return true
```
ikushum, line 226: I think `props.readonly` should not be added to this condition.

We should probably add `pointer-events: none` to `VDatePickerMonth` if `readonly` instead of showing a grayed out calendar (which makes more sense for `disabled`).

What do you think?

packages/vuetify/src/components/VDatePicker/VDatePickerControls.tsx
```diff
@@ -52,6 +52,10 @@ export const makeVDatePickerControlsProps = propsFactory({
     type: [Boolean, String, Array] as PropType<boolean | string | string[] | null>,
     default: null,
   },
+  readonly: {
+    type: Boolean,
+    defauly: null,
+  },
```
ikushum: looks like this prop doesn't really impact `VDatePickerControls.tsx`, does it? If this is the case, it should be moved to `VDatePicker`

ikushum:
Alternative solution - https://github.com/vuetifyjs/vuetify/pull/22433

packages/vuetify/src/composables/calendar.ts
Haviles04: I think you're probably right

Haviles04:
Closed by better solution - https://github.com/vuetifyjs/vuetify/pull/22433

PR #22393, opened 2025-12-02
docs: New Sponsor Popup

Haviles04:
New sponsorship popup

### DIFF
--- a/packages/docs/src/components/sponsor/SnackbarPopup.vue
+++ b/packages/docs/src/components/sponsor/SnackbarPopup.vue
@@ -9,7 +9,7 @@
     <template #default>
       <div class="pl-4">
         <div class="d-flex flex-row justify-space-between w-100 mb-6">
-          <h5 class="text-headline-medium">Love Vuetify?</h5>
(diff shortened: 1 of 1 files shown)


KaelWD:
Popups are yucky 🤢
We already have these ad spots all over the place
image

J-Sek:
Minor UI improvements + reordered links.

I don't mind experimenting with new placement. Those "ads via vuetify" do not show up on Brave.

<details>
<summary>Preview</summary>

image

</details>

Haviles04:
The idea was to mimic the lil popup that coolify has

johnleider:
> Popups are yucky 🤢 We already have these ad spots all over the place image

Most of them are adblocked though.

KaelWD:
uBlock doesn't block them, only 5% of our traffic uses brave. What makes you think brave won't block this too?

KaelWD:
If you're going to add this anyway the dismiss button needs to be remembered, and it probably shouldn't appear until either navigation or focus loss.

KaelWD:
I've updated this to only show after route change if you've visited the site at least 5 times before, and clicking dismiss is remembered.

PR #22396, opened 2025-12-04
feat(styles): add separate entry points

### LINKED ISSUE #20100: [Feature Request] Split up the global style into required and opt in styling.
### Problem to solve
Current vuetify global css includes a lot of optional selectors which in most use cases aren't used at all. It would be nice to opt in for these functionalities and keep this dry (and reduce bundle size).

### Proposed solution
Split up the 'vuetify/styles' into smaller chunks like e.g.:
- 'helpers.css' (paddings, margins, overflow, flex, text helpers, around 279 kB)
- 'colors.css' (colors around 40 kB)
- 'ress.css'
- 'components.css' (this is one is a bit weird as the components already load its own style)

To be non breaking you can add the 'bare' essentials with someting like 
```js 
import 'vuetify/core'
```

And opt in for the colors or helpers with
```js 
import 'vuetify/helpers' 
import 'vuetify/colors' 
```

KaelWD:
closes #20100

### DIFF
--- a/packages/docs/src/pages/en/getting-started/upgrade-guide.md
+++ b/packages/docs/src/pages/en/getting-started/upgrade-guide.md
@@ -23,6 +23,7 @@ This page contains a detailed list of breaking changes and the steps required to
 - The CSS reset has been mostly removed, with style normalisation being moved to individual components instead.
   - `<button>`, `<input>`, `<select>` have their browser native borders and background colors.
   - `<ul>`, `<ol>` and headings have padding and margins.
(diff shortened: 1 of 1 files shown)


J-Sek:
So many years I was bumping into issues because of `@use 'vuetify' with (...)` used with `@use 'vuetify/settings' with (...)` in the same app. We could mention something like

> When using on `@use 'vuetify/settings' with (...)`, remove `@use 'vuetify'` to avoid rewriting variables. Instead add `import 'vuetify/styles';` in JavaScript or pass it `css: ['vuetify/styles'],` when using Nuxt.

KaelWD:
You can keep them both, just without `with` on one:

```css
@use 'vuetify/settings' with (...)
@use 'vuetify'
```

Not really relevant to this change though.

PR #22405, opened 2025-12-04
feat(styles): possibility to opt-out from misc styles

J-Sek:
## Description

- introduces `$misc` map to opt-out from utilities that are not covered by `$utilities`.

**Context**:
Imagine a project where you don't want to disable all `$utilities` (because you like `border`, `rounded` or want to keep some styles pre-generated for VRow), but you also want modern or standardized shadows. You can override them, but they are still added to the bundle.

```scss
// it feels wrong, I should only need elevation-0 here
.elevation-0,
.elevation-4,
.elevation-5,
.elevation-6,
.elevation-7,
.elevation-8,
.elevation-9,
.elevation-10,
.elevation-11,
.elevation-12 { box-shadow: none !important }

// standardized levels 0-3
.elevation-1 { ... }
.elevation-2 { ... }
.elevation-3 { ... }
```

related to #20100

## Markup:

```scss
@use 'vuetify/settings' with (
  $misc: (
    elevation: false,
    hidden: false,
    'sr-only': false,
    'pointer-events': false,
  ),
);
```

### DIFF
--- a/packages/docs/src/pages/en/features/sass-variables.md
+++ b/packages/docs/src/pages/en/features/sass-variables.md
@@ -127,7 +127,21 @@ Keep in mind that to obtain settings from Vuetify, you must forward its variable
 
 ## Disabling utility classes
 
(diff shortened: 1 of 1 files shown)


J-Sek:
Needs follow up:
- ~~split `VGrid.sass` to avoid loading VRow/VCol styles (over 9kB) when using only VContainer or VSpacer~~ [`f899803`](https://github.com/vuetifyjs/vuetify/commit/f899803)
- let devs opt out from `grid-offset`.. responsive styles? something to think about later

PR #22406, opened 2025-12-05
fix(VInput): prior prepend, append slots over props

### LINKED ISSUE #22332: [Bug Report][3.10.11] Date Input prepend inconsistencies
### Environment
**Vuetify Version:** 3.10.11
**Vue Version:** 3.6.0-alpha.4
**OS:** Linux undefined (current)

### Steps to reproduce
1. Take a default v-date-input and add an icon using the prepend template slot
2. observe the two icons next to each other

### Expected Behavior
When I use the prepend template slot I expect it to replace the calendar icon, not add another icon next to it

### Actual Behavior
The prepend slot shows to the left of the calendar icon, so I have both

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrNk8FOxCAQhl9lxEM12bbbaDyYXRNvvoEH6wHL1CWhhQCtMZt9dwfauq3Zi0YTb8zH8PPPwDztmbNVfm9M1nfIbtnGY2MU93hXtgCbPuXGxGUMKt16Llu0I4pQUHYqW9N56NNGC1TbkjXutWSg+EuMBNa8U55IfvrkBOGrxHFj0nqTfgfGosFWgCRH86SRp4EHCSHTiisi3B7T/sADcFpMwClNpX7T1KcncjW9AZyPR2eb0XS88qTcrDjKzBevOaB5yb/TCLrDv/+4Hf+jBQEuf3cgw+dfaFDoKiuNB4e+G2ZDNkZbD3uwWMMBaqsbSGiekpAOQLrOA3UQtiHjInlApTQ8aqvEWXIZLhgkSYwdVuwqu8nWKVdmx7NrFuJinRUFW9VcOVwtJnZgzx/j5Sm9)

### Other comments
The reason I am even doing this is because I am trying to streamline how tabbing through our form works and for v-date-inputs, I first focus the icon, then the textfield next to it. I don't need to focus the same component twice when both icon and textfield lead to the exact same input. Since just replacing the icon does not get rid of the onClick action of it, I decided to override it using the prepend slot. If I unset the prepend-icon to go along with it, I get the behavior I want, but it feels like prepend not overwriting the prepend-icon is not the intended behavior.

### Images
Image

SonTT19:
## Description

fixes #22332

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-input v-model="msg" label="default" />
      <v-date-input
        v-model="msg"
        label="with prepend icon"
        prepend-icon="mdi-calendar"
      />
      <v-date-input
        v-model="msg"
        label="with prepend icon and prepend slot"
        prepend-icon="mdi-calendar"
      >
        <template #prepend>
          <v-icon icon="mdi-calendar" />
        </template>
      </v-date-input>
      <v-date-input
        v-model="msg"
        label="with empty prepend icon and prepend slot"
        prepend-icon=""
      >
        <template #prepend>
          <v-icon icon="mdi-calendar" />
        </template>
      </v-date-input>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

</script>
```

### DIFF
--- a/packages/vuetify/src/components/VField/VField.tsx
+++ b/packages/vuetify/src/components/VField/VField.tsx
@@ -281,15 +281,16 @@ export const VField = genericComponent<new <T>(
 
           { hasPrepend && (
             <div key="prepend" class="v-field__prepend-inner">
-              { props.prependInnerIcon && (
-                <InputIcon
-                  key="prepend-icon"
-                  name="prependInner"
-                  color={ iconColor.value }
-                />
-              )}
-
-              { slots['prepend-inner']?.(slotProps.value) }
+              { slots['prepend-inner']
+                ? slots['prepend-inner'](slotProps.value)
+                : (props.prependInnerIcon && (
+                  <InputIcon
+                    key="prepend-icon"
+                    name="prependInner"
+                    color={ iconColor.value }
+                  />
+                ))
+              }
             </div>
           )}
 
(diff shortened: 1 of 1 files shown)


J-Sek:
Looks fine. This would be in line with v2.

Should probably cover `prepend-inner` and `append-inner` in `VField.tsx`

SonTT19:
> Looks fine. This would be in line with v2.
> 
> Should probably cover `prepend-inner` and `append-inner` in `VField.tsx`

Thanks, i've updated it

PR #21181, opened 2025-03-29
fix(VSelect): add aria-label a11y

jcjp:
## Description

Missing `aria-label` on `VList` inside `VSelect`. Used axe developer tool to check if now fixed.

Fixes: https://github.com/vuetifyjs/vuetify/issues/21164

## Markup:

```vue
<template>
  <v-app>
    <v-select
      label="Select"
      :menu="true"
      :items="['California', 'Colorado', 'Florida', 'Georgia', 'Texas', 'Wyoming']"
    />
  </v-app>
</template>
```

PR #22418, opened 2025-12-08
fix(fields): clean up aria attributes

KaelWD:
Slightly better than #20339 but still not great

- Removed the aria-label from select list checkboxes, orca and voiceover already read "checked" for selected items so this is redundant. NVDA only reads "not selected" but I guess users would be used to that?
- Also marked the checkboxes aria-hidden to prevent orca from reading item labels twice
- I can't get the field label to not be read twice for some reason. Even removing all the labels but the base one still does it in orca, and talkback needs aria-labelledby to be not aria-hidden or hidden by CSS.
- Removed aria-label and title from the field element, this was causing orca to read "open open" instead of the actual label.
- Replaced #21181 with aria-labelledby so "list" isn't read twice.

### DIFF
--- a/packages/vuetify/src/components/VAutocomplete/__tests__/VAutocomplete.spec.browser.tsx
+++ b/packages/vuetify/src/components/VAutocomplete/__tests__/VAutocomplete.spec.browser.tsx
@@ -416,7 +416,7 @@ describe('VAutocomplete', () => {
       const items = ['Item 1', 'Item 2', 'Item 3', 'Item 4']
       const selectedItems = ['Item 1', 'Item 2']
 
-      render(() => (
+      const { element } = render(() => (
         <VAutocomplete
           items={ items }
           modelValue={ selectedItems }
@@ -425,8 +425,7 @@ describe('VAutocomplete', () => {
         />
       ))
 
-      const menuIcon = screen.getByRole('button', { name: /open/i })
-      await userEvent.click(menuIcon)
+      await userEvent.click(element)
       await commands.waitStable('.v-list')
 
       const listItems = screen.getAllByRole('option')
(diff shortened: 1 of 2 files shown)


KaelWD:
### Orca

My label clickable (does nothing)
My label my label opens popup
Expanded
My label multi dash select list with 2 items hello
World not selected
World

### NVDA

Combobox collapsed clickable opens list (still doesn't read label)
My label combobox expanded editable opens list hello
My label list hello 1 of 2
World not selected 2 of 2
Selected

### Talkback

My label my label double tap to activate
Expanded hello editbox hello expanded actions available
Selected hello 1 of 2 in listbox my label 2 items double tap to activate
World 2 of 2 double tap to activate
World

J-Sek:
### JAWS

3.11.2

_Close_ edit combo expanded
Hello (To set the value use the Arrow keys or type the value.)
Hello (To move to an item press the Arrow keys.)
World (2 of 2)
selected

3.11.2-master.2025-12-08

**My label** edit combo collapsed
Hello (To set the value use the Arrow keys or type the value.)
expanded
**Checked** Hello (To move to an item press the Arrow keys.)
**Not Checked** World (2 of 2)
**Checked** World
selected
**Checked** World (2 of 2)

This PR

**My label** edit combo collapsed
Hello (To set the value use the Arrow keys or type the value.)
expanded
**My label ExtendedSelect ListBox**
Hello (To move to an item press the Arrow keys.)
**Not Selected** World (2 of 2)
selected

KaelWD:
Yeah I'm ok with that, jaws probably intentionally doesn't read selected for selected items just like nvda: https://github.com/nvaccess/nvda/issues/12893

J-Sek:
Thanks for taking care of it 🥇

PR #22446, opened 2025-12-16
fix(VInput): make all input icons unfocusable

### LINKED ISSUE #22333: [Feature Request] Add a way to disable the on-click event of the date input prepend icon (or make it not focusable)
### Problem to solve
With the date field specifically, clicking the icon will do the exact same thing as clicking into the field itself. This means that people navigating a form via keyboard will get the same field focused twice, once the icon, and once the field. The ability to disable the onClick action completely, or just force it to not be focused would allow for easier tab navigation, and it would allow for a clear difference between "user-set" on click actions that actually do something unique, and the date field

### Proposed solution
Add a prop along the way of "disable-prepend-icon-action" or "disable-prepend-icon-focus" so that it is possible to ensure the icon will not receive focus when it does not do anything other than what clicking into the field does. This also allows to keep the focus (or click action) for the use-cases when the user actually changes what the icon does on click.

The only way I found to do this right now is to set the prepend icon to an empty slot and add it back via #prepend template slot (so it is not clickable) and this seems a bit convulsed, especially when there are many date fields.

ikushum:
fixes #22333

## Description
Alternative 2 for https://github.com/vuetifyjs/vuetify/issues/22333
All input icons are unfocusable

## Markup:
```vue
<template>
  <v-app>
    <v-container>
      <v-date-input label="Prepend Icon cannot be focused" prepend-icon="$vuetify" />
      <v-text-field label="Prepend Icon cannot be focused" prepend-icon="$vuetify" @click:prepend="() => {}" />
      <v-select label="Prepend Icon cannot be focused" prepend-icon="$vuetify" @click:prepend="() => {}" />
    </v-container>
  </v-app>
</template>

```

### DIFF
--- a/packages/vuetify/src/components/VField/VField.tsx
+++ b/packages/vuetify/src/components/VField/VField.tsx
@@ -288,6 +288,7 @@ export const VField = genericComponent<new <T>(
                 ? slots['prepend-inner'](slotProps.value)
                 : (props.prependInnerIcon && (
                   <InputIcon
+                    tabindex={ -1 }
                     key="prepend-icon"
                     name="prependInner"
                     color={ iconColor.value }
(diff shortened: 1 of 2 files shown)


KaelWD:
I don't think doing it for all inputs is a great idea without any data on how people are actually using them, let's go with just #22445

PR #22463, opened 2025-12-19
fix(VDataTable): fix typing on the align attribute (replace center with middle)

AlexandreFournier:
## Description

This patch fixes the TypeScript typing of the `align` prop on the `VDataTableColumn` component in Vuetify. The previously allowed value `'center'` has been replaced with `'middle'` in the prop’s union type (`'start' | 'middle' | 'end'`), aligning the type definition with the actual, supported alignment values used by the component.

The change improves type safety and developer experience by preventing invalid or misleading alignment values in TypeScript, while ensuring consistency between the component’s API and its internal alignment behavior.

### DIFF
--- a/packages/vuetify/src/components/VDataTable/VDataTableColumn.tsx
+++ b/packages/vuetify/src/components/VDataTable/VDataTableColumn.tsx
@@ -6,7 +6,7 @@ import type { PropType } from 'vue'
 
 export const VDataTableColumn = defineFunctionalComponent({
   align: {
-    type: String as PropType<'start' | 'center' | 'end'>,
+    type: String as PropType<'start' | 'middle' | 'end'>,
     default: 'start',
   },
   fixed: {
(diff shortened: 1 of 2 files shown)


J-Sek:
Why?

https://github.com/vuetifyjs/vuetify/blob/3c3f667525ebb5be4a63956d53111bb2c181130f/packages/vuetify/src/components/VDataTable/VDataTable.sass#L38

AlexandreFournier:
@J-Sek My bad, you are right, the correct value should be "center" according to the SASS. I also checked the behavior and the SASS is correct. 

It just does not apply to checkboxes (i.e. [Simple checkboxes](https://vuetifyjs.com/en/components/data-tables/basics/#simple-checkbox)). For a strange reason, the checkboxes get centered when using the value "center" for the column...

Temporary fix ?

```
<v-data-table-server...>
    ...
    <template v-slot:item.some_boolean="{ item }" class="text-center">
      <v-checkbox-btn v-model="item.some_boolean" class="d-inline-flex" />
    </template>
    ...
</v-data-table-server>
```

KR

J-Sek:
If you have access to the checkbox `style="justify-self: center"` might help (and does not require anything extra on the `<td>`

PR #22469, opened 2025-12-24
fix(VOverlay): force real scrollbar visibility to avoid shift

J-Sek:
Noticable shift by ~1px:
- on docs [Dialogs](https://next.vuetifyjs.com/en/components/dialogs/#usage)
- with 150% scaling on the OS level

Also larger shift when using `<v-main scrollable>` (see example below)

Cause: `window.innerWidth - document.documentElement.offsetWidth` returns wrong value (verified on vanilla Chrome and Brave, on Linux) unless scaling is exactly 100%.

**Note**: `scrollbar-gutter` requires Safari 18.1 (we claim 16.4 is gonna be supported by v4)

```vue
<template>
  <v-app theme="dark">
    <v-main scrollable>
      <v-container>
        <v-dialog max-width="500" scroll-strategy="block">
            <v-btn
              v-bind="activatorProps"
              color="surface-variant"
              text="Open Dialog"
              variant="flat"
            ></v-btn>
          </template>

          <v-card title="Dialog">
            <v-card-text>
              Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
              eiusmod tempor incididunt ut labore et dolore magna aliqua.
            </v-card-text>
          </v-card>
        </v-dialog>
      </v-container>

      <v-img
        src="https://cdn.cosmicjs.com/68b46aa0-d4fe-11ee-9ce5-59949019255e-VUETIFY-ONE-3.png"
        width="200"
        height="300"
        class="position-absolute right-0"
      />

      <v-card width="200" height="200vh" />
    </v-main>
  </v-app>
</template>
```

### DIFF
--- a/packages/vuetify/src/components/VAppBar/VAppBar.sass
+++ b/packages/vuetify/src/components/VAppBar/VAppBar.sass
@@ -10,6 +10,3 @@
 
       &:not(.v-toolbar--flat)
         @include tools.elevation($app-bar-elevation)
-
-    &:not(.v-toolbar--absolute)
-      padding-inline-end: var(--v-scrollbar-offset)
--- a/packages/vuetify/src/components/VDialog/VDialog.sass
+++ b/packages/vuetify/src/components/VDialog/VDialog.sass
@@ -21,7 +21,6 @@
 
         > .v-card,
         > .v-sheet
-          --v-scrollbar-offset: 0px
           border-radius: $dialog-border-radius
           overflow-y: auto
           flex: 1 1 100%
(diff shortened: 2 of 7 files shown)


KaelWD:
I did actually try this first but the blank space when a scrim is up is pretty fugly.

J-Sek:
Maybe `overscroll-behavior: contain` could help instead?
It would be kinda hacky... let's say scrim would have a pseudo-element making it scrollable. If we hide scrim scrollbar, user wheel events won't reach the page.

Issue #21778, opened 2025-07-17
[Bug Report][3.9.1] VTreeview - items is no longer reactive


J-Sek:
### Environment
**Vuetify Version:** 3.9.1
**Vue Version:** 3.5.17
**Browsers:** Chrome 138.0.0.0
**OS:** Linux x86_64

### Steps to reproduce
- bind `load-children` prop to a method

(basically doing the same thing as this [example in the docs](https://vuetifyjs.com/en/components/treeview/#selectable-icons))

### Expected Behavior
method should be called (works in 3.9.0)

### Actual Behavior
method is not exectuted

### Reproduction Link

[Playground](https://play.vuetifyjs.com/#eNptU8GOnDAM/RUrl2EkYIat2qqIGWnbH+ihN+AQgdmNFAJKArRC/HudMNnZWe0J2+/Z8XsW5cqMbk7P45jOE7KcFRb7UXKL10oBFHPScN362Gd6WKBFZXb4VmwGCY3kxlwq1iadxL/ApXhRSYPKoq4YEMOB2RPFpqfoW8XeJvgZViPOApd7EWBO+qFFmRuU2FhsqS+EFXtPzAVt7R7w3w+YHHibNK9CthoVcVz+SHFdiRVWIsGK9/gJPHM5OVg89F6L0331uyVUJMXBNcrItpCNGq9BRQ7rCiFJJaoX+wrbVpwcyfvvJu0HKE7vLkOpabQYLXXbafRc0Y+DtrCCxg426PTQw4GOenB0oBMo4+j7Y3BxtKisj3fMmxeAdV9X0I5ZvMfOmRwOz1LCT40LaoHmcMOCvzmUtS9tbrILuPmnGugm1VgxKHDuQ3SEML+DyL+beoPLc52GUTc/jrSPnTTN2jv4woUFhQv8JoXCYKThcnU+/BE9DpONdAzZ+Xw+emn7+T4ZT0LLcLDV63yKg8RfjuMbD7DtpNqLcmfYjSfL2RazL+nXNPvOXPAjzVjccWkwfvin9lr9H/gXESk=)

### Comments

regression after [96a4d95](https://github.com/vuetifyjs/vuetify/commit/96a4d95aeaf34b76e3ee218aa43b65eb9e78e99d)

KaelWD:
[...]

PR #22483, opened 2025-12-29
chore: replace vue/babel-jsx with vue-jsx-vapor

KaelWD:
- [ ] check reactivity
- [ ] check performance
- [ ] babel -> esbuild

This looks like a lot but it's mostly`slots.foo?.() ?? fallback` -> `renderSlot(slots, 'foo', () => fallback)`

- `slots.foo?.()` in optimized mode fails on handwritten render function passed to slots
- need to change this at some point anyway for #18308
	- there are still some `slots.foo ? <div>{ renderSlot(slots, 'foo') }</div> : fallback` so that isn't fully fixed yet


zhiyuanzmj:
Oh, Maybe i should convert LogicalExpression to ConditionExpression.

zhiyuanzmj:
https://github.com/vuejs/vue-jsx-vapor/commit/0f48e4c625905c0ebc896b254e7ebf66481d3183#diff-8ffcde0cce9ebaa70dfe58c84117c47ef6da11b6af147a2efd5285e7d57bd884

LogicalExpression `||` and `??` operator will be covert to ConditionExpression.

KaelWD:
Nice, tyvm. I think you made a mistake with `??` though I left a comment on the commit.

zhiyuanzmj:
Now, all of LogicalExpression has been cached, for render number 0 and prevent capture ref twice.
https://github.com/vuejs/vue-jsx-vapor/commit/f3e14b32b6f56f3c1cae7f49f60969e8e3bf3316#diff-8ffcde0cce9ebaa70dfe58c84117c47ef6da11b6af147a2efd5285e7d57bd884

KaelWD:
Lots of reactivity bugs again :(
Looks like maybe the same problem as #21778, slots don't update on external value changes.

zhiyuanzmj:
if user using map expression then i can't collect items to detected whether children slots is dynamic and whether should cached events.
So the slots should default to dynamic and don't cached events by default. 
Because i always use v-for and v-slot to writing vue-jsx, so I didn't realize this problem.

zhiyuanzmj:
v3.1.9 released. if you want mark slots as stable, use `_: 1` instead of `$stable: true`.

KaelWD:
It isn't only with .map()

Here https://github.com/vuetifyjs/vuetify/blob/c30a8d7ff66648d990426ee1e636a86404ee0838/packages/vuetify/src/components/VHotkey/VHotkey.tsx#L413 we have essentially
```js
const content = (
  <>{ something.value }</>
)

return <div>{ condition ? content : <div>{ content }</div> }</div>
```

which breaks reactivity on `something` afaict due to emitting `_normalizeVNode(content)` instead of `_normalizeVNode(() => content)`

Something similar is happening with `menu.value` in VSelect but I haven't got around to tracking down the cause yet https://github.com/vuetifyjs/vuetify/blob/c30a8d7ff66648d990426ee1e636a86404ee0838/packages/vuetify/src/components/VSelect/VSelect.tsx#L453

KaelWD:
There also seems to be a lot of redundant `_normalizeVNode` calls on nodes that are created by jsx-vapor:

https://github.com/vuetifyjs/vuetify/blob/c30a8d7ff66648d990426ee1e636a86404ee0838/packages/vuetify/src/components/VHotkey/VHotkey.tsx#L372-L374

```js
(_temp = props.prefix, _temp)
  ? (_openBlock(),
    _createBlock(_Fragment, {
        key: 0
    }, [_normalizeVNode((_openBlock(),
    _createElementBlock("span", _hoisted_1, [_normalizeVNode( () => props.prefix)])))], 64))
  : _normalizeVNode((_openBlock(),
    _createBlock(_Fragment, {
        key: 1
    }, [_normalizeVNode(_temp)], 64)))
```

The only one here should be on `props.prefix`

zhiyuanzmj:
https://github.com/vuejs/vue-jsx-vapor/commit/6e4613961c0ca82541af792c289c25f8873dd43a
Thank for your reports. I have been removed one normalizeVNode. but the Fragment is still needed. [repl](https://repl.zmjs.dev/vuejs/vue-jsx#eNqtVUtv1DAQ/itWODSR0oSXOISWd5HgABKgXkip3GR269axLdvZbqny3xk7j/VmtxUSHJraM988vpnx7F1kdJUzUcM6u7QNj4q7qJI1REV05O6vSkHI0YWsb/0Jz6bSTFlibxUcl1Ej65ZDGQ1aQlijpLbkjlQaqIW3SpGOLLRsyMGqhYMZzKl7ZZbjObNmPUEmBzH+JVkjW2Hjg0dUqYNkyCXvk8Hgg6BmK8JqzAtRmNRRjoKeQj5wOMoHWlEaXbK6BhEVC8oNdKkvxZBFUIeJkYZFSmpYMAHvJQoFCDuSKyNkV0YvSzHBhdQN5ew3nH5BR4Qacj4ThbaHV2Z9uKJou+3lo6bLxgVy9uMlHYrzjsvq2muC+6g84eCwc0wonkE3ie6KUyIViI2zzW2nArD2uWOlaMv9/7BicZyQ41fkDntmKymMJZXrLDl29Y0TtMfWO0m2orwFlD/xMgP2B2tAtoEHPyZb2BYn2YWre+WO0cyzH6QOv6XVYFstSA8LkywtB0vOLTQKE3H3AbqpQZykWy2Ig06JlvOU/BynOvaOMHaQCRo7YUJGECGvEbjPfdi9uIyMoqKMUpyTa7gtyGPSYajZmO0yT86SMFYxH8w+x4TkOblhnBPQGsdywqO4uCe9+9jvD3CWkhfPk+Ex9xdf4S7xc9C5z55numIWMhydBVviUw1eKib2xppDIatLqK6nR4Rz+dmsT93b2myizXvLNb3JrgzunZ3R9SOmeLtkwhRjDwN38TCDuM+EBS1VQazGfo7ShlZaoqUT9rIu8coz/HZ7yWH+3up/ElxJTvU/UwzJDAz/gs1A48pIFE9EvNMSb41iHPRXZRmugjIqxleNG5xzefPZy4KIaOOo75EjXScrI6XBgF7hLgp1n3ypvstWV6hxuNnSncCW6iXYHnPy/Qus8Twpx9+7B5TfwEjeOkI97AJ3EnIMcEKeNMxH2GIAKPwAFfaKOuuvgt/uYOqNftT5FXZfAxStrukS7qm/oM1YDFeEnXK4ePjK0WPFYLs9ftk7y2fZi+zxIeXqkmZPA5Kz8jror2fZk+zpa/RpjlH9Zmb6AJHuD/Tj3Z4=)

Resolved https://github.com/vuetifyjs/vuetify/pull/22483#issuecomment-3788477904 by https://github.com/vuejs/vue-jsx-vapor/commit/a6c0cb53481089f53b421134cb8d6bff427d83db

zhiyuanzmj:
> Lots of reactivity bugs again :( Looks like maybe the same problem as #21778, slots don't update on external value changes.

I have an idea, find the direct function scope of JSX.
 1. if the function don't have params then should to optimize slots and event handlers. 
    ```tsx
    defineComponent((props) => {
      return () => ( // <- none params, should to cache event.
        <div onClick={() => alert(props.foo)} />
      )
    })
    ```
2. if the function has params then don't to optimize.
   ```tsx
   defineComponent(() => {
     return () => [1,2,3].map(
       (item) => { // <- has params, should not mark the slots as stable.
         let resolvedItem = item
         return <Comp>dynamic slots {resolvedItem}</Comp>
      })
   })

zhiyuanzmj:
https://github.com/zhiyuanzmj/vuetify/pull/1
I tried to migrate it successfully. only need modify a few files.

PR #22491, opened 2026-01-02
feat(VDatePicker): add prefixed day events handler

### LINKED ISSUE #22428: [Feature Request] VDatePicker add add click:date and dblclick:date events
### Problem to solve
This are really needed feature:
the double click event is needed when you want to also close the dialog/menu
the click is needed for example when you want to select time in another input so you want to focus there

update:modelValue is not enought for this, because you can detext doubleclick here, and also you can change model value externally, you really need in this case the click and dblclick event

note: i saw old requests and PR where click:date was maybe implemented but its not in current latest source code of component

### Proposed solution
just emit these two events: click:date and dblclick:date

ikushum:
fixes #22428

## Description
Emit any events from the day button.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-date-picker @click:day="onClickDay" @dblclick:day="onDblClickDay" />
    </v-container>
  </v-app>
</template>

<script setup>
  const onDblClickDay = (e, day) => {
    console.log('dblclick', day)
  }

  const onClickDay = (e, day) => {
    console.log('click', day)
  }
</script>

```


KaelWD:
Calendar and datatable use `getPrefixedEventHandlers` for this which works with any event https://github.com/vuetifyjs/vuetify/blob/29e3f09a51764cb96e233008e2ac4ecba3de3496/packages/vuetify/src/components/VDataTable/VDataTableRows.tsx#L127

PR #22514, opened 2026-01-13
feat: user dashboard

Haviles04:
https://github.com/vuetifyjs/one/pull/46


johnleider (dismissed):

Haviles04:
@J-Sek  Can you take a look at this :)

packages/docs/src/components/dashboard/DashboardActivityTable.vue
```diff
@@ -0,0 +1,141 @@
...
+      name: 'Play',
+      url: 'https://play.vuetifyjs.com/playgrounds',
+      img: 'vplay.svg',
+    },
+    'default::Link': {
+      name: 'Link',
+      url: 'https://link.vuetifyjs.com/links',
+      img: 'vlink.svg',
+    },
+    'default::Studios': {
```
J-Sek:
```suggestion
    'default::StudioProject': {
```

packages/docs/src/components/dashboard/DashboardActivityTable.vue
```diff
@@ -0,0 +1,141 @@
...
+            color="warning"
+            icon="mdi-star"
+            size="small"
+          />
+        </a>
+      </template>
+
+      <template #item.property="{ item }">
+        <v-chip
+          :href="properties[item.property].url"
```
J-Sek: I don't think it makes sense for the chip to be a general product link.
If you think otherwise, we have to extract clean domain name as URL prefix we use leads to non-existing page that is rarely handled.

J-Sek (changes requested):

packages/docs/src/components/dashboard/DashboardActivityTable.vue
```diff
@@ -0,0 +1,141 @@
...
+    { title: 'Property', key: 'property' },
+    { title: 'Updated', key: 'updatedAt' },
+  ]
+
+  const itemsPerPageOptions = [{ value: 10, title: '10' }, { value: 25, title: '25' }]
+
+  const properties: Record<string, { name: string, url: string, img: string }> = {
+    'default::Bin': {
+      name: 'Bin',
+      url: 'https://bin.vuetifyjs.com/bins',
```
KaelWD: None of these urls exist, they're all actually /user/dashboard/

packages/docs/src/components/dashboard/DashboardActivityTable.vue
```diff
@@ -0,0 +1,141 @@
...
+    <v-data-table
+      v-model:search="search"
+      :headers="headers"
+      :items="items"
+      :items-per-page-options="itemsPerPageOptions"
+      :loading="one.isLoading"
+      item-value="id"
+    >
+      <template #item.title="{ item }">
+        <Jdenticon :value="item.id" />
```
KaelWD: These don't match the ones on the individual dashboards because they transform slug to id and don't actually return the internal id at all.

KaelWD:
Should display VoAuthCard when not logged in instead of an empty table and error snackbar.

packages/docs/src/components/dashboard/DashboardActivityTable.vue
Haviles04: Changed it to urlPrefix, and only using it to link to specific items

packages/docs/src/components/dashboard/DashboardActivityTable.vue
Haviles04: Makes sense, good catch

J-Sek (approved):

johnleider (approved):

Issue #10722, opened 2020-03-02
[Feature Request] Add keyboard controls to v-timepicker


jacekkarczmarczyk:
### Problem to solve
Currently user's need to use the mouse to select time. Which is not very quick and efficient.

### Proposed solution
Use the arrow keys to move around the clock when it is open. Press enter to select the time the user has currently selected.

Forked from https://github.com/vuetifyjs/vuetify/issues/5629

stuft2:
Why not also allow for entering of time by typing in numbers?

jacekkarczmarczyk:
because that's another issue https://github.com/vuetifyjs/vuetify/issues/3883

PR #22544, opened 2026-01-23
fix(VTimePicker): add aria-disabled attribute

### LINKED ISSUE #21885: [Feature Request] Add aria-disabled to disabled hours/minutes in time picker
### Problem to solve
When you set allowed-minutes on time picker some values get disabled and that works fine, but it is hard to target it in testing.
I wan to test that certain hours and minutes are disabled.

### Proposed solution
Add aria-disabled to disabled hour/minutes divs.
```
<div class="v-time-picker-clock__item v-time-picker-clock__item--disabled" style="left: 75%; top: 7%;"><span>05</span></div>
```
this is how it currently looks like, it has `1disabled` in class value but you could also add `aria-disabled` to it.
it would also help with screen readers since it would be more informative to the end user that some values are disabled.

ikushum:
fixes #21885

## Description
Add `aria-disabled` to disabled clock item

## Markup:

```vue
<template>
  <v-container>
    <v-time-picker v-model="model" :allowed-hours="allowedHours" />
  </v-container>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref(new Date())
  const allowedHours = hour => hour % 2 === 0
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VTimePicker/VTimePickerClock.tsx
+++ b/packages/vuetify/src/components/VTimePicker/VTimePickerClock.tsx
@@ -268,14 +268,16 @@ export const VTimePickerClock = genericComponent()({
             {
               genChildren.value.map(value => {
                 const isActive = value === displayedValue.value
+                const isDisabled = props.disabled || !isAllowed(value)
 
                 return (
                   <div
+                    aria-disabled={ isDisabled || undefined }
                     class={[
                       {
                         'v-time-picker-clock__item': true,
                         'v-time-picker-clock__item--active': isActive,
-                        'v-time-picker-clock__item--disabled': props.disabled || !isAllowed(value),
+                        'v-time-picker-clock__item--disabled': isDisabled,
                       },
                       isActive && backgroundColorClasses.value,
                     ]}
--- a/packages/vuetify/src/components/VTimePicker/__tests__/VTimePickerClock.spec.browser.tsx
+++ b/packages/vuetify/src/components/VTimePicker/__tests__/VTimePickerClock.spec.browser.tsx
@@ -1,0 +1,86 @@
+// Components
+import { VTimePickerClock } from '../VTimePickerClock'
+
+// Utilities
+import { render, screen } from '@test'
+
+describe('VTimePickerClock', () => {
+  describe('aria-disabled', () => {
+    it('should set aria-disabled on items disabled by allowedValues', () => {
+      render(() => (
+        <VTimePickerClock
+          min={ 0 }
+          max={ 11 }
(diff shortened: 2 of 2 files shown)


J-Sek:
I don't observe any meaningful difference. Clock is not focusable, so screen reader ignores it. Accessibility tree does not seem to improve either. All we really get is a fancy selector for tests.

Smrtnyk:
> I don't observe any meaningful difference. Clock is not focusable, so screen reader ignores it. Accessibility tree does not seem to improve either. All we really get is a fancy selector for tests.

You have a point, but it is cumbersome to test without it

J-Sek:
Correct title: "_chore(VTimePicker): mark disabled clock digits for testing_".

Would it make any difference if we set `disabled` (without `aria-*`)?

Smrtnyk:
> Correct title: "_chore(VTimePicker): mark disabled clock digits for testing_".
> 
> Would it make any difference if we set `disabled` (without `aria-*`)?

I am using vitest browser mode and expect.element().toBeDisabled()
I can try locally to see if it would work

Smrtnyk:
> Correct title: "_chore(VTimePicker): mark disabled clock digits for testing_".
> 
> Would it make any difference if we set `disabled` (without `aria-*`)?

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled#overview 
per mdn spec disabled should not go to divs
maybe this div should actually be a button since it is acting as a button in the end
then it could also be targeted via `clockContainer.getByRole("button", { name: "22" })`

ikushum:
The proper fix definately is to convert these divs to buttons

J-Sek:
When we enable focus within the clock, it should handle up/down arrows to change value by 1 to access minutes and seconds that do not have visible buttons.

I would also consider special handling for Tab
- Where does the initial focus land when the time value is 12:44 and we don't have "44" button on minutes clock?
- to exit with single click rather than forcing user to cycle until the last button. For 24h hour clock user would need to click 23 times to exit.

ikushum:
> When we enable focus within the clock, it should handle up/down arrows to change value by 1 to access minutes and seconds that do not have visible buttons.
> 
> I would also consider special handling for Tab
> 
> * Where does the initial focus land when the time value is 12:44 and we don't have "44" button on minutes clock?
> * to exit with single click rather than forcing user to cycle until the last button. For 24h hour clock user would need to click 23 times to exit.

@J-Sek The changes you are proposing should probably go to dev. Will create a separate PR for it targeting dev. 

There's also a separate issue - https://github.com/vuetifyjs/vuetify/issues/10722

PR #22573, opened 2026-02-01
fix(styles): utilities should override responsive typography

J-Sek:
`.font-weight-bold` should override `.text-md-display-large` (but it cannot do it within the same layer)

### DIFF
--- a/packages/docs/src/components/home/Entry.vue
+++ b/packages/docs/src/components/home/Entry.vue
@@ -171,12 +171,6 @@
 </script>
 
 <style lang="sass" scoped>
-  @use 'sass:map'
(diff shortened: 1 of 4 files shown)


KaelWD:
> but it cannot do it within the same layer

Can't it? If they have the same specificity and the same layer then the last one wins, and the typography classes are before font-weight and text-mono.

J-Sek:
`.text-md-display-large` lands wrapped in `@media` query, so it "gains" +1 to order, in a way.

You can verify it on `next.vuetifyjs.com` hero » title text

KaelWD:
Yeah nvm I'm dumb they're grouped by breakpoint not by name.

PR #22574, opened 2026-02-01
feat(VRow): smaller density steps

J-Sek:
Effective change:
- density steps: 24px » 16px » 8px (instead of 24px » 8px » 0px)
- restores `no-gutter`
- removes `$form-grid-gutter`
- removes `$grid-gutters` to avoid confusion

image

- can be verified on the new homepage (2nd section » Settings)

### DIFF
--- a/packages/vuetify/src/components/VGrid/VRow.ts
+++ b/packages/vuetify/src/components/VGrid/VRow.ts
@@ -54,7 +54,7 @@ function breakpointClass (type: keyof typeof classMap, prop: string, val: string
 }
 
 export const makeVRowProps = propsFactory({
-  /** @deprecated use density="comfortable" instead */
+  /** @deprecated use density="compact" instead */
   dense: Boolean,
   /** @deprecated use align-* class instead */
   align: { type: String as PropType<typeof ALIGN_VALUES[number]>, default: null, validator: alignValidator },


KaelWD:
Why 18,12 instead of 16,8?

J-Sek:
> Why 18,12 instead of 16,8?

Tested on the home page and it looks better with 12.

<details>
<summary>Compact (8px) - label is almost touching the field above</summary>

image

</details>

<details>
<summary>Compact (12px)</summary>

image

</details>

Other than this, I don't have much more to back it. We can go with 16/8 if you feel it is going to match with other use cases better.

If we keep it, the only minor change would be to drop recommending `dense` » `density=compact` and leave only `gap=8`

KaelWD:
Looks like you have something weird going on with your fonts there, labels should be centered on the outline.
image

That isn't even Roboto, compare the J in John and U in Username.

J-Sek:
Oh.. right, local override for Roboto.

image

Either way, it still renders label closer to the previous field outline than content within. The example looks worse without icons.

KaelWD:
My main justification for 8px is that's what we already had with `dense` and I didn't see anyone complaining about it. Yeah it doesn't look amazing in that example but if you go with compact you probably want an *actually compact* layout. In other components compact has generally been almost as tight as we can get it without clipping.

J-Sek:
- changed to 24/16/8
- removed duplicate variable declarations on .v-row (those are defined by `*--density-default` now)

packages/vuetify/src/components/VGrid/VRow.ts
```diff
@@ -56,8 +56,6 @@ function breakpointClass (type: keyof typeof classMap, prop: string, val: string
 export const makeVRowProps = propsFactory({
   /** @deprecated use density="comfortable" instead */
```
KaelWD:
```suggestion
  /** @deprecated use density="compact" instead */
```

PR #22598, opened 2026-02-06
fix(VTable): apply $table-header-font-size to th elements

### LINKED ISSUE #21002: [Bug Report][3.7.13] VTable: SASS variable `$table-header-font-size` not working
### Environment
**Vuetify Version:** 3.7.13
**Vue Version:** 3.5.13
**Browsers:** Google Chrome, Mozilla Firefox, Microsoft Edge
**OS:** Linux

### Steps to reproduce
1. Specify any font size to override `$table-header-font-size` in `src/styles/settings.scss`. Like  [this](https://github.com/mt-u/vtable_sass_variable_table-header-font-size_not_working/blob/master/src/styles/settings.scss#L10).
2. Then [use VTable](https://github.com/mt-u/vtable_sass_variable_table-header-font-size_not_working/blob/master/src/components/Reproduce.vue).

### Expected Behavior
The font size of the header row is displayed with the overridden value.

### Actual Behavior
The overridden value of `$table-header-font-size` is not working, both header rows and regular rows are displayed with the overridden `$table-font-size` value.

### Reproduction Link
[https://github.com/mt-u/vtable_sass_variable_table-header-font-size_not_working](https://github.com/mt-u/vtable_sass_variable_table-header-font-size_not_working)

nemanjamalesija:
## Summary

- `$table-header-font-size` was defined in `_variables.scss` but never referenced in `VTable.sass`, so overriding it via `settings.scss` had no effect
- Added `font-size: $table-header-font-size` to the `> th` selector, right next to the existing `font-weight: $table-header-font-weight`

Fixes #21002

### DIFF
--- a/packages/vuetify/src/components/VTable/VTable.sass
+++ b/packages/vuetify/src/components/VTable/VTable.sass
@@ -112,6 +112,7 @@
             > th
               height: var(--v-table-header-height)
               font-weight: $table-header-font-weight
+              font-size: $table-header-font-size
               user-select: none
               text-align: start
 
--- a/packages/vuetify/src/components/VTable/_variables.scss
+++ b/packages/vuetify/src/components/VTable/_variables.scss
@@ -9,7 +9,7 @@ $table-color: rgba(var(--v-theme-on-surface), var(--v-high-emphasis-opacity)) !d
 $table-density: ('default': 0, 'comfortable': -2, 'compact': -4) !default;
 $table-header-height: 56px !default;
 $table-header-font-weight: 500 !default;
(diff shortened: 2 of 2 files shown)


J-Sek:
For v3.x it is too late - while the drop from .875rem to .75rem (with default styles) does not seem like much, it cannot be backed by any spec (MD3 does not cover tables) and would be perceived as breaking by people bumping the framework version. I can only see it being merged if we change the variable to `inherit` instead of `$typography`»`caption`/

nemanjamalesija:
Ok. I changed the default value of `$table-header-font-size` to inherit so it doesn't affect the current rendered output. Users can still override it via `settings.scss`

J-Sek (approved):

PR #22604, opened 2026-02-07
feat(VTimePicker): Add keyboard controls

### LINKED ISSUE #10722: [Feature Request] Add keyboard controls to v-timepicker

ikushum:
fixes #10722, #21885

## Description
- [x] Convert the hour, second and minute divs to buttons
- [x] Add keyboard controls and navigation

## Markup:

```vue
<template>
  <v-container>
    <v-time-picker
      v-model="model"
      use-seconds
    />
  </v-container>
</template>

<script setup>
  import { ref } from 'vue'

  const model = ref()
</script>

```


J-Sek:
The immediate problem becomes - assistive tools will show the clock has 12 options for minutes and seconds with gaps. Not sure if VBtn approach is correct here. Maybe the clock should have role `spinbutton` and accept up/down keys to change, while the visual representation is more of a guide for people relying on pointer-based interaction.

packages/vuetify/src/components/VTimePicker/VTimePickerClock.sass
```diff
@@ -74,8 +70,9 @@
     &--readonly
       pointer-events: none
 
-    .v-time-picker-clock__item--disabled
-      opacity: var(--v-disabled-opacity)
+    &:focus-visible
+      outline: none
```
J-Sek: Outline would definitely help convey the clock is focused. Background color difference is barely visible.

packages/vuetify/src/components/VTimePicker/VTimePickerClock.tsx
```diff
@@ -268,24 +326,27 @@ export const VTimePickerClock = genericComponent()({
             {
               genChildren.value.map(value => {
                 const isActive = value === displayedValue.value
+                const isDisabled = props.disabled || !isAllowed(value)
 
                 return (
-                  <div
+                  <VBtn
```
J-Sek, line 331: Existing projects might have some customization on `.v-btn` that could unintentionally leak here. Since the styles would need to be slightly reworked for v4, maybe we should push it all to `next`?

J-Sek:
The claim "because tests look ugly" is too weak to justify the risk unintentional styling accidents. E.g. user can make all VBtn have strong shadow and even supplement custom CSS, just to later realize it leaks into this component.

Everything else looks good and we could merge without VBtn.

ikushum:
> The claim "because tests look ugly" is too weak to justify the risk unintentional styling accidents. E.g. user can make all VBtn have strong shadow and even supplement custom CSS, just to later realize it leaks into this component.
> 
> Everything else looks good and we could merge without VBtn.

@J-Sek Looks like all the v-btn issues you are explaining are not relevant anymore because this VBtn now has `_as="VTimePickerClockBtn"`. Global defaults set to `VBtn` won't impact the clock buttons.

Let me know if I'm wrong

J-Sek:
@ikushum, thanks for the answer. I think I had CSS and Sass variables in mind - e.g. solid "brutalist" shadow is easier to pull off with custom CSS.

And I did not notice we have added `!important`s as well. It is a bit backwards after cleanup we did to support CSS `@layers`.

Also, when both parts are visible (inputs and the clock), the interaction feels a bit weird... I mean, the user [Tab]s over to reach the clock, but can only change the last segment (minutes or seconds). Theoretically we could re-assign left/right arrows to toggle hours/minutes/seconds. Sorry for not bringing it in the very first feedback batch. Fixed it in last commit.

PR #22636, opened 2026-02-21
feat(VVideo): add `hide-progress-bar` prop

### LINKED ISSUE #22633: [Feature Request] VVideo: allow hiding progress bar
### Problem to solve
It would be useful to have to hide the seek bar without having to use the "controls" slot.
The seek bar is the only default control which is not hideable with a component prop.

### Proposed solution
Add a new component prop, similar to hide-fullscreen, hide-play, hide-volume etc.

Abdulqudus001:
Resolves #22633 

Added an extra prop called `hide-progress-bar` that allows us to hide the seek bar without having to use the `controls` slot

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-video
        aspect-ratio="16/9"
        image="https://cdn.jsek.work/cdn/vt-sunflowers.jpg"
        hide-progress-bar
        src="https://cdn.jsek.work/cdn/vt-sunflowers.mp4"
      />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VVideoControls.json
+++ b/packages/api-generator/src/locale/en/VVideoControls.json
@@ -5,6 +5,7 @@
     "trackColor": "Passed to the main slider `color` prop.",
     "playing": "Applies correct icon of the default play button.",
     "hidePlay": "Hides default play button.",
+    "hideSeek": "Hides default seek slider.",
     "hideVolume": "Hides default volume control.",
     "hideFullscreen": "Hides default fullscreen button.",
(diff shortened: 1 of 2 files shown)


J-Sek:
`hide-seek` feels a bit off. Would you be able to come up with some alternative naming?

Abdulqudus001:
Good point @J-Sek . I changed it to `hideSeekbar`

J-Sek:
small fix [`1be0091`](https://github.com/vuetifyjs/vuetify/commit/1be0091) - new prop should be false by default

PR #22662, opened 2026-02-28
feat(VProgressLinear): add `split` variant to align with MD3

Haviles04:
## Description
Adds a "split" option to chunks which matches md3 design

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container class="pa-md-12" max-width="1600" fluid>
      <v-card class="py-5 px-6 mb-6" rounded="lg">
        <div class="d-flex flex-column ga-3">
          <div class="d-flex align-start">
            <v-scroll-y-transition mode="out-in">
              <div :key="progressValue" class="text-h2">
                {{ progressValue }}<span class="text-h4">%</span>
              </div>
            </v-scroll-y-transition>
            <v-btn-toggle
              v-model="group"
              class="ml-auto border"
              density="compact"
              max-width="200"
              variant="solo-filled"
              flat
              hide-details
              mandatory
              single-line
            >
              <v-btn text="Transactions" />
              <v-btn text="Fees" />
            </v-btn-toggle>
          </div>
          <div class="mb-1">
            <v-progress-linear
              :buffer-value="bufferValue"
              :model-value="progressValue"
              buffer-opacity=".5"
              chunk-gap="6"
              color="blue"
              height="18"
              variant="split"
              rounded
            />
          </div>
          <div class="mb-1">
            <v-progress-linear
              :model-value="progressValue"
              variant="split"
              chunk-gap="6"
              color="blue"
              height="18"
              indeterminate
              rounded
              rounded-bar
            />
          </div>
        </div>
      </v-card>
    </v-container>
  </v-app>
</template>

<script setup>
  import { shallowRef, toRef } from 'vue'

  const group = shallowRef(0)
  const progressValue = toRef(() => [78, 35][group.value])
  const bufferValue = toRef(() => [90, 55][group.value])
</script>


```


J-Sek:
- [x] alternative API to `chunk-count="split"`
  - `variant="split"` (with "default" preserving MD2 style)

J-Sek:
- [x] list new prop in `new-in.json`
- [x] describe the prop in VProgressLinear.json
- [x] a 1 or 2 tests for visual regression (in `stories` object)

packages/vuetify/src/components/VProgressLinear/chunks.ts
```diff
@@ -64,11 +73,47 @@ export function useChunks (
...
 
+  const splitStyles = computed(() => {
+    if (!isSplit.value) return undefined
+
+    const h = toValue(height)
+    const isRounded = toValue(rounded)
+    const isReversed = toValue(reversed)
+    const halfGap = convertToUnit(chunkGap.value / 2)
+    const r = isRounded ? convertToUnit(h / 2) : undefined
+    const pos = isReversed ? 'right' : 'left'
```
J-Sek: single-letter variables and cut names belong in lambda expressions (except `val` on rare occasions). It would make this part more readable if we would use full words.

Haviles04: fixed this

J-Sek (approved):

packages/docs/src/examples/v-progress-linear/prop-variant.vue
```diff
@@ -0,0 +1,50 @@
+<template>
```
J-Sek: The examples on this page are pretty monotonous.
It will be better to remove this one and add `options = ['split']` in `usage.vue`

J-Sek (changes requested):

packages/docs/src/examples/v-progress-linear/prop-variant.vue
Haviles04: updated

J-Sek (approved):

PR #22670, opened 2026-03-02
feat(VVideo): add `src-object` prop for MediaStream/WebRTC

### LINKED ISSUE #22568: [Feature Request] V-Video set srcObject
### Problem to solve
I would need the ability to set the native video [srcObject ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/srcObject) property using the v-video component. 
The srcObject is useful, for example, when attaching a webrtc streaming as video streaming source. 

As I can see, now there isn't this possibility (using the src property doesn't work).

### Proposed solution
Expose an attribute to easily attach an srcObject to the <v-video> element.

mixelburg:
## Description

Adds a `srcObject` prop to `v-video`, allowing users to attach a `MediaStream`, `MediaSource`, or `Blob` as the video source without using the `src` attribute.

This is particularly useful for WebRTC streaming where a `MediaStream` is obtained from `navigator.mediaDevices.getUserMedia()` or an `RTCPeerConnection`.

## Changes

- Added `srcObject` prop (`MediaStream | MediaSource | Blob`) to `makeVVideoProps()`
- Added a watcher that sets `videoRef.value.srcObject` when the prop changes, and triggers the video element to render if needed
- Added a watcher on `videoRef` to apply a pending `srcObject` when the element lazily mounts
- Updated API locale (`VVideo.json`) with a description for the new prop

## Usage

```vue
<script setup>
  import { onMounted, ref } from 'vue'

  const stream = ref(null)
  const streamAccessError = ref('')

  async function loadCameraStream () {
    await new Promise(r => setTimeout(r, 1000))
    return navigator.mediaDevices.getUserMedia({ video: true })
      .then(v => stream.value = v)
      .catch(err => streamAccessError.value = err)
  }

  async function retry () {
    streamAccessError.value = false
    await loadCameraStream()
  }

  onMounted(async () => {
    await loadCameraStream()
  })
</script>

<template>
  <v-app theme="dark">
    <v-container max-width="600">
      <v-card>
        <v-video
          :error="!!streamAccessError"
          :src-object="stream"
          height="400"
          autoplay
          muted
          @retry="retry"
        >
          <template #header>
            <v-slide-y-transition>
              <v-alert
                v-if="streamAccessError"
                :text="streamAccessError.message"
                title="No access to the video stream"
                type="error"
              />
            </v-slide-y-transition>
          </template>
        </v-video>
      </v-card>
    </v-container>
  </v-app>
</template>
```

## Closes

Closes #22568

### DIFF
--- a/packages/docs/src/examples/v-video/slot-error.vue
+++ b/packages/docs/src/examples/v-video/slot-error.vue
@@ -1,17 +1,46 @@
 <template>
-  <v-video
-    class="mx-auto"
-    max-width="450"
-    src="https://cdn.jsek.work/cdn/vt-sunflowers.mp4"
-    error
-  >
-    <template v-slot:error-details>
-      <div>
-        <div>
-          <div class="font-weight-bold">Video unavailable</div>
-          <div class="text-body-medium">Report problem to our <a href="" @click.stop @mousedown.prevent @pointerdown.stop>support</a></div>
+  <v-container class="d-flex flex-wrap justify-center align-center ga-8">
+
+    <v-video
+      max-width="450"
+      src="https://www.pexels.com/download/video/3620220"
+      error
+    >
+      <template v-slot:error>
+        <div class="d-flex align-center ga-4">
+          <v-icon class="opacity-60" color="error"></v-icon>
+          <div>
+            <div class="font-weight-bold mb-1">Video unavailable</div>
+            <div class="text-body-medium">Report problem to our <a href="">support</a></div>
+          </div>
         </div>
-      </div>
-    </template>
-  </v-video>
+      </template>
+    </v-video>
+
+    <v-video
+      ref="secondVideoRef"
+      max-width="450"
+      src="https://www.pexels.com/download/video/3620220"
+      error
+    >
+      <template v-slot:error>
+        <div
+          class="d-flex align-center ga-4 pl-4 pr-8 py-3 rounded-pill border cursor-pointer"
+          @click="secondVideoRef.retry()"
+        >
+          <v-icon class="opacity-60" icon="$loading"></v-icon>
+          <div>
+            <div class="font-weight-bold mb-1">Connection timout</div>
+            <div class="text-body-medium">Tap to reload</div>
+          </div>
+        </div>
+      </template>
+    </v-video>
+  </v-container>
 </template>
+
+<script setup>
+  import { ref } from 'vue'
+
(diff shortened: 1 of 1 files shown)


packages/vuetify/src/labs/VVideo/VVideo.tsx
```diff
@@ -222,6 +223,16 @@ export const VVideo = genericComponent<VVideoSlots>()({
...
 
+    watch(() => props.srcObject, async v => {
+      if (v) triggered.value = true
+      await nextTick()
+      if (videoRef.value) videoRef.value.srcObject = v ?? null
+    })
+
+    watch(videoRef, v => {
+      if (v && props.srcObject) v.srcObject = props.srcObject
+    }, { immediate: false })
```
J-Sek: It is `false` by default
```suggestion
    })
```

J-Sek (changes requested):
I have upgraded the demo in the description. Would it make sense for the VVideo to show an error state instead of spinning infinitely?

mixelburg:
Great point, @J-Sek! I've just pushed a fix for this — the video element now has an `onError` handler that transitions state to `'error'` and emits a new `error` event. In error state:

- The loading spinner is hidden
- The poster overlay is hidden
- An error icon (`$error`) is shown in the center overlay
- Clicking the video retries the load via `videoRef.load()`

The `v-video--error` CSS class is also applied to the root element so consumers can style the error state as needed.

J-Sek:
If the device does not have a camera, this `getUserMedia` fails and the `src-object` is not set. Error state is helpful, but the initial failure is a case where we would need a binding `error`«-»`retry`.

I have updated the demo in the PR description again. Let me know if you can test it against disconnecting device or interrupting the stream somehow (after it started successfully). I want to make sure the `onVideoError` is triggered in scenarios involving `src-object`.

packages/vuetify/src/labs/VVideo/VVideo.tsx
mixelburg: Removed — thanks for the catch.

mixelburg:
Updated the `error` prop to support `v-model:error` for bi-directional binding. Now when `getUserMedia` fails on the consumer side, setting `:error="true"` transitions VVideo to error state, and clicking to retry emits `update:error` with `false`. Internal video errors also sync back via `update:error`.

PR #22673, opened 2026-03-03
fix(VCheckboxBtn): restore full opacity for indeterminate state icon

### LINKED ISSUE #22039: [Bug][VDataTable] Unnecessary opacity for VCheckbox in indeterminate state
### Problem to solve
This extra style makes the checkbox look as it is disabled for some reasons and It is not aligned with material design

### Proposed solution
Just remove the opacity

mixelburg:
## Description

When a `VCheckboxBtn` (and by extension `VCheckbox`) is in the **indeterminate** state, the checkbox icon was rendered with reduced opacity (`var(--v-medium-emphasis-opacity)`, typically ~0.6), making it appear as if the checkbox is disabled.

This happens because `v-selection-control--dirty` (which sets icon opacity to 1) is only applied when `model.value` is truthy. In indeterminate state, the model value is `false` (unchecked), so the icon remains at medium-emphasis opacity.

## Changes

- Added `v-checkbox-btn--indeterminate` CSS class to the `VSelectionControl` wrapper in `VCheckboxBtn` when the checkbox is in indeterminate state.
- Added a CSS rule in `VCheckbox.sass` to set `opacity: 1` for the icon when this class is present.

## Before / After

| Before | After |
|--------|-------|
| Indeterminate icon shown at ~60% opacity (looks disabled) | Indeterminate icon shown at 100% opacity (matches Material Design) |

Fixes #22039

### DIFF
--- a/packages/vuetify/src/components/VCheckbox/VCheckbox.sass
+++ b/packages/vuetify/src/components/VCheckbox/VCheckbox.sass
@@ -10,3 +10,7 @@
 
     .v-selection-control
       min-height: var(--v-input-control-height)
+
+  .v-checkbox-btn--indeterminate
+    .v-selection-control__input > .v-icon
+      opacity: 1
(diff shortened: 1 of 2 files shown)


J-Sek:
Overall good change, but we should include an alignment to MD3 spec so the `color` applies to the indeterminate state as well. Otherwise it stands out too much and could be reported by QA as bug:

<details>
<summary>Preview</summary>

image

</details>

Ideally, we would make it possible to restore original behavior with `indeterminate-color`.

PR #22679, opened 2026-03-04
fix(styles): add missing `.rounded-md` utility class

juanjcardona13:
## What
Adds `rounded-md` to the `$rounded` Sass map so that the `rounded-md` utility class and `rounded="md"` prop work as shown in the docs (e.g. border-radius / misc-rounded-corners example).

## Changes
- **packages/vuetify:** Add `'md': $border-radius-root` to `$rounded` in `src/styles/settings/_variables.scss`.
- **packages/docs:** Document `rounded-md` and all directional variants in the border-radius page; update SASS example and copy.

## Note for reviewers
When using the **rounded** prop: `rounded` (no value) applies a component-specific class (e.g. `v-sheet--rounded`) that uses the default radius from Sass, while `rounded="md"` applies the value for `md` from the `$rounded` map. They can look the same when `md` equals the default, but the underlying mechanism is different. Worth keeping in mind when comparing usages.


J-Sek:
Included minor change in `md2.ts` blueprint - it used to set `md` value which made the app generate useless `rounded-md` in components. Minor fix, should not affect existing apps unless devs supplemented the class definition and did not bother to let us know about the issue.

J-Sek (approved):

PR #22682, opened 2026-03-05
feat(VProgress): create new component

### LINKED ISSUE #22256: [Feature Request] VProgressLinear - slots for label and value
### Problem to solve
Dedicated slots for label and value could help developers avoid crafting those on the fly. When label slot is provided, it could supply `aria-labelledby` to improve accessibility.

### Proposed solution

Expose new slots.

Potential challenge:
- some prop or an easy way to put those slots either above or below the bar
- needs a `<div>` wrapper, so will land in v4.0.0

Nice to have: new `label` prop that could land as `aria-label` if (yet another) `hide-label` prop is present.

Design idea from MD3 spec:

Image

Similar to [react-aria](https://react-spectrum.adobe.com/react-aria/ProgressBar.html)

Haviles04:
Fixes #22256 

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="800">
      <v-progress
        buffer-value="70"
        class="my-10"
        details-position="bottom"
        label="Bottom Label"
        model-value="60"
        stream
      />

      <v-progress
        buffer-value="70"
        class="my-10"
        details-position="top"
        label="Top Label"
        model-value="60"
        stream
      />

      <v-progress
        class="my-10"
        details-position="bottom"
        label="Bottom Label"
        model-value="60"
        type="circular"
      />

      <v-progress
        class="my-10"
        details-position="top"
        label="Top Label"
        model-value="60"
        type="circular"
      />

      <v-progress
        buffer-value="70"
        class="my-10"
        details-position="top"
        label="Top Label"
        model-value="60"
        type="circular"
        hide-value
      />

      <v-progress
        buffer-value="70"
        class="my-10"
        details-position="bottom"
        label="Bottom Label"
        model-value="60"
        type="circular"
        hide-label
      />
    </v-container>
  </v-app>
</template>
```

### DIFF
--- a/packages/docs/src/examples/v-progress/misc-card-loader.vue
+++ b/packages/docs/src/examples/v-progress/misc-card-loader.vue
@@ -1,6 +1,6 @@
 <template>
   <v-card
-    border="sm opacity-50"
+    border="md opacity-30"
     class="mx-auto pa-1"
     height="300"
     rounded="xl"
@@ -11,9 +11,9 @@
       <v-img
         v-if="done"
         key="image"
-        height="300"
+        height="292"
         src="https://cdn.vuetifyjs.com/images/carousel/sky.jpg"
-        style="border-radius: 20px;"
+        style="border-radius: 18px;"
         cover
       ></v-img>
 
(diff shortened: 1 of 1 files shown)


Haviles04:
@J-Sek Let me know your thoughts and I can write up some test and doc files

J-Sek:
Good foundation.
- [x] move to `./labs` folder
- [x] add to `nav.json`
- [x] add to `page-to-api.json`
- [x] add `VProgress.json` with API descriptions
- [x] add `valueFormat` prop (similar to `titleFormat` in `VPieTooltip.tsx` and getting us what `valueLabel` do in react-aria)
- [x] add `aria-*` attributes (align with react-aria)
- [x] docs page, usage.vue and 2-3 examples
- ~~maybe useDimension?~~ if requested by the community

J-Sek:
If you asked about mixing `makeVProgressLinearProps` and `makeVProgressCircularProps`... it does not feel right. TS will suggest props that make no sense.

I think we need to try something different - e.g limit to just a subset of common props:
- modelValue
- indeterminate
- color
- bgColor
- rounded
- theme

... and add slot `default` (with props full of aria attributes) for the user to customize it further if needed. Ofc it would also be possible to use global defaults to pass props within, so it is not like we force people to use this slot.

packages/vuetify/src/labs/VProgress/index.ts
```diff
@@ -0,0 +1 @@
+export { VProgress, makeVProgressProps } from './VProgress'
```
J-Sek: be careful when exporting stuff in index.ts - it should include only components. No types, no helpers, as API generator will error during build and it is not clearly pointing to this place.

PR #22703, opened 2026-03-10
feat(VAvatarGroup): add strictLimit prop to make limit include overflow avatar

### LINKED ISSUE #22702: [Feature Request] Clarify VAvatarGroup limit prop to consistently limit rendered avatars
### Problem to solve
The current limit prop on VAvatarGroup (labs) limits datasets processed before creating an overflow avatar, but doesn’t limit total rendered avatars. With 3 datasets and limit="2", it renders 3 avatars (2 individual + “+1”), which confuses users expecting exactly 2 visible avatars max. This mismatch between prop name/intent and behavior leads to unexpected layouts. Also the third rendered avatar that contains "+1" seems wasteful, as that avatar could've been used to just render the third dataset, if there is already space for 3 rendered avatars.

### Proposed solution
Consider making the limit behavior more predictable by either:
1. Always applying it as a strict cap on total rendered avatars (so  limit="2"  with 3 datasets shows 1 avatar + “+2” overflow), or
2. Adding a configurable prop (e.g.  strictLimit  boolean) to toggle between current dataset-limiting logic and strict rendered-avatar limiting. This would align better with user expectations while maintaining flexibility.

Happy to discuss or contribute a PR!

levibeeck:
## Description

Added a prop to enforce the limit as the max total number of avatars actually rendered (including the overflow indicator) while preserving current limit logic rendering the overflow indicator additionally to the limit.

resolves #22702

## Markup:

```vue
<template>
    <v-app>
        <v-container class="d-flex flex-column gr-4">
            <h3>Example using limit of "4"</h3>
            <h4>Without strict (renders 5 avatars, of which the last is useless containing "+1")</h4>
            <div class="d-flex align-center gc-4">
                <v-avatar-group
                    border="md surface opacity-100"
                    :items="items"
                    :limit="4"
                />
            </div>
            <h4>With strict (renders 4 avatars)</h4>
            <div class="d-flex align-center gc-4">
                <v-avatar-group
                    border="md surface opacity-100"
                    :items="items"
                    :limit="4"
                    strict-limit
                />
            </div>
        </v-container>
    </v-app>
</template>

<script setup>
import { VAvatarGroup } from '../src/labs/VAvatarGroup'
const items = [
    { image: 'https://vuetifyjs.b-cdn.net/docs/images/one/snips/avatars/3.jpg' },
    { image: 'https://vuetifyjs.b-cdn.net/docs/images/one/snips/avatars/4.jpg' },
    { icon: 'mdi-account', color: 'info' },
    { text: 'JD', color: 'primary' },
    { icon: 'mdi-account', color: 'success' },
]
</script>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VAvatarGroup.json
+++ b/packages/api-generator/src/locale/en/VAvatarGroup.json
@@ -4,10 +4,11 @@
     "gap": "Sets the overlap gap between avatars. Negative values cause avatars to overlap.",
     "hoverable": "Enables a hover animation on child avatars.",
     "itemProps": "Props object that will be applied to each item component. `true` will treat the original object as raw props and pass it directly to the component.",
-    "limit": "Limits the number of visible avatars. Remaining items are represented by an overflow indicator.",
+    "limit": "Limits the number of items rendered into avatars. Remaining items are represented by an additional overflow indicator.",
     "overflowText": "Custom text to display in the overflow avatar. Defaults to `+N` where N is the number of hidden items.",
     "reverse": "Reverses the stacking order of the avatars.",
     "size": "Sets the size of all child avatars.",
+    "strictLimit": "Enforces the limit as the maximum total number of rendered avatars, including the overflow indicator.",
     "vertical": "Stacks avatars vertically instead of horizontally."
   },
   "slots": {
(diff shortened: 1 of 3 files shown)


J-Sek:
Two very similar props might end up hurting us more than helping. Since the component is still in labs, we are allowed to introduce breaking changes. Can you adapt the PR to make it fix the original `limit` prop?

Issue #176, opened 2017-03-03
9.0 - modal has a horiz scrollbar with long text


sgehrman:
![image](https://cloud.githubusercontent.com/assets/287137/23540964/47509a56-ff99-11e6-9b24-4e7f248ed11a.png)

Wasn't happening in 8.9

dohomi:
@sgehrman could you post a small codepen for this issue? maybe can you manually try to set this CSS property to the inner div:
```
                word-wrap: break-word;
```
Just curious if this would fix it..

johnleider:
That is on the container

sgehrman:
I added the word-wrap: break-word; and it did break the word once, but it was still sticking out.  It needs to break multiple times I guess.  I'm experimenting now.  I'm 100% sure it was working which is weird.

sgehrman:
adding overflow-x: hidden; to .modal
gets rid of the horizontal scroll, but my text is clipped, but it does look nicer.  Maybe that's what changed?
anyway, you can choose not to fix this since it's really an odd case in my app.

sgehrman:
OK, I think the best fix is adding overflow: hidden; text-overflow: ellipsis;  to my card and it's the best looking solution.

           <v-card-text style='text-align: left; overflow: hidden; text-overflow: ellipsis; background: rgba(255, 255, 255, .02);'> {{item.value}} </v-card-text>

PR #22710, opened 2026-03-12
refactor: migrate display composable to v0 createBreakpoints

johnleider:
## Summary

- Updates `@vuetify/v0` to `^0.1.6`
- Replaces hand-rolled breakpoint computation in `display.ts` with `createBreakpoints()` from `@vuetify/v0`
- Deletes ~90 lines of duplicated breakpoint logic (watchEffect, getClientWidth/Height, defaultDisplayOptions, parseDisplayOptions)
- Replaces manual resize listener with `useWindowEventListener` from v0
- ~~Adds `xxlAndUp` and `xxlAndDown` to `DisplayInstance`~~
- ~~Zero changes to consumer components — `useDisplay()` API is unchanged~~
- DisplayInstance properties are now readonly

## Test plan

- [x] `pnpm lint` passes (typecheck + eslint)
- [x] Browser tests pass (742/743 — 1 pre-existing flaky VCombobox test)
- [x] Manual: resize browser, verify breakpoint classes update on VNavigationDrawer, VDataTable

## Playground

```vue
<template>
  <v-app theme="dark" v-resize="onResize">
    <v-container ref="container" class="mt-12 outline-dashed">
      <v-sheet class="mx-auto pa-6">
        <pre>{{ $vuetify.display.thresholds }}</pre>
        <v-table>
          <tbody>
            <tr>
              <td>Viewport width</td>
              <td class="text-right">
                <v-code class="text-body-large">{{ $vuetify.display.width }}px</v-code>
              </td>
            </tr>
            <tr>
              <td>Breakpoint</td>
              <td class="text-right">
                <v-code class="text-body-large">{{ $vuetify.display.name }}</v-code>
              </td>
            </tr>
            <tr>
              <td>VContainer width</td>
              <td class="text-right">
                <v-code class="text-body-large">{{ containerWidth }}px</v-code>
              </td>
            </tr>
          </tbody>
        </v-table>
        <br>
        <v-code>cols: xs:12 / sm:6 / md:4 / lg:3 / xl:2 / xxl:1</v-code>
        <br>
        <v-row gap="4" wrap>
          <v-col
            v-for="i in 12"
            :key="i"
            cols="12"
            lg="3"
            md="4"
            sm="6"
            xl="2"
            xxl="1"
          >
            <div class="bg-red text-center py-2">{{ i }}</div>
          </v-col>
        </v-row>
        <br>
        <div :key="$vuetify.display.width" class="d-flex flex-wrap ga-2">
          <template v-for="key in flags" :key="key">
            <v-chip
              v-if="$vuetify.display[key]"
              :color="key.endsWith('AndUp') ? 'yellow' : key.endsWith('AndDown') ? '' : 'success'"
              :text="key"
              prepend-icon="mdi-check"
              label
            />
          </template>
        </div>
      </v-sheet>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { type ComponentPublicInstance, ref, useTemplateRef } from 'vue'
  import { type DisplayInstance } from '@/types'

  const container = useTemplateRef<ComponentPublicInstance>('container')

  const containerWidth = ref(0)
  function onResize () {
    containerWidth.value = container.value?.$el.clientWidth
  }

  const flags = [
    'mobile',
    'xs',
    ...'sm md lg xl xxl'.split(' ').flatMap(v => [`${v}AndDown`, v, `${v}AndUp`]),
  ] as (keyof DisplayInstance)[]
</script>
```


packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
 }
 
 export function createDisplay (options?: DisplayOptions, ssr?: SSROptions): DisplayInstance {
```
MatthewAry, line 122: Am I missing something? I can't seem to find where DisplayOptions is defined as a type.

J-Sek: Line 27?

J-Sek:
> Adds `xxlAndUp` and `xxlAndDown`

It is not clear what we gain by doing it. One is redundant to `xxl` and the other is always `true`. Looking at the code briefly it technically allows configuring with `xxxl: 3400` that would make those work differently, but it is probably not intended.

see [v0#156](https://github.com/vuetifyjs/0/issues/156)

johnleider:
~~Realistically all sizes could be defined by the user, like theme colors.~~ Keeping the sizes as predefined.

J-Sek:
It would be nice - I tend to limit the options (3 elevation levels, 3 surface shades, 3 rounding levels). Having 3 breakpoints instead of 5 - with custom naming mobile/tablet/desktop might fit some design guidelines.

Do we have a precedence for type generation? Without it, the TS suggestions/validation will be precise.

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
-  const { thresholds, mobileBreakpoint } = parseDisplayOptions(options)
+  const ssrOptions = isObject(ssr)
+    ? { clientWidth: ssr.clientWidth, clientHeight: ssr.clientHeight }
+    : undefined
+
+  const breakpoint = createBreakpoints({
+    ...!isNullOrUndefined(options?.mobileBreakpoint) && { mobileBreakpoint: options.mobileBreakpoint },
+    ...!isNullOrUndefined(options?.thresholds) && { breakpoints: options.thresholds },
```
KaelWD: Ewwwwww, why doesn't createBreakpoints accept undefined values?

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
 }
 
 export function createDisplay (options?: DisplayOptions, ssr?: SSROptions): DisplayInstance {
-  const { thresholds, mobileBreakpoint } = parseDisplayOptions(options)
+  const ssrOptions = isObject(ssr)
+    ? { clientWidth: ssr.clientWidth, clientHeight: ssr.clientHeight }
+    : undefined
```
KaelWD: What's the purpose of this bit, the types are the same.

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
-    onScopeDispose(() => {
-      window.removeEventListener('resize', updateSize)
-    }, true)
+    useWindowEventListener('resize', () => breakpoint.update(), { passive: true })
   }
 
-  return { ...toRefs(state), update, ssr: !!ssr }
+  // v0 returns Readonly<ShallowRef<T>> which doesn't satisfy Ref<T> at the
+  // type level (Readonly removes the setter). ShallowRef extends Ref at runtime
+  // so the cast is safe. DisplayInstance keeps Ref<T> to avoid public API break.
```
KaelWD: This might be the least consequential potential breaking change I've ever seen, what's the problem with making these readonly in a major version?

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
+  type R<T> = Ref<T>
+  const r = <T>(v: unknown) => v as R<T>
```
KaelWD: `as any` with extra steps

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,62 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
+    mdAndDown: r<boolean>(breakpoint.mdAndDown),
+    lgAndDown: r<boolean>(breakpoint.lgAndDown),
+    xlAndDown: r<boolean>(breakpoint.xlAndDown),
+    xxlAndUp: r<boolean>(breakpoint.xxlAndUp),
+    xxlAndDown: r<boolean>(breakpoint.xxlAndDown),
+    name: r<DisplayBreakpoint>(breakpoint.name),
+    height: r<number>(breakpoint.height),
+    width: r<number>(breakpoint.width),
+    mobile: r<boolean>(breakpoint.isMobile),
+    mobileBreakpoint: toRef(() => options?.mobileBreakpoint ?? 'lg'), // 'lg' matches v0's default
```
KaelWD: So why the fallback here instead of just using the one in v0?

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,34 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
-    state.width = width.value
-    state.mobile = mobile
-    state.mobileBreakpoint = mobileBreakpoint
-    state.platform = platform.value
-    state.thresholds = thresholds
-  })
-
   if (IN_BROWSER) {
-    window.addEventListener('resize', updateSize, { passive: true })
+    breakpoint.update()
```
KaelWD: This is called here: https://github.com/vuetifyjs/vuetify/blob/605cd8097ba60ca6bc17b5c4083128c1ff40453a/packages/vuetify/src/framework.ts#L97-L111

v0 also has this but I don't understand how it doesn't cause hydration mismatch: https://github.com/vuetifyjs/0/blob/f6f6933dab0636735bb79f52f9da2b5f6a5a2b69/packages/0/src/composables/useBreakpoints/index.ts#L273

`ssr: true` was originally a hint for the *client* that the server did not know its width. The initial render should use the same value as the server (0px by default) to avoid hydration mismatch, then re-render with the correct value.

packages/vuetify/test/setup/unit-setup.ts
```diff
@@ -6,6 +6,23 @@ if (!globalThis.visualViewport) {
   vi.stubGlobal('visualViewport', new EventTarget())
 }
 
+// Stub matchMedia for v0's createBreakpoints in JSDOM
```
KaelWD: v0 should use SUPPORTS_MATCH_MEDIA for this.

packages/docs/src/pages/en/features/display-and-platform.md
```diff
@@ -162,6 +164,8 @@ In the following example, we use a switch statement and the current breakpoint n
   lgAndUp: boolean // > 1144
   xlAndDown: boolean // < 2138
   xlAndUp: boolean // > 1544
+  xxlAndDown: boolean // >= 0 (always true)
+  xxlAndUp: boolean // >= 2138
```
KaelWD: These don't exist and would be redundant if they did.

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +122,32 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
-  return { ...toRefs(state), update, ssr: !!ssr }
+  return {
+    ...breakpoint,
+    mobile: breakpoint.isMobile,
+    mobileBreakpoint: toRef(() => breakpoint.mobileBreakpoint),
+    platform: readonly(platform),
+    thresholds: toRef(() => breakpoint.breakpoints as DisplayThresholds),
+    update,
+    ssr: !!ssr,
+  } as unknown as DisplayInstance
```
KaelWD: https://github.com/vuetifyjs/0/issues/176

This hides the missing properties.

KaelWD:
Is there a reason why v0 uses `initialWidth = 0` even if not SSR?

johnleider:
Addressed review feedback:

- Removed `xxlAndUp`/`xxlAndDown` from docs and `DisplayInstance`
- Removed `as unknown as DisplayInstance` cast so type gaps are visible

The underlying v0 fixes land in vuetifyjs/0@066b700e and need a release before this PR is mergeable:

- SSR hydration: both server and client now use `ssr.clientWidth` for initial values, avoiding mismatch
- `SUPPORTS_MATCH_MEDIA` guard on `matchMedia` calls
- Removed unsafe generic type assertions (`as E`) per #176
- `context.ssr` now means "SSR was configured" (not "am I on the server")

J-Sek (dismissed):

packages/vuetify/src/composables/display.ts
```diff
@@ -150,72 +120,32 @@ function getPlatform (ssr?: SSROptions): DisplayPlatform {
...
-    state.thresholds = thresholds
-  })
-
   if (IN_BROWSER) {
-    window.addEventListener('resize', updateSize, { passive: true })
-
-    onScopeDispose(() => {
-      window.removeEventListener('resize', updateSize)
-    }, true)
+    useWindowEventListener('resize', () => breakpoint.update(), { passive: true })
```
johnleider: Can remove the IN_BROWSER check, useWindowEventListener already checks it.

KaelWD (approved):

J-Sek (approved):

PR #22713, opened 2026-03-13
fix(VListItem): prevent navigation when disabled link is clicked

### LINKED ISSUE #22172: [Bug Report][3.10.5] v-list-item navigates when disabled if click occurs on append slot content
### Environment
**Vuetify Version:** 3.10.5
**Vue Version:** 3.5.22

### Steps to reproduce
1. Click the badge inside the append slot of the disabled list item (“Item 2”).

### Expected Behavior
No navigation and no click handling when the list item is disabled.

### Actual Behavior
Clicking the badge in the append slot triggers navigation as if the item were enabled.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNq1UrtuwzAM/BVBsx+NgwwN3AId+w11B8ZmAwEKJYiMOxT99zJ2HFjo0qHZdCfydCfc25fl1NcvMVbjGe3etoKn6EHwuSNj2rGEGKfjBAhGdwRxgcohwScmEzGdgJDkOjSNecdidPbG3djSqb4RJx6fOvt6AZvOGgmK6s6uFnRlsWLGkn2QvVpBGrKZSfgAwxFNH3xIKoMphaSafSBRX8psFTnyjnTIA7NSAxKjvtfW1/X85Tr7hYVcRfhTsmaVzAyO4eAxt/9vER/vGHFhlxoo/NWDuS16M9clU1fIfXJRDKOcoxqaoV7Z78Juq13VNPZy2DxUO1t8gGcssl7O3PsPM/LTng==)

mixelburg:
Fixes #22172

When a `v-list-item` has `href` or `to` (renders as `<a>`) and is `disabled`, clicking content inside a slot (e.g. a badge in the `append` slot) still causes the browser to follow the link.

The root cause: the `onClick` handler returns early when `!isClickable.value` (which covers the disabled case), but native anchor click behaviour fires independently of our handler — `preventDefault()` was never called.

**Fix:** call `e.preventDefault()` at the start of `onClick` when `props.disabled` is true. This is the same pattern used by other Vuetify components (e.g. `VBtn`) to block navigation on disabled link elements.

Added a browser test verifying that clicking a disabled list item with `href` does not navigate (`event.defaultPrevented === true`).

### DIFF
--- a/packages/vuetify/src/components/VList/VListItem.tsx
+++ b/packages/vuetify/src/components/VList/VListItem.tsx
@@ -233,6 +233,7 @@ export const VListItem = genericComponent<VListItemSlots>()({
     } satisfies ListItemSlot))
 
     function onClick (e: MouseEvent) {
+      if (props.disabled) e.preventDefault()
       emit('click', e)
       if (['INPUT', 'TEXTAREA'].includes((e.target as Element)?.tagName)) return
 
--- a/packages/vuetify/src/components/VList/__tests__/VListItem.spec.browser.tsx
+++ b/packages/vuetify/src/components/VList/__tests__/VListItem.spec.browser.tsx
@@ -1,0 +1,26 @@
+// Components
+import { VList, VListItem } from '..'
+
+// Utilities
+import { render, screen, userEvent } from '@test'
+
+describe('VListItem', () => {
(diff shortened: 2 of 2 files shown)


J-Sek:
I can still focus the list item and click `Enter` to navigate

PR #22725, opened 2026-03-17
feat(VExpansionPanelTitle): add hover prop to disable hover effect

### LINKED ISSUE #21615: [Feature Request] Add a way to disable the hover effect on <v-expansion-panels>
### Problem to solve
This feature allows developers to disable the hover effect from `<v-expansion-panels>` without relying on CSS rules. Right now the desired result can be achieved with this:
```
.v-expansion-panel-title:hover > .v-expansion-panel-title__overlay {
  --v-hover-opacity: 0;
}
```

### Proposed solution
Add a prop or a SASS variable to control above mentioned behavior.

mixelburg:
## Summary

Closes #21615

There was no built-in prop to disable the hover overlay effect on `v-expansion-panels`; the only option was a CSS workaround:

```css
.v-expansion-panel-title:hover > .v-expansion-panel-title__overlay {
  --v-hover-opacity: 0;
}
```

This PR adds a `hover` boolean prop (default: `true`) to both `VExpansionPanelTitle` and `VExpansionPanels`.

## Usage

```html
<!-- Individual panel title -->
<v-expansion-panel-title :hover="false">Title</v-expansion-panel-title>

<!-- All panels in a group -->
<v-expansion-panels :hover="false">
  <v-expansion-panel>...</v-expansion-panel>
</v-expansion-panels>
```

## Changes

- `VExpansionPanelTitle.tsx` — adds `hover: { type: Boolean, default: true }` prop and applies the `v-expansion-panel-title--no-hover` modifier class when disabled
- `VExpansionPanel.sass` — adds CSS rule that forces overlay opacity to `0` when the modifier class is present
- `VExpansionPanels.tsx` — adds `'hover'` to the `pick()` list and to `provideDefaults` so the value cascades to all child panels
- `VExpansionPanelTitle.json` — adds documentation for the new prop


J-Sek:
- [ ] PRs  adding new props need to target `dev`
- [ ] needs entry in `new-in.json`
- [ ] cannot use `!important`, should either add `--hover` class to "set" hover effect or make use of `tools.layer('trumps')` if there is no better way and we have to override the CSS from lower layers

I can see the changes in the dedicated tab, no need to repeat those - better to leave PR description for information that is not clear upon inspecting the code (like motivation behind some less-obvious of the changes).

PR #22738, opened 2026-03-20
feat(VRating): expose activeColor in item slot props

### LINKED ISSUE #22273: [Feature Request] Add `activeColor` to `#item` slot props for `VRating`
### Problem to solve
Currently, on the `VRating` component, only `color` is passed to `#item` slot. It would be nice to have `activeColor` there for simpler configuration.
 
### Proposed solution
Pass `activeColor` down as slot prop to `#item` in `VRating` component.

### EDIT: 
I did notice now that `activeColor` comes down as `color` inside `props`. It is slightly counter-intuitive, and could at least be specified in documentation.

mixelburg:
Fixes #22273

Currently only `color` is passed to the `#item` slot, even though `activeColor` is a distinct prop that gets resolved separately. This makes it hard to use `activeColor` for custom rendering without manually re-deriving it from props.

The `activeColor` is already computed in `itemState` — this just includes it in the returned object so it shows up in the slot payload alongside `color`.

Before:
```vue
<VRating>
  <template #item="{ color, props, isFilled }">
    <!-- activeColor not available here -->
  </template>
</VRating>
```

After:
```vue
<VRating active-color="red" color="grey">
  <template #item="{ color, activeColor, isFilled, props }">
    <VIcon :color="isFilled ? activeColor : color" v-bind="props" />
  </template>
</VRating>
```

### DIFF
--- a/packages/vuetify/src/components/VRating/VRating.tsx
+++ b/packages/vuetify/src/components/VRating/VRating.tsx
@@ -29,6 +29,7 @@ type VRatingItemSlot = {
   isHovered: boolean
   icon: IconValue
   color?: string
+  activeColor?: string
   props: Record<string, unknown>
   rating: number
 }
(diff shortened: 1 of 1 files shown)


J-Sek:
- [ ] features go to `dev` branch

PR #22739, opened 2026-03-20
feat(VCarousel): add stopOnHover prop to pause cycling on hover

### LINKED ISSUE #10544: [Feature Request] Add mouse hover to stop cycle(change of slide) in carousel.
### Problem to solve
currently, there is no way to stop slide when hovering over the slide in v-carousel

### Proposed solution
well i don't have a proposed solution, sorry...also thanks you for making this wonderful component framework.. :)

mixelburg:
Fixes #10544

Adds a `stop-on-hover` prop to `VCarousel`. When set alongside `cycle`, the auto-cycling pauses while the user hovers over the carousel and resumes when they move away.

The implementation adds an `isHovered` ref and `onMouseenter`/`onMouseleave` handlers on the root `VWindow`. The handlers are no-ops unless `stopOnHover` is true, so there's no runtime cost for the default case. The `startTimeout` guard also checks `isHovered.value` so a programmatic restart (e.g. slide change) won't queue the next tick while hovering.

Usage:
```vue
<v-carousel cycle :interval="3000" stop-on-hover>
  <v-carousel-item v-for="item in items" :key="item.src" :src="item.src" />
</v-carousel>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VCarousel.json
+++ b/packages/api-generator/src/locale/en/VCarousel.json
@@ -10,6 +10,7 @@
     "nextIcon": "The displayed icon for forcing pagination to the next item.",
     "prevIcon": "The displayed icon for forcing pagination to the previous item.",
     "progress": "Displays a carousel progress bar. Requires the **cycle** prop and **interval**.",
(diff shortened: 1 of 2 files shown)


J-Sek:
new prop qualifies this for `dev` branch

PR #22743, opened 2026-03-21
feat(VSkeletonLoader): add chip-group type

### LINKED ISSUE #20695: [Feature Request] Skeleton loader for chip group
### Problem to solve
It is complicated to make a good looking skeleton loader for a chip group

### Proposed solution
On a VSkeletonLoader, adding a new "type" prop value "chip-group" to mimic a small v-chip-group

mixelburg:
Closes #20695

Adds a `chip-group` type to `VSkeletonLoader` that renders 4 chip-shaped skeleton bones in a row, mimicking the appearance of a `v-chip-group`.

**Before:** Users had to compose their own chip skeletons using `type="chip@4"` manually.

**After:** `<v-skeleton-loader type="chip-group" />` provides a convenient shorthand.

The new type follows the same pattern as existing composed types like `paragraph` (`text@3`) and `sentences` (`text@2`).

### DIFF
--- a/packages/vuetify/src/components/VSkeletonLoader/VSkeletonLoader.tsx
+++ b/packages/vuetify/src/components/VSkeletonLoader/VSkeletonLoader.tsx
@@ -28,6 +28,7 @@ export const rootTypes = {
   card: 'image, heading',
   'card-avatar': 'image, list-item-avatar',
   chip: 'chip',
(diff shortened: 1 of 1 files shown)


J-Sek:
I don't see how it addresses the claim from issue author
> _It is complicated to make a good looking skeleton loader for a chip group_

PR #22764, opened 2026-03-25
fix(SSR): avoid errors when rendering with latest Vue

### LINKED ISSUE #22762: [Bug Report][4.0.4] SSR crash in useFocusTrap: onScopeDispose accesses document without IN_BROWSER guard
### Environment
**Vuetify Version:** 4.0.4
**Vue Version:** 3.5.31+
**OS:** Any (server-side rendering context)

### Steps to reproduce
1. Use Nuxt (or any SSR framework) with Vuetify 4.0.4 **and Vue 3.5.31+**
2. Render any component using `useFocusTrap` (e.g. `VDialog`, `VMenu` with `retainFocus`, `VSelect`)
3. Request the page server-side

### Expected Behavior
SSR rendering completes without errors.

### Actual Behavior
SSR crashes with:

```
Error: Cannot read properties of undefined (reading 'removeEventListener')
  at Array.<anonymous> (node_modules/vuetify/lib/composables/focusTrap.js:129:14)
  at EffectScope.stop (@vue/reactivity/dist/reactivity.cjs.js:124:25)
  at cleanupContext (@vue/server-renderer/dist/server-renderer.cjs.js:662:15)
  at renderToString (@vue/server-renderer/dist/server-renderer.cjs.js:1014:5)
```

### Root cause

Vue 3.5.31 fixed a bug where `EffectScope.stop()` was not executing `onScopeDispose` callbacks during SSR ([vue-core#14548](https://github.com/vuejs/core/pull/14548)). With the fix, Vue now correctly calls those callbacks on the server — but Vuetify's `onScopeDispose` in `composables/focusTrap.js` calls `document.removeEventListener` directly without an `IN_BROWSER` guard:

```js
if (IN_BROWSER) {
  // ✅ setup listeners correctly guarded
}

onScopeDispose(() => {
  registry.delete(trapId);
  clearTimeout(focusTrapSuppressionTimeout);
  document.removeEventListener('pointerdown', onPointerdown);   // ❌ not guarded
  document.removeEventListener('focusin', captureOnFocus);
  document.removeEventListener('keydown', captureOnKeydown);
  if (--subscribers < 1) {
    document.removeEventListener('keydown', onKeydown);
  }
});
```

On the server, `document` is `undefined`, causing the crash.

**Workaround:** pin Vue to `3.5.30` until a fix is released.

**Expected fix:** wrap the `document.removeEventListener` calls inside `onScopeDispose` with the same `IN_BROWSER` check used in the setup block.

J-Sek:
fixes #22762

steps to verify:

1. scaffold new project with Nuxt (Vite project needs more manual work to setup SSR, and it behaves the same anyway)
2. run it locally
3. inject code below into HelloWorld.vue

```html
<v-select :items="[1,2,3]" />
<v-number-input />
<v-time-picker />
<v-slider v-model="sliderValue" />
<v-virtual-scroll renderless :height="300" :items="Array.from({ length: 999 }, (_, i) => i)">
  <template v-slot:default="{ item }">Item {{ item }}</template>
</v-virtual-scroll>
```

```ts
const sliderValue = shallowRef(10)
useHotkey('shift+right', () => sliderValue.value++)
useHotkey('shift+left', () => sliderValue.value--)
```

Note: tested without changes to `hotkey.ts` and it does not break

### DIFF
--- a/packages/vuetify/src/components/VNumberInput/hold.ts
+++ b/packages/vuetify/src/components/VNumberInput/hold.ts
@@ -1,5 +1,6 @@
 // Utilities
 import { onScopeDispose } from 'vue'
+import { IN_BROWSER } from '@/util'
 
 const HOLD_REPEAT = 50
 const HOLD_DELAY = 500
@@ -21,6 +22,7 @@ export function useHold ({ toggleUpDown }: { toggleUpDown: (increment: boolean) 
   }
 
   function holdStop () {
+    if (!IN_BROWSER) return
     window.clearTimeout(timeout)
     window.clearInterval(interval)
     window.removeEventListener('pointerup', holdStop)
--- a/packages/vuetify/src/components/VSlider/slider.ts
+++ b/packages/vuetify/src/components/VSlider/slider.ts
@@ -7,7 +7,7 @@ import { makeRoundedProps } from '@/composables/rounded'
 
 // Utilities
 import { computed, nextTick, onScopeDispose, provide, ref, shallowRef, toRef } from 'vue'
-import { clamp, createRange, getDecimals, propsFactory } from '@/util'
+import { clamp, createRange, getDecimals, IN_BROWSER, propsFactory } from '@/util'
 
 // Types
 import type { ExtractPropTypes, InjectionKey, PropType, Ref } from 'vue'
@@ -316,6 +316,7 @@ export const useSlider = ({
   }
 
   onScopeDispose(() => {
+    if (!IN_BROWSER) return
     window.removeEventListener('touchmove', onMouseMove)
     window.removeEventListener('mousemove', onMouseMove)
     window.removeEventListener('mouseup', onSliderMouseUp)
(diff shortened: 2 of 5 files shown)


alkihis:
If I can help,
It might be useful to also look at those usages that might be invalid too:
- https://github.com/vuetifyjs/vuetify/blob/131e6593b07796377e3baa53b629610aa2ee4a2f/packages/vuetify/src/components/VVirtualScroll/VVirtualScroll.tsx#L91
- https://github.com/vuetifyjs/vuetify/blob/131e6593b07796377e3baa53b629610aa2ee4a2f/packages/vuetify/src/components/VTimePicker/VTimePickerClock.tsx#L234
- https://github.com/vuetifyjs/vuetify/blob/131e6593b07796377e3baa53b629610aa2ee4a2f/packages/vuetify/src/components/VSlider/slider.ts#L318

This is no need to handle cases for `onUnmounted` or `onBeforeUnmount`, because they're not called during server-side rendering ([according to Vue documentation](https://vuejs.org/api/composition-api-lifecycle.html#onbeforeunmount))

PR #22765, opened 2026-03-26
refactor: migrate theme system to @vuetify/v0

johnleider:
## Summary

- Replaces Vuetify's internal theme runtime with `@vuetify/v0`'s `createTheme()` for theme selection, `isDark` state, and reactive registry
- Creates `VuetifyThemeAdapter` for Vuetify-specific CSS generation (`.v-theme--*` selectors, utility classes, `@layer`, variables, overlay multipliers)
- Extracts color definitions and generators into `theme/colors.ts`
- System theme detection delegated to v0's `usePrefersDark()`
- Consumer API (`useTheme`, `provideTheme`, `makeThemeProps`, `ThemeSymbol`) fully preserved — 50+ components unchanged

## Breaking Changes

- `ThemeInstance.styles` ref removed — CSS injection handled by adapter
- `ThemeInstance.isDisabled` removed — themes always enabled
- `ThemeInstance.isSystem` removed — check `name.value === 'system'`
- New theme registration via `register()` instead of `themes.value.custom = { ... }`

## Test plan

- [x] `pnpm build:lib` — 554 files compiled
- [x] `pnpm lint` — tsc + eslint clean
- [x] `pnpm test:unit -- --run` — 661 passed (4 pre-existing date.spec.ts failures)
- [x] Verify theme switching in browser (light/dark/system)
- [x] Verify VThemeProvider scoped themes in browser
- [x] Verify utility classes (.bg-primary, .text-primary) in browser


J-Sek:
while verifying [this demo](https://play.vuetifyjs.com/playgrounds/ENpm6Q) locally, I noticed `cycle()` goes through "system".

I personally don't like it - given it is attached to a simple dual-state button, it feels broken 1/3 of the times user clicks. It can also result in missing colors (`current.dark` does not resolve, so app falls back to default theme colors). It needs a better callout in the Upgrade Guide if we intend to keep it.

PR description needs a quick human check. I don't think this point applies
> - ThemeInstance.isSystem removed — check name.value === 'system'

J-Sek (changes requested):
- PR description needs to be reflect intended changes to avoid confusing devs
- `current.dark` with system theme - breaking change needs fix or better callout
- `cycle()` - (without arguments) breaking change needs fix or better callout
- (nice to have) some hint on what "adapter" might enable... quick info in `features/theme.md`

packages/docs/src/pages/en/getting-started/upgrade-guide.md
```diff
@@ -13,593 +13,59 @@ related:
...
-### VDatePicker
-
-`multiple="range"` emits only the start and end dates instead of everything in between. Use something like [date-fns' `eachDayOfInterval`](https://date-fns.org/v4.1.0/docs/eachDayOfInterval) if you need all dates in the range.
+Assigning new themes directly to `themes.value` is replaced by `register()`:
 
 ```diff
-- ['2023-09-28', '2023-09-29', '2023-09-30', '2023-10-01', '2023-10-02']
-+ ['2023-09-28', '2023-10-02']
+- theme.themes.value.custom = { dark: true, colors: { primary: '#ff5722' } }
++ theme.register({ id: 'custom', dark: true, colors: { primary: '#ff5722' } })
```
J-Sek, line 43: `useTheme` does not expose `register` yet

J-Sek: I assume this surgically adds specific fields to the theme. Let's say one piece of code sets primary, another one picks surface color - they should not interfere.. My previous demo does not work yet, so I am speculating.

J-Sek (changes requested):

packages/docs/src/pages/en/getting-started/upgrade-guide.md
```diff
@@ -13,593 +13,59 @@ related:
...
 ```
 
-### VFileInput
+Mutating existing theme colors continues to work:
```
J-Sek, line 46: not true at the moment

packages/vuetify/src/composables/theme/adapter.ts
```diff
@@ -0,0 +1,218 @@
...
+        '@layer theme-foreground {\n',
+        ...fgLines.map(v => `  ${v}`),
+        '}\n',
+      )
+    }
+
+    return '@layer vuetify-utilities {\n' + lines.map(v => `  ${v}`).join('') + '\n}'
+  }
+
+  // @ts-expect-error Vue types mismatch between v0 and vuetify packages
```
KaelWD, line 140: So fix the types mismatch? Why is this even a generic function? If v0 defined it as `context: ThemeAdapterSetupContext` you wouldn't have to do anything and it could just inherit from ThemeAdapter.setup.

packages/docs/src/pages/en/getting-started/upgrade-guide.md
```diff
@@ -21,3 +21,30 @@ This page contains a detailed list of breaking changes and the steps required to
...
+## Theme
+
+The theme system now uses `@vuetify/v0` under the hood. The consumer API (`useTheme`, `VThemeProvider`) is unchanged for most users.
+
+### ThemeInstance
+
+Several properties have been removed from the `ThemeInstance` type:
+
+- `styles` — CSS injection is now handled internally by the theme adapter
+- `isDisabled` — themes are always enabled
```
KaelWD, line 34: isDisabled still exists in parsedOptions but isn't exposed in ThemeInstance because...?

packages/vuetify/src/composables/theme.ts
```diff
@@ -1,592 +0,0 @@
...
-  readonly computedThemes: DeepReadonly<Ref<Record<string, InternalThemeDefinition>>>
-  readonly prefix: string
-
-  readonly themeClasses: Readonly<Ref<string | undefined>>
-  readonly styles: Readonly<Ref<string>>
-
-  readonly global: {
-    readonly name: Ref<string>
-    readonly current: DeepReadonly<Ref<InternalThemeDefinition>>
-  }
```
KaelWD, line 121: This removal is missing from the upgrade guide.

Without global it is not possible to read or modify the application theme from within a theme provider.

PR #22769, opened 2026-03-26
fix(types): add type declarations for CSS-only style imports to resolve TS2882

### LINKED ISSUE #22766: [Bug Report][4.0.4] error TS2882: Cannot find module or type declarations for side-effect import of 'vuetify/styles'.
### Environment
**Vuetify Version:** 4.0.4
**Vue Version:** 3.5.31
**OS:** Linux

### Steps to reproduce
1. create new vuetify project
2. update `typescript` dependency from 5.x.x to 6.x.x
3. Run `pnpm vue-tsc -b`

```
src/main.ts:1:8 - error TS2882: Cannot find module or type declarations for side-effect import of 'vuetify/styles'.

1 import "vuetify/styles";
         ~~~~~~~~~~~~~~~~

```

### Reproduction Link
<https://github.com/Jisu-Woniu/digital-signature/tree/73b316175925b0dbf036d2efe6946cd33d92e4af>

mixelburg:
## Summary

Fixes #22766

### Problem

TypeScript 5.2 introduced error TS2882: *"Cannot find module or type declarations for side-effect import of 'vuetify/styles'"*. This affects any project that imports `vuetify/styles` (or the sub-paths `vuetify/styles/colors`, `vuetify/styles/core`, `vuetify/styles/utilities`) with strict type checking enabled (e.g. via `vue-tsc` or `tsc`).

The root cause is that the `./styles` (and related) entries in `package.json`'s `exports` map don't have a `"types"` condition, so TypeScript doesn't know where to find type declarations.

### Fix

- Add stub `export {}` declaration files in `src/styles/` for each affected entry point (`main.d.ts`, `colors.d.ts`, `core.d.ts`, `utilities.d.ts`). The existing build step (`babel src --out-dir lib --copy-files`) already copies `.d.ts` files from `src/` to `lib/`, so the built package will include them automatically.
- Add `"types"` conditions pointing to those files in the `./styles`, `./styles/colors`, `./styles/core`, and `./styles/utilities` exports in `package.json`.

### Tested

With this patch applied and `lib/styles/*.d.ts` present, `import "vuetify/styles"` resolves without TS2882 in both `tsc` and `vue-tsc`.


J-Sek:
it is about TS 6.0

how did you verify this fix is working?

KaelWD:
2098fb1

PR #22777, opened 2026-03-28
feat(VFileUpload): add allowPaste prop to support pasting files from clipboard

### LINKED ISSUE #22732: [Feature Request] Add area to v-file-upload to allow pasting from clipboard
# Problem:

In the bug/feature tool I would like to add a screenshot. That screenshot by default is put into my clipboard. I have no way to use that clipboard and must instead paste it into some image program and save it manually (or use another screenshot tool)

# Solution:

- Variant A: Enable a focusable element (e.g. textarea) where one can paste something the usual way
- Variant B: Add a button "Insert from clipboard" which uses the Clipboard API to read the content

## Context

This made my previous feature request more annoying to create, and whatever I did still did not work (drag and drop). Having proper copy-paste support would have likely avoided this issue.

> PS: Because they seem to got lost...
> 
> UI:
> 
> Image
> 
> Fullcalendar usage:
> 
> Image 

 _Originally posted by @septatrix in [#22731](https://github.com/vuetifyjs/vuetify/issues/22731#issuecomment-4091106417)_

mixelburg:
Fixes #22732

## Problem

There's no way to paste files (e.g. screenshots) directly into `VFileUpload`. Users have to paste into an image editor first, save the file, then drag-and-drop or browse for it.

## Solution

Added an `allowPaste` boolean prop. When enabled, a `paste` event handler intercepts `ClipboardEvent.clipboardData.files` and passes them through the existing `selectAccepted()` flow — same path used by drag-and-drop and file browse selection.

```vue
<v-file-upload allow-paste multiple />
```

Users can now Ctrl+V / Cmd+V anywhere on the page when the component is focused to add files.

## Implementation notes

- Uses the same `selectAccepted()` path as drag-drop → file type filtering and `rejected` events work as expected
- `allowPaste` defaults to `false` → fully backward-compatible
- Guard: no-op when `disabled` is set
- No paste handler is attached at all when `allowPaste` is false (no unnecessary event listener)

### DIFF
--- a/packages/api-generator/src/locale/en/VFileUpload.json
+++ b/packages/api-generator/src/locale/en/VFileUpload.json
@@ -1,5 +1,6 @@
 {
   "props": {
+    "allowPaste": "Enables pasting files directly into the dropzone via Ctrl+V / Cmd+V.",
     "browseText": "Text for the browse button.",
     "dividerText": "Text in the divider.",
     "hideBrowse": "Hides the browse button.",
--- a/packages/vuetify/src/labs/VFileUpload/VFileUpload.tsx
+++ b/packages/vuetify/src/labs/VFileUpload/VFileUpload.tsx
@@ -62,6 +62,7 @@ export const makeVFileUploadProps = propsFactory({
     type: IconValue,
     default: '$upload',
   },
+  allowPaste: Boolean,
   clearable: Boolean,
   insetFileList: Boolean,
   hideBrowse: Boolean,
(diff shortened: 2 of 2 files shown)


J-Sek:
1. read Docs » contributing » [choosing-a-base-branch](https://vuetifyjs.com/en/getting-started/contributing/#choosing-a-base-branch)
2. switch target to `dev`
3. should support pasting folders

PR #22811, opened 2026-04-22
Add warning for v-app-bar placement in v-app

bartduisters:
Added a warning about the placement of the v-app-bar component.

### DIFF
--- a/packages/docs/src/pages/en/components/app-bars.md
+++ b/packages/docs/src/pages/en/components/app-bars.md
@@ -26,6 +26,12 @@ The `v-app-bar` component is pivotal to any graphical user interface (GUI), as i
 
 <PageFeatures />
 
+::: warning
+
+Must be placed within a `v-app` component.
+
+:::
+
 ## Usage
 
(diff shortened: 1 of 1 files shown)


J-Sek:
Not sure if it helps when it is so vague. What was the problem? Missing `v-app` or placing `v-app-bar` deeper in the DOM tree? It should also work with `v-layout`, but going with too much detail can confuse readers.

bartduisters:
A student was trying out components, so the v-app-bar was copy pasted as shown in the example. There then was a warning/error in the console about layout. I then took a look and assumed v-app-bar should be placed within v-app, which fixed it.

This was my attempt to make the documentation page self containing. Someone new to Vuetify doesn't know it should live within a v-app or v-layout, is there a specific page/piece of information I missed?

KaelWD:
The app components' pages should probably all link to https://vuetifyjs.com/en/features/application-layout/ somehow. (Also that page barely even mentions v-app and v-layout or the difference between them)

PR #22903, opened 2026-06-05
fix(VTreeview): more aria attributes and correct keyboard navigation

### LINKED ISSUE #10796: [Feature Request] v-treeview: Keyboard support
### Problem to solve
Add keyboard support: 
- navigate using Arrow Up & Arrow Down keys,
- (de)select item using Enter or Space keys,
- delete item using Del key,
-  etc.

Existing components:
- https://amsik.github.io/liquor-tree/#Keyboard-Navigation
- https://element.eleme.io/#/en-US/component/tree
- https://vue-treeselect.js.org/

### Proposed solution
Add props:
- deleteable: boolean

Add events:
- keyup: Emits the standard vue events https://vuejs.org/v2/guide/events.html#Key-Modifiers
- delete: Emits the array of deleted items

### LINKED ISSUE #21585: [Feature Request] Add Aria-labels to VTreeview Buttons
### Problem to solve

### Proposed solution

So if you have a tree like:

```
> Red
V Green
   > Orange
```

J-Sek:
- aria-level/posinset/setsize
- aria-expanded only for expandable nodes
- arrow left/right should collapse/expand and move focus to parent/child (RTL-aware)
- toggle button gets `tabindex=-1` so the row owns expansion

resolves #10796
resolves #21585

## Markup:

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="500">
      <button>before</button>
      <v-treeview
        :items="items"
        item-value="id"
        select-strategy="classic"
        open-all
        selectable
      >
        <template #append="{ item }">
          <v-select
            v-if="item.id === 2"
            v-model="selected"
            :items="['a', 'b', 'c', 'd']"
            density="compact"
            style="width: 100px"
            hide-details
          />
          <v-switch
            v-if="item.id === 3"
            v-model="switched"
            hide-details
          />
        </template>
      </v-treeview>
      <button>after</button>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const selected = ref('a')
  const switched = ref(false)

  const items = [
    {
      id: 1,
      title: 'Root',
      children: [
        { id: 2, title: 'Child 1' },
        { id: 3, title: 'Child 2', children: [{ id: 4, title: 'Grandchild' }] },
      ],
    },
  ]
</script>
```


docpelux:
Test with Accessibility Insights for Web now has another issue, see comment below.

=========================================================================

Title: WCAG 4.1.2: Ensure buttons have discernible text (.v-btn.v-btn--icon.v-btn--density-compact)
Tags: Accessibility, WCAG 4.1.2, button-name

Issue: Ensure buttons have discernible text (button-name - https://accessibilityinsights.io/info-examples/web/button-name)

Target application: Vuetify Dev Playground - http://localhost:8090/

Element path: #v-list-group--id-1 > .v-list-item__prepend > .v-list-item-action.v-list-item-action--start > .v-btn.v-btn--icon.v-btn--density-compact

Snippet: <button tabindex="-1" type="button" class="v-btn v-btn--icon v-theme--dark v-btn--density-compact v-btn--size-default v-btn--variant-text v-treeview-item__toggle">

How to fix: 
Fix any of the following:
  Element does not have inner text that is visible to screen readers
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element has no title attribute
  Element does not have an implicit (wrapped) <label>
  Element does not have an explicit <label>
  Element's default semantics were not overridden with role="none" or role="presentation"

Environment: Chrome version 149.0.0.0

====

This accessibility issue was found using Accessibility Insights for Web 2.47.0 (axe-core 4.11.3), a tool that helps find and fix accessibility issues. Get more information & download this tool at http://aka.ms/AccessibilityInsights.

docpelux:
The buttons have tabindex -1, so they are not reachable, and is ok.

They need to have `aria-hidden=true`.

docpelux:
If you set `selectable` in the treeview, when targeting checkboxes with tab, screen reader simply says "checkbox not checked" but never reads the row, i expect something like this "Grandchild not checked" or "checkbox not checked row Grandchild".

If i navigate the tree with arrows, screen reader never tells me that Child1 or Grandchild are selectable items.

Screen reader is **orca** from Ubuntu.


The following is partially AI generated:

>   Expected behavior (per WAI-ARIA Treeview pattern):
>   
>   1. The treeitem element itself should carry aria-checked="true|false" so that navigating with
>   arrow keys announces both the label and the checked state (e.g. "Child 1, not checked, tree
>   item").
>   2. The internal <input type="checkbox"> should have tabindex="-1" so it is not an independent tab
>   stop. Selection should be toggled via Space on the focused treeitem.
>   3. If the checkbox must remain focusable independently, it needs aria-labelledby pointing to the
>   associated v-list-item-title element so the screen reader announces the correct label.
>   
>   Suggested fix:
>   
>   - Add aria-checked to [role="treeitem"] elements that have a checkbox, reflecting the current
>   selection state.
>   - Remove the checkbox <input> from the tab order (tabindex="-1") since focus management in a tree
>   should be handled at the treeitem level, and add aria-hidden="true".
>  - Add aria-label="select" to the checkbox <input>, to mute wave (false positive), screen readers never reads it.
>   - Handle Space keypress on the treeitem to toggle selection.
>   
>   This aligns with the WAI-ARIA Treeview pattern
>   (https://www.w3.org/WAI/ARIA/apg/patterns/treeview/).

J-Sek:
Is there any UI library that ships tree component with such extensive labels? I am trying to align with react-aria that does not pollute labels with selection status, but it also has selection strategy equivalent to the "independent", so it is semantically a selection list and does not really need them.

Just curious. Is it something you'd use or are we trying to satisfy the a11y requirements without consulting with actual end users?

docpelux:
Having to comply with the EAA regulations from the EU directive.
However I don't use a screen reader directly, so perhaps there's a better solution.
I think that when I target a selectable tree element, it should say that I targeted a selectable tree element.

However the solution I have proposed was written by AI, so I don't know...

===============

Non related, addendum

The simple solution I use for now is this:
Is a little abrupt but hides accessibility errors from automatic checks, i don't know what happens if tree changes or page is reloaded, and if causes performance issues. i leave code here for anyone has problems with trees.

Note that when this pr is merged this code conflicts with its new aria attributes.
Is a composable.

edited: 11 July

```ts
import { nextTick, onMounted } from 'vue';

/**
 * Workaround to fix/hide treeview accessibility issues.
 */
export function useTreeviewAccessibilityFix() {
    const treeviewAccessibilityFix = () => {
        nextTick(() => {
            // Fix missing aria-labels for treeview checkboxes
            const checkboxes = document.querySelectorAll('.v-treeview-item input[type="checkbox"]');
            checkboxes.forEach((checkbox, index) => {
                const item = checkbox.closest('.v-treeview-item');
                const title = item?.querySelector('.v-treeview-item__title')?.textContent || `item ${index + 1}`;
                checkbox.setAttribute('aria-label', `Select ${title}`);
            });

            // Fix missing aria-labels for treeview expand/collapse buttons
            const expandButtons = document.querySelectorAll('.v-treeview-item .v-btn--icon');
            expandButtons.forEach((button) => {
                if (!button.getAttribute('aria-label')) {
                    const item = button.closest('.v-treeview-item');
                    const title = item?.querySelector('.v-treeview-item__title')?.textContent || 'item';
                    button.setAttribute('aria-label', `Expand/Collapse ${title}`);
                }
            });
        });
    };
    onMounted(treeviewAccessibilityFix);
}
```

docpelux:
any new news/ideas?

J-Sek:
Still cooking
- [enter] key to expand collapse
- [space] to toggle selected state
- checkbox will get the same treatment as expand button - i18n labels can be reverted entirely
- will follow up with docs overhaul

It will need to land in `dev` branch - too much disruption to ship in patch release.

docpelux:
Everything seems ok, for me it can go.

mmustafasenoglu:
Hi @KaelWD, great work on this PR! I reviewed the changes and have a few observations:

### Positives
- The `useTreeviewKeyboard.ts` composable is well-structured and cleanly separates concerns
- RTL support via `useRtl()` is properly handled for arrow key navigation
- The `aria-busy`, `aria-hidden`, and `tabindex=-1` additions follow WAI-ARIA best practices
- New a11y tests are comprehensive and cover positional attributes

### Suggestions

1. **useTreeviewKeyboard.ts:86** - The `onItemActionKeydown` function has good Enter/Space handling, but the `*` (asterisk) key shortcut for expanding all nodes isn't documented in the treeview docs. Consider adding this to the keyboard shortcuts table.

2. **VTreeviewItem.tsx:135** - The `tabindex={-1}` on the toggle button is correct for keyboard navigation, but should we also add `aria-hidden="true"` to the icon inside the button? Currently only the button gets `aria-hidden`.

3. **useTreeviewKeyboard.ts:42** - The `visibleTreeitems` function uses `focusableChildren` which is good, but what happens if the tree is virtualized? Might be worth noting this limitation.

4. **Test coverage** - The a11y tests are solid. Consider adding a test case for when `loading=true` to verify `aria-busy` behaves correctly.

Overall this is a clean, well-tested implementation. Nice work! 🎉

PR #22909, opened 2026-06-08
feat(VFileInput,VFileUpload): unify dragging effect

### LINKED ISSUE #20576: [Feature Request] Add Drag and Drop Visual Feedback to `v-file-input` Component.
### Problem to solve
When local files are dragged onto the v-input-file component, there is an absence of clear visual feedback, which might cause confusion. Users will not be certain whether drag and drop functionality is operational.

### Proposed solution
I propose adding visual changes, such as a distinct border color when a file is being dragged over the input area, similar to what the [Quasar framework](https://quasar.dev/vue-components/file#design) does. This would render it more intuitive for users.

image

mahsarajabpour:
This improves drag-and-drop feedback in `v-file-input`.

resolves #20576

## Markup

```vue
<template>
  <v-sheet class="mb-8 pa-6" border rounded>
    <h2 class="text-h6 mb-4">
      VFileInput drag feedback test
    </h2>

    <v-row>
      <v-col v-for="variant in ['outlined', 'filled', 'solo', 'solo-inverted', 'solo-filled']" :key="variant" cols="12"
             md="6">
        <v-file-input
          :label="variant"
          :variant="variant"
          multiple
        />
      </v-col>
    </v-row>
  </v-sheet>
</template>
<script setup lang="ts">
</script>

### DIFF
--- a/packages/vuetify/src/components/VFileInput/VFileInput.sass
+++ b/packages/vuetify/src/components/VFileInput/VFileInput.sass
@@ -37,17 +37,22 @@
       z-index: 0
 
     &--dragging
-      .v-field__overlay
-        background-color: rgb(var(--v-theme-primary))
-        opacity: .08
+      .v-field:not(.v-field--variant-outlined)
+        outline: 2px dashed rgba(var(--v-theme-on-surface), var(--v-medium-emphasis-opacity))
+        outline-offset: 2px
 
-      .v-field__outline
-        --v-field-border-opacity: 1
+      .v-field--variant-outlined
+        .v-field__outline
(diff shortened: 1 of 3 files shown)


J-Sek:
- when dragging, leaving to the side - the dragging state is stuck
- the effect could is almost not visible for the filled variant, solo-inverted flashes solid color, solo-filled does not respond at all
- we cannot use primary theme color here

I wonder if replacing the effect with simple dashed outline would help work around some of the problems.

BTW. PR needs some playground so we can easily paste it locally to `dev/Playground.vue` for verification.

PR #22938, opened 2026-06-20
fix(VVirtualScroll): call calculateVisibleItems after updateOffsets

### LINKED ISSUE #22937: [Bug Report][4.0.3] VVirtualScroll: paddingBottom not updated after item height changes (stale until scroll)
### Environment
**Vuetify Version:** 4.0.3
**Vue Version:** 3.5.30
**OS:** Windows

### Steps to reproduce
1. Use VVirtualScroll without the `item-height` prop (dynamic row height mode).
2. Render 50 items at 48px each, with `height="400"`.
3. Read `paddingBottom` on `.v-virtual-scroll__container` (value A).
4. Collapse item #0 from 48px → 12px (e.g. an accordion close), **without scrolling**.
5. Read `paddingBottom` again after layout settles (value B).

### Expected Behavior
Shrinking item #0 by 36px reduces the total content height, so `paddingBottom` must be recalculated: **B should differ from A** (B < A).

### Actual Behavior
**B equals A** — `paddingBottom` does not change after the collapse. The virtual list keeps stale padding (and a stale visible range) until some *unrelated* trigger — a scroll, a viewport resize, or an `items` change — happens to call `calculateVisibleItems()`.

Note: I intentionally do **not** assert on absolute padding values here.
`_calculateVisibleItems()` only updates "the side we're scrolling towards" (direction-based guard), which makes the absolute `paddingBottom` at scrollTop=0 history-dependent. The reliable, position-independent signal is that `paddingBottom` **does not change** in response to the resize.

### Reproduction Link
https://stackblitz.com/edit/vitejs-vite-jvo7yuqr
(The demo measures paddingBottom before vs after collapsing item #0, with no scrolling. On 4.0.3 the two values are equal — stale.)

### Other comments
`updateOffsets()` (debounced, triggered by `handleItemResize`) correctly rebuilds `offsets[]`, but never calls `calculateVisibleItems()`, so `paddingBottom`, `paddingTop`, and the visible `first`/`last` range stay stale until the next unrelated trigger.

The other paths that rebuild offsets — the initial-render watcher and the `items` watcher — already follow up with `calculateVisibleItems()`. 
The resize path is the only one missing it, so this looks like an oversight.

Fix: add `calculateVisibleItems()` at the end of the `updateOffsets` callback.

pinichi:
## Description

resolves vuetifyjs#22937

When `handleItemResize` is called (e.g. after an accordion-like collapse), it triggers the debounced `updateOffsets()` which correctly rebuilds `offsets[]`.

However, `calculateVisibleItems()` was never called afterwards, leaving `paddingBottom` stale until the next scroll event — causing the scrollbar to display an incorrect size.

## Markup:

```vue
<template>
  <div style="max-width: 600px; margin: 0 auto; padding: 16px">
    <h3>VVirtualScroll — paddingBottom stale after item resize</h3>

    <div
      style="
        margin: 12px 0;
        font-size: 13px;
        line-height: 1.9;
        background: #f5f5f5;
        padding: 10px;
        border-radius: 6px;
      "
    >
      <div>
        50 items × 48px. We measure <code>paddingBottom</code> of
        <code>.v-virtual-scroll__container</code> at <b>scrollTop = 0</b>,
        <b>before</b> and <b>after</b> collapsing item&nbsp;#0 (48px → 12px).
        <b>No scrolling is performed.</b>
      </div>

      <div style="margin-top: 8px">
        paddingBottom before collapse (A): <b>{{ fmt(before) }}</b>
      </div>
      <div>
        paddingBottom after collapse (B): <b>{{ fmt(after) }}</b>
      </div>

      <div
        v-if="done"
        style="margin-top: 8px; font-weight: bold"
        :style="{ color: after !== before ? '#2e7d32' : '#c62828' }"
      >
        <template v-if="after !== before">
          ✓ paddingBottom changed after collapse → visible items were
          recalculated (FIXED build)
        </template>
        <template v-else>
          ✗ paddingBottom did NOT change after collapse → stale (BUG present on
          current master)
        </template>
      </div>
    </div>

    <div style="display: flex; gap: 8px; margin-bottom: 12px">
      <button @click="run" :disabled="running">
        {{ collapsed ? 'Expand #0' : 'Collapse #0 (run test)' }}
      </button>
    </div>

    <p style="font-size: 12px; color: #666; line-height: 1.6">
      Click to toggle item #0 between 48px and 12px. Each click measures
      paddingBottom before and after the size change.<br />
      Expected: on current published Vuetify, B equals A (stale — the bug). With
      PR #22938 applied, B differs from A (paddingBottom is recalculated on
      resize). This compares before/after state rather than any absolute value.
    </p>

    <v-virtual-scroll ref="vsRef" :items="items" height="400">
      <template #default="{ index }">
        <div
          :style="{
            height: index === 0 && collapsed ? '12px' : '48px',
            background: index === 0 ? '#bbdefb' : '#c8e6c9',
            display: 'flex',
            alignItems: 'center',
            padding: '0 8px',
            borderBottom: '1px solid #e0e0e0',
          }"
        >
          #{{ index }} —
          {{ index === 0 && collapsed ? '12px (collapsed)' : '48px' }}
        </div>
      </template>
    </v-virtual-scroll>
  </div>
</template>

<script setup>
import { nextTick, ref } from 'vue';

const items = Array.from({ length: 50 }, (_, i) => i);
const collapsed = ref(false);
const vsRef = ref(null);
const before = ref(null);
const after = ref(null);
const running = ref(false);
const done = ref(false);

const fmt = (v) => (v === null ? '—' : v + 'px');

function settle() {
  return nextTick().then(
    () =>
      new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r)))
  );
}

function readPaddingBottom() {
  const el = vsRef.value?.$el?.querySelector('.v-virtual-scroll__container');
  return el ? parseFloat(el.style.paddingBottom || '0') : null;
}

async function run() {
  running.value = true;
  done.value = false;
  before.value = null;
  after.value = null;

  await settle();
  before.value = readPaddingBottom(); // A: before the size change

  collapsed.value = !collapsed.value; // toggle item #0 size (fires resize)
  await settle();
  after.value = readPaddingBottom(); // B: after the size change

  done.value = true;
  running.value = false;
}
</script>
```


J-Sek (approved):

J-Sek:
caused regression when opening long list in any dropdown

```html
<v-select
  :items="Array.from({ length: 1000 }, (_, i) => i)"
  :model-value="102"
/>
<!-- should open list showing current selection, but the list window is way off -->
```

PR #22946, opened 2026-06-26
feat(VTreeview): add aria-label and aria-expanded to expand/collapse buttons

### LINKED ISSUE #21585: [Feature Request] Add Aria-labels to VTreeview Buttons

mmustafasenoglu:
## Description
Adds accessibility attributes to the expand/collapse buttons in VTreeview. Screen readers will now announce "Expand [item]" / "Collapse [item]" instead of just "Button".

Fixes #21585

### Motivation
Currently, screen readers announce the expand/collapse buttons in VTreeview simply as "Button" without any context about what the button does. This makes the treeview inaccessible for users relying on assistive technologies.

### Changes
- Added `isOpen` prop to `VTreeviewItem` to receive the open state from parent
- `VTreeviewGroup` now passes its `isOpen` state to `VTreeviewItem` via defaults
- Added `aria-label` ("Expand [title]" / "Collapse [title]") to the toggle button
- Added `aria-expanded` attribute to indicate the current state

### Before
```html
<button>▶</button>  <!-- Screen reader: "Button" -->
```

### After
```html
<button aria-label="Expand Red" aria-expanded="false">▶</button>
<!-- Screen reader: "Expand Red, collapsed" -->
```

### DIFF
--- a/packages/vuetify/src/components/VTreeview/VTreeviewGroup.tsx
+++ b/packages/vuetify/src/components/VTreeview/VTreeviewGroup.tsx
@@ -30,6 +30,7 @@ export const VTreeviewGroup = genericComponent<VListGroupSlots>()({
         prependIcon: undefined,
         appendIcon: undefined,
         toggleIcon: toggleIcon.value,
+        isOpen: vListGroupRef.value?.isOpen ?? false,
       },
     }))
 
--- a/packages/vuetify/src/components/VTreeview/VTreeviewItem.tsx
+++ b/packages/vuetify/src/components/VTreeview/VTreeviewItem.tsx
@@ -31,6 +31,7 @@ export const makeVTreeviewItemProps = propsFactory({
   hasCustomPrepend: Boolean,
   indentLines: Array as PropType<IndentLineType[]>,
   toggleIcon: IconValue,
+  isOpen: Boolean,
 
   ...makeVListItemProps({ slim: true }),
 }, 'VTreeviewItem')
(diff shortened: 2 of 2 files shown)


mmustafasenoglu:
Thanks for the review consideration! 

I've added `aria-label` and `aria-expanded` attributes to the VTreeview expand/collapse buttons so screen readers can properly announce the item name and state (e.g., "Expand Red" / "Collapse Red").

Let me know if you need any adjustments to the accessibility attributes.

J-Sek:
Please compare against #22903 and let me know if that PR can be improved.

mmustafasenoglu:
Hi @J-Sek, thanks for pointing me to #22903.

After comparing the two:

**#22903** is more comprehensive — it adds:
- `aria-level`, `aria-posinset`, `aria-setsize` for tree structure
- Keyboard navigation (arrow left/right to collapse/expand)
- `tabindex=-1` on toggle button
- Tests for keyboard and a11y behavior

**#22946** (my PR) only adds:
- `aria-label` ("Expand/Collapse [title]")
- `aria-expanded` on the toggle button

Since #22903 already covers `aria-expanded` and provides a more complete a11y solution, I'm happy to close #22946 in favor of #22903. 

However, one thing #22903 doesn't have is the explicit `aria-label` with the item title — it relies on the button content for screen reader context. If that's useful, I could add it as a small enhancement to #22903 instead.

Let me know how you'd like to proceed!

J-Sek:
closing in favor of #22903

PR #22947, opened 2026-06-26
feat(VAlert): add timeout prop to auto-dismiss alerts

### LINKED ISSUE #20409: [Feature Request] Add duration Prop to Automatically Dismiss v-alert
### Problem to solve
Currently, v-alert components remain visible until manually dismissed, requiring developers to implement a setTimeout() to hide the alert after a certain period. This can lead to repetitive code.

### Proposed solution
 Introduce a duration prop to the v-alert component. This prop accepts a number representing the duration (in milliseconds) for which the alert should remain visible before automatically closing
**usage example :**
```
<v-alert :duration="5000" text=" This alert will automatically close after 5 seconds."/>
```

eldardada:
## Description

Adds a `timeout` prop to `VAlert` that automatically dismisses the alert after a given time in milliseconds, matching `VSnackbar`'s `timeout`. Currently developers have to wire up their own `setTimeout` to hide an alert after a delay.

- `timeout` accepts a number (or numeric string) of milliseconds.
- `-1` (the default) keeps the alert visible indefinitely, so existing alerts are unaffected. `0` dismisses immediately and `> 0` dismisses after the given delay, matching `VSnackbar`'s semantics.
- The timer pauses while the alert is hovered or focused, so it doesn't vanish while someone is reading it or reaching for the close button / a link in a slot. It restarts when the alert becomes visible again or when `timeout` changes, and is cleared on manual close and on unmount.

The auto-dismiss timeout and pause-on-hover/focus handling is shared with `VSnackbar` through a new `useAutoDismiss` composable, so the two components stay consistent and the hover/focus pause logic that used to live only in `VSnackbar` is no longer duplicated. Pausing while hovered or focused also satisfies WCAG 2.2.1 (Timing Adjustable). `VSnackbar` keeps driving its visual countdown timer through the composable's `onStart`/`onClear` hooks, so its behavior is unchanged.

Unit tests cover auto-dismiss, the pause-on-hover case, and the persistent default. The API description and a docs example have been added.

resolves #20409

> #22952 addresses the same issue. This PR was opened first and also covers the requested shared composable, the hover/focus pause, and tests for both `VSnackbar` and `VAlert`. Happy to consolidate into whichever the team prefers.

## Markup:

```vue
<template>
  <v-btn @click="visible = true">Show alert</v-btn>
  <v-alert
    v-model="visible"
    :timeout="3000"
    closable
    type="success"
    text="I will dismiss myself after 3 seconds"
  />
</template>

<script setup>

  const visible = shallowRef(true)
</script>
```

### DIFF
--- a/packages/api-generator/src/locale/en/VAlert.json
+++ b/packages/api-generator/src/locale/en/VAlert.json
@@ -6,6 +6,7 @@
     "closable": "Adds a close icon that can hide the alert.",
     "closeIcon": "Change the default icon used for **closable** alerts.",
     "closeLabel": "Text used for *aria-label* on **closable** alerts. Can also be customized globally in [Internationalization](/customization/internationalization).",
+    "duration": "Time (in milliseconds) to wait until the alert is automatically dismissed. A value of `0` keeps the alert visible.",
     "height": "Sets the height for the component.",
     "maxHeight": "Sets the maximum height for the component.",
     "maxWidth": "Sets the maximum width for the component.",
--- a/packages/api-generator/src/locale/en/VAlert.json
+++ b/packages/api-generator/src/locale/en/VAlert.json
@@ -6,7 +6,6 @@
     "closable": "Adds a close icon that can hide the alert.",
     "closeIcon": "Change the default icon used for **closable** alerts.",
     "closeLabel": "Text used for *aria-label* on **closable** alerts. Can also be customized globally in [Internationalization](/customization/internationalization).",
-    "duration": "Time (in milliseconds) to wait until the alert is automatically dismissed. A value of `0` keeps the alert visible.",
     "height": "Sets the height for the component.",
     "maxHeight": "Sets the maximum height for the component.",
     "maxWidth": "Sets the maximum width for the component.",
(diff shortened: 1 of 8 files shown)


eldardada:
On the prop name: I used `duration` as asked in #20409, though `VSnackbar` already calls the same thing `timeout`. The "off" value differs too: `0` (the default) means "never dismiss" here so existing alerts keep working, whereas `VSnackbar` defaults to `5000` and uses `-1` for never. Happy to rename to `timeout` and match that convention if you'd rather the two line up.

eldardada:
On pausing: I copied the hover/focus handling from `VSnackbar` (pointerenter/leave + focusin/out), so an alert with a close button or links in a slot doesn't vanish while someone's reading it or reaching to click. Covers WCAG 2.2.1 as a side effect.

eldardada:
On reuse: I kept the timer inline rather than extracting a shared helper. `VSnackbar`'s `useCountdown` is local and not exported, so sharing it would mean touching `VSnackbar` and pushing this past a good first issue. Can split that out separately if you want the two unified.

eldardada:
On edge values: `Infinity` and values past the 32-bit `setTimeout` limit dismiss right away, same as `VSnackbar` today. Left it for parity, can add a guard if you'd prefer.

J-Sek:
> _Can split that out separately if you want the two unified._

I think it would make sense. Including prop rename to `timeout` so we don't cause cognitive load by inconsistency when both props deliver basically the same behavior.

eldardada:
@J-Sek updated. Renamed the prop to `timeout` and moved the shared auto-dismiss + hover/focus pause into a `useAutoDismiss` composable used by both `VSnackbar` and `VAlert`. Ready for another look when you have a moment, thanks!

PR #22953, opened 2026-06-28
fix(VTreeview): add aria-label to expand/collapse toggle buttons

### LINKED ISSUE #21585: [Feature Request] Add Aria-labels to VTreeview Buttons

sridhar-3009:
Fixes #21585

## Problem

Screen readers announce VTreeview expand/collapse toggle buttons as just "Button" with no context about which node they control or what action they perform.

## Solution

- Added `expanded` prop to `VTreeviewItem` to track the open/collapsed state of a tree node
- Added `useLocale` to `VTreeviewItem` and set `aria-label` on the toggle `VBtn` using i18n keys
- Added `treeview.expand` and `treeview.collapse` locale keys to `en.ts` with `{0}` placeholder for the node title
- Pass `expanded: isOpen` from the `VTreeviewChildren` activator slot into `listItemProps`

With this change, screen readers will announce something like **"Expand Fruits"** or **"Collapse Fruits"** instead of just **"Button"**.

### DIFF
--- a/packages/vuetify/src/components/VTreeview/VTreeviewChildren.tsx
+++ b/packages/vuetify/src/components/VTreeview/VTreeviewChildren.tsx
@@ -217,6 +217,7 @@ export const VTreeviewChildren = genericComponent<new <T extends InternalListIte
                 ...activatorProps,
                 value: treeItemProps?.value,
                 ariaExpanded: isOpen,
+                expanded: isOpen,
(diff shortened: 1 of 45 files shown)


J-Sek:
closing in favor of #22903

PR #22906, opened 2026-06-07
feat(VCarousel): add wheel navigation

mahsarajabpour:
## Description

Adds mouse wheel navigation support for `VCarousel`.

Wheel handling is implemented as an opt-in `wheel` prop on `VWindow`. `VCarousel` can use it, but it is not enabled by default. The handler respects direction, so horizontal carousels react to horizontal wheel input and vertical carousels react to vertical wheel input. It also prevents repeated slide jumps during a continuous wheel gesture.

Wheel handling follows the intent of the older Vuetify 0 fix, but uses gesture locking to avoid repeated jumps on modern trackpads.

Fixes #22893

## Markup

```vue
<template>
  <v-app theme="dark">
    <v-container max-width="600">
      <div class="mb-1 mt-6">
        <v-code class="bg-purple-darken-2">default, slower</v-code>
      </div>
      <div class="my-1 reduced-motion-info">
        <v-code class="bg-red">duration change suppressed - prefers-reduced-motion: reduce</v-code>
      </div>
      <v-carousel height="200" transition-duration="600">
        <v-carousel-item v-for="(src, i) in items" :key="i" :src="src" cover />
      </v-carousel>

      <div class="mb-1 mt-6">
        <v-code class="bg-purple-darken-2">opt-in wheel</v-code>
      </div>
      <v-carousel height="200" wheel>
        <v-carousel-item v-for="(src, i) in items" :key="i" :src="src" cover />
      </v-carousel>

      <div class="mb-1 mt-6">
        <v-code class="bg-purple-darken-2">vertical wheel</v-code>
      </div>
      <v-carousel height="200" direction="vertical" wheel vertical-delimiters="right">
        <v-carousel-item v-for="(src, i) in items" :key="i" :src="src" cover />
      </v-carousel>
    </v-container>
  </v-app>
</template>

<script setup>
  const items = [
    'https://cdn.vuetifyjs.com/images/carousel/squirrel.jpg',
    'https://cdn.vuetifyjs.com/images/carousel/sky.jpg',
    'https://cdn.vuetifyjs.com/images/carousel/bird.jpg',
    'https://cdn.vuetifyjs.com/images/carousel/planet.jpg',
  ]
</script>

PR #22955, opened 2026-06-28
feat(VCarousel): add mousewheel prop for scroll navigation

### LINKED ISSUE #22893: [Feature Request] Add mouse wheel gesture support for navigation in v-carousel / v-window
### Problem to solve
Currently, users can navigate the v-carousel using touch swipe gestures, navigation arrows, or pagination dots. However, desktop users frequently rely on their mouse wheel or trackpad scroll to navigate through horizontal galleries or sliding content.

When a desktop user hovers over a large carousel and instinctively scrolls their mouse wheel, nothing happens (or the whole page scrolls, depending on the layout). This breaks the expected user experience, especially in full-screen image galleries, presentations, or horizontal content sliders where mouse-wheel navigation is an industry standard. While touch devices are well-supported, desktop interaction feels incomplete without wheel support.

### Proposed solution
I propose adding a native wheel event listener to the v-carousel (or the underlying v-window component) to allow users to scroll through items.

sridhar-3009:
Closes #22893

## Summary

Adds a `mousewheel` prop to `v-carousel` that enables mouse-wheel (and trackpad scroll) navigation between slides. Disabled by default to preserve existing behavior.

- Scroll down / right → next slide
- Scroll up / left → previous slide  
- 300ms cooldown prevents accidental multi-slide jumps per scroll gesture
- Uses `{ passive: false }` so `preventDefault()` suppresses page scrolling while hovering over the carousel
- Handler is registered on mount and cleaned up on unmount; responds reactively if the prop is toggled at runtime

## Usage

```html
<v-carousel mousewheel>
  <v-carousel-item v-for="item in items" :key="item" />
</v-carousel>
```

## Changes

- `packages/vuetify/src/components/VCarousel/VCarousel.tsx` — add `mousewheel` prop and wheel event handler
- `packages/docs/src/examples/v-carousel/prop-mousewheel.vue` — new example
- `packages/docs/src/pages/en/components/carousels.md` — document the prop

### DIFF
--- a/packages/docs/src/examples/v-carousel/prop-mousewheel.vue
+++ b/packages/docs/src/examples/v-carousel/prop-mousewheel.vue
@@ -1,0 +1,20 @@
+<template>
+  <v-carousel
+    height="300"
+    mousewheel
+  >
+    <v-carousel-item
+      v-for="(color, i) in colors"
+      :key="i"
+      :color="color"
+    >
+      <div class="d-flex fill-height align-center justify-center text-h5 text-white">
+        Slide {{ i + 1 }}
+      </div>
+    </v-carousel-item>
+  </v-carousel>
+</template>
+
+<script setup>
+  const colors = ['primary', 'secondary', 'success', 'info', 'warning', 'error']
+</script>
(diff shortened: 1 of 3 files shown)


J-Sek:
closing in favor of #22906

PR #22957, opened 2026-06-28
fix(VStepper): make step items clickable when non-linear prop is set

### LINKED ISSUE #19275: [Bug Report][3.5.2] v-stepper "non-linear" property not working as expected
### Environment
**Vuetify Version:** 3.5.2
**Vue Version:** 3.4.15
**Browsers:** Chrome 121.0.0.0
**OS:** Windows 10

### Steps to reproduce
With "non-linear" property, ideally, user should be allowed to move through the steps when they click on the v-stepper-header. However, the property seems to have no effect. Only when "editable" is added, user is allowed to move non-linearly by clicking on the header.

### Expected Behavior
Adding "non-linear" alone to the v-stepper should allow user to move from one step to another by clicking on the v-stepper-header.

### Actual Behavior
Only when "editable" property is added, user can click on the steps to move non-linearly. "non-linear" seems to have no effect.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrtVMFKxDAQ/ZUxlyrYlt11L9IteBA8qBcPHqyH2M5qIG1CkhZl2X83SbclhYIK4kH2NvPmzXtpZ5inHdGqTK+kTLoWySXJDNaSU4N50QBkXUyl9KFPStEYyhpUB6gHqarGfERiwwzHHO5FA7e2hSroYm1QSlSQpRPSTDO+mwD2haG5EU3MveKEEXLiN6RV8MoZBrMfCt5+U5C7D3iwOCwKAh3lrcNsnLt3hg1zghXrmPNy3CH+qe8y8LXxn/muAl8bf+Ublmf+cFAORzrOOpzpgH5/ka4rZugLx99aIzzoHZfovyyRTycXyiH9AcvS4LDZVJeKSQMaTdvfN1ZLoQzsQOEW9rBVoobI3sTI0QGsrjZQ61fYOMZpdIOcC3gUilcn0Zkz6CWtGNmfk1VykSzWxAXrZEmePwGZZp1v)

sridhar-3009:
Fixes #19275

## Root cause

`VStepper` provides defaults to `VStepperItem` (editable, icons, etc.) but never provided `nonLinear`. As a result, in `VStepperItem`, `isClickable` was computed as `!props.disabled && props.editable` — meaning the `nonLinear` prop had no effect on step navigation.

## Fix

- In `VStepper.tsx`: include `nonLinear` in the `provideDefaults` for `VStepperItem`
- In `VStepperItem.tsx`: add `nonLinear` prop and update `isClickable` to `!disabled && (editable || nonLinear)`
- The `disabled` attribute and ripple on the button now use `isClickable` so non-linear steppers properly enable clicking without requiring `editable`

`editable` and `nonLinear` remain independent: `editable` also shows the edit icon on the active step, `nonLinear` only enables click navigation.

### DIFF
--- a/packages/vuetify/src/components/VStepper/VStepper.tsx
+++ b/packages/vuetify/src/components/VStepper/VStepper.tsx
@@ -105,7 +105,7 @@ export const VStepper = genericComponent<new <TModel>(
   setup (props, { slots }) {
     const { items: _items, next, prev, selected } = useGroup(props, VStepperSymbol)
     const { displayClasses, mobile } = useDisplay(props)
-    const { completeIcon, editIcon, errorIcon, color, editable, prevText, nextText } = toRefs(props)
+    const { completeIcon, editIcon, errorIcon, color, editable, nonLinear, prevText, nextText } = toRefs(props)
 
     const items = computed(() => props.items.map((item, index) => {
       const title = getPropertyFromItem(item, props.itemTitle, item)
@@ -141,6 +141,7 @@ export const VStepper = genericComponent<new <TModel>(
     provideDefaults({
       VStepperItem: {
         editable,
+        nonLinear,
         errorIcon,
         completeIcon,
         editIcon,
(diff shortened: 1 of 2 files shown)


J-Sek:
behavior might be revised in v5, but in v4 we only need docs change as described [here](https://github.com/vuetifyjs/vuetify/issues/19275#issuecomment-3619551099)

PR #22960, opened 2026-06-28
feat(VFileInput): add placeholder and persistent-placeholder props

sridhar-3009:
## Summary

Fixes #19837 — adds `placeholder` and `persistent-placeholder` props to `v-file-input`, similar to `v-text-field`.

### Changes
- `VFileInput.tsx`: add `placeholder: String` and `persistentPlaceholder: Boolean` props; render placeholder text when no file is selected; include `persistentPlaceholder` in `isActive` so the label floats
- `VFileInput.sass`: add `.v-file-input__placeholder` style
- `_variables.scss`: add `$file-input-placeholder-color` SASS variable

### Usage

```vue
<v-file-input
  label="Upload documents"
  placeholder="Choose one or more files"
  persistent-placeholder
></v-file-input>
```

## Test plan
- [ ] Add `placeholder` to `v-file-input` — text should appear when no file is chosen
- [ ] Verify text disappears after a file is selected
- [ ] Verify `persistent-placeholder` keeps the label floated so placeholder remains visible
- [ ] Verify no visual regression when `placeholder` is omitted

PR #22964, opened 2026-06-28
feat(VTextarea): add placeholder slot for custom placeholder content

### LINKED ISSUE #20886: [Feature Request] Slot `placeholder` for `VTextarea`
### Problem to solve
`v-textarea` has a template slot for `label` which is useful to customize the label.

My current use case is to have a smooth transition if the label changes based on previous user input:

```vue
<v-textarea v-model="inputMessage">
  <template #label>
    <v-fade-transition mode="out-in">
      <div :key="inputMessageLabel">{{ inputMessageLabel }}</div>
    </v-fade-transition>
  </template>
</v-textarea>
```

`inputMessageLabel` changes based on previous user input.

This is cool, but that type of changing message would better be placed as a placeholder. Unfortunately I can't use a transition in the placeholder like I'm doing here, because there's no slot for it.

### Proposed solution
Have a slot for `placeholder` that works analog to the slot for `label`

sridhar-3009:
Closes #20886

## Changes

Added a `placeholder` slot to `VTextarea` that allows users to provide custom HTML content as the placeholder, enabling Vue transitions and dynamic content.

## Usage

```vue
<v-textarea v-model="message" label="Message">
  <template #placeholder>
    <v-fade-transition mode="out-in">
      <span :key="currentHint">{{ currentHint }}</span>
    </v-fade-transition>
  </template>
</v-textarea>
```

## How it works

- When the `placeholder` slot is provided, the native `placeholder` attribute is removed from the `<textarea>` element
- The slot content is rendered as an absolutely positioned overlay div (`.v-textarea__placeholder`)
- The slot is visible when the textarea has no value and is not focused (respects `persistent-placeholder` prop)
- The overlay is `pointer-events: none` so it doesn't interfere with user interaction

### DIFF
--- a/packages/docs/src/examples/v-alert/prop-duration.vue
+++ b/packages/docs/src/examples/v-alert/prop-duration.vue
@@ -10,10 +10,10 @@
     <v-alert
       v-else
       v-model="visible"
-      type="info"
+      closable
       :duration="3000"
       text="This alert will automatically dismiss after 3 seconds."
-      closable
+      type="info"
     ></v-alert>
   </div>
 </template>
--- a/packages/docs/src/examples/v-alert/prop-duration.vue
+++ b/packages/docs/src/examples/v-alert/prop-duration.vue
@@ -1,0 +1,25 @@
+<template>
+  <div>
+    <v-btn
+      v-if="!visible"
+      @click="visible = true"
+    >
+      Show alert
+    </v-btn>
+
+    <v-alert
+      v-else
+      v-model="visible"
+      type="info"
+      closable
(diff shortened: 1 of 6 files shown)


J-Sek:
[same](https://github.com/vuetifyjs/vuetify/pull/22960#pullrequestreview-4587942423) problem as in PR to VFileInput

sridhar-3009:
Closing to rework. J-Sek flagged the same issue as #22960 — needs to target `dev` branch and use an absolute-positioned overlay so existing `::placeholder` styles continue to work.

Issue #22361, opened 2025-11-25
[Bug Report] VImg: Cannot destructure property 'naturalHeight' of undefined


Timbological:
# [Bug Report] VImg: Cannot destructure property 'naturalHeight' of undefined

## Environment

- **Vuetify Version:** 3.8.12 (also affects 3.8.x - 3.11.x based on open issues)
- **Vue Version:** 3.x
- **Browsers:** All
- **OS:** All

## Description

The VImg component throws a `TypeError: Cannot destructure property 'naturalHeight' of 'img' as it is undefined` error when the `pollForSize()` function is called with an undefined `image.value`.

## Steps to Reproduce

This error occurs in several scenarios:

1. Using VImg with remote images during page transitions
2. Fast navigation between pages with VImg components (especially with cache disabled)
3. Using VImg in async components
4. Any scenario where the onLoad handler fires before `image.value` is properly set

## Expected Behavior

The VImg component should handle cases where the internal image element is not yet available and gracefully skip the `pollForSize()` call or wait until the element is ready.

## Actual Behavior

The application crashes with:

```
TypeError: Cannot destructure property 'naturalHeight' of 'img' as it is undefined.
    at poll (VImg.js:162:26)
    at pollForSize (VImg.js:175:7)
    at onLoad (VImg.js:139:7)
```

## Root Cause

In `VImg.js`, the `onLoad()` function and other places call `pollForSize(image.value)` without checking if `image.value` is defined:

```javascript
// Line 136-142 (v3.8.12)
function onLoad() {
  if (vm.isUnmounted) return;
  getSrc();
  pollForSize(image.value); // ← image.value can be undefined here
  state.value = 'loaded';
  emit('load', image.value?.currentSrc || normalisedSrc.value.src);
}
```

Similarly at lines 127 and 130:
```javascript
if (!aspectRatio.value) pollForSize(image.value, null); // Line 127
if (!aspectRatio.value) pollForSize(image.value); // Line 130
```

The `pollForSize()` function then tries to destructure properties from the undefined value:

```javascript
// Line 158-164
const poll = () => {
  clearTimeout(timer);
  if (vm.isUnmounted) return;
  const {
    naturalHeight: imgHeight, // ← Destructuring fails here
    naturalWidth: imgWidth
  } = img; // ← img is undefined
```

## Proposed Fix

Add null checks before calling `pollForSize()`:

```diff
function onLoad() {
  if (vm.isUnmounted) return;
  getSrc();
[...]

PR #23006, opened 2026-07-13
fix(VImg): don't poll for size when image ref is missing

Faustze:
## Description

The native `load` event can fire while `VImg` is kept alive by a leave transition (a fade between routes is the common case). At that point the `image` template ref is already cleared, but `vm.isUnmounted` is still `false`, so `onLoad` calls `pollForSize(image.value!)` and the destructuring throws:

```
TypeError: Cannot destructure property 'naturalHeight' of 'img' as it is undefined
```

In a Nuxt app this reproduces every time you navigate away from a page with several `v-img` cards before their images finish loading — one uncaught error per in-flight image.

The fix widens `pollForSize` to accept `undefined` and bails out early, which also removes the two non-null assertions at the call sites.

No test added: `VImg` has no existing spec and the race (load event landing during a leave transition) needs a browser-mode test to reproduce honestly. Happy to add one if you want it.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-btn text="toggle" @click="show = !show" />

      <v-fade-transition>
        <div v-if="show">
          <v-img
            v-for="n in 8"
            :key="n"
            :src="`https://picsum.photos/seed/${n}/1600/900`"
            width="200"
          />
        </div>
      </v-fade-transition>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const show = ref(true)
</script>
```

Throttle the network (e.g. Fast 4G), then click toggle while the images are still loading — each in-flight image throws once the leave transition starts.

### DIFF
--- a/packages/vuetify/src/components/VImg/VImg.tsx
+++ b/packages/vuetify/src/components/VImg/VImg.tsx
@@ -193,7 +193,7 @@ export const VImg = genericComponent<VImgSlots>()({
             if (!aspectRatio.value) pollForSize(image.value, null)
             if (state.value === 'loading') onLoad()
           } else {
-            if (!aspectRatio.value) pollForSize(image.value!)
+            if (!aspectRatio.value) pollForSize(image.value)
             getSrc()
           }
         })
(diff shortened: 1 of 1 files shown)


J-Sek:
Maybe related to [#22361](https://github.com/vuetifyjs/vuetify/issues/22361#issuecomment-4416380385)

Faustze:
Reopened as #23012 with an updated description. I now have a controlled reproduction from a real router-driven page transition (see #23011), plus this branch was rebased onto current master

Issue #22948, opened 2026-06-27
[Feature Request] Prop to disable transition on VMain


frederikheld:
### Problem to solve
This bugs me for ages and I finally figured out how to fix it:

I usually put individual VAppBars into each View because I use very different app bars (different menu items, some use the #extension slot, some don't) in each view and some views even don't have an app bar.

The different height of those app bars causes the content height to glitch on view transitions which looks bad.

I finally figured out the cause and how to fix it: `<v-main style="transition: none" />`

### Proposed solution

I would very much prefer a simple prop to turn the transition off: `<v-main disable-transition />`

Or maybe a string prop to choose from different transitions (including "none"): `<v-main :transition="none" />`

The prop would provide better DX over the custom style because the docs would have told me that there is a transition on that component and how to turn it off. Figuring out what causes the glitch took me way longer than fixing it.

J-Sek:
`$main-transition` is recommended, until we learn about some specific scenario where a dedicated prop has significant advantage.

frederikheld:
Is this a SASS variable? I've never used SASS in any of my VuetifyJS projects. SASS adds another layer of complexity to the stack that tends to break on upgrades and Vuetify already has enough of those. Most issues that could be fixed with changing SASS variables can also be fixed with custom CSS. But over all of them, I prefer props.

J-Sek:
It is understandable. I wish props could come for free, but unfortunately it does not and some people already complain about amount of props Vuetify ships with. This specific component is lean, but `.v-main { transition: none }` is what we want to set here - a project-wide static piece of CSS – i.e. additional prop looks like too much. Having more use cases might increase its value someday.

PR #23020, opened 2026-07-20
feat(VMain): default transition prop to true

UllernJ:
Hello, my first PR :)
## Description
Added a transition flag to vmain component, with default value of true to avoid breaking changes.

closes https://github.com/vuetifyjs/vuetify/issues/22948

```vue
<template>
  <v-app>
    <v-app-bar height="64" />
    <v-navigation-drawer v-model="drawer" width="240" />
    <v-main :transition="transition">
      <v-container>
        <v-switch v-model="transition" color="primary" label="transition prop" />
        <v-switch v-model="drawer" color="primary" label="toggle drawer" />
        <p>
          Toggle the drawer switch above. When <code>transition</code> is
          <strong>on</strong>, the content should slide/smoothly resize as
          the drawer opens and closes. When <code>transition</code> is
          <strong>off</strong>, it should snap instantly with no animation.
        </p>
      </v-container>
    </v-main>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const transition = ref(true)
  const drawer = ref(true)
</script>

```

### DIFF
--- a/packages/vuetify/README.md
+++ b/packages/vuetify/README.md
@@ -17,7 +17,7 @@
     <img src="https://img.shields.io/npm/l/vuetify.svg" alt="License">
   </a>
   <a href="https://discord.gg/vuetify">
-    <img src="https://discordapp.com/api/guilds/340160225338195969/widget.png" alt="Chat">
+    <img src="https://img.shields.io/discord/1513968811047522396?label=Chat&logo=discord" alt="Chat">
   </a>
   <br>
   <a href="https://www.npmjs.com/package/vuetify">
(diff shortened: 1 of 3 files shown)


J-Sek:
while being small in scope, some devs already complain that Vuetify has too much props. This looks like a problem where `$main-transition` Sass variable should be enough, unless you can explain a use case that requires on/off changes or multiple `v-main` side-by side having different behaviors.

UllernJ:
Honestly just a PoC, dont really have a use case other than that. 

Maybe a good idea to just close this MR and the issue as well?

J-Sek:
It never hurts to see an idea fleshed out into PR 😄 
Thanks for taking interest in improving Vuetify

PR #23024, opened 2026-07-22
fix(VList): resolve keyboard navigation inside an open Shadow DOM

johnleider:
## Problem
VMenu / VList / VSelect keyboard navigation (ArrowUp/Down) breaks when the component renders inside a custom element with an **open Shadow DOM** (`defineCustomElement`). `document.activeElement` returns the shadow **host**, not the focused descendant — so `getNextElement`'s `elements.indexOf(document.activeElement)` returns `-1` and navigation can't advance. Tab behavior is inconsistent for the same reason.

Reported with a repro by @jdalk; root cause confirmed against a live repro (`document.activeElement` === the host, the focused button lives at `host.shadowRoot.activeElement`).

## Fix
Add a shadow-piercing `getActiveElement()` to `util/helpers.ts` that walks `document.activeElement` down through open shadow roots, and use it in `getNextElement` + `focusChild`. Light-DOM behavior is unchanged (nothing to walk). **Open** shadow roots only — closed roots can't be traversed (platform limit).

## v3 patch
The identical bug exists in v3 (`util/helpers.ts` is the same). A ready-to-apply v3 patch + `patch-package` instructions: **https://bin.vuetifyjs.com/bins/Ya05xQ**

## Scope / follow-ups
- Fixes the reported arrow-nav (`getNextElement`) + the `focusChild` containment check.
- For *complete* shadow-DOM focus coverage, the same `getActiveElement()` should also replace raw `document.activeElement` reads in `focusTrap.ts`, `VOverlay` (`ownsFocus`), `useActivator`, `VSelect/useFocusRepair`, and `hotkey` — follow-up.
- **Test:** the jsdom unit env doesn't reliably reproduce shadow-DOM focus retargeting, so a browser-based shadow-DOM nav spec (mirroring `click-outside-shadow-dom.spec.ts`) is the right home for a regression test — recommended follow-up.

### DIFF
--- a/packages/vuetify/src/components/VOverlay/VOverlay.tsx
+++ b/packages/vuetify/src/components/VOverlay/VOverlay.tsx
@@ -39,6 +39,7 @@ import {
   convertToUnit,
   focusableChildren,
   genericComponent,
+  getActiveElement,
   getCurrentInstance,
   getScrollParent,
   IN_BROWSER,
@@ -238,7 +239,7 @@ export const VOverlay = genericComponent<OverlaySlots>()({
 
       if (contentEl.value?._clickOutside?.lastMousedownWasOutside) return
 
-      const activeEl = document.activeElement
+      const activeEl = getActiveElement()
       const focusWasInOverlay =
         ((!activeEl || activeEl === document.body) && openedWithActivatorFocus) ||
         activeEl === el ||
@@ -260,7 +261,7 @@ export const VOverlay = genericComponent<OverlaySlots>()({
 
     watch(isActive, val => {
       if (val) {
-        const activeEl = document.activeElement
+        const activeEl = getActiveElement()
         const el = activatorEl.value
         openedWithActivatorFocus = !!el && (activeEl === el || el.contains(activeEl))
       } else {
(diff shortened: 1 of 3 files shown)


johnleider:
### Applying this fix in Vuetify 3 today (no upgrade / no fork)

For anyone hitting this on v3 before it's released: the same root cause (`document.activeElement` returning the shadow **host** instead of the focused descendant) can be patched at runtime. Drop this in your app entry, **before** Vuetify is used — it makes `document.activeElement` descend through open shadow roots, which is the one seam every internal Vuetify focus read observes:

```js
function installShadowActiveElementFix () {
  const desc = Object.getOwnPropertyDescriptor(Document.prototype, 'activeElement')
  const nativeGet = desc.get
  Object.defineProperty(Document.prototype, 'activeElement', {
    configurable: true,
    enumerable: desc.enumerable,
    get () {
      let el = nativeGet.call(this)
      while (el?.shadowRoot?.activeElement) el = el.shadowRoot.activeElement
      return el
    },
  })
}
installShadowActiveElementFix()
```

Patching `getNextElement`/`focusChild` directly doesn't work — they're ESM module-internal bindings that `VList` calls via its own reference, so an external wrapper can't intercept them. Overriding the native getter does.

**Caveat:** open shadow roots only — a closed root exposes no `.shadowRoot`.

Full write-up (cause, verified before/after arrow-key sequences, and a source-patch diff form for older v3 lines) is on the Bin: https://bin.vuetifyjs.com/bins/raHF1w

J-Sek (approved):

PR #23030, opened 2026-07-23
fix(VStepperWindow/VTouch): only remove touchHandlers when   not null…

Orange-Rabbit:
## Description
VStepperWindow uses the VWindow component internally. VWindow always renders with the VTouch directive, even when touch is disabled (which it is in the VStepperWindow usage).

In my Application, due to a mixture of router-view, suspense and other side-effects, the unmounted hook is called **twice**. The first time, the handler is removed. For the second call the assignment `const handlers = target._touchHandlers[uid]` returns `null` which later, together with the `keys(handler)` call, causes the error `TypeError: Cannot convert undefined or null to object`

This has the potential to break VStepper usage (for me it already does).

The fix makes the unmounted lifecycle hook idempotent, so that it does not fail on subsequent runs, by guarding against the handlers being null/undefined.

## Markdown
Unfortunately I cannot reproduce the issue in the Playground.vue 
I tried many different combinations, but i cannot re-create it in this environment.

### DIFF
--- a/packages/vuetify/src/directives/touch/index.ts
+++ b/packages/vuetify/src/directives/touch/index.ts
@@ -139,11 +139,11 @@ function unmounted (el: HTMLElement, binding: TouchDirectiveBinding) {
   if (!target?._touchHandlers || uid === undefined) return
 
   const handlers = target._touchHandlers[uid]
-
-  keys(handlers).forEach(eventName => {
-    target.removeEventListener(eventName, handlers[eventName])
-  })
-
+  if (handlers) {
+    keys(handlers).forEach(eventName => {
+      target.removeEventListener(eventName, handlers[eventName])
+    })
+  }
   delete target._touchHandlers[uid]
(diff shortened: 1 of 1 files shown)


J-Sek:
In Vue, I already saw that hydration errors can sometimes impact some lifecycle events and lead into "unrelated" errors. Patching them does not really solve the problem. The risk is: we do not get error when we should. This piece of code (prior to your change) assumed that when _touchHandlers exist, the [uid] matches. I will try to walk back from the place you changed and attempt to find a scenario that may cause a mismatch. (edit: did not find one, so leaving comments instead)

J-Sek (approved):

Orange-Rabbit:
@J-Sek sorry, did not get around checking Github before today. I am kind of glad, that it's not just me that was struggling to reproduce it 👀 

Thanks for investigating, and approving my PR!

PR #22768, opened 2026-03-26
refactor(date): delegate to v0 createDate as runtime

johnleider:
## Summary

- Migrate Vuetify's date system to use `@vuetify/v0`'s `createDate()` as the runtime
- Replace `VuetifyDateAdapter` (native Date) with v0's `Vuetify0DateAdapter` (Temporal)
- Add `VuetifyDateBridge<T>` to translate Vuetify's `string|number` param types to v0's `number`
- Add `LegacyDateAdapterCompat<T>` for backwards compat with old custom adapters
- Rewrite `StringDateAdapter` over v0's Temporal adapter
- Consolidate two injection symbols (`DateOptionsSymbol` + `DateAdapterSymbol`) into single `DateSymbol`
- Update `framework.ts` to use `date.install(app)` pattern

Closes vuetifyjs/0#111

## Breaking Changes

- Default date type shifts from `Date` to `Temporal.PlainDateTime`
- Custom adapter authors must implement v0's `DateAdapter<T>` interface (compat shim provided with deprecation warning)
- `DateAdapterSymbol` deprecated — use `useDate()` instead
- Third-party date lib adapters will come from `@vuetify/v0` (once shipped)

Issue #744, opened 2017-06-11
Vuetify with Laravel


T0miii:
Install latest version of laravel 5.4
Install vuetify via npm.
Create a master.blade.php layout file.
Insite this file copy the default layout provided by vuetify. (the blue one with anymated toolbar https://vuetifyjs.com/layout/pre-defined)
Create a Vue instance el: #app

```
<!DOCTYPE html>
<html lang="{{ config('app.locale') }}">
<head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    <meta name="csrf-token" content="{{ csrf_token() }}">

    <title>{{ config('app.name', 'Laravel') }}</title>

    <link href="{{ asset('css/app.css') }}" rel="stylesheet">
    <link href="{{ asset('css/vuetify.min.css') }}" rel="stylesheet">

    <script>
        window.Laravel = {!! json_encode([
            'csrfToken' => csrf_token(),
        ]) !!};
    </script>
</head>
<body>
<v-app id="app" top-toolbar footer v-cloak>
    <v-navigation-drawer persistent light :mini-variant.sync="mini" v-model="drawer">
        <v-list class="pa-0">
            <v-list-item>
                <v-list-tile avatar tag="div">
                    <v-list-tile-avatar>
                        <img src="myimg.jpg" />
                    </v-list-tile-avatar>
                    <v-list-tile-content>
                        <v-list-tile-title>Tamás Barta</v-list-tile-title>
                    </v-list-tile-content>
                    <v-list-tile-action>
                        <v-btn icon @click.native.stop="mini = !mini">
                            <v-icon>chevron_left</v-icon>
                        </v-btn>
                    </v-list-tile-action>
                </v-list-tile>
            </v-list-item>
        </v-list>
        <v-list class="pt-0" dense>
            <v-divider></v-divider>
            <v-list-item v-for="item in items" :key="item">
                <v-list-tile>
                    <v-list-tile-action>
                        <v-icon>@{{ item.icon }}</v-icon>
                    </v-list-tile-action>
                    <v-list-tile-content>
                        <v-list-tile-title><a href="/partners">@{{ item.title }}</a></v-list-tile-title>
                    </v-list-tile-content>
                </v-list-tile>
            </v-list-item>
        </v-list>
    </v-navigation-drawer>
    <v-toolbar fixed class="primary darken-4" light>
[...]

PR #723, opened 2017-06-08
update translatable.js

lerte:
🇨🇳

PR #23039, opened 2026-07-27
feat(util): consume @vuetify/v0 utilities in core

johnleider:
## Description

Starts consuming `@vuetify/v0` in core. Now that v0 v1 is public, this takes the real dependency and adopts what we can through 4.x minors rather than holding it for v5.

This is the utilities slice: type guards, `range`, and the `toHighlight` shim.

> [!IMPORTANT]
> This adds **the first runtime dependency** to the `vuetify` package — it previously shipped with peers only.

## Changes

**`src/util/v0.ts`** — new, the single place core pulls from v0. Re-exports `isArray`, `isBoolean`, `isElement`, `isFunction`, `isNaN`, `isNull`, `isNullOrUndefined`, `isNumber`, `isObject`, `isString`, `isSymbol`, `isThenable`, `isUndefined`, and `range`. Grow this file as more of v0 lands.

**Adoption** — ~170 raw `typeof` / null comparisons across 43 files in `src/util` and `src/composables` now use the guards. Core had no type guards of its own before this.

**Duplicates deleted:**
- `createRange` in `helpers.ts` — now aliases v0's `range`
- `isObject` in `helpers.ts` — now v0's
- `labs/VHighlight/toHighlight.ts` — the ~70-line mirror ("temporary shim, to be replaced in v5.0") is now a re-export. Its existing 15-test spec runs against v0's implementation unchanged and passes, which proves the mirror was faithful.

## Two boundary decisions worth reviewing

**`isObject` is re-widened.** v0 narrows to `Record<string, unknown>`; core narrows to `Record<string, any>`. The stricter value type propagates `unknown` into every downstream property access and broke `mask.ts`, `VTabs.tsx`, `transition.ts` and `directives/tooltip`. `src/util/v0.ts` keeps v0's runtime implementation but restores the wider narrowing. Possibly worth reconciling upstream in v0.

**`range` is exported as `createRange`.** v0's name shadows local `range` variables in VPagination, VRating and VSlider. Renaming call sites produced real compile errors, so core keeps its own name for the import.

## Not adopted, deliberately

| Left alone | Why |
|---|---|
| `clamp` | v0 returns `min` for non-finite input; core returns `max` for `Infinity` and `NaN` for `NaN`. Used in slider/color math — needs a deliberate behavior decision. |
| `mergeDeep` | v0's signature has no `arrayFn`/`targetCondition`, which the defaults system depends on. Needs an upstream v0 change first. |
| `isPrimitive` | v0 drops `bigint`; core's includes it and `deepEqual` relies on that. |
| bare `typeof x === 'object'` | v0's `isObject` also excludes arrays, so a blanket swap changes behavior at 10 sites. Only the two spots already spelling out the full `typeof` + null + array triple were collapsed. |
| `Array.isArray` | TS's built-in narrows better than v0's `value is unknown[]`, which erases element types and forces casts. |

`typeof window === 'undefined'` global-existence checks are untouched.

## Consumer impact

- **Bundle size is a net win.** `dist/vuetify.min.js` **shrinks 918 bytes** (650,900 → 649,982) — calling `isString(x)` compresses better than 82 inlined `typeof x === 'string'` comparisons. The unminified `vuetify.esm.js` grows 7.4 KB, but that is entirely v0's JSDoc, which minification strips.
- **No peer-dependency warnings.** Every exotic v0 peer (`@adobe/leonardo-contrast-colors`, `@ant-design/colors`, `posthog-js`, `launchdarkly-js-client-sdk`, `@flagsmith/flagsmith`, `@js-temporal/polyfill`, `@material/material-color-utilities`, `vue-i18n`) is marked `optional` in v0's `peerDependenciesMeta`. Only `vue` is required, which core already peers on.
- **CDN builds unaffected.** v0 is fully inlined into `dist/` — zero surviving `import`/`require` of `@vuetify/v0` in the esm/umd/cjs bundles.
- **Types resolve.** `lib/util/v0.d.ts` references `@vuetify/v0/utilities` as an external module, resolvable via the new dependency. Public `framework.d.ts` surface is unchanged.

## Follow-ups

- `wrapInArray` is **runtime-identical** to v0's `toArray` — a drop-in, though it needs the same return-type preservation as `isObject`.
- Further overlap not yet touched: `toElement`, `useId`, `useToggleScope`, `useHydration`, `toReactive`.
- `src/components` and `src/labs` still hold ~440 raw checks, left out to keep this reviewable.
- Beyond utilities, the composable-level v0 surface (selection, validation, registries, observers) is the bigger prize.

## Verification

- `eslint src --max-warnings 0` — clean
- `tsgo -p tsconfig.checks.json --noEmit` — clean (confirmed non-vacuous via a deliberate-error smoke test)
- `vitest --project unit` — 804 passed. The 4 failures in `composables/date/__tests__/date.spec.ts` (`getWeek`) are **pre-existing on `dev`** — verified by building the base tree and re-running.
- `rollup --config build/rollup.config.js` — succeeds, before and after.
- Declaration emit + types bundle — succeed.

Two things that look alarming but are **pre-existing on `dev`**, confirmed by a baseline build: the `src/util/index.ts` circular-dependency warning, and labs' `VHighlight`/`toHighlight` appearing in the main bundle.

`pnpm-lock.yaml` also picks up ~10 lines of unrelated `deprecated:` metadata refresh from the registry, which any install regenerates.


packages/vuetify/src/composables/date/adapters/vuetify.ts
```diff
@@ -1,5 +1,5 @@
 // Utilities
-import { consoleWarn, createRange, padStart } from '@/util'
+import { consoleWarn, createRange, isFunction, isNullOrUndefined, isString, isUndefined, padStart } from '@/util'
```
J-Sek: since this file is gonna be removed by #22768, we could skip the changes and save ourselves a conflict. not a big deal either way

packages/vuetify/src/composables/defaults.ts
```diff
@@ -83,7 +84,7 @@ export function provideDefaults (
     }
 
     return properties.prev
-      ? mergeDeep(properties.prev, properties, undefined, (_, v) => v !== undefined)
+      ? mergeDeep(properties.prev, properties, undefined, (_, v) => !isUndefined(v))
```
J-Sek: if we are not going to be bothered by linter rules, I would skip it for the hot paths as it does impact performance.

packages/vuetify/src/labs/VHighlight/toHighlight.ts
```diff
@@ -1,96 +1,3 @@
-// Utilities
-import { toValue } from 'vue'
-import { wrapInArray } from '@/util'
+export { toHighlight } from '@vuetify/v0'
```
J-Sek: I would move it to `VHighlight.tsx` and drop `toHighlight.spec.ts`

packages/vuetify/src/composables/nested/nested.ts
```diff
@@ -176,7 +176,7 @@ export const useNested = (
 
   const activeStrategy = computed(() => {
     if (typeof props.activeStrategy === 'object') return props.activeStrategy
```
J-Sek: is this a good place to put `isObject` or do we strictly keep it for "is object and not array"?

Note: I assume we would add `if (!props.activeStrategy) return null` one line above

J-Sek: It is only an opportunity to improve clarity of the author intent, not much else

J-Sek:
> `src/components` and `src/labs` still hold ~440 raw checks, left out to keep this reviewable.

I wish we could scope by utilities or package, no "up to X files touched", but let's keep what is done for the sake of avoiding another review for the same pieces.
- proceed with 77 places in `packages/vuetify/src` applying `isBoolean`, `isNumber`, `isString`
- proceed with 38 places in `packages/vuetify/src` applying `isFunction`
- proceed with ~86 places in `packages/vuetify/src` applying `isArray`
- apply `isElement` in 2 places calling `instanceof Element`

leaving ~200 places deferred for null checks would give us a chance of ensuring those do not land on a hot path and we can also replace them with implicit checks where it makes sense.

`isNaN` is a tricky thing to apply without a linting rule. Another PR maybe.

johnleider:
Scoping looks right — batching by guard rather than by file keeps each pass mechanically reviewable. Go ahead on three of the four:

- ✅ `isBoolean` / `isNumber` / `isString` — 77 sites
- ✅ `isFunction` — 38 sites
- ✅ `isElement` — 2 sites replacing `instanceof Element`

**Hold the `isArray` batch (~86 sites)** — that one is blocked on v0, not on review appetite. It's in this PR's "not adopted" table for a concrete reason: `isArray (item: unknown): item is unknown[]` discards the element type, so `isArray(xs)` on a `string[] | string` narrows to `unknown[]` where `Array.isArray(xs)` gives you `string[]`. Converting those 86 sites today buys 86 casts.

That's a real defect in v0 rather than a style preference, and it's now being fixed there — tracked as a sibling to [vuetifyjs/0#723](https://github.com/vuetifyjs/0/issues/723). Once `isArray` preserves the element type, this batch becomes a clean mechanical pass with no casts. Worth waiting for.

Agreed on deferring the ~200 null checks, and there's a second reason beyond hot paths: [vuetifyjs/0#723](https://github.com/vuetifyjs/0/issues/723) is open on exactly that surface — `isObject` narrows to `Record<string, unknown>`, which structurally cannot match interface-typed values, because TypeScript never grants interfaces an implicit index signature. That's why `src/util/v0.ts` re-widens it here. Mass-applying the null-ish guards before that settles risks a second rewrite.

Agreed on `isNaN` needing a lint rule first. It shadows the global and the two disagree on strings — `isNaN('x')` is `true`, `Number.isNaN('x')` is `false` — so a mechanical swap is the one in this set that can change behavior silently.

johnleider:
Correction to my previous comment, and good news attached.

**My example was wrong.** I said `isArray(xs)` on a `string[] | string` narrows to `unknown[]` where `Array.isArray` gives `string[]`. That's false — a type predicate narrows a union by *filtering* it, and `string[]` is assignable to `unknown[]`, so the constituent survives intact. Verified against tsc 6.0.3: that case has always worked. Apologies for the bad example.

The defect is real but narrower than I described. Three actual failure modes:

1. **`readonly` unions** — `readonly string[] | string` narrows to `(string | readonly string[]) & unknown[]`, which is not an array of anything, and the `else` branch fails to subtract the array constituent at all.
2. **Arrays with `any` elements** — `Record<string, any>[] | Record<string, any>` narrows to `unknown[] | Record<string, any>[]`, so `c[0]` is `unknown`. This is the real "erases element types" case, and it's the one that's pervasive in core.
3. **`any` input** → `unknown[]` instead of `any[]`, which kills contextual typing on callbacks.

**Fixed in [vuetifyjs/0#745](https://github.com/vuetifyjs/0/pull/745)** (issue [#744](https://github.com/vuetifyjs/0/issues/744)), non-breaking, no `any` spent. It was measured rather than argued: all 87 `Array.isArray` sites here were converted by codemod against a locally built v0 and typechecked per candidate signature. The status quo produces **6** errors across those sites; the shipped signature produces **1**.

**That last error is worth your attention, because it's a pre-existing bug this surfaces:**

```
src/util/helpers.ts(87,54): error TS4104: The type 'readonly (string | number)[]'
  is 'readonly' and cannot be assigned to the mutable type '(string | number)[]'.
```

`SelectItemKey` is `readonly (string | number)[]`, but `getNestedValue` takes `path: (string | number)[]`. Core has been passing a readonly array into a mutable parameter — both the old guard and `Array.isArray` hide it, for different reasons. Widening that parameter to `readonly (string | number)[]` is a one-word change, and with it all 87 sites typecheck clean.

So the `isArray` batch is unblocked once #745 lands and v0 republishes.

J-Sek:
Thoughts on adding `.git-blame-ignore-revs` right after we get merge commit hash?

J-Sek:
Accidentally covered `src/util/helpers.ts(87,54)` problem by adding missing `readonly` to `getNestedValue` param. Does not matter :)

johnleider:
@J-Sek

Yes on `.git-blame-ignore-revs` — this is exactly the kind of mechanical mass rewrite it's for. Repo doesn't have the file yet, so first entry goes in with the merge commit hash right after it lands.

Thanks for the `getNestedValue` `readonly` — that was the one remaining typecheck failure the `isArray` batch would have surfaced. Good it's already in.

Migration commit looks complete. Status on the v0 blockers from earlier:

- **isArray** ([vuetifyjs/0#745](https://github.com/vuetifyjs/0/pull/745)) — merged. Not yet published (npm latest is still `1.0.1` from before it). Once a patch cuts, `^1.0.1` picks it up.
- **isObject** ([vuetifyjs/0#746](https://github.com/vuetifyjs/0/pull/746) / [#723](https://github.com/vuetifyjs/0/issues/723)) — still open. The local re-widen in `src/util/v0.ts` stays until that lands; no action needed here.

Anything else you want eyes on before merge, or are we good?

J-Sek:
I have reviewed everything we have so far. Left 4 comments already.

johnleider:
@J-Sek **`@vuetify/v0@1.0.2` is live** ([npm](https://www.npmjs.com/package/@vuetify/v0/v/1.0.2) · [release](https://github.com/vuetifyjs/0/releases/tag/v1.0.2)).

Both upstream blockers from earlier are in that cut:

- **isArray** — [vuetifyjs/0#745](https://github.com/vuetifyjs/0/pull/745) preserves element types / tuples / readonly
- **isObject** — [vuetifyjs/0#746](https://github.com/vuetifyjs/0/pull/746) narrows to `Record<string, any>` (closes [#723](https://github.com/vuetifyjs/0/issues/723))

This PR already depends on `^1.0.1`, so a lockfile refresh picks up `1.0.2`. Once that lands, the local re-widen in `src/util/v0.ts` for `isObject` is optional cleanup (runtime was always the same; types now match upstream). The `isArray` batch is unblocked if anything is still on `Array.isArray`.

Will look at the 4 review comments you left next if they aren't already resolved.

packages/vuetify/src/composables/date/adapters/vuetify.ts
johnleider: Done — reverted this file so #22768 can remove it cleanly.

packages/vuetify/src/composables/defaults.ts
johnleider: Agreed. Reverted the hot paths in `defaults.ts`: the `mergeDeep` callback and the proxy `get` trap both keep raw `!== undefined` / `!= null`. Left `isString(root)` — that is not on the per-prop path.

packages/vuetify/src/labs/VHighlight/toHighlight.ts
johnleider: Done — `VHighlight.tsx` imports `toHighlight` from `@vuetify/v0` directly. Dropped `toHighlight.ts` and the unit mirror (`toHighlight.spec.ts`); it was only proving the old shim was faithful, which 1.0.2 already ships.

packages/vuetify/src/composables/nested/nested.ts
johnleider: Kept `typeof === 'object'` here on purpose.

`isObject` is the right runtime for "plain object, not array/null", but its type predicate (`item is Record<string, any>`) is too wide for this `string | function | strategy-object` union — after `if (isObject(...))`, the `isFunction` branch no longer narrows to a callable (TS2349). `typeof` discriminates the union cleanly.

Also no need for `if (!props.activeStrategy) return null` above: `undefined` falls through to the `default` case and gets the mandatory-aware built-in strategy, which is the intended behavior.

johnleider:
Addressed the open review threads + the 1.0.2 bump in 6fa78c5:

| Thread | Action |
|--------|--------|
| `date/adapters/vuetify.ts` | Reverted (defers to #22768) |
| `defaults.ts` hot paths | Raw `!== undefined` / `!= null` restored on mergeDeep + proxy get |
| `toHighlight` shim | Import from `@vuetify/v0` in `VHighlight.tsx`; deleted shim + unit mirror |
| `nested` `isObject` | Kept `typeof === 'object'` — see thread for why the predicate breaks the function branch |

Also:
- `@vuetify/v0` → `^1.0.2` (lockfile surgical)
- dropped the local `isObject` re-widen in `src/util/v0.ts` — 1.0.2 already returns `Record<string, any>`
- `@vuetify/v0` added to `minimumReleaseAgeExclude` so first-party cuts install without waiting out the 3-day age gate

packages/vuetify/src/composables/nested/nested.ts
johnleider: Correction — it *does* drop in. My earlier read was wrong.

The failure was order, not the 1.0.2 predicate. `isObject` first left a residual union that made the `isFunction` branch uncallable (TS2349). Flip it:

```ts
if (isFunction(props.activeStrategy)) return props.activeStrategy(props.mandatory)
if (isObject(props.activeStrategy)) return props.activeStrategy
```

…and typecheck is clean on `1.0.2`. That *is* what the isObject widen was for — plain-object strategy configs without the `typeof === 'object'` null/array quirk. Pushed.

johnleider:
Swept the remaining production `typeof === 'object'` sites (22 files).

**Converted to `isObject`:** options bags (legend/tooltip/animation/reveal/transition/ssr), calendar category leftovers, color utils / VColorPicker, filter items, VDataTable `selectStrategy`, VImg `src`, sparkline/pie/heatmap.

**Directives (function | object unions):** use `isFunction` first, then object field access. Runtime `isObject` is correct for functions, but the type predicate `item is Record<string, any>` cannot exclude callables — TS treats functions as assignable to `Record`. Same ordering rule as the nested strategy fix.

**Left alone:** `helpers.ts` plain-object probe + vnode walk (implementing object checks / VNode trees, not consumer option bags).

`tsgo -p tsconfig.checks.json --noEmit` clean. Production `typeof === 'object'` is down to those two helpers lines.

packages/vuetify/src/components/VDataTable/composables/select.ts
```diff
@@ -119,7 +119,7 @@ export function provideSelection (
   const currentPageSelectable = computed(() => toValue(currentPage).filter(item => item.selectable))
 
   const selectStrategy = computed(() => {
-    if (typeof props.selectStrategy === 'object') return props.selectStrategy
+    if (isObject(props.selectStrategy)) return props.selectStrategy
```
J-Sek, line 122: without `if (isNull(props.selectStrategy)) return null` it changes the behavior and computed lambda returns fallback `pageSelectStrategy` instead. Is this correct/intended?

packages/vuetify/src/components/VImg/VImg.tsx
```diff
@@ -131,7 +132,7 @@ export const VImg = genericComponent<VImgSlots>()({
     let deferredLoadEmit = false
 
     const normalisedSrc = computed<srcObject>(() => {
-      return props.src && typeof props.src === 'object'
+      return props.src && isObject(props.src)
```
J-Sek: should be possible to fold that first part. Nitpick

packages/vuetify/src/components/VDataTable/composables/select.ts
J-Sek: 3 places like this one can be found in packages/vuetify/src/composables/nested/nested.ts

J-Sek:
Reviewed all files, leaving just 2 comments left to resolve

packages/vuetify/src/components/VImg/VImg.tsx
johnleider: Folded — `isObject` already rejects `null`/`undefined`/`''`, so the truthiness guard was dead weight from the `typeof null === 'object'` era. Pushed in 467c453959.

packages/vuetify/src/components/VDataTable/composables/select.ts
johnleider: Intentional — and applies to all 4 sites (here + the 3 in nested.ts). The behavior change only triggers on an explicit `:select-strategy="null"` (Vue skips the prop default for `null`). Previously that returned `null` from the computed, and every consumer dereferences it immediately — `selectStrategy.value.select(...)`, `showSelectAll` during header render, and in nested.ts the model transforms call `.in(...)` on mount — so the old "behavior" for `null` was a guaranteed TypeError, not a usable state. Falling through to the default strategy is strictly more robust.

Restoring `if (isNull(...)) return null` would also widen the computeds to `| null` and break typecheck at every consumer, forcing assertions to preserve a crash path. Leaving it as-is.

J-Sek (approved):

PR #23047, opened 2026-07-30
feat(VMonthPicker): add `fluid` prop for responsive design and update documentation

### LINKED ISSUE #23044: [Feature Request] Support scrolling/touch interactions in VMonthPicker
### Problem to solve
In current version of VMonthPicker it uses a fixed width about ~300px, and i want to show a row of months.
If you set 12 cols for `VMonthPicker` months and set `class="w-100"`, it displays a row of months, but is cutted when the parent width is too short, and is non scrollable.
See the following image.

Image

Template vue
```vue
<v-month-picker
   v-model="selectedMonths"
   hide-header
   hide-title
   :months-columns="12"
   multiple="range"
   class="w-100"
   @update:model-value="searchReservations">
</v-month-picker>
```

### Proposed solution
Set width to 100% to the month picker root, to display the full row of months.
In months, set width to 100% and overflow x to auto.
In months, remove flex-grow-1 (optional).

Translated, this custom style works as solution, so I propose to add a property maybe called 'fill' or 'fluid' or other name to apply this config automatically. This is useful to display a month selector in top of the page. In my case to filter reservations from january to august for example.

Maybe there is a better solution, perhaps the month grid could adapt to the size of parent and wrap the months in multiple rows if the width is not enough instead of using scrollbars, so the user don't need to scroll.

```vue
<template>
  <v-month-picker
     v-model="selectedMonths"
     hide-header
     hide-title
     :months-columns="12"
     multiple="range"
     class="month-picker-fill"
     @update:model-value="searchReservations">
  </v-month-picker>
</template>

<style>
.month-picker-fill.v-picker.v-month-picker {
    width: 100%;
    .v-month-picker__months {
        overflow-x: auto;
        width: 100%;
    }
}
</style>
```

Image

it still adapts to the cols system.

Image

nquangtrung:
## Description

Add fluid prop along with documentation and examples

resolves #23044

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        fluid>
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        class="w-100">
      </v-month-picker>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```


J-Sek:
It definitely improves the experience.

My thoughts after playing with the playground:
- would it make sense to make it "default" rather than opt-in via dedicated prop?
- (mobile) does not show bottom scrollbar, so on certain widths it might look like picker shows subset of months and it there is no indicator suggesting scrollable area.. introducing icons (like those in control area) might be a good choice if we can make them slightly smaller
- (desktop) when picker is wide enough to show all columns without bottom scrollbar, it will briefly show (expanding height) when changing years
- when opening with active selection that happens to be off-screen, we could automatically adjust the scroll (similar to how we handle it for years picker)

nquangtrung:
Thank you, I appreciate your feedback.

>- (mobile) does not show bottom scrollbar, so on certain widths it might look like picker shows subset of months and it there is no indicator suggesting scrollable area.. introducing icons (like those in control area) might be a good choice if we can make them slightly smaller
>- (desktop) when picker is wide enough to show all columns without bottom scrollbar, it will briefly show (expanding height) when changing years

To solve this, I basically hide the scrollbar, and add 2 small nav buttons on 2 sides if the view is scrollable. These would be hidden if not.
Since the scrollbar takes up different amount of space in different devices, hiding it would be better to make the component behave the same across devices. Do you think it makes sense as I approached it this way?

Updated Playground.vue

```vue
<template>
  <v-app>
    <v-container>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        class="w-50">
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="6"
        class="w-50">
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        class="w-100">
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="6"
        class="w-100">
      </v-month-picker>
      <!-- First selected month should be scrolled into view -->
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        model-value="2026-06"
        class="w-50">
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="12"
        multiple
        :model-value="['2026-10', '2026-06']"
        class="w-50">
      </v-month-picker>
      <v-month-picker
        hide-header
        hide-title
        :months-columns="6"
        multiple
        :model-value="['2026-10', '2026-06']"
        class="w-25">
      </v-month-picker>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>

```

J-Sek:
To be honest I barely recognize the component 😄
- it would need a resize observer, to show/hide additional buttons without  the need of hard refresh
- the vertical scrollbar is missing for `:month-columns="1"` and whenever user would constraint the picker height
- the keyboard navigation is something that would need a final polish - it is not trivial

I'd rather suggest we stop the experiment and look at it from the distance. Maybe there is something a slot could help with. If you have a specific design in mind or could point me to a component that supports this kind of view in any other framework or platform, I am interested in seeing any reference.

nquangtrung:
After quickly looking at some other frameworks for inspiration, my impression is that they all have the same problem with limited width space, which is a bit disappointing.
One potential solution might be giving the user access to the internal months grid similar to Reka UI:
```vue
<v-month-picker v-slot="{ grid }">
  <v-month-picker-grid class="...">
    <v-month-picker-grid-cell class="..." v-for="month in grid.months">...</v-month-picker-grid-cell>
  </v-month-picker-grid>
</v-month-picker>
```
Then I guess the user can have more freedom with styling and how they can make it fit their own use-case.

docpelux:
mmm in current status it cuts the months and is not possible to reach all the buttons.

can we have a prop to limt the number of year rows? or disable the year button completely and allow to change the year with the arrows only? it will be useful for compact views to avoid stretching years in all the page.

or maybe we can introduce a custom slot for the year header so user can customize it (for example replacing by a v-select).
```html
<v-month-picker v-model="date">
    <template #years>
        <v-select v-model="year" @update:model-value="updateDateWithNewYear" />
    </template>
<v-month-picker/>
```

Schermata del 2026-08-06 10-40-31
Schermata del 2026-08-06 10-40-51

PR #23061, opened 2026-08-02
fix(VBtn): set explicit tabindex=0 to restore keyboard focus on Safari

### LINKED ISSUE #21905: [Bug Report][3.9.4] V-Toolbar button tabbing and keyboard focus broken on Safari
### Environment
**Vuetify Version:** 3.9.4
**Vue Version:** 3.5.18
**Browsers:** Safari
**OS:** Mac OS 10.15.7

### Steps to reproduce
From Safari try to tab to the icons, notice the tab key skips over them and the user can't 'focus' them via the keyboard.

### Expected Behavior
The user can navigate from button to button via the keyboard's tab key.

### Actual Behavior
Pressing the tab key the user is taken to the browsers address field.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqNkUuOwjAQRK9i9TpOmJ80gzIjzR3YERYmboglx7acJgIh7k7nQ0QQC3burqonu7w+QxPL7D+EtD0gLCEnrINVhH+FEyJvZamiFhWafUW/BbwvFgX0Ui+S93arosAjodOoR+Vek2TIoiC2cH41LJmRZw8Wzk7p2yVEKxvraalCYP5E7/lbcsKU3jG11kbWau/M7jSSWbzjPfNXqCK97NaeGtliJFMqO0uNiWzWW7eYnjc0yXNXJQ8zL1wS+Ei/0rdv6A4/6SckO2UbTGYfM+w2V2EfiPY=)

### Other comments
The example code attached is from the Vuetify website / Toolbars / Extended section
https://vuetifyjs.com/en/components/toolbars/#extended

waterWang:
## Description

Fixes #21905

Safari requires an explicit `tabindex` attribute on `<button>` elements to include them in sequential keyboard navigation. Without it, toolbar buttons (VBtn rendered inside VToolbar) are skipped when tabbing, making the toolbar inaccessible via keyboard on Safari.

## Root Cause

The VBtn component sets `tabindex` to `undefined` when the button is not loading or readonly. For `<button>` elements, this means the `tabindex` attribute is not rendered in the DOM. While Chrome and Firefox correctly handle this by defaulting to `tabindex="0"`, Safari requires the explicit attribute to include the button in the tab order.

## Fix

Changed `tabindex={ props.loading || props.readonly ? -1 : undefined }` to `tabindex={ props.loading || props.readonly ? -1 : 0 }`.

This is a safe change — for `<button>` elements, `tabindex="0"` is the default behavior. Making it explicit only fixes the Safari edge case without changing behavior on other browsers.

## Testing

- ✅ Icon buttons inside VToolbar (append/prepend/default slots) are keyboard-focusable on Safari
- ✅ Loading buttons still have `tabindex="-1"` (not focusable)
- ✅ Readonly buttons still have `tabindex="-1"` (not focusable)
- ✅ Disabled buttons still have `disabled` attribute (not focusable regardless of tabindex)
- ✅ Non-interactive non-button elements (VCard, VChip) unchanged

### DIFF
--- a/packages/vuetify/src/components/VBtn/VBtn.tsx
+++ b/packages/vuetify/src/components/VBtn/VBtn.tsx
@@ -228,7 +228,7 @@ export const VBtn = genericComponent<VBtnSlots>()({
           ]}
           aria-busy={ props.loading ? true : undefined }
           disabled={ (isDisabled.value && Tag !== 'a') || undefined }
-          tabindex={ props.loading || props.readonly ? -1 : undefined }
+          tabindex={ props.loading || props.readonly ? -1 : 0 }
           onClick={ onClick }
           value={ valueAttr.value }
         >


J-Sek:
fabricated facts about Safari

PR #23067, opened 2026-08-03
fix(VList): properly trigger reactivity when updating disabled state

### LINKED ISSUE #21883: [Bug Report][3.9.0] List items are not interactive in classic mode if they are disabled on mount.
### Environment
**Vuetify Version:** 3.9.0
**Last working version:** 3.8.12
**Vue Version:** 3.5.13
**Browsers:** Chrome 138.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Steps to reproduce the issue:
Create a <v-list> with the select-strategy="classic" prop.
Initially set the disabled prop on the list items to true.
Later, switch the disabled prop to false.
Visually, the items appear enabled, but clicking on them does not trigger selection.
It seems like the component still treats all items as disabled internally, even though they’re no longer visually disabled.

### Expected Behavior
Elements should be selectable. 

### Actual Behavior
Items are not selectable

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqFU8tu2zAQ/JUtUSAtYEp2XLWWKwcp0EvPPVo+SCRVs6EeICklhaB/75KyHgEC5ETtzuxyZ7g698RoFv5omqBrBTmSxIqyUZkVD2kFkHSUZZoDU5kxp5SULzRrbZ0SKLMX+iy5vWJ2v92mxPN9RW4reGRKsifEuDRZrgSHE3yYvpEMv5+lZVeYUkno66YmXHYPfQ9CiVJUFoYhCV1qvkJJY8cA4GiQxqzgeNutICUzKFGPm9yfS96FtMtUKxyGE03A2Iwaq9GDP/8Q9dolWyha2FZXtM7/InNK1o2o1lVlq6xslFiNMmldubKgj23DsfYNMejcR9GtVN1smI2gTgx0tKj1TSjIyitEwXCcVWIicFLfmQQvQOs9u8pK4c1fXTTRpse4YeO6YOD2BYMkXC0ShoZp2Vi017aN58qyqbWFHu0sYIBC1yXc4Q7efXcoqytj5/c/OdKn8+UzYgu6Wi0HW92K1wRvAZzGQc89jC99xJNdBXv69TMlGwycyjFtBVbt0KFhg4Mt9GIbF/tD9JWKXf6NfolzTvPDfUQ5PxTZIYq3RRS7Xq7T3Ofe9bngQEk4ikfZZNiQfRAFuz1xH3GAZ5EpIzavfsMxd/kPchYmDQ==)

waterWang:
## Description

When using the classic select strategy, list items that are initially disabled are added to the `disabled` Set. When later re-enabled via `updateDisabled()`, the item was removed from the Set using `Set.delete()`, but since `disabled` is a `shallowRef`, mutating the Set in place does not trigger Vue's reactivity system. This means any component or select strategy relying on the `disabled` set doesn't see the change, and the item remains non-interactive.

## Root Cause

In `packages/vuetify/src/composables/nested/nested.ts`, the `updateDisabled` function:

```typescript
updateDisabled: (id, isDisabled) => {
    if (isDisabled) {
      disabled.value.add(id)  // Mutates the Set in place — no reactivity trigger
    } else {
      disabled.value.delete(id)  // Same issue
    }
}
```

`disabled` is a `shallowRef(new Set())`. Using `.add()`/`.delete()` on the Set mutates it in place, but `shallowRef` only triggers when the reference (`disabled.value`) changes, not when the internal state of the Set changes.

## Fix

Create a new Set on every update, so the `shallowRef` reference changes and triggers reactivity:

```typescript
updateDisabled: (id, isDisabled) => {
    const newDisabled = new Set(disabled.value)
    if (isDisabled) newDisabled.add(id)
    else newDisabled.delete(id)
    disabled.value = newDisabled  // New reference → triggers reactivity
}
```

Closes #21883

### DIFF
--- a/packages/vuetify/src/composables/nested/nested.ts
+++ b/packages/vuetify/src/composables/nested/nested.ts
@@ -383,17 +383,13 @@ export const useNested = (
         itemsUpdatePropagation()
       },
       updateDisabled: (id, isDisabled) => {
+        const newDisabled = new Set(disabled.value)
         if (isDisabled) {
-          disabled.value.add(id)
+          newDisabled.add(id)
         } else {
-          disabled.value.delete(id)
+          newDisabled.delete(id)
         }
-        // classic selection requires refresh to re-evaluate on/off/indeterminate but
-        // currently it is only run for selection interactions, so it will set new disabled
-        // to "off" and the visual state becomes out of sync
-        // -- selected.value = new Map(selected.value)
-        // it is not clear if the framework should un-select when disabled changed to true
-        // more discussion is needed
+        disabled.value = newDisabled
       },
       open: (id, value, event) => {
         vm.emit('click:open', { id, value, path: getPath(id), event })


J-Sek:
It takes 30 seconds to verify the problem using the VPlay reproduction link

PR #23069, opened 2026-08-03
fix(VField): correct outline label position for tile variant

### LINKED ISSUE #22269: [Bug Report][3.10.8] Outline VTextField Label position
### Environment
**Vuetify Version:** 3.10.8
**Vue Version:** 3.5.21

### Steps to reproduce
1. Open the link

### Expected Behavior
The label should be a little bit more to the left on Outlined Tile VTextField

### Actual Behavior
The label appears moved to the right

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNrFkc9OhDAQxl9l7AVN+OO6MTEb1sSbB4/Gi3jowiA1pTTtQNYQ3t1SQFhfwFPn+33TzJeZ955ZkyddiyTK7/jLsgMTtW4MQQ+5QU74NnkwQGmaGoK5N8hUpvDsW/NGWYLZgOPlx+s+UwBUYY0H8DVAgSVvJb1OMJDis6IgnLwkIUS0VI5ycHC4yRQLfc4nrWM3xoVMCWst3ZTHsS3tIq61L71wgYgLhWZGHhKeKSoFymKBAHVToIw6Lls8ZuxZXGVsNSU/oXT4ZXy3RseN4Iqc1bQk3Zxi61aiwKhAF0DalZKQuKhkjXUyW/W/IZcgafJ3gyOZFpy66/wung0h28f38d3O3Wcf727jBxaWXFq8PNfEPn4ArlS/fw==)

waterWang:
## Description

Fixes #22269

The `[class^="rounded-"]` CSS selector was also matching the `rounded-0` class (applied by the `tile` prop), which caused the outline `__start` element to get a wider `flex-basis` intended for rounded corners.

## Root Cause

In `VField.sass`, the outline `__start` element has these rules:

1. Default: `flex: 0 0 $field-control-affixed-padding` (12px)
2. Override for rounded: `flex-basis: calc(var(--v-input-control-height) / 2 + 2px)` (≈30px)

The override selector `[class^="rounded-"]` matches `rounded-0` (tile), causing the `__start` to be 30px wide when it should be 12px. This shifts the notch and label to the right.

## Fix

Added an explicit override for `.rounded-0` to restore the correct `$field-control-affixed-padding` (12px) `flex-basis`, so the label properly aligns with the outline border on tiled fields.

## Reproduction

```vue
<template>
  <v-app>
    <v-container>
      <v-text-field
        model-value="Hi!"
        label="Label"
        variant="outlined"
        hide-details
        tile
      />
      <v-text-field
        model-value="Hi!"
        label="Label"
        variant="outlined"
        hide-details
      />
    </v-container>
  </v-app>
</template>
```

The first field (tiled) has the label slightly misaligned (too far right). After the fix, both fields have correctly aligned labels.

### DIFF
--- a/packages/vuetify/src/components/VField/VField.sass
+++ b/packages/vuetify/src/components/VField/VField.sass
@@ -400,6 +400,11 @@
           #{selector.append('[class*=" rounded-"]', &)}
             flex-basis: calc(var(--v-input-control-height) / 2 + 2px)
 
+        @at-root
+          #{selector.append('.rounded-0', &)},
+          #{selector.append('[class*=" rounded-0"]', &)}
+            flex-basis: $field-control-affixed-padding
+
         @at-root #{selector.append('.v-field--reverse', &)}
           border-start-start-radius: 0
           border-start-end-radius: inherit


J-Sek:
Thank you for the PR. Unfortunately, utilities leaking into component is not the direction I want to take here.

PR #23079, opened 2026-08-03
fix(VTooltip): close tooltip when activator moves on click

### LINKED ISSUE #21915: [Bug Report][3.9.5] VTooltip does not close after activator position changes
### Environment
**Vuetify Version:** 3.9.5
**Last working version:** 2.7.2
**Vue Version:** 3.5.18
**Browsers:** Google Chrome
**OS:** Windows

### Steps to reproduce
1. Set up test case where tooltip activator moves out from under cursor on click.
2. Hover over activator, see tooltip.
3. Click activator.
4. Tooltip does not close even though cursor is no longer over the activator.

### Expected Behavior
Tooltip should close if activator moves out from under the cursor.

### Actual Behavior
Tooltip does not close.

### Reproduction Link
[https://github.com/simshaun/vuetify-v3-vtooltip-closing-issue](https://github.com/simshaun/vuetify-v3-vtooltip-closing-issue)

### Other comments
This seems to affect Chromium-based browsers. Tested buggy functionality in Chrome and Microsoft Edge.

It works fine in Firefox.

waterWang:
## Description

Fixes #21915

Chromium-based browsers don't fire a `mouseleave` event when an element moves under the cursor. When a tooltip activator has a click handler that changes its position (e.g. moving it out from under the cursor), the tooltip stays open because the `mouseleave` event never fires in Chromium.

This fix adds a click handler (`onClickHover`) that checks if the cursor is still over the activator element after the click event. If the activator has moved away from the cursor, the tooltip is closed.

## Root Cause

In `useActivator.tsx`, the `onMouseleave` handler is responsible for closing the tooltip when the cursor leaves the activator. However, when an element's position changes (e.g. due to a click handler), Chromium-based browsers do not fire `mouseleave` on the element. Firefox correctly fires the event in this scenario.

## Solution

When `openOnHover` is true and `openOnClick` is false (the VTooltip default), a new `onClickHover` handler is registered. After the click event, using `nextTick` to wait for DOM updates, it checks whether the cursor position (captured during the click) is still within the activator's bounding rect. If not, `isHovered` is set to `false` and `runCloseDelay` is called.

## Testing

- Verified TypeScript compilation passes
- Existing unit tests pass (62 test files passed, pre-existing failures in locale/composables tests unrelated to this change)
- Browser tests require Playwright which is not installed in this environment

### DIFF
--- a/packages/vuetify/src/components/VOverlay/useActivator.tsx
+++ b/packages/vuetify/src/components/VOverlay/useActivator.tsx
@@ -122,6 +122,22 @@ export function useActivator (
       }
       isActive.value = !isActive.value
     },
+    onClickHover: (e: MouseEvent) => {
+      e.stopPropagation()
+      nextTick(() => {
+        const el = e.currentTarget as HTMLElement
+        if (el && isActive.value) {
+          const rect = el.getBoundingClientRect()
+          if (
+            e.clientX < rect.left || e.clientX > rect.right ||
+            e.clientY < rect.top || e.clientY > rect.bottom
+          ) {
+            isHovered = false
+            runCloseDelay()
+          }
+        }
+      })
+    },
     onMouseenter: (e: MouseEvent) => {
       isHovered = true
       activatorEl.value = (e.currentTarget || e.target) as HTMLElement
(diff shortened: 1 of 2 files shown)


J-Sek:
It is not that simple. Also, does not work against the repro from the linked issue

PR #23080, opened 2026-08-03
fix(VOverlay): use clientWidth when visualViewport is zoomed

### LINKED ISSUE #22124: [Bug Report][3.10.3] V-Overlay in MobileDevices with Mode "as DesktopWindow" is positioned incorrectly
### Environment
**Vuetify Version:** 3.10.3
**Vue Version:** 3.5.13
**OS:** Android

### Steps to reproduce
1. Create a vueapp direct for <body> to fit the hole Page.
2. use a <v-select>
3. use this app with android->chrome->in mode "desktopview"

### Expected Behavior
The “new Box()” on document.documentElement fits to the entire document.

### Actual Behavior
The “new Box()” on document.documentElement does not fit the entire document. It is additionally manipulated with visualViewport.scale, which always makes it appear too small.

As a result, the selection is recalculated several times due to overFlow and appears in the wrong place.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNp1k02L2zAQhv/KYCiGEn80iw913cCypT220EIPmz0o9thRkSUhyW6CyX/vWFonTmlP1sw8M/O+wnqeImvq7FHrdBwwKqPKYa8Fc7jbS4BqTBrOhOqgZ6fkN2/c8eM+KvJ8H/k6EQsPY2KFciWrHR+ZU4bACbRR2pZwTX6bY7hc2/2Kg5NLBDTmwGVDzfc9++iG1Er48XYwLasxGZnhTLo14vDkiPiqUcIn72BdfW0goCXpt8quyryaxVu2uoz/2G2wZYOYR03A7eOsGe/9NXyEWjBrialROjSrqvdvUWDtbikAwQ4oiP8eKivpACUnDfOw5/iJCd4qIzmLNxA/zdfCGjWfP9ORNz79BZXpAvEDT8zOh59n1XPZxS/r0d590HJTn5H8f13HHC4/B4V3tcq6swhQGhzDFEY03BJ0LqEVePoQcr8G63h7TmpFpHQlhJbXKjnsZOIt/1U5Iu+OhG/zXIdZl/XGHbxdtl7/3RKK/M2CVtkis7K14dqBRTdoL5v3WhkHExhs4QKtUT3E9EBi3+Zp4qLLJnpIizSP5u+7PH1IemZpebrNt0WSv0+2RbRpmbC4uXtmIffyB6tgGG4=)

Unfortunately, I can't reproduce this with the tool, as document.documentElement is never used as an overlay.

### Other comments
I suspect that the problem is caused by manipulating the visualViewport.scale. If you remove this, everything seems to work as desired.

waterWang:
Fixes #22124

## Problem
When using VOverlay on Android Chrome in desktop view mode with pinch zoom, the `getElementBox` function for `document.documentElement` could produce incorrect layout viewport dimensions. The `visualViewport.width * visualViewport.scale` calculation was unreliable for determining the actual layout viewport size when zoomed, causing the overlay to be positioned incorrectly and triggering unnecessary overflow recalculations.

## Root Cause
The `getElementBox` function in `packages/vuetify/src/util/box.ts` used `visualViewport.width * visualViewport.scale / pageScale` for the viewport dimensions regardless of whether the page was zoomed. When `visualViewport.scale > 1` (pinch zoomed) or on WebKit browsers, the calculation could produce values that didn't match the actual layout viewport, making the overlay appear too small and in the wrong position.

## Fix
When `visualViewport.scale > 1` or `IS_WEBKIT` is true, use `document.documentElement.clientWidth` and `clientHeight` directly instead of the visualViewport-derived calculation. These properties always return the correct layout viewport dimensions regardless of zoom level or browser type.

### DIFF
--- a/packages/vuetify/src/util/box.ts
+++ b/packages/vuetify/src/util/box.ts
@@ -70,11 +70,16 @@ export function getElementBox (el: HTMLElement) {
       })
     } else {
       const pageScale = document.body.currentCSSZoom ?? 1
+      const isZoomed = visualViewport.scale > 1 || IS_WEBKIT
       return new Box({
-        x: visualViewport.scale > 1 || IS_WEBKIT ? 0 : visualViewport.offsetLeft,
-        y: visualViewport.scale > 1 || IS_WEBKIT ? 0 : visualViewport.offsetTop,
-        width: visualViewport.width * visualViewport.scale / pageScale,
-        height: visualViewport.height * visualViewport.scale / pageScale,
+        x: isZoomed ? 0 : visualViewport.offsetLeft,
+        y: isZoomed ? 0 : visualViewport.offsetTop,
+        width: isZoomed
+          ? document.documentElement.clientWidth
+          : visualViewport.width * visualViewport.scale / pageScale,
+        height: isZoomed
+          ? document.documentElement.clientHeight
+          : visualViewport.height * visualViewport.scale / pageScale,
       })
     }
   } else {


J-Sek:
Did not work – as `visualViewport.scale` is apparently `1`.

PR #23082, opened 2026-08-03
fix(VSelectionControl): remove redundant aria-label when visible label element is present

### LINKED ISSUE #21928: [Bug Report][3.6.13] v-radio-group inline screenreader reads labels letter by letter
### Environment
**Vuetify Version:** 3.6.13
**Vue Version:** 3.5.17
**Browsers:** Firefox 141.0
**OS:** Ubuntu undefined

### Steps to reproduce
- Start screenreader Orca
- Navigate via Tab to the Radio buttons
- Try to navigate with left/right arrow

### Expected Behavior
I guess either read the whole label or skip it entirely

### Actual Behavior
The screenreader will read the labels letter by letter as you press left/right arrow

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqtU1FLwzAQ/itHXqawtogMdHQDYeCjz+J8iOm1BtMkJGlxjP13r01X16Hiwx5Kk7vv7vu+5PKyZ96J7MHatG2QLVkesLaKB1xvNUDeJtzaftlvhNGBS40OStXIYkj0KccLaZLKmcaC1IpA0Ca1KVCttqxP+i0bC6jkSEQwr0xYKv6G6gRAkEK262fTEBtv6ScJnPvgjK7WHrkT74C6IqY8G6J51pWckGQTO1Ox0HLVIKl7NKZSOFH3L32DwmkI4KkEQWo9ggwzf1QMQnHviS3gZ0h8IwT67kQi+ejgrP+Zn98cZYOlv1xuGvFR0FeZizndYCk13Yra/ezSOllztyO+b/JLOx0jcfaOs0rRcVjjKFMkzvKkMW29cNIG8BiGcllb4wLswWEJByidqWFGz2PWwYEuV/sAcaRh1YGurruusQ91YIc5u00X6c0d6xb36YLNS648ziePLcZevwC2bRNH)

### Other comments
It would be great to be able to pass down the `aria-hidden="true"` to the slot.

waterWang:
## Description

When a VSelectionControl (used by VRadio, VCheckbox, etc.) has both an `aria-label` attribute on the `<input>` element AND a visible `<label>` element linked via `for`/`id`, screenreaders may read the label text incorrectly. This is especially problematic for Orca on Linux, where the label is read letter by letter when navigating inline radio buttons with arrow keys.

## Fix

Only set `aria-label` on the `<input>` when there is no visible label element present. When a visible `<label for="...">` exists, the explicit label association provides the accessible name, making `aria-label` redundant.

This follows the WCAG best practice: prefer visible labeling over `aria-label` when both are possible.

## Related Issue

Fixes #21928


J-Sek:
duplicate of #23081

PR #23083, opened 2026-08-03
fix(VOverlay): cache aria-owns lookups

### LINKED ISSUE #23056: [Bug Report][4.1.7] VOverlay: click-outside include() runs a full-document querySelector per overlay on every mousedown
### Environment
**Vuetify Version:** 4.1.7
**Vue Version:** 3.5.40
**Browser:** Chromium 150.0.7871.186 (current)

### Steps to reproduce
1. Open the playground link.
2. Click 1. open this menu and leave it open.
3. Click 2. measure a mousedown — baseline, ~0 ms.
4. Click 3. add 100 tooltips, then measure again. Repeat.

The table in the repro records how long a single synthetic `mousedown` blocks the main thread.

### Expected Behavior
Dispatching a mousedown costs roughly the same whether the page holds 0 or 500 idle overlays.

### Actual Behavior
Cost grows super-linearly with the number of overlays whose content has been mounted, and (for tooltips) linearly with total document size on top of that.

Measured in the playground (20 000 filler nodes):

| tooltips | `.v-overlay__content` | DOM nodes | blocked main thread |
|---|---|---|---|
| 0 | 1 | 20 087 | 0.0 ms |
| 100 | 101 | 20 402 | 17.9 ms |
| 200 | 201 | 20 707 | 37.6 ms |

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlWW+O2zYWvwrrLGDNxpJnJskG63qmCXYb7ALNZtEW+TIOWlqibXYoUStS9hgDL3qIPcNeod97lJ6kv0dSsqRxJgE6gGCJ7/H9/0fOzf1I5qWubJzzMvnJ6GI0G90vCsYWAWAWoxlzK7S2rQV9L0Yba0szm07TrMC2TCi5rZJC2GlR5tNXQJtWdWFlLuJM56+eJS+S5xfTTBrbXU+EyeNlpXdGVKCyGE06fKZY3IoqrkSRiUpUn8t3sK3HewB7lL+Vq/1n8AyYr54nF8lLzyUsxYovDbFwpInyYVEcRpORksWtGRg7Nc7QN40Af4inI0aUPniGpkqnr8syARI4zq3IS8WtuCaU+TbmZcnsRuTiajHKeHW7GDmIg+VcFuHLfae6sFgSVbuI5c0lSxU3BvutuLPx5i8sX8aXLR3/9/4dbK/4fsYEXvYs17URmd4VzKS8MCQCy3Ra56KwTBepYKWo2FJrKzKm/d4O0+nmEuQ7C2VfiKXO9vElCfJ8IMg3gm8F4wyMaqZLUTBeZExaAwIyvY11bY3MBJunOhPXskhVnYnobD513wwBbCBfl+JAoRmIsVQrJVIQ9UBPLNnGQZMffiBTQtWGrOxR7BmDxEu5UiZQAQvzBkBzFEoXTPB0M2G7jUw3TBpTC9Ml6Hc2JJP/1BDqO0ES6ioa3/BK8pjo/hf2++3n/y9GH8Yt8d9+/h9kYHUhkTd3IuvS5dZWcllbwUygNmFGQ56O/Duubp0zg+4Je1eoji8b6d6/hUcapkbAdn69FS7AHIeVrpjVWllZBht3CTqWtoKjKyN8bO02Wg2MmnObbgAutN3IYp10w6vsR5dWHw0vlpsHMTZX8vodhRZxdoG2FErvHFfl4g8hQrGXzKdAHW59K7ipK+Esv5N2AwmPyu6FBVlp4GVHfslheuTkaVKvs+y41SkdaPM1Ejlh34pScDvcO59q1TdAJretBbJ4pcQdW/P4GeNKros4hUnhYFqOdxUvT+UdlROyBH7gRQU69ElWWozYLFXaiFiDlM+L2CUjkFZcGXQdCiCDUgdQjyroNgWNPeGplVuOIMS+e1ZWGjofBmIEUZa2gCRLhDRwHaaTQiu3uZGMfcXGpk5TYcyYzdi4rGTOq/34BE3G7u+dq5t9Tln4iNzsNl8kvtw45xF0zA7oCQPRpk62oY7TXtXuaZLyKjulIa3HFKzXvuLloom4LmzI5wQ9WiRxewHRGvFV46gQWDDNZXKMsmNZPKXZkEYbqU+v2MX5OWg9SxhHCOOjDeNHCG2pVhQ2ZCk8+pDyFSOylUCBORIakHIpvpJC9Woda0I3Kep8KShOVhJVHrNJHy0ThZF2D3iq8xJBOURAm3YJ4LcjuzNU6wGOsXtFLTnnSCmZ2c2MXbw4L++GeHZfEpoXaQjcoJHFmUDPVr12MO3lOnK7n+wYCiCX7e7YxnIFNl8cc7YLbSqDT/su5BO2IEsD9i9fgZt0cRW+7bysErau0HUFr9TeFcWQR6G2rmRlbDIg7M2y4xXa1roLI917ulq+RGcICiI3UT/2iRLF2m4QQX3VTunTD0RIxYf5OLfdgalFvD5GND5OIZwcGD6C+/d3b30kfRRjqXR6i2mKpjpYr4KgJ3Cx1JeWcIY6zS11wIdqwozozLBOhEFAnsGJiEFnUSqvt4IMJ09VZJtdo35WSZunhwP4nihsDWIwx6cRnU0ewaLe411cJblh1+zFOZVvVwNEVWHOWIFVvBNyvaHGrzJXzT/aA4gK1HgjMSpFF2fgjAnhYZl/IMwpuw+MTAXLReujo2/KSysxoQ4C081Ogq9F1cxYfiahnMM8hlJtNjjxpTXGC4wbNFn44dtY9J1ZlxalH6Urb2qyH9gRUG6eor2Go9+4nX5+RXt3CW4xgvmRq0uwLjIkvHW5Tbs/OS6D8h4TkC8CzVDXpejmQ8CUXKHvrdw7ZSwm08K6+uLacKoNnTag/poOgm5EJUug4cOEj02E29hshOjxbLxQcjr9dCHdYr5xgYRqfo5q/qXjuIIfZozXVve3LXWFk2p3pdIwVXcEH2RlMwo1aSidjUJOPZaCc1Nixkd4edyZ1SXwfmz8++RP9/LwI3Y9QYRLl0204SSRXjDieEoBm2Hrr7/8+suJbSdnG4p0Z9/j6ZNGk8Hx000mzRGVPnCWxXuPIj5NWsnS0pGiJjBj/m6DIVfFih3YqtI5G+N4PPYeBhtERWdiAFp0fnYEhb7tAZfn+OsA2xnQg9382gGHchigNx8AIuCqLjC84hwXJie0vXDt4re5Y+cVu8GZvgnvSOEkIGmg+RI/c/YCP0+ftvtaPc6BgukZW3KONEVB3EVOINefmyMhlZoE3irpUPT1FkvR0R2F2LG3NMh5wLgd6sYTWHFZL+FiM8OJqxbscBbuURhruZD0SVmbTfRAEBZDwoAYpmGHbeChKOITtjxjV9fIyRhvTQY2XXrLVS084Vbrxm8QJ7x5tFaqHCBicXP5oV1rusnsaJDeGfm1UtH4RD0an4VJoaXk2s2jZP483HRw6kN5ZIcLVQRpuLeh6E6swb1NG7MpyqwVuNAZRG6LEC6DWnDSXA8dkWh3AwxXQ0cg1lyquGzwMUR3RFdHzhEeyIxVasMruU5co/wHKr3PC0Euo604c3tYRDngtyB0oiBRu4aAorh6gvcxFoP2TgyvfiVy2P47qqPf8zXF778hvzQiwc1IdKM0z97AJdHZhNH7P4l5dOazq82tPhHW5FjIE6xcfcRvkKy5YnNXZbEr6K5AkbioKCsWfQEKzbBKa/hMSo7bRvu1wh7Q9Pz/tpEqiwgZ93M9+UhyJ52jzCKMaKsgoyfrEjEoHkU4xGi1FRMAf4KYLk96RYOuGrs6eQcGaaIxgb34OJLgHfIpoI87ynWAJAyg9NNZxeHM1YRv3NlcwFKkBMpCEK5L/iGuCw2H7BTwuP2ShIjAfW1EBHyanLQa+b71Z7BV35rRuLlUpXHOJGut17iKKSUuTHU+xaXpVyueS7W/+lYvtdUz9OfJMzzP8bzA8xLPX8/PyWAP+bt4+1wBTt9e55l0or16mdyROMh93KlIrlBPcMciiUOSoxwAFoQIaRJC82SdeD8oBsdSsCjEnUP1sdJUjSbNw0ZfV13Qt/8IyMSK18p+7xcRRphnxqGWTafOSfgiR40Ok5G/gYeo7r56NHENsX8r3awFGUaT+/Z1FrYdPvwOkiXMeQ==)

### Root cause

the click-outside binding:

https://github.com/vuetifyjs/vuetify/blob/94e94a92a3edf8afd8b9dcd7e56fcbd4961661fe/packages/vuetify/src/components/VOverlay/VOverlay.tsx#L408-L420

and `ownsFocus`, called once per candidate:

https://github.com/vuetifyjs/vuetify/blob/94e94a92a3edf8afd8b9dcd7e56fcbd4961661fe/packages/vuetify/src/components/VOverlay/VOverlay.tsx#L228-L229

`[aria-owns~=…]` is an unindexed attribute selector. A tree walk, with no fast path. One walk per `.v-overlay__content` in the document, on every mousedown.

`packages/vuetify/src/directives/click-outside/index.ts` is what makes it run so often:

https://github.com/vuetifyjs/vuetify/blob/94e94a92a3edf8afd8b9dcd7e56fcbd4961661fe/packages/vuetify/src/directives/click-outside/index.ts#L87-L92

The `true` is `ignoreActive`. Note the comment a few lines above `include` in the same file:

> The include element callbacks below can be expensive so we should avoid calling them when we're not active.

The mousedown path deliberately bypasses exactly that guard. `VOverlay`'s own `if (!isActive.value) return []` still short-circuits idle overlays, so the multiplier is *active* overlays. But a single open dialog or menu is enough, and that is the normal state while a user works.

An `id` on the overlay is required to reach the query at all: without one, `el.closest('.v-overlay')?.id` is `''` and the loop exits early. `VMenu` and `VTooltip` both assign one unconditionally (`v-menu-${uid}` / `v-tooltip-${uid}`), so stock components always qualify.

### Other comments

I was investigating this, because of user reports of slow pages in a booking-management screen: a non-virtualised infinite scroll list where each row carries a hover popover, with a details sheet open on top. When the list was filled with ~ 100 items it was unusable.
I reproduced that slowness with a DevTools performance measurement: Successive clicks in one session while scrolling the list measured **107 ms → 290 ms → 1031 ms → 3033 ms**. 
The DevTools profile attributed **98% of the 3 s to a single native `querySelector`**, under `ownsFocus ← include ← click-outside`. 
Style recalculation and layout across the whole 31 s recording totalled 265 ms and 72 ms. The page was not rendering, it was searching.

I fixed that slowness in the app by virtualising the list and only arming the hover popover when the row is actually hovered. (So reducing the amount of `.v-overlay__content` actually in the DOM)

cyphercodes:
## Description
Fixes #23056

`VOverlay` now builds the `aria-owns` lookup table once while collecting click-outside include elements, instead of running a document-wide selector for each mounted overlay content node on every mousedown. This keeps submenu/activator ownership behavior intact while avoiding the repeated selector work reported in the issue.

Added browser coverage that mounts unrelated overlay owners and asserts the outside mousedown path does not issue per-overlay `[aria-owns~=...]` queries.

## Markup:

```vue
N/A — covered by VMenu browser regression test.
```

### DIFF
--- a/packages/vuetify/src/components/VOverlay/VOverlay.tsx
+++ b/packages/vuetify/src/components/VOverlay/VOverlay.tsx
@@ -217,7 +217,21 @@ export const VOverlay = genericComponent<OverlaySlots>()({
 
     let openedWithActivatorFocus = false
 
-    function ownsFocus (activeElement: Element | null): boolean {
+    function getAriaOwners (): Map<string, Element> {
+      const owners = new Map<string, Element>()
+
+      for (const el of document.querySelectorAll('[aria-owns]')) {
+        const ids = el.getAttribute('aria-owns')?.trim().split(/\s+/) ?? []
+
+        for (const id of ids) {
+          if (id) owners.set(id, el)
+        }
+      }
+
+      return owners
+    }
+
+    function ownsFocus (activeElement: Element | null, ariaOwners?: Map<string, Element>): boolean {
       let current = activeElement
       const visited = new Set<Element>()
       while (current) {
@@ -226,7 +240,7 @@ export const VOverlay = genericComponent<OverlaySlots>()({
         if (el === contentEl.value) return true
         visited.add(el)
         const ownerId = el.closest('.v-overlay')?.id
-        current = ownerId ? document.querySelector(`[aria-owns~="${CSS.escape(ownerId)}"]`) : null
+        current = ownerId ? ariaOwners?.get(ownerId) ?? document.querySelector(`[aria-owns~="${CSS.escape(ownerId)}"]`) : null
       }
       return false
     }
@@ -410,11 +424,13 @@ export const VOverlay = genericComponent<OverlaySlots>()({
                     closeConditional,
                     include: () => {
                       if (!isActive.value) return []
+                      const ariaOwners = getAriaOwners()
+
                       return [
                         activatorEl.value,
                         // Submenu clicks count as "inside"; clicks in ancestor overlays (e.g. a host dialog) don't.
                         ...Array.from(document.querySelectorAll('.v-overlay__content'))
-                          .filter(ownsFocus) as HTMLElement[],
+                          .filter(el => ownsFocus(el, ariaOwners)) as HTMLElement[],
                       ]
                     },
                   }}
--- a/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.browser.tsx
+++ b/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.browser.tsx
@@ -467,6 +467,59 @@ describe('VMenu', () => {
       await expect.poll(() => screen.queryByTestId('l1')).toBeNull()
     })
 
+    it('does not query aria-owns once for every mounted overlay on mousedown', async () => {
+      render(() => (
+        <div>
+          <VBtn data-testid="opener">
+            Open
+            <VMenu activator="parent" closeOnContentClick={ false }>
+              <VList>
+                <VListItem link>Item</VListItem>
+              </VList>
+            </VMenu>
+          </VBtn>
+          <div data-testid="outside" style="position: fixed; bottom: 0; right: 0; width: 120px; height: 120px;">out</div>
+        </div>
+      ))
+
+      const fixtures: Element[] = []
+      for (let i = 0; i < 8; i++) {
+        const owner = document.createElement('button')
+        const overlay = document.createElement('div')
+        const content = document.createElement('div')
+
+        overlay.id = `stale-overlay-${i}`
+        overlay.className = 'v-overlay'
+        content.className = 'v-overlay__content'
+        owner.setAttribute('aria-owns', overlay.id)
+        overlay.append(content)
+        document.body.append(owner, overlay)
+        fixtures.push(owner, overlay)
+      }
+
+      const querySelector = vi.spyOn(document, 'querySelector')
+
+      try {
+        await userEvent.click(screen.getByTestId('opener'))
+        await expect.poll(() => screen.queryByText('Item')).toBeVisible()
+
+        const outside = screen.getByTestId('outside')
+        outside.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }))
+        outside.dispatchEvent(new MouseEvent('mouseup', { bubbles: true }))
+        outside.dispatchEvent(new MouseEvent('click', { bubbles: true }))
+
+        const ariaOwnsQueries = querySelector.mock.calls.filter(call => {
+          const selector = call[0]
+          return typeof selector === 'string' && selector.startsWith('[aria-owns~=')
+        })
+
+        expect(ariaOwnsQueries).toHaveLength(0)
+      } finally {
+        querySelector.mockRestore()
+        for (const fixture of fixtures) fixture.remove()
+      }
+    })
+
     it('should close the parent menu when clicking outside an autocomplete child', async () => {
       render(() => (
         <div>


packages/vuetify/src/components/VOverlay/VOverlay.tsx
```diff
@@ -226,7 +240,7 @@ export const VOverlay = genericComponent<OverlaySlots>()({
         if (el === contentEl.value) return true
         visited.add(el)
         const ownerId = el.closest('.v-overlay')?.id
-        current = ownerId ? document.querySelector(`[aria-owns~="${CSS.escape(ownerId)}"]`) : null
+        current = ownerId ? ariaOwners?.get(ownerId) ?? document.querySelector(`[aria-owns~="${CSS.escape(ownerId)}"]`) : null
```
d--j, line 243: doesn't this fallback to the `document.querySelector('[aria-owns~="${CSS.escape(ownerId)}"]')` even when `ariaOwners` is non-null but the `ownerId` has no `aria-owns`? Then each negative lookup would still execute the expensive CSS selector lookup.


```suggestion
        current = ownerId && ariaOwners ? ariaOwners.get(ownerId) : ownerId ? document.querySelector(`[aria-owns~="${CSS.escape(ownerId)}"]`) : null
```

J-Sek:
closing in favor of [a57ec6e](https://github.com/vuetifyjs/vuetify/commit/a57ec6e)

PR #23092, opened 2026-08-10
fix(VTable): stripe expanded data table rows with their parent row

### LINKED ISSUE #23091: [Bug Report][4.1.8] VDataTable: Striping does not work when using expansion with expanded-slot
### Environment
**Vuetify Version:** 4.1.8
**Vue Version:** 3.5.41
**OS:** Linux

### Steps to reproduce
1. Open the reproduction link
2. Observe that all main items share the same striping color (no striping visible)
3. Expand one or more items
4. Observe that all expansion areas share the same striping color (opposite color of all main items)

### Expected Behavior
Striping should always happen on the level of main items. If a main item is expanded, the expanded area should share the same striping as its related main item.
The next main item (together with its expanded area) should then receive the opposite striping-color.

### Actual Behavior
All main items have the same striping color and all expansion areas have the opposite striping color.
Therefore, there is no striping between main items (especially when all items are closed / not expanded).

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNq1V21v2zYQ/iuc9sEKZkvJmqKrkXQptg4LUGDFWuxLXRS0dLbZSKRG0m6MwP99D1/0lrhBi6FBBIv3+twd70i9v0tE3ShtZzVvsk9GyWSe3C0kY4vIMItkzjzF0XZbcutFsrG2MfM8L0oJtZIqsdOZJJvLps6vIJbrrbSiplmp6qsn2dPs/CwvhbFDekamni21+mxIw8oimQ785CDuSM80yZI06a/1e09t5Pse71H/Vqz2X+EzSl6dZ2fZL8FLJM0qvjTOhTftLB8WEv/JNKmEvDH30l0Yn+r3/9OfN/OhdWR0kb9smgwi8HRhqW4qbumFc3mxm/Gm8a9+UShpuZCkI8kTS275zPJlRWy+IY60mUsgDG+LhM0FbDqS/wXBbNTnGd02XJbMWC0aKsFVZblIOruw3CJhu5mplJ0HDS97x5wtJGuoAJUWavv3biMMw7/dEPPqRijJnDWmVsHGXbCViZIdDhk0yPPHdoS0pCWvqj3bGjKMM+AQBcCVzOoZVVSTtNkISz4G8wDbNWK5dTB6A0wTAhZyzawCZgAfg0aXxTbrgHnoztJHq9lP7AxBfBOKlyaUwDl1uULKPRQEthMc2ObSbmbFRlTlFN6AsCCxCykQqxVaRN7L1ZIXN2uttihuoSqlEQiXvgKDQH32DauxmUIMoVQFd+kd23Oakm5tL+yys+E7GHTF4jU99MlDzRtNO6G29ywODG0QwHLPlvuGG+MTDy1XbbfR+swAbqdkHinzRT5qn0AadkjbSiCOe8lRQquNbGBpCoCwzJDdBldQNJbF/mKXGAfe5h2zwlY0Z5PrcjJlN7THqygn7BBHVi/whmth951QE5ZHBK/bTPQJ7k1H3sfl+pjqy8JueXVMkXvOF9SOdMXAY9jnrR5GWJ8PX5pBNtr8i3LOzqIbxkKosPVXCUCdTB8KWL9zfdPzOrQPOC2c3n4XztD5zw+dv9qR/JL312K9Qcjf4v7JY+5b5veK/eljzs+/e+zPHnPfYvtesT8fOcd2vMhDt6JP48nqxkZmDU7WcFfCVi80oblx5LIDW2lVswmO3slCdgLxsO7YWXt890JOu2XGw7tnguanBSgLGboDowW90XlO8ZwsJKgZ+Cuxzkhrpf/E+VqRhiSxyxe+sVRFgZcSFIIKZnQaEXW0Gi1u08mPeJ+AGKP3MEL4mmq1o7d2X9E7vk4h8wb4haEMx2r6vlK8/AMTMT2ZMvd+7ZynJx+8g9VWFtYdgmMjLD0J9Q4xWlAuWamKrT+M/92S3r/FDCks0E9+bK9AmPC4Whpnw2yIrIOLoq5Y+gMsnMCH3Wo/WLDMsHFg7FU84IP/39xpmDphf4ca4HPIPTpvmaUbTauIMZjFSfaZxcDTVBPyu6MpmJ8A88QlPW7gEJK7BA5jCgWMaNKJYwf4zIsCXwXxySC4AdOBAdf9DKi8LF1L2te4JhJOI1hFEJi5EdzQ/ENZvzW8sA8gyHZwl6rcZ9gRaLLUGfD8w9Gsudp39Yy5GmcznbSX3hVkTbZWal0RbwQutKrOcan9dcVrUe0v/1ZLZdX87PR0+gTPOZ6neJ7heX566hL20L/fb18L4PiXRV0KD+3qWXbr4KD3cWsUvCrJiLUUzkNWYxyAF0HENolb8+ic+OfeMOhHwQK3Ii8a9ko7Ndo2j4qpD8hv+u4jraQV31YWd11HxDYajN0890XCyhUqOUyT8HUEqP57IpmueGVo/N3Q0iKGZHrXvc6j2uHDf+hRoSk=)

### Other comments
The expanded-slot internally is rendered as a dedicated tr-element. As striping is implemented via tr:nth-child, it receives the opposing striping color than the tr-element of its main item. This causes the next main item to have the same striping color again as the previous main item thereby bypassing the intended striping of main items.

lazerg:
## Description

`VDataTable` renders the `expanded` slot as its own `<tr>` right after the row it belongs to, and that `<tr>` is in the DOM even while the row is collapsed. Striping is done with `tr:nth-child(odd/even)`, so each of those extra rows flips the parity: every main row ends up the same color and every expansion row gets the other one, which reads as no striping at all.

The stripe selectors now count with `nth-child(... of :not(.v-data-table__tr--expanded))` and paint an expansion row the same as the row above it. Plain tables are unaffected, since every row there matches the `:not()`.

fixes #23091

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-data-table
        :headers="headers"
        :items="items"
        striped="odd"
        show-expand
      >
        <template #expanded="{ item }">
          <div class="pa-4">{{ item.name }}</div>
        </template>
      </v-data-table>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        headers: [
          { title: 'Dessert', key: 'name' },
          { title: 'Calories', key: 'calories' },
        ],
        items: [
          { name: 'Frozen Yogurt', calories: 159 },
          { name: 'Ice cream sandwich', calories: 237 },
          { name: 'Eclair', calories: 262 },
          { name: 'Cupcake', calories: 305 },
        ],
      }
    },
  }
</script>
```

### DIFF
--- a/packages/vuetify/src/components/VDataTable/VDataTableRows.tsx
+++ b/packages/vuetify/src/components/VDataTable/VDataTableRows.tsx
@@ -13,11 +13,11 @@ import { useLocale } from '@/composables/locale'
 import { MaybeTransition } from '@/composables/transition'
 
 // Utilities
-import { Fragment, mergeProps } from 'vue'
+import { cloneVNode, Fragment, isVNode, mergeProps } from 'vue'
 import { genericComponent, getPrefixedEventHandlers, pick, propsFactory, useRender } from '@/util'
 
 // Types
-import type { Component, PropType, TransitionProps } from 'vue'
+import type { Component, PropType, TransitionProps, VNode, VNodeArrayChildren } from 'vue'
 import type { Group, GroupSummary } from './composables/group'
 import type { CellProps, DataTableItem, DataTableLoading, GroupHeaderSlot, GroupSummarySlot, ItemSlot, RowProps } from './types'
 import type { VDataTableGroupHeaderRowSlots } from './VDataTableGroupHeaderRow'
@@ -32,6 +32,17 @@ export type VDataTableRowsSlots<T> = VDataTableGroupHeaderRowSlots & VDataTableR
   'no-data': never
   'expanded-row': ItemSlot<T>
   expanded: ItemSlot<T>
+}
+
+// rows in the expanded-row slot belong to the item above, so striping skips them
+export function expansionNodes (nodes?: VNodeArrayChildren): VNode[] {
+  return (nodes ?? []).flatMap(node =>
+    Array.isArray(node) ? expansionNodes(node)
+    : !isVNode(node) ? []
+    : node.type === Fragment ? expansionNodes(node.children as VNodeArrayChildren)
+    : typeof node.type === 'symbol' ? []
+    : [cloneVNode(node, { class: 'v-data-table__tr--expansion' })]
+  )
 }
 
 export const makeVDataTableRowsProps = propsFactory({
@@ -200,9 +211,9 @@ export const VDataTableRows = genericComponent<new <T>(
                 )}
 
                 { slots['expanded-row']
-                  ? isExpanded(item) && slots['expanded-row'](slotProps)
+                  ? isExpanded(item) && expansionNodes(slots['expanded-row'](slotProps))
                   : slots.expanded && (
-                    <tr class="v-data-table__tr--expanded">
+                    <tr class="v-data-table__tr--expanded v-data-table__tr--expansion">
                       <td colspan={ columns.value.length }>
                         { props.expandTransition
                           ? (
--- a/packages/vuetify/src/components/VDataTable/VDataTableVirtual.tsx
+++ b/packages/vuetify/src/components/VDataTable/VDataTableVirtual.tsx
@@ -2,7 +2,7 @@
 import { makeDataTableProps } from './VDataTable'
 import { VDataTableHeaders } from './VDataTableHeaders'
 import { VDataTableRow } from './VDataTableRow'
-import { VDataTableRows } from './VDataTableRows'
+import { expansionNodes, VDataTableRows } from './VDataTableRows'
 import { VTable } from '@/components/VTable'
 import { VVirtualScrollItem } from '@/components/VVirtualScroll/VVirtualScrollItem'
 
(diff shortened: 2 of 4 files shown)


J-Sek:
Can we make it support using `<v-data-table-row>` in `#expanded-row` example? (this [example](https://github.com/vuetifyjs/vuetify/blob/master/packages/docs/src/examples/v-data-table/virtual-expanded-rows.vue))?

lazerg:
Added in the last commit.

Rows from the `#expanded-row` slot now get a `v-data-table__tr--expansion` class, and striping skips them, so they no longer shift the stripes of the rows below. That covers `<v-data-table-row>`, a plain `<tr>`, or anything else the slot returns, in both `VDataTable` and `VDataTableVirtual`, so the example itself needs no change.

One limit though. CSS can carry the parent stripe onto only one expansion row, the one directly after the parent. `nth-child(An+B of S)` never matches a row outside `S`, and there is no previous-sibling combinator, so a run of expansion rows cannot be selected. So a lone expansion row still gets the parent stripe, and a group of two or more is left unpainted rather than tinting only the first one. In your example the child rows keep their own `bg-surface-light`.

If you want the whole group painted, the class has to come from JS with the parity baked in. Say the word and I will do it that way.

lazerg:
Hey @J-Sek, I added support for v-data-table-row in the expanded-row example. Rows from that slot now get their own class, and striping skips them so they don't shift the rows below. Does that cover what you had in mind?

PR #23095, opened 2026-08-11
fix(VBadge): make wrapper element use tag prop for valid HTML nesting (Fixes #23093)

### LINKED ISSUE #23093: [Bug Report][4.1.8] VBadge renders divs inside VBtn's span, producing invalid HTML
### Environment

**Vuetify Version:** 4.1.8
**Vue Version:** 3.5.38
**Browsers:** Chrome
**OS:** Linux

### Steps to reproduce

Put a `v-badge` inside a `v-btn` — the pattern shown in the VBadge documentation — and run the rendered page through the W3C validator (https://validator.w3.org/nu):

```vue
<v-toolbar>
  <v-toolbar-items>
    <v-btn title="Notifications" stacked>
      <v-badge content="3" color="warning">
        <v-icon icon="mdi-bell" />
      </v-badge>
    </v-btn>
  </v-toolbar-items>
</v-toolbar>
```

### Expected Behavior

Valid HTML. `VBtn` wraps its default slot in `<span class="v-btn__content">`, which is phrasing content, so what is rendered inside it should be phrasing content too.

### Actual Behavior

`VBadge` renders a `div` root plus a hardcoded `div.v-badge__wrapper`:

```html
<span class="v-btn__content">
  <div class="v-badge">
    <div class="v-badge__wrapper">
      <i class="mdi-bell v-icon"></i>
      <span class="v-badge__badge bg-warning">3</span>
    </div>
  </div>
</span>
```

The W3C validator reports:

> Element “div” not allowed as child of element “span” in this context. (Suppressing further errors from this subtree.)

The suppression matters in practice: the validator stops checking that subtree, so any other error below it goes unreported until this one is fixed.

The public `tag` prop only changes the outer element. `v-badge__wrapper` is hardcoded in `VBadge` and has no API, so `tag="span"` alone is not enough to make the output valid.

### Reproduction Link

https://play.vuetifyjs.com/playgrounds/798dTg

The playground asserts the nesting at runtime and prints the rendered button markup.

### Other comments

**The usual workaround does not apply here.** The natural answer is "wrap the button in the badge instead", but our use case is a notification button inset in a `v-toolbar`, and inverting the nesting breaks the inset styling in three ways:

- `.v-toolbar-items > .v-btn { border-radius: 0 }` no longer matches, because `div.v-badge` becomes the toolbar-items child — the button gets its rounded corners back.
- The button no longer stretches to the toolbar height: its parent goes from `.v-toolbar-items` (`height: inherit; align-self: stretch`) to `.v-badge__wrapper`, whose height is auto, so it collapses to content height.
- `.v-badge__badge` is positioned against `.v-badge__wrapper`, which goes from hugging the icon to wrapping the whole button — the counter jumps to the button's top-right corner and can be clipped by the toolbar.

So for a toolbar-inset button, badge-inside-button is the only nesting that renders correctly, and it is the one that produces invalid HTML.

**Suggested fix** — in `packages/vuetify/src/components/VBadge/VBadge.tsx`:

1. render `v-badge__wrapper` as a `span` instead of a `div`;
2. optionally default `makeTagProps({ tag: 'span' })`, so the documented badge-on-a-button pattern is valid without the consumer having to pass `tag="span"`.

This cannot change rendering: Vuetify already declares the display of both elements explicitly — `.v-badge { display: inline-block }` and `.v-badge__wrapper { display: flex }` — so the tag name carries no layout meaning. I checked the same markup with `span` in both places against the W3C validator: 0 errors, identical rendering.

Happy to open a PR if that would help.

waterWang:
## Summary

When `VBadge` is used inside `VBtn` (as shown in the VBadge documentation), the rendered HTML is invalid because `VBadge` hardcodes `<div class="v-badge__wrapper">` inside `VBtn`'s `<span class="v-btn__content">` wrapper. The W3C validator reports:

> Element "div" not allowed as child of element "span" in this context.

## Fix

Make the `v-badge__wrapper` element use the `tag` prop instead of being hardcoded as a `<div>`. When `tag="span"` (or any phrasing-valid tag) is passed, both the root element and the wrapper element produce valid phrasing content.

### Before (invalid HTML)
```html
<span class="v-btn__content">
  <div class="v-badge">
    <div class="v-badge__wrapper">   <!-- INVALID: div inside span -->
```

### After (valid HTML with tag="span")
```html
<span class="v-btn__content">
  <span class="v-badge">
    <span class="v-badge__wrapper">  <!-- VALID: span inside span -->
```

### Backward compatibility
Default `tag="div"` produces identical output to before: `div > div > ...`

Fixes #23093

### DIFF
--- a/packages/vuetify/src/components/VBadge/VBadge.tsx
+++ b/packages/vuetify/src/components/VBadge/VBadge.tsx
@@ -112,7 +112,7 @@ export const VBadge = genericComponent<VBadgeSlots>()({
           { ...attrs }
           style={ props.style }
         >
-          <div class="v-badge__wrapper">
+          <props.tag class="v-badge__wrapper">
             { ctx.slots.default?.() }
 
             <MaybeTransition transition={ props.transition }>
@@ -150,7 +150,7 @@ export const VBadge = genericComponent<VBadgeSlots>()({
                 }
               </span>
             </MaybeTransition>
-          </div>
+          </props.tag>
         </props.tag>
       )
     })
--- a/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
+++ b/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
@@ -71,6 +71,13 @@ describe('VBadge', () => {
       const el = wrapper.find('custom-tag').element
       expect(el).toHaveTextContent('tag')
     })
+
+    it('uses the tag for the wrapper element so phrasing content stays valid', () => {
+      const { wrapper } = render(<VBadge tag="span" content="3">child</VBadge>)
+      const root = wrapper.find('span.v-badge').element
+      const wrapperEl = root.querySelector('.v-badge__wrapper')
+      expect(wrapperEl?.tagName).toBe('SPAN')
+    })
   })
 
   showcase({ stories, props, component: VBadge })


packages/vuetify/src/components/VBadge/VBadge.tsx
```diff
@@ -112,7 +112,7 @@ export const VBadge = genericComponent<VBadgeSlots>()({
           { ...attrs }
           style={ props.style }
         >
-          <div class="v-badge__wrapper">
+          <props.tag class="v-badge__wrapper">
```
J-Sek, line 115: This would be fine if we are convinced `tag` on `<v-badge>` is never used in real life. Otherwise it is too risky

J-Sek:
We have badges within buttons and tabs across handful of examples. Assuming we don't bail on the risk of regression, those should be updated in the same PR.

BatLeDev:
The three examples with a badge inside a button or a tab, for the update @J-Sek asked for — I checked every `<v-badge>` nested in a `<v-btn>`/`<v-tab>` across `packages/docs/src`, there are exactly these:

- `examples/v-badge/prop-content.vue` (3 badges, inside `v-btn stacked`)
- `examples/v-badge/prop-dot.vue` (2 badges, inside `v-btn stacked`)
- `examples/v-badge/misc-tabs.vue` (3 badges, inside `v-tab`)

All eight wrap only a `v-icon` or plain text, so adding `tag="span"` makes them valid with no other change.

Worth noting for anyone tempted to change the default instead: it would not work. `examples/v-badge/misc-customization.vue` wraps a `v-avatar` (a `div`), so a `span` default would just move the invalidity there. Keeping `tag="div"` as the default and opting in to `span` inside phrasing contexts is the right call.

The same defect on `VChip` is now tracked separately in #23103, with a fix proposed in #23106 — there it needs no `tag` discussion, since the chip root is already a `span`.

@waterWang I have those three example updates ready in waterWang/vuetify-fork#1 — a single commit on top of this PR's branch, so merging it into `fix/23093-vbadge-wrapper-tag` lands everything here in one piece.

Issue #18229, opened 2023-09-07
[Bug Report][3.3.13] a11y: v-divider hr should not have role of separator


kmmerc18:
### Environment
**Vuetify Version:** 3.3.13
**Vue Version:** 3.1.5
**Browsers:** Chrome 116.0.0.0
**OS:** Mac OS 10.15.7

### Steps to reproduce
Using the Chrome aXe Accessibility plugin, scan the reproduction link page and see the error "Certain aria roles must contain certain children". The role of 'separator' is not allowed in the list, and is redundant anyways given the implicit role in all hr elements.

### Expected Behavior
There should not be a role of separator on the hr element

### Actual Behavior
There is a role of separator on the hr element

### Reproduction Link
[...]

PR #23107, opened 2026-08-12
fix(VDivider): stop emitting the implicit separator role on hr

### LINKED ISSUE #23104: [Bug Report][4.1.8] a11y: v-divider emits a redundant role="separator" on hr
### Environment

**Vuetify Version:** 4.1.8
**Vue Version:** 3.5.38
**Browsers:** Chrome
**OS:** Linux

### Steps to reproduce

Render any divider and run the page through the [W3C validator](https://validator.w3.org/nu):

```vue
<v-divider />
```

### Expected Behavior

No redundant ARIA. `hr` already maps to the `separator` role in [ARIA in HTML](https://www.w3.org/TR/html-aria/#el-hr), so the computed role is `separator` whether or not the attribute is serialized.

### Actual Behavior

`VDivider` always renders an `hr` and hardcodes the role:

```js
role={ `${attrs.role || 'separator'}` }
```

The validator reports, once per divider:

> The “separator” role is unnecessary for element “hr”.

Severity is **info/warning**, not error — nothing is invalid. But it is noise in every accessibility report; six of them on one page of our portal.

### On the previous report

This was raised in #18229 (Sept 2023) and closed as *not planned* by the stale bot, on a single comment stating *"Everything correct, this is default behaviour"* and linking the [dividers accessibility docs](https://vuetifyjs.com/en/components/dividers/#accessibility).

That objection does not conflict with this request. The documented contract is:

> By default, `v-divider` components are assigned the WAI-ARIA role of **separator** […] To override the default separator role, simply add a `role="presentation"` prop.

Both halves survive dropping the attribute:

- *"assigned the role of separator by default"* — still true, delivered by `hr`'s implicit role. The computed role in the accessibility tree is unchanged.
- *"add `role="presentation"` to override"* — still works, as long as the component forwards `attrs.role` instead of defaulting it. `aria-orientation` keeps its existing guard, so `role="presentation"` still leaves it undefined, exactly as documented.

Nothing about the documented behaviour changes; only the serialized markup gets shorter. The issue was never argued on its merits, hence this re-report.

### Suggested fix

```diff
-role={ `${attrs.role || 'separator'}` }
+role={ attrs.role as string | undefined }
```

Unrelated but worth noting so the two are not confused: `VDateRangePicker` renders `role="separator"` on a `div`, where it is required and correct. This does not touch it.

I have a PR ready for this.

BatLeDev:
## Description

fixes #23104

`VDivider` always renders an `hr`, and `hr` already maps to the `separator` role in [ARIA in HTML](https://www.w3.org/TR/html-aria/#el-hr), so hardcoding `role="separator"` is redundant. The W3C validator reports it once per divider:

> The “separator” role is unnecessary for element “hr”.

It is a warning rather than an error, so nothing is invalid — but it is noise in every accessibility report, six of them on one page of our own portal.

`role` is now emitted only when the caller supplies one, so the [documented behaviour](https://vuetifyjs.com/en/components/dividers/#accessibility) is unchanged on both halves:

- *"assigned the WAI-ARIA role of separator by default"* — still true, now delivered by the implicit role, with the same computed role in the accessibility tree.
- *"add `role='presentation'` to override"* — still honoured, and still drops `aria-orientation`, which the existing guard already handled.

This was raised in #18229 and closed as *not planned* by the stale bot, on a comment that the separator role is the documented default. It is — and it stays the default here; only the redundant attribute goes away.

Worth noting so the two are not confused: `VDateRangePicker` renders `role="separator"` on a `div`, where it is required and correct. This PR does not touch it. I also checked that nothing in `packages/vuetify/src` or `packages/docs/src` selects on `[role=separator]` or asserts the attribute.

Added tests for the three cases: default (no `role` attribute, `aria-orientation="horizontal"`), `vertical`, and an explicit `role="presentation"` (honoured, `aria-orientation` dropped).

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <p>above</p>
      <v-divider />
      <p>between</p>
      <v-divider role="presentation" />
      <p>below</p>
    </v-container>
  </v-app>
</template>

<script>
  export default {
    name: 'Playground',
    setup () {
      return {
        //
      }
    },
  }
</script>
```

Run the rendered page through https://validator.w3.org/nu — one "unnecessary role" warning per divider before, none after. The second divider keeps `role="presentation"`.

### DIFF
--- a/packages/vuetify/src/components/VDivider/VDivider.tsx
+++ b/packages/vuetify/src/components/VDivider/VDivider.tsx
@@ -74,6 +74,9 @@ export const VDivider = genericComponent()({
     })
 
     useRender(() => {
+      const role = attrs.role as string | undefined
+      const hasSlot = !!slots.default
+
       const divider = (
         <hr
           class={[
@@ -95,15 +98,15 @@ export const VDivider = genericComponent()({
             props.style,
           ]}
           aria-orientation={
-            !attrs.role || attrs.role === 'separator'
+            !hasSlot && !role
               ? props.vertical ? 'vertical' : 'horizontal'
               : undefined
           }
-          role={ attrs.role as string | undefined }
+          role={ !hasSlot ? role : undefined }
         />
       )
 
-      if (!slots.default) return divider
+      if (!hasSlot) return divider
 
       return (
         <div
--- a/packages/vuetify/src/components/VDivider/VDivider.tsx
+++ b/packages/vuetify/src/components/VDivider/VDivider.tsx
@@ -99,7 +99,7 @@ export const VDivider = genericComponent()({
               ? props.vertical ? 'vertical' : 'horizontal'
               : undefined
           }
-          role={ `${attrs.role || 'separator'}` }
+          role={ attrs.role as string | undefined }
         />
       )
 
--- a/packages/vuetify/src/components/VDivider/__tests__/VDivider.spec.browser.tsx
+++ b/packages/vuetify/src/components/VDivider/__tests__/VDivider.spec.browser.tsx
@@ -228,4 +228,31 @@ describe('VDivider', () => {
       }
     })
   })
+
+  describe('role', () => {
+    it('relies on the implicit separator role of hr', () => {
+      render(() => <VDivider />)
+
+      const divider = screen.getByCSS('.v-divider')
+
+      expect(divider.tagName).toBe('HR')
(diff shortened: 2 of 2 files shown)


J-Sek:
should we set the role on the div (instead?) when using slot

J-Sek (approved):

PR #23113, opened 2026-08-13
fix(VTextField): prevent double role="combobox" on VSelect

### LINKED ISSUE #23105: [Bug Report][4.1.8] VSelect renders two nested role="combobox", the outer one invalid during SSR
### Environment

**Vuetify Version:** 4.1.8
**Vue Version:** 3.5.38
**Browsers:** Chrome
**OS:** Linux

### Steps to reproduce

Server-render a `v-select` and run the **served** HTML through the [W3C validator](https://validator.w3.org/nu). The defect is not visible after hydration, so it has to be checked on the SSR output.

```vue
<v-select :items="['a', 'b']" label="Sort by" variant="outlined" />
```

### Expected Behavior

One combobox per select, carrying its own `aria-expanded`, per the [ARIA 1.2 combobox pattern](https://www.w3.org/TR/wai-aria-1.2/#combobox) — the role belongs on the text input, not on a wrapper.

### Actual Behavior

`VTextField` applies its `role` prop to **both** the `VField` root and the `input`, and `VSelect` defaults that prop to `combobox` (`makeVTextFieldProps({ modelValue: null, role: 'combobox' })`). Two nested comboboxes result:

```html
<div class="v-field" role="combobox">        <!-- no aria-expanded in SSR -->
  …
  <input role="combobox" aria-expanded="false" aria-controls="menu-…">
</div>
```

The inner one is correct and carries `aria-expanded`/`aria-controls` in the server output. The outer one gets them only at runtime, from `VMenu`'s `activator="parent"` binding, so the server-rendered markup has a bare `role="combobox"` and the validator reports:

> Element “div” is missing required attribute “aria-expanded”.

Beyond validity, the nesting is a semantics defect after hydration too: a combobox whose accessible subtree contains another combobox.

### Suggested fix

Apply `props.role` to the input only and drop it from the `VField` root.

Note that removing the role alone is not sufficient: `VMenu` would then leave its runtime `aria-expanded`/`aria-owns` on a roleless `div`, which is invalid in a different way. The activator's ARIA wiring needs to target the same element that carries the role — which is why I am reporting this rather than opening a PR, unlike the sibling reports #23103 and #23104.

waterWang:
## Description

Remove the `role` prop from the `VField` root element in `VTextField`. The `role` attribute should only be set on the `<input>` element, not on the outer `VField` wrapper.

## Problem

When `VSelect` sets `role: 'combobox'` (via `makeVTextFieldProps`), the `VTextField` applies this role to **both** the `VField` root div and the `<input>` element. This results in two nested `role="combobox"` elements:

```html
<div class="v-field" role="combobox">        <!-- invalid — no aria-expanded in SSR -->
  <input role="combobox" aria-expanded="false" aria-controls="menu-…">
</div>
```

Per the [ARIA 1.2 combobox pattern](https://www.w3.org/TR/wai-aria-1.2/#combobox), the `combobox` role belongs on the text input, not on a wrapper. The nested `combobox` is semantically invalid and causes W3C validator warnings during SSR.

## Fix

Remove `role={ props.role }` from the `<VField>` component in `VTextField.tsx`. The `role` prop is still applied to the `<input>` element (line 251), which is the correct placement.

Fixes #23105

### DIFF
--- a/packages/vuetify/src/components/VTextField/VTextField.tsx
+++ b/packages/vuetify/src/components/VTextField/VTextField.tsx
@@ -216,7 +216,6 @@ export const VTextField = genericComponent<VTextFieldSlots>()({
                 onMousedown={ onControlMousedown }
                 onClick={ onControlClick }
                 onClick:clear={ (e: MouseEvent) => onClear(e, reset) }
-                role={ props.role }
                 { ...omit(fieldProps, ['onClick:clear']) }
                 id={ id.value }
                 labelId={ `${id.value}-label` }


J-Sek:
I wish it was that simple

PR #23120, opened 2026-08-14
fix(VMenu/VSelect/etc.): only set aria-controls while open

### LINKED ISSUE #23119: [Bug Report][4.1.9] a11y: VMenu and VSelect/VAutocomplete/VCombobox reference an aria-controls/aria-owns id absent from the document while closed
### Environment

**Vuetify Version:** 4.1.9
**Vue Version:** 3.5.38
**Browsers:** Chrome
**OS:** Linux

### Steps to reproduce

Render a closed select or menu and inspect the served or hydrated DOM — no interaction needed, the defect is the **closed** state:

```vue
<v-select :items="['a', 'b']" label="Sort by" />

<v-menu>
    <v-btn v-bind="props" text="Open" />
  </template>
  <v-list :items="['a']" />
</v-menu>
```

Run the page through the W3C validator (https://validator.w3.org/nu) or axe.

### Expected Behavior

ARIA id references resolve, or are not emitted. The [APG combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) is explicit that the reference is only needed while the popup is up:

> Note that `aria-controls` only needs to be set when the popup is visible. However, it is valid to reference an element that is not visible.

Referencing a *hidden* element is valid — referencing an id that is not in the document is not. In the [menu button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/), `aria-controls` is optional altogether.

### Actual Behavior

The references are emitted unconditionally:

- `VMenu.tsx` — the `activatorProps` computed puts `'aria-controls': id.value, 'aria-owns': id.value` on the activator.
- `composables/menuActivator.ts` — `ariaControls = toRef(() => menuId.value)`, rendered as `aria-controls` on the input of `VSelect`, `VAutocomplete` and `VCombobox`.

But the element carrying that id only exists while the overlay has content: `useLazy`'s `hasContent` is `isBooted || eager || active`, and `onAfterLeave` resets `isBooted` when not eager. So before the first open — **and again after every close** — the id is absent from the document, and every closed select/menu on the page fails validation:

> The `aria-controls` attribute must point to an element in the same document.

(severity: **error**, one per attribute per closed select/menu; axe reports the same as `aria-valid-attr-value`. European accessibility audits — RGAA / EN 301 549 — flag each occurrence.)

#### Prior reports

- #19250 reported exactly this (axe flagging `aria-owns="v-menu-759"` on the activator) and was closed as a duplicate of #19054 — but #19054 only gave the id to the **open** menu. The closed state was never addressed.
- #22540 (open) reports that `aria-owns` on the activator breaks VoiceOver's announcement of menu item labels — one more reason not to advertise it while nothing is open.

### Suggested fix

Emit the references only while the popup is open, which is the APG-documented pattern:

```diff
 // composables/menuActivator.ts
-  const ariaControls = toRef(() => menuId.value)
+  const ariaControls = toRef(() => toValue(isOpen) ? menuId.value : undefined)
```

```diff
 // VMenu.tsx
-        'aria-controls': id.value,
-        'aria-owns': id.value,
+        'aria-controls': isActive.value ? id.value : undefined,
+        'aria-owns': isActive.value ? id.value : undefined,
```

One existing test pins the current behaviour and would need its closed-state assertion moved after the open: `VSelect.spec.browser.tsx` › *"should have reactive accessibility attributes"* asserts `aria-controls` while `expanded: false` — i.e. it asserts the dangling reference.

`VTooltip` has the same shape (`'aria-describedby': id.value` unconditional while the tooltip content is unmounted) — happy to cover it in the same PR or leave it for a follow-up.

Our current userland workaround is `eager` defaults on the four components so the overlay stays mounted and the ids always resolve — viable because the lists are virtualised, but it mounts overlays nobody may ever open.

Happy to open a PR if this approach sounds right.

lazerg:
## Description

`VMenu` puts `aria-controls` and `aria-owns` on its activator, and `useMenuActivator` puts `aria-controls` on the input of `VSelect`, `VAutocomplete` and `VCombobox`. Both reference the overlay id, but the element carrying that id only exists while the overlay has content: `useLazy` mounts it on first open and drops it again on close. So before the first open, and after every close, those attributes point at an id that is not in the document, and the W3C validator flags it as an error, once per attribute per closed menu.

They are now emitted only while the menu is open. The APG combobox pattern only asks for `aria-controls` while the popup is visible, and in the menu button pattern it is optional.

The existing `VSelect` test that asserted `aria-controls` on a closed select now asserts it after opening instead.

Fixes #23119

## Markup:

```vue
<template>
  <v-app>
    <v-main>
      <v-container>
        <v-select :items="['a', 'b']" label="Sort by" />

        <v-menu>
            <v-btn v-bind="props" text="Open" />
          </template>
          <v-list :items="['a']" />
        </v-menu>
      </v-container>
    </v-main>
  </v-app>
</template>
```

Inspect the select input and the button before opening anything. On `master` both carry an id reference that resolves to nothing, with this change neither does until the menu opens.

### DIFF
--- a/packages/vuetify/src/components/VMenu/VMenu.tsx
+++ b/packages/vuetify/src/components/VMenu/VMenu.tsx
@@ -168,8 +168,8 @@ export const VMenu = genericComponent<OverlaySlots>()({
       mergeProps({
         'aria-haspopup': 'menu',
         'aria-expanded': String(isActive.value),
-        'aria-controls': id.value,
-        'aria-owns': id.value,
+        'aria-controls': isActive.value ? id.value : undefined,
+        'aria-owns': isActive.value ? id.value : undefined,
         onKeydown: onActivatorKeydown,
       }, props.activatorProps)
     )
--- a/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.browser.tsx
+++ b/packages/vuetify/src/components/VMenu/__tests__/VMenu.spec.browser.tsx
@@ -340,6 +340,38 @@ describe('VMenu', () => {
     expect(model.value).toBe(false)
   })
 
+  it('should only reference the menu id on the activator while it is open', async () => {
+    render(() => (
+      <VMenu>
+        {{
+          activator: ({ props }: any) => <VBtn { ...props } data-testid="btn">Open</VBtn>,
+          default: () => (
+            <VSheet class="pa-3" data-testid="menu-content">
+              <p>Content</p>
+            </VSheet>
+          ),
+        }}
+      </VMenu>
+    ))
+
+    const btn = screen.getByTestId('btn')
+    expect(btn).not.toHaveAttribute('aria-controls')
+    expect(btn).not.toHaveAttribute('aria-owns')
+
+    await userEvent.click(btn)
+    await expect.poll(() => screen.queryByTestId('menu-content')).toBeVisible()
+
+    const id = btn.getAttribute('aria-controls')
+    expect(btn).toHaveAttribute('aria-owns', id)
+    expect(document.getElementById(id!)).not.toBeNull()
+
+    await userEvent.keyboard('{Escape}')
+    await expect.poll(() => screen.queryByTestId('menu-content')).toBeNull()
+
+    expect(btn).not.toHaveAttribute('aria-controls')
+    expect(btn).not.toHaveAttribute('aria-owns')
+  })
+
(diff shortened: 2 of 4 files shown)


lazerg:
Closing as no one reviewed and this has been open more than a month with no activity. Happy to reopen if there's interest — thanks for your time.

PR #23127, opened 2026-08-16
fix(VProgressLinear): animate left/right directly instead of @property to fix Firefox indeterminate animation throttle

### LINKED ISSUE #23125: [Bug Report][4.1.9] Animation issue for indeterminate VProgressLinear (Firefox only)
### Environment
**Vuetify Version:** 4.1.9
**Vue Version:** 3.5.40
**OS:** Linux

### Steps to reproduce
1. Use Firefox.
2. Go to https://vuetifyjs.com/en/components/progress-linear/ and view the example usage. Check the "indeterminate" box.
3. Keep the mouse still in the window and watch the animation. Some iterations animate properly and others appear cut off. There appears to be a pattern to the animation errors.
4. Move the mouse out of the window. The animation does not play at all.
5. Move the mouse back into the window and keep it moving. The animation will play properly.

### Expected Behavior
Indeterminate animation repeats regardless of mouse position/movement

### Actual Behavior
Indeterminate animation will only play properly if the mouse is moving within the window. This happens in Firefox (I'm using version 153.0.4) but not Chrome.

### Reproduction Link
[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/#eNqlVU1v2zAM/Suae7ACJHKLtigatF2GocMG7DCsxS51D0pMJ+pkyZPktEGQ/z7qI58thgI7GHHIR75HiaQflploWm3coOEte7JaZcNsWSpCyuSwZTYkweJt8w78/zKbOdfaYVFMKoVhFUgxN0yBK1TbFCOEFaZTTjQwqHQzOmXn7OykqIR1u3YGthmMjX62YDBLmfV3eAo0zsEMDKgKDJj38h6E7XEf+P7J70S9eAdnQo7O2Am7jCzJNJB8bD1FSO0zr0q1yvqZFOq3PTjsiQ0H/bAW8F+cIZnP9BgJrZkUn9qWIQgZrxw0reQObjzkaj5ojZ4asHaAwoAbIvB0HJhGKA+6Kl4hMPCq2MmSKBouFHMWKWLrkCWZGEAIcpMVqY1uSI4a8lJtAEn3xs3WlWxBPnrtTFVsnWiz4LoWLaWaaGUd4Yi/3jJTfHqlQitDfy2mDIzR5itXlQSDSCDXN8SHagnRRwEDYkhngSZFG1ujsYVpfoTvORpT9UFGLN9Ao+dw5xYS7vmUIuYH6hcWGJeSPkjNqy8aU/T6xL9/8+S09xgI6k5NnNCK7CchtBenMNbo0HJNKj3pGlCO/enALO5AwsSh+vxo3Q1uBjhp1uewMwDn5RIiakI/YIYecrjOICnxCVnLcTDcrcQYzBn5P8+ErKgHYyvt6fPKg7qQmdCZgTppjGmJgmeSCqcU20fLOfTR+YQye/7Q01qJJfmp2K0pXmBSQ3PvjvJJgKI+ifB8p7gdpxeDXv+zY+VVdTvHbN9xYkABnpQvIveagrjd9K+xoTUCOBQQsRu5Y10tGHYErhbqEwT/6s1T83e/uc90VvunSfP1/NeItWyq9VQCbwXOtm4KnO+PNW+EXFz/1GPt9PDk+Lh/is8ZPuf4XOBzeXzsD+w1f+i39wp4e9E2lQjSRhfsxcvB2ceNIbiswIqpEp6B4QLxqyiJSGOSWvPNPfHrYBlsV0Gp4CVAY6+st8Z6zFMgDQWFpt98syqoeSfdfTRiG01nLk97vijCJeE/f1HZqp/FjwVKDas169dcWthfoGtb0pD1l5vXYQpbPf4Fw8KM+Q==)

### Other comments
The behavior in the playground I linked is slightly different than the behavior at https://vuetifyjs.com/en/components/progress-linear/. In the playground, the animation will only play if the mouse is moving inside the preview pane.

### Images
[github.com/user-attachments/assets/301a987e-5b65-455b-b9d5-e0ddae1d37a8](https://github.com/user-attachments/assets/301a987e-5b65-455b-b9d5-e0ddae1d37a8)

waterWang:
## Problem

In Firefox, the indeterminate animation of VProgressLinear pauses when the mouse is not moving. The issue is that Firefox throttles `@property`-registered custom property animations when there is no user interaction, stopping the indeterminate progress bar animation from playing.

## Root Cause

The current implementation uses `@property`-registered custom properties (`--v-progress-indeterminate-long-left`, etc.) animated via `@keyframes`, with `.long` and `.short` elements reading these values via `left: var(...)`. Firefox has a known issue where registered custom property animations are throttled (paused) when the page is idle — the animation only runs when the user is actively interacting with the page.

## Fix

Instead of relying on `@property` custom property interpolation for the `.long` and `.short` indicator bars, animate `left` and `right` directly on these elements using dedicated `@keyframes`:

- `indeterminate-ltr-long` / `indeterminate-ltr-short` for the default LTR direction
- `indeterminate-rtl-long` / `indeterminate-rtl-short` for the reverse RTL direction

This bypasses the Firefox `@property` throttling entirely because `left` and `right` are standard CSS properties that browsers handle on the compositor thread without user-interaction gating.

The `@property` declarations and the original `indeterminate-ltr`/`indeterminate-rtl` keyframes are preserved for the `variant="split"` background chunk elements, which still need custom property interpolation for their gap-based width calculations.

## Key changes

- `.long` and `.short` now use `animation-name: indeterminate-ltr-long` / `indeterminate-ltr-short` instead of reading animated custom properties
- `.long` and `.short` use `animation-play-state: inherit` so they pause/resume with the parent `--active` class
- `.long` and `.short` have explicit initial `left`/`right` values matching the 0% keyframe positions
- Added 4 new `@keyframes` rules: `indeterminate-ltr-long`, `indeterminate-ltr-short`, `indeterminate-rtl-long`, `indeterminate-rtl-short`
- The same fix applied to the RTL/reverse variant

## Fixes

Fixes #23125

### DIFF
--- a/packages/vuetify/src/components/VProgressLinear/VProgressLinear.sass
+++ b/packages/vuetify/src/components/VProgressLinear/VProgressLinear.sass
@@ -98,6 +98,9 @@
     animation-name: indeterminate-ltr
 
     .long, .short
+      animation-duration: inherit
+      animation-iteration-count: inherit
+      animation-play-state: inherit
       bottom: 0
       height: inherit
       position: absolute
@@ -105,12 +108,14 @@
       width: auto
 
     .long
-      left: var(--v-progress-indeterminate-long-left)
-      right: var(--v-progress-indeterminate-long-right)
+      animation-name: indeterminate-ltr-long
+      left: -90%
+      right: 107%
 
     .short
-      left: var(--v-progress-indeterminate-short-left)
-      right: var(--v-progress-indeterminate-short-right)
+      animation-name: indeterminate-ltr-short
+      left: -200%
+      right: 107%
 
     > .v-progress-linear__background
       height: 100%
@@ -153,6 +158,16 @@
 
     .v-progress-linear__indeterminate
       animation-name: indeterminate-rtl
+
+      .long
+        animation-name: indeterminate-rtl-long
+        left: 107%
+        right: -90%
+
+      .short
+        animation-name: indeterminate-rtl-short
+        left: 107%
+        right: -200%
 
       > .v-progress-linear__background
         &:nth-child(1)
@@ -234,6 +249,25 @@
       --v-progress-indeterminate-short-left: 107%
       --v-progress-indeterminate-short-right: -8%
 
+  @keyframes indeterminate-ltr-long
+    0%
+      left: -90%
+      right: 107%
+    100%
+      left: 107%
+      right: -35%
+
+  @keyframes indeterminate-ltr-short
+    0%
+      left: -200%
+      right: 107%
+    60%
+      left: 107%
+      right: -8%
+    100%
+      left: 107%
+      right: -8%
+
   @keyframes indeterminate-rtl
     0%
       --v-progress-indeterminate-long-left: 107%
@@ -250,6 +284,25 @@
       --v-progress-indeterminate-long-right: 107%
       --v-progress-indeterminate-short-left: -8%
       --v-progress-indeterminate-short-right: 107%
+
+  @keyframes indeterminate-rtl-long
+    0%
+      left: 107%
+      right: -90%
+    100%
+      left: -35%
+      right: 107%
+
+  @keyframes indeterminate-rtl-short
+    0%
+      left: 107%
+      right: -200%
+    60%
+      left: -8%
+      right: 107%
+    100%
+      left: -8%
+      right: 107%
 
   @keyframes stream
     to


J-Sek:
you need to verify the slop you generate

PR #23132, opened 2026-08-18
docs: stable anchor scrolling and TOC active states

userquin:
## Description

This PR completely overhauls the hash navigation, TOC scrolling UX, and layout offset handling in the documentation to fix race conditions, janky long-distance scrolls, and CSS offset overrides.

**Key changes include:**

- **Native CSS Offsets:** Replaced manual JS layout math (`getComputedStyle`) with native `scroll-margin-top: var(--v-layout-top)` on heading anchors (`main.v-main section section[id]`).
- **Stable Initial Load:** Replaced the arbitrary 1.5s timeout for `START_LOCATION` with a bounded `requestAnimationFrame` polling (`waitForElementStable`) that waits for DOM position stability (capped at 3s) before jumping to the anchor.
- **TOC Scrolling UX (`Toc.vue`):** 
  - Dropped the redundant `scrolling` flag from the global store.
  - Replaced arbitrary timeouts with native `scrollend` event listeners (with a fallback) for releasing the internal scroll lock exactly when the animation finishes.
  - Added smart scroll behavior: fallbacks to `instant` scroll behavior for long distances (> 500px) or when `prefers-reduced-motion: reduce` is enabled.
  - Removed early return guards in TOC `onClick` to allow manual re-triggering of the active anchor.
- **Router Scroll Behavior (`main.ts`):**  
  - Bypassed `vue-router`'s native `scrollTo` (which inherently ignores CSS `scroll-margin-top`) by manually calling `el.scrollIntoView()` and returning `false` for hash navigations, allowing the CSS rule to do its job. Included a fallback to native router behavior if the element is not found.
  - During the initial load's rAF poll (which waits up to 3s for DOM stability on heavy API pages), a user might start scrolling manually. If the auto-scroll triggers after they've moved, it yanks them back to the anchor, creating a jarring UX. We add one-time, passive capture listeners for `wheel`, `touchmove`, and `keydown` during the polling phase. If any user input is detected, the automatic `scrollIntoView` is immediately aborted to respect the user's manual navigation.
  - **Reload & Abort Safety:** Ensured browser reload conventions by checking `savedPosition` first on initial load, and returning `false` instead of `{ top: 0 }` on aborts or timeouts so the user is never unexpectedly yanked back to the top of the page.

## Motivation and Context

The previous implementation relied on arbitrary timeouts and JS-calculated offsets which led to race conditions, missed anchors on slow initial loads (especially noticeable with eager API-table rendering), and jarring smooth scrolls over massive distances (e.g., in the `v-date-input` API page). 

By delegating the layout offset to CSS, utilizing native DOM events, and implementing bounded polling, the scrolling experience is now stable, accessible, and much more performant.

## Markup:

```vue

```

### DIFF
--- a/packages/docs/src/components/app/Toc.vue
+++ b/packages/docs/src/components/app/Toc.vue
@@ -122,7 +122,7 @@
 </template>
 
 <script setup lang="ts">
-  const { toc: tocDrawer, scrolling } = storeToRefs(useAppStore())
+  const { toc: tocDrawer } = storeToRefs(useAppStore())
 
   const route = useRoute()
   const router = useRouter()
@@ -145,7 +145,6 @@
   }, { rootMargin: '-10% 0px -75%' })
 
   async function observeToc () {
-    scrolling.value = false
     activeStack.length = 0
     activeItem.value = ''
     observer.disconnect()
@@ -165,11 +164,10 @@
   })
 
   let internalScrolling = false
-  let timeout = -1
+
   watch(activeItem, async val => {
     if (!val || internalScrolling) return
 
-    scrolling.value = true
     const query = route.query
 
     if (val === frontmatter.value?.toc?.[0]?.to.slice(1) && route.hash) {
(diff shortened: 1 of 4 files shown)


packages/docs/src/main.ts
```diff
@@ -107,31 +166,61 @@ const router = createRouter({
...
-
-    if (!main) {
-      // For initial page load
-      wait = 1500
-      main = document.querySelector('main')
-    } else if (to.path !== from.path && to.hash) {
-      // For cross page navigation
-      wait = 500
+    // 1: Table of Contents (TOC) rewrites should never scroll: Toc.vue will scroll.
+    if (to.path === from.path && to.hash !== from.hash) return false
```
KaelWD, line 170: This breaks internal links like https://github.com/vuetifyjs/vuetify/blob/74a26b42957298a1dae7f85c6e05a5e2b528bd54/packages/docs/src/pages/en/features/global-configuration.md#L344

Would moving the scrolling code from TOC to here instead of returning a `ScrollPosition` work?

packages/docs/src/components/app/Toc.vue
```diff
@@ -180,20 +178,37 @@
         await router.replace({ path: route.path, hash: toc.to, query })
```
KaelWD, line 178: This watcher triggers after onScrollEnd, so internalScrolling doesn't skip it. scrollBehavior has no condition matching `to.path === from.path && to.hash === from.hash` so it falls back to `return { top: 0 }`, causing TOC clicks to flicker to the target heading then right back to the top of the page.

KaelWD (changes requested):
Doesn't work at all for me in either chrome or firefox.

PR #23134, opened 2026-08-19
docs: optimize api-table rendering and resolve scroll behavior

userquin:
## Description

The `ApiTable` component was causing significant main-thread blocking (up to 1.4s frames) on long pages (e.g., `v-date-input`) due to synchronous rendering of hundreds of rows within a table layout. Additionally, navigation via the TOC was experiencing layout thrashing, causing UI flickering and scroll anchoring issues when navigating between hashes on the same path.

## Solution

- **Performance**: Refactored `.v-table.api-table` from standard `display: table` to a dynamic `display: grid` layout using `subgrid`. This eliminates the browser's heavy auto-table-layout calculations, reducing scripting/painting time significantly (~70% improvement).
- **Accessibility**: Added explicit `role="table"`, `rowgroup`, `row`, and `cell` ARIA roles to maintain screen-reader semantic integrity, effectively compensating for the CSS display shift.
- **TOC & Scroll Fixes:**
  - Updated `scrollBehavior` in `main.ts` to skip automatic browser scroll when navigating within the same path, delegating all scroll control to the TOC component.
  - Optimized `Toc.vue` click handler with a sequenced `rAF -> setTimeout -> rAF` chain to ensure element positioning is measured only after Vue applies class updates and the browser stabilizes the grid layout.
  - Resolved an edge-case where clicking the TOC would trigger an unintended "jump-to-top" by refining hash navigation logic.

## Impact

- Drastic reduction in main-thread blocking time during API page interactions.
- Smooth, reliable TOC navigation without layout flickering or "rewind" effects.
- Maintained full a11y compatibility for assistive technologies.

Stacked on #23132

### DIFF
--- a/packages/docs/src/main.scss
+++ b/packages/docs/src/main.scss
@@ -3,4 +3,39 @@
 /* sections for toc */
 main.v-main section section[id] {
   scroll-margin-top: var(--v-layout-top);
+  .v-table.api-table {
+    --cols: 1;
+    table {
+      display: grid;
+      grid-template-columns: repeat(var(--cols), 1fr);
+      overflow: hidden;
+      tbody, thead, tfoot {
+        display: contents;
+      }
+
+      tbody tr, thead tr, tfoot tr {
+        display: grid;
+        grid-column: 1 / -1;
+        grid-template-columns: subgrid;
+        align-items: center;
+        border-bottom: thin solid rgba(var(--v-border-color), var(--v-border-opacity));
+      }
+
+      tbody tr:last-of-type {
+        --v-border-opacity: 0;
+      }
+
+      th, td {
+        display: block;
+        align-content: center;
+        min-height: calc(var(--v-table-header-height));
+        height: 100%;
+        border-bottom: none;
+      }
+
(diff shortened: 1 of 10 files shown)


packages/docs/src/main.scss
```diff
@@ -3,4 +3,39 @@
...
+        grid-template-columns: subgrid;
+        align-items: center;
+        border-bottom: thin solid rgba(var(--v-border-color), var(--v-border-opacity));
+      }
+
+      tbody tr:last-of-type {
+        border-bottom: none;
+      }
+
+      th, td {
```
userquin, line 28: we should add `[role="cell"]`  here (not all the entries using `td`)

packages/docs/src/main.scss
```diff
@@ -3,4 +3,39 @@
...
+        display: block;
+        align-content: center;
+        min-height: calc(var(--v-table-header-height));
+        height: 100%;
+        border-bottom: none;
+      }
+
+      td[colspan] {
```
userquin, line 36: same here: `[role="cell"]`

PR #23139, opened 2026-08-21
fix(locale): correct mistranslations and russianisms in uk.ts

klivak:
Audited `packages/vuetify/src/locale/uk.ts` against `en.ts` — no structural gaps (all 137 keys present, parity test passes), but found 34 keys worth fixing across two categories.

## Commit 1 — outright bugs (12 keys)
- `carousel.next`: typo, "Наступий" → "Наступний"
- `input.otp`: stale translation left a literal unfilled `{0}` placeholder visible in the UI (the English string was updated upstream and no longer has a placeholder) — corrected to a placeholder-free translation
- `input.prependAction`/`appendAction`: translated as "previous/next action" instead of "prepended/appended action" — wrong sense entirely
- `fileInput.counter`/`counterSize`: "{0} файлів" breaks Ukrainian numeral agreement at 1 and 2–4 — reworded to avoid the agreement requirement
- `fileUpload.browse`: "Переглянути файли" (view files) instead of the file-picker action
- `timePicker.minute`/`second`: plural forms for what should be singular labels
- `pagination.ariaLabel.currentPage`: clause order inverted relative to English
- `video.seek`: "Пошук" (search) instead of the scrub/seek action
- `colorPicker.ariaLabel.lightnessInput`: "Яскравість" (brightness) instead of lightness — different HSL properties

## Commit 2 — russianisms, register, and consistency (22 keys)
A cluster of entries tracked `ru.ts` closely enough to share its specific defects (inverted clause order, dropped words, wrong register) rather than being independent translations. Also includes:
- `дійсний`/`недійсний` (validity calque) → `коректний`/`неправильний` across `rules.email`, `rules.strictLength`, `rules.pattern`, etc.
- `pagination.ariaLabel.root`: "Навігація по сторінках" (ru "по + locative" pattern) → "Посторінкова навігація"
- Imperative register fixes for `dataTable` sort-activation hints (infinitive → imperative, matching their function as control instructions)
- Normalized `вибр-`/`обр-` stem inconsistency to the dominant `вибр-` form (3 keys)
- Disambiguated 6 `colorPicker` aria labels where two distinct controls (e.g. `hueSlider` vs `hueInput`) shared identical text, an accessibility issue for screen reader users

Left unchanged (out of scope): `dataIterator.loadingText` (duplicates the top-level `loading` string but that's a pre-existing pattern), `carousel` "visual" vs "slide" wording, `timePicker` AM/PM (Latin form is the norm across this locale set), and the `hotkey` block's key ordering (parity test is order-insensitive; reordering would just add diff noise).

## Verification
- `pnpm vitest --project unit --run src/locale`: both locale tests pass (`should have listed all available locales`, `should have same structure for all translations`)
- Full unit suite: 816 passed, 0 failed
- `tsgo` type-check: exit 0
- `eslint --max-warnings 0`: exit 0


J-Sek:
2 suggestions after rough verification I did. Could you please check if any of them improves translations:

- **`video.seek`**: `Перемотування` → translates to 'en' as "Rewind". Would it match the video slider better as `Позиція відтворення`?
- **`badge`**: `Значок` → translates to 'en' as "Icon". Should it be `Бейдж` or `Індикатор` instead? Or do we keep it as-is?

klivak:
@J-Sek , thanks for reviewing!

video.seek: agreed — "Позиція відтворення" describes the scrub-bar aria-label better than "Перемотування" (which implies the rewind action specifically). I'll update it.

badge: that key wasn't part of this PR's scope, but you're right that "Значок" reads as "icon" rather than "badge".

klivak:
@J-Sek 

Updated video.seek to "Позиція відтворення" - folded directly into the original commit (fix(locale): correct mistranslations and errors) rather than a new commit, so the diff you're seeing now already reflects the change. Thanks again for catching that.

Still open on badge whenever you get a chance to weigh in - happy to include it here or as a follow-up, whichever you prefer.

J-Sek:
I think it would make sense to cover badge in this PR as well

J-Sek (approved):

PR #23140, opened 2026-08-21
fix(VBadge): guard against NaN in aria-label when content is absent

klivak:
`VBadge`'s `aria-label` renders the literal string "NaN" when a consumer uses the documented `{0}` extension point in `label` but the badge has no `content` prop (e.g. before async content resolves, or a conditionally-omitted count).

Screen reader users hear "Notifications NaN" instead of the intended count.

## Root cause
`value = Number(props.content)` at `VBadge.tsx:87` evaluates to `NaN` when `content` is undefined. The `content` variable on the next line does guard with `isNaN(value)` for display purposes, but that guard isn't applied on the accessibility path — `value` is passed unguarded into `t(props.label, value)` at line 140, and the locale adapter's `replace()` (`packages/vuetify/src/locale/adapters/vuetify.ts:17-21`) stringifies parameters unconditionally (`String(NaN)` → `"NaN"`).

## Reproduction
```vue
<v-badge label="Notifications {0}">
  <v-btn>Inbox</v-btn>
</v-badge>
```
Confirmed live via play.vuetifyjs.com before filing this PR: renders `aria-label="Notifications NaN"`.

`value` and the `aria-label` interpolation are computed unconditionally in `useRender`, independent of `dot` — the bug reproduces the same with or without it. Also directly reachable through `VAvatar`'s public API: `<v-avatar :badge="{ label: 'Notifications {0}' }">` with no badge slot hits this exact path today.

## Fix
```tsx
aria-label={ t(props.label, isNaN(value) ? '' : value) }
```

## Test
Added a regression test in `VBadge.spec.browser.tsx`: without the fix it fails with `expected 'Notifications NaN' not to contain 'NaN'`; with the fix, all 5 tests in that describe block pass.

## Verification
- `pnpm --project browser --run VBadge`: 1 new test passing (5 total in the `label` describe block)
- `tsgo` type-check: exit 0
- `eslint --max-warnings 0`: exit 0

### DIFF
--- a/packages/vuetify/src/components/VBadge/VBadge.tsx
+++ b/packages/vuetify/src/components/VBadge/VBadge.tsx
@@ -137,7 +137,7 @@ export const VBadge = genericComponent<VBadgeSlots>()({
                   roundedStyles.value,
                 ]}
                 aria-atomic="true"
-                aria-label={ t(props.label, value) }
+                aria-label={ t(props.label, isNaN(value) ? '' : value) }
                 aria-live="polite"
                 role="status"
                 { ...badgeAttrs }
--- a/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
+++ b/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
@@ -58,7 +58,7 @@ describe('VBadge', () => {
     })
 
     it('should not interpolate NaN into the aria label when content is absent', () => {
-      const { wrapper } = render(<VBadge label="Notifications {0}" dot />)
+      const { wrapper } = render(<VBadge label="Notifications {0}" />)
       const label = wrapper.find('.v-badge__badge').attributes('aria-label')
 
       expect(label).not.toContain('NaN')
--- a/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
+++ b/packages/vuetify/src/components/VBadge/__tests__/VBadge.spec.browser.tsx
@@ -56,6 +56,14 @@ describe('VBadge', () => {
       render(<VBadge label="label-badge">label</VBadge>)
       await expect(screen.findByLabelText('label-badge')).resolves.toBeDefined()
     })
+
+    it('should not interpolate NaN into the aria label when content is absent', () => {
+      const { wrapper } = render(<VBadge label="Notifications {0}" dot />)
+      const label = wrapper.find('.v-badge__badge').attributes('aria-label')
+
+      expect(label).not.toContain('NaN')
+      expect(label).toBe('Notifications ')
+    })
   })
 
   describe('max', () => {


J-Sek:
This combination of props looks like a dev error. I don't see enough justification to this fix.

klivak:
You're right that `dot` was a weak part of the original repro — it's unrelated to the bug. The NaN leak reproduces with plain `<v-badge label="Notifications {0}">` and no `content`: `value = Number(props.content)` and the aria-label interpolation run unconditionally in `useRender`, regardless of `dot`.

Realistic trigger: a badge count that's async-loaded or conditionally omitted (see `misc-dynamic.vue`'s pattern of binding `content` to reactive state that starts unset). It's also directly reachable through `VAvatar`'s own public API — `badgeProps` defaults `dot: !slots.badge` and spreads `props.badge` when it's an object, so `<v-avatar :badge="{ label: 'Notifications {0}' }">` with no badge slot hits this exact path today.

Updated the test to drop `dot` and use the minimal repro.

klivak:
Live repro without `dot`: https://play.vuetifyjs.com/?one=subscribe#eNqlVU1v2zgQ/SusehAN2FSCpijWiLtZLLroAntYbIvtIcpBEUc2W36oJOXGCPzfO/yQorhBEaAHwuLM8L03w+H4+r4QqjfWr1TTs8/O6GJd3NeakDo7XF2sSbQE236AsK+Lnfe9W1dVyzUe4yDF3jINvtK9qq4wrLKD9kLBiht19Yq9ZhfnFRfOz+0MnFrdWvPNgUWUuljOeCo07sGuLGgOFuxzeU+OPeI+8f2U34vu8AzOHHl1wc7ZeabJtpVsbl3giNgB+ljrY7EspNBf3Em1WxcrfT0q+DXSiBagbhKjs231R98zDELKSw+ql42HtyHkcr9q+j5+xk1rtG+EBptN0ejhzq86AZKT/UoZ1LOpC+W2dUGq8Wh1ejZYEvRlNaPErWut6D1x4IfEnJqN3BMLHTmSzhpFSlRbhnBCENd5gnxkEyJo+R6kNOSTsZK/KBeBIEEiWE5XoQ7mHaY7YbcWUAHW4YRhCsg1nNxsrOpDUDg9OnNFH5xoiylF1Ukz5o+aJ2aKC+WilaG/E1sG1hr7vtFcgsVIIJu3MV0jIfko4IF0ZHBAs6LJpgw+KFq+xO9QiJx9lJHSt6DMHj74g4SPzZZizL+oXzhgjZT0WpqG/4XXRhdLEr7/DuR0cRMJukG3XhiNRZ+DELpIMyHl6NGyIdy0gwLt2dcB7OEDSGg9qi9fjp3pd4Dv3gUMtwPwQS7efEfoC0RYIIcfbLxu3LK+wWfq30k8g5iJ/8+dkJyGYGzrR/qC8qguIhO6wy7JGhMs0fCN5MQptYD13cMSnZ9R5iIUPQ+5lFJ4ovOc0gVmNbQM7iSfxFDUJzG8nCU3cwYx6A0/M2vD+bs9ov2DrxfwySAqJlEGTVHcHP7H2NgaMTgmkGInubeGHxh2BA46GgCi//hk1cLdT/eZa/W4mrQch1GHsY5tjdlKaHqBc8aoCmfN712jhDxs/jO3xpv1+dnZ8hWuC1yvcb3B9dvZWSjYj/yx354r4Omxr7iI0q7esLsgB9++BysaycGJrRaBgSkcB+jLIvIzya355Jz4/2QYPIyCWsNdDE29Mk6N8ZnngzQmFJt++gfl0DWD9B+TEdtou/Nl/tepqnhJuAsXVRyXRfrrQqlpzhfLrpEOHk/z0ZZFFMv76XM9njvefAdI/Ljp

J-Sek:
The link leads to default VPlay content.
Regardless, are we looking for a bug or is it something that hurts DX in a real project?

klivak:
Fair question — I don't have a confirmed production issue to point to, this API corner (custom `label` with `{0}`) doesn't seem to be widely used in the wild.

But the underlying bug is real and cheap to fix: the `isNaN` guard already exists one line above for the visual `content` display, this just extends the same guard to the aria-label path that was missed. Confirmed live: https://codepen.io/khlopchyna/pen/01a029e9-5ce8-7640-8c35-4b43fd01746c — inspect `span.v-badge__badge`, `aria-label="Notifications NaN"`.

Happy to close this if the bar here is a confirmed real-world report — entirely your call.

J-Sek:
For "_async-loaded or conditionally omitted_" user can just provide `content` fallback.

Not merely questioning this PR too push back on unconfirmed issue, but rather to point out that our limited resources we can spent elsewhere. I would appreciate if you could lookup issues with assigned milestones for a low hanging fruit there instead. People that burn tokens for fun pumping PRs get 1/20 merged and waste maintainers' time to review and reject the other 19.

I appreciate your understanding.

PR #23146, opened 2026-08-25
fix(VPagination): flex bug #23145

### LINKED ISSUE #23145: [Bug Report][4.1.10] VPagination ellipsis flickers indefinitely in a flex container at length 3
### Environment

**Vuetify Version:** 4.1.10
**Vue Version:** 3.x (Vuetify Play default)
**OS:** Not specified

### Steps to reproduce

1. Place a `v-pagination` with `:length="3"` inside a `<div class="d-flex">`
2. Load the page and observe the component without interacting with it

```vue
<template>
  <div class="text-center">
    <div class="d-flex">
      <v-pagination v-model="page" :length="3" rounded="circle"></v-pagination>
    </div>
  </div>
</template>

<script setup>
  import { ref } from 'vue'

  const page = ref(1)
</script>
```

### Expected Behavior

A stable set of page buttons. With `length="3"` and no width constraint there is nothing to truncate, so no ellipsis should appear and the rendered output should not change while the component is idle.

### Actual Behavior

The ellipsis flickers continuously. The component oscillates between renders on its own, with no user interaction and no external state change.

### Reproduction Link

[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/playgrounds/7R28SQ)

### Other comments

Reported by @hankthetank4 in Discord; filed on their behalf.

Root cause appears to be a measurement/render feedback loop, from reading the source (not yet confirmed under a debugger):

`VPagination` measures itself with a `ResizeObserver` and derives how many buttons to show from that measurement:

https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VPagination/VPagination.tsx#L159-L175

`contentRect.width` → `getMax()` → `maxButtons` → `totalVisible` → `range`. In a normal block container the width is imposed by the parent, so the measurement is stable. As a flex item the component shrink-wraps to its own content instead, so changing the number of rendered buttons changes the measured width, which re-runs the observer and can select a different button count — the measurement depends on the content it determines.

This looks adjacent to #22907 / #22912. That fix guards the same path:

```ts
if (props.totalVisible == null && length.value < 3) {
  return createRange(length.value, start.value)
}
```

`length < 3` short-circuits before any measurement is used, which is why lengths 1 and 2 are stable in a flex container. `length="3"` is the first value that falls through to the measurement path, and it is exactly the value in this reproduction — so the guard may stop one case short of the underlying problem.

If that reading is right, raising the threshold would only move the boundary; the durable fix is to stop letting the rendered content feed back into the width the component measures.

Haviles04:
## Description
fixes #23145 
## Markup:

```vue
<template>
  <div class="text-center">
    <div class="d-flex w-50">
      <v-pagination v-model="page" :length="3" rounded="circle" />
    </div>
  </div>
</template>

<script setup>
  import { ref } from 'vue'

  const page = ref(1)
</script>

```

### DIFF
--- a/packages/vuetify/src/components/VPagination/VPagination.tsx
+++ b/packages/vuetify/src/components/VPagination/VPagination.tsx
@@ -161,7 +161,8 @@ export const VPagination = genericComponent<VPaginationSlots>()({
 
       const { target, contentRect } = entries[0]
 
-      const firstItem = target.querySelector('.v-pagination__list > *') as HTMLElement
+      const buttons = target.querySelectorAll('.v-pagination__list > *')
+      const firstItem = buttons[0] as HTMLElement
 
       if (!firstItem) return
 
@@ -169,6 +170,8 @@ export const VPagination = genericComponent<VPaginationSlots>()({
       const itemWidth =
         firstItem.offsetWidth +
         parseFloat(getComputedStyle(firstItem).marginRight) * 2
+      const usedWidth = itemWidth * buttons.length
+      if (usedWidth <= totalWidth + 1 && totalWidth - usedWidth < itemWidth) return
 
       maxButtons.value = getMax(totalWidth, itemWidth)
     })
--- a/packages/vuetify/src/components/VPagination/VPagination.tsx
+++ b/packages/vuetify/src/components/VPagination/VPagination.tsx
@@ -170,8 +170,16 @@ export const VPagination = genericComponent<VPaginationSlots>()({
       const itemWidth =
         firstItem.offsetWidth +
         parseFloat(getComputedStyle(firstItem).marginRight) * 2
-      const usedWidth = itemWidth * buttons.length
-      if (usedWidth <= totalWidth + 1 && totalWidth - usedWidth < itemWidth) return
+      // Assumes every button is as wide as the first one, which is a nav control
+      // (first/prev), not a page item. Custom slots can break that and skew the total.
+      const slack = totalWidth - itemWidth * buttons.length
+
+      // A shrink-to-fit container (flex item, inline-block, table cell) is only ever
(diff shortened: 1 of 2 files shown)


packages/vuetify/src/components/VPagination/VPagination.tsx
```diff
@@ -161,14 +161,19 @@ export const VPagination = genericComponent<VPaginationSlots>()({
...
       if (!firstItem) return
 
       const totalWidth = contentRect.width
       const itemWidth =
         firstItem.offsetWidth +
         parseFloat(getComputedStyle(firstItem).marginRight) * 2
+      const slack = totalWidth - itemWidth * buttons.length
+
+      // only react to a container that is genuinely too small, or has room for another button.
+      if (slack >= -1 && slack < itemWidth) return
```
J-Sek, line 176: first item can be a custom prev button of specific width that breaks the calculation even if it would work in flexbox.

If the goal was only to prevent loop, it succeeds. Just feels brittle and does not really fix the pagination in flexbox in general.

```vue
<template>
  <v-app theme="dark">
    <v-container class="border ml-0" width="50vw">
      <div v-for="length in [1,2,3,4,5,14]" :key="length" class="d-flex align-center">
        <v-pagination
          v-model="page"
          :length
          class="border border-dashed"
        />
        <v-btn>something</v-btn>
      </div>
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const page = ref(1)
</script>
```

I am thinking about something more like #23149.

Haviles04: yea I was soley focused on the ellipsis bug from the play.. but I agree yours is better considering flex containment is broken

PR #23148, opened 2026-08-26
fix(VCheckbox,VRadio,VRadioGroup,VSwitch,VColorInput): expose focus

### LINKED ISSUE #23147: [Bug Report][4.1.10] focus() is not exposed on VCheckbox, VRadioGroup, VSwitch or VColorInput
### Environment

**Vuetify Version:** 4.1.10
**Vue Version:** 3.x
**OS:** Not specified

### Steps to reproduce

1. Put a template ref on a `v-checkbox`, `v-radio-group`, `v-switch` or `v-color-input`
2. Call `ref.value.focus()`

```vue
<template>
  <v-checkbox ref="checkbox" label="VCheckbox" />
</template>

<script setup>
  import { useTemplateRef, onMounted } from 'vue'

  const checkbox = useTemplateRef('checkbox')

  onMounted(() => checkbox.value.focus())
</script>
```

### Expected Behavior

The control receives focus, as it does on every sibling input — `VTextField`, `VTextarea`, `VFileInput`, `VSelect`, `VAutocomplete`, `VCombobox`, `VNumberInput`, `VDateInput`, `VSlider`, `VRangeSlider` and `VOtpInput` all expose a working `focus()`.

### Actual Behavior

`TypeError: focus is not a function` — the method does not exist.

Probed in a real browser against `4.1.10`:

```
[VCheckbox]   typeof focus = undefined | THREW TypeError | exposes: reset, validate, resetValidation, isValid
[VRadioGroup] typeof focus = undefined | THREW TypeError | exposes: reset, validate, resetValidation, isValid
[VSwitch]     typeof focus = undefined | THREW TypeError | exposes: reset, validate, resetValidation, isValid
[VColorInput] typeof focus = undefined | THREW TypeError | exposes: NOTHING
[VTextField]  typeof focus = function  | ok, activeElement = <input type=text> | exposes: focus, blur, reset, validate, resetValidation, isValid
```

### Reproduction Link

[https://play.vuetifyjs.com/#...](https://play.vuetifyjs.com/playgrounds/bnThuw)

### Other comments

Reported by @userquin in Discord; filed on their behalf. Four separate defects, three sharing one cause.

**1–3. `VCheckbox`, `VRadioGroup`, `VSwitch` forward refs to a dead end.**

All three end with:

```ts
return forwardRefs({}, inputRef)   // inputRef is the VInput wrapper
```

and `VInput`'s setup returns only `{ reset, resetValidation, validate, isValid, errorMessages }` — no `focus`, no element. So there is nothing for `focus()` to resolve to. That matches the probe exactly: the four validation members are present, `focus`/`blur` are not.

The fix has a precedent in-repo — `VSlider`/`VRangeSlider`:

```ts
return forwardRefs({ focus: () => thumbContainerRef.value?.$el.focus() }, inputRef)
```

For checkbox/switch no DOM query is needed: `VSelectionControl` already returns `{ isFocused, input }`, where `input` is the native `<input>`. `VRadioGroup` needs group semantics — focus the **checked** radio, falling back to the first, since that is the group's tab stop.

**4. `VColorInput` exposes nothing at all.** Its `setup` (`VColorInput.tsx:76`) ends after `useRender` with no return statement and the file contains no refs, so it wraps a `VTextField` and discards that component's entire public API — not only `focus()` but `reset()`, `validate()`, `resetValidation()`, `isValid` and `errorMessages`. `VDateInput` and `VNumberInput` are the same wrapper shape and do `forwardRefs({}, vTextFieldRef)`; one line would fix all of it.

**Related: `VRadioGroup` does not emit `update:focused`.** `VCheckbox` (`:44`) and `VSwitch` (`:87`) both declare it; `VRadioGroup` does not, so there is no way to observe its focus state either.

**Related: programmatic focus cannot restore the focus ring.** `VSelectionControl` gates `.v-selection-control--focus-visible` (`VSelectionControl.sass:96`) on the JS flag `isFocusVisible` (`VSelectionControl.tsx:267`), which is only ever set inside `onFocus` (`:192-199`). Calling `.focus()` on an already-focused input fires no focus event, so the ring never reappears — even with `focus({ focusVisible: true })`. If the new `focus()` accepts `FocusOptions`, it should set `isFocusVisible` directly when `options.focusVisible` is true rather than infer it from the event. The browser heuristic should stay in place for user-initiated focus; this is only about making an explicit programmatic call authoritative.

### Source references

- `packages/vuetify/src/components/VCheckbox/VCheckbox.tsx:101`
- `packages/vuetify/src/components/VRadioGroup/VRadioGroup.tsx:139`
- `packages/vuetify/src/components/VSwitch/VSwitch.tsx:303`
- `packages/vuetify/src/components/VColorInput/VColorInput.tsx:76`
- `packages/vuetify/src/components/VInput/VInput.tsx:263-269`
- `packages/vuetify/src/components/VSelectionControl/VSelectionControl.tsx:192-199`, `:267`, `:330`
- `packages/vuetify/src/components/VSlider/VSlider.tsx` — the pattern to copy

userquin:
## Description

fixes #23147

This PR fixes a widespread issue where the imperative API (`focus()`, `blur()`, `validate()`, etc.) was inaccessible on selection controls and `VColorInput` due to missing or incorrectly ordered `forwardRefs`.

### Changes made:
* **VCheckbox, VSwitch, VRadioGroup:** Reordered `forwardRefs` to expose `focus()` and `blur()` from the control while preventing shadowing of `VInput` methods (`validate`, `reset`, etc.).
* **VCheckboxBtn, VRadio:** Added missing `forwardRefs` to expose the underlying `VSelectionControl` API.
* **VColorInput:** Implemented `forwardRefs` to properly expose the wrapped `VTextField` API (which was previously returning nothing).
* **VSelectionControlGroup:** Exposed `blur()` internally to allow `VRadioGroup` to clear focus unconditionally.

### Testing:
* Added comprehensive browser specs for `VCheckbox`, `VRadio`, `VRadioGroup`, `VSwitch`, and `VColorInput` to guarantee the API is exposed and functional.
* Verified no double-emissions occur on `update:focused` when navigating `VRadioGroup` via keyboard.
* Verified `focus({ focusVisible: true })` correctly applies the focus ring.

### Local Testing

To run the browser tests from packages/vuetify folder:
- VCheckbox: `pnpm vitest src/components/VCheckbox/__tests__/VCheckbox.spec.browser.tsx --project browser`
- VColorInput: `pnpm vitest src/components/VColorInput/__tests__/VColorInput.spec.browser.tsx --project browser`
- VSwitch: `pnpm vitest src/components/VSwitch/__tests__/VSwitch.spec.browser.tsx --project browser`
- VRadioGroup: `pnpm vitest src/components/VRadioGroup/__tests__/VRadioGroup.spec.browser.tsx --project browser`
- VRadio: `pnpm vitest src/components/VRadio/__tests__/VRadio.spec.browser.tsx --project browser`

## Markup:

<details>
<summary>packages/vuetify/dev/Playground.vue</summary>

```vue
<template>
  <v-app>
    <v-container>
      <h1 class="text-h4 mb-6">🎯 Focus API Playground</h1>

      <v-row>
        <!-- VCheckbox -->
        <v-col cols="12" md="6">
          <v-card class="pa-4" variant="outlined">
            <v-card-title>VCheckbox</v-card-title>
            <v-card-text>
              <v-checkbox
                ref="checkboxRef"
                v-model="checkboxModel"
                label="Checkbox test"
                @update:focused="checkboxFocused = $event"
              />
              <div class="text-caption font-weight-bold">
                Model: {{ checkboxModel }} | Focused: {{ checkboxFocused }}
              </div>
            </v-card-text>
            <v-card-actions style="flex-flow: wrap;">
              <v-btn color="primary" @click="checkboxRef?.focus()">.focus()</v-btn>
              <v-btn color="primary" @click="checkboxRef?.focus({ focusVisible: true })">.focus({ focusVisible: true })</v-btn>
              <v-btn color="error" @click="focusAndBlurCheckbox()">
                .focus({ focusVisible: true }) + timeout(1500) + .blur()
              </v-btn>
            </v-card-actions>
          </v-card>
        </v-col>

        <!-- VSwitch -->
        <v-col cols="12" md="6">
          <v-card class="pa-4" variant="outlined">
            <v-card-title>VSwitch</v-card-title>
            <v-card-text>
              <v-switch
                ref="switchRef"
                v-model="switchModel"
                label="Switch test"
                @update:focused="switchFocused = $event"
              />
              <div class="text-caption font-weight-bold">
                Model: {{ switchModel }} | Focused: {{ switchFocused }}
              </div>
            </v-card-text>
            <v-card-actions style="flex-flow: wrap;">
              <v-btn color="primary" @click="switchRef?.focus()">.focus()</v-btn>
              <v-btn color="primary" @click="switchRef?.focus({ focusVisible: true })">.focus({ focusVisible: true })</v-btn>
              <v-btn color="error" @click="focusAndBlurSwitch()">
                .focus({ focusVisible: true }) + timeout(1500) + .blur()
              </v-btn>
            </v-card-actions>
          </v-card>
        </v-col>

        <!-- VRadioGroup -->
        <v-col cols="12" md="6">
          <v-card class="pa-4" variant="outlined">
            <v-card-title>VRadioGroup</v-card-title>
            <v-card-subtitle>
              Try moving between the spokes using the arrow keys on your keyboard!
            </v-card-subtitle>
            <v-card-text>
              <v-radio-group
                ref="radioGroupRef"
                v-model="radioModel"
                @update:focused="radioGroupFocused = $event"
              >
                <v-radio label="Opción 1" value="op-1" />
                <v-radio label="Opción 2" value="op-2" />
                <v-radio label="Opción 3 (Disabled)" value="op-3" disabled />
                <v-radio label="Opción 4" value="op-4" />
              </v-radio-group>
              <div class="text-caption font-weight-bold">
                Model: {{ radioModel }} | Group Focused: {{ radioGroupFocused }}
              </div>
            </v-card-text>
            <v-card-actions style="flex-flow: wrap;">
              <v-btn color="primary" @click="radioGroupRef?.focus()">.focus()</v-btn>
              <v-btn color="primary" @click="radioGroupRef?.focus({ focusVisible: true })">.focus({ focusVisible: true })</v-btn>
              <v-btn color="error" @click.prevent="focusAndBlurRadioGroup" @mousedown.prevent="">
                .focus({ focusVisible: true }) + timeout(1500) + .blur()
              </v-btn>
            </v-card-actions>
          </v-card>
        </v-col>

        <!-- VColorInput -->
        <v-col cols="12" md="6">
          <v-card class="pa-4" variant="outlined">
            <v-card-title>VColorInput</v-card-title>
            <v-card-subtitle>
              Verifying that it inherits the entire VTextField API
            </v-card-subtitle>
            <v-card-text>
              <v-color-input
                ref="colorInputRef"
                v-model="colorModel"
                label="Color picker"
                @update:focused="updateColorModelFocused"
              />
              <div class="text-caption font-weight-bold mt-2">
                Model: {{ colorModel }} | Focused: {{ colorInputFocused }}
              </div>
            </v-card-text>
            <v-card-actions style="flex-flow: wrap;">
              <v-btn color="primary" @click="colorInputRef?.focus()">.focus()</v-btn>
              <v-btn color="primary" @click="colorInputRef?.focus({ focusVisible: true })">.focus({ focusVisible: true })</v-btn>
              <v-btn color="error" @click="focusAndBlurColorInput">
                .focus({ focusVisible: true }) + timeout(1500) + .blur()
              </v-btn>
              <v-btn color="warning" @click="colorInputRef?.reset()">.reset()</v-btn>
            </v-card-actions>
          </v-card>
        </v-col>
      </v-row>
    </v-container>
  </v-app>
</template>

<script setup lang="ts">
  import { ref, shallowRef, useTemplateRef } from 'vue'

  const checkboxRef = useTemplateRef('checkboxRef')
  const switchRef = useTemplateRef('switchRef')
  const radioGroupRef = useTemplateRef('radioGroupRef')
  const colorInputRef = useTemplateRef('colorInputRef')

  const checkboxModel = shallowRef(false)
  const switchModel = shallowRef(false)
  const radioModel = shallowRef('op-1')
  const colorModel = shallowRef('#FF0000')

  const checkboxFocused = ref(false)
  const switchFocused = ref(false)
  const radioGroupFocused = ref(false)
  const colorInputFocused = ref(false)

  function focusAndBlurCheckbox () {
    checkboxRef.value?.focus({ focusVisible: true })
    setTimeout(() => checkboxRef.value?.blur(), 1500)
  }

  function focusAndBlurSwitch () {
    switchRef.value?.focus({ focusVisible: true })
    setTimeout(() => switchRef.value?.blur(), 1500)
  }

  function focusAndBlurRadioGroup () {
    radioGroupRef.value?.focus({ focusVisible: true })
    setTimeout(() => radioGroupRef.value?.blur(), 1500)
  }
  function focusAndBlurColorInput () {
    colorInputRef.value?.focus({ focusVisible: true })
    setTimeout(() => colorInputRef.value?.blur(), 1500)
  }
  function updateColorModelFocused (value: boolean) {
    // VColorInput must not emit @update:focused twice for a SINGLE focus change
    // (see VColorInput.spec.browser.tsx). A false/true *pair* per button click is
    // expected and not that bug: the v-btn takes focus (blur), then focus() restores
    // it — two real changes.
    // eslint-disable-next-line no-console
    console.log('calling @update:focused with value:', value)
    colorInputFocused.value = value
  }
</script>
```
</details>


packages/vuetify/src/components/VCheckbox/VCheckbox.tsx
```diff
@@ -46,8 +46,9 @@ export const VCheckbox = genericComponent<new <T>(
 
   setup (props, { attrs, slots }) {
     const model = useProxiedModel(props, 'modelValue')
-    const { isFocused, focus, blur } = useFocus(props)
```
johnleider, line 49: A lot of unnecessary changes. only needed changes are 51 and 103

packages/vuetify/src/components/VRadioGroup/VRadioGroup.tsx
```diff
@@ -66,13 +67,22 @@ export const VRadioGroup = genericComponent<new <T>(
...
   emits: {
     'update:modelValue': (value: any) => true,
+    'update:focused': (focused: boolean) => true,
   },
 
   setup (props, { attrs, slots }) {
     const uid = useId()
     const id = computed(() => props.id || `radio-group-${uid}`)
     const model = useProxiedModel(props, 'modelValue')
+    const { isFocused, focus: setIsFocused, blur: setBlurred } = useFocus(props)
```
johnleider: Same in this file, lots of unnecessary changes

packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
```diff
@@ -103,8 +116,47 @@ export const VSelectionControlGroup = genericComponent<new <T>(
...
+          controls.add(control)
+          onScopeDispose(() => {
+            controls.delete(control)
+          })
+        }
+      },
     })
 
+    function focus (options?: FocusOptions) {
+      for (const control of controls) {
```
johnleider, line 132: Should any of this be gated by an IN_BROWSER check?

packages/vuetify/src/components/VSwitch/VSwitch.tsx
```diff
@@ -93,7 +93,7 @@ export const VSwitch = genericComponent<new <T>(
     const indeterminate = useProxiedModel(props, 'indeterminate')
     const model = useProxiedModel(props, 'modelValue')
     const { loaderClasses } = useLoader(props)
-    const { isFocused, focus, blur } = useFocus(props)
+    const { isFocused, focus: setIsFocused, blur: setBlurred } = useFocus(props)
```
johnleider: Unnecessary changes

johnleider (changes requested):

packages/vuetify/src/components/VSwitch/VSwitch.tsx
userquin: The aliasing here is actually strictly necessary to avoid scope shadowing.

The `useFocus(props)` composable returns a state setter named focus (which only emits `update:focused` and never touches the DOM). If I don't alias it to `setIsFocused`, it collides with the actual imperative `focus()` method we need to expose via `forwardRefs`.

If we leave it as `const { focus } = useFocus()`, passing focus down will silently do nothing to the DOM when the consumer calls `ref.value.focus()`. Bash actually highlighted this exact trap in the original issue (#23147).

packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
userquin: Since this is an imperative method exposed via `defineExpose/forwardRefs`, it will only be called by the consumer after the component is mounted (e.g., inside an event handler or `onMounted`), which inherently only runs in the browser.

Defining the function itself during SSR is perfectly safe. However, if you prefer defensive programming here just in case a consumer does something weird, I can wrap the inner logic with `if (!IN_BROWSER) return`.

packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
```diff
@@ -103,8 +116,47 @@ export const VSelectionControlGroup = genericComponent<new <T>(
...
+        }
+      }
+
+      first?.focus(options)
+    }
+
+    function blur () {
+      for (const c of controls) {
+        const el = c.el()
+        if (el && el === document.activeElement) {
```
J-Sek: we should probably use `getActiveElement()` here

packages/vuetify/src/components/VSwitch/VSwitch.tsx
KaelWD: That would only be true if you were calling `forwardRefs({ isFocused, focus, blur })`, just having a variable inside `setup()` doesn't shadow anything.

packages/vuetify/src/components/VCheckbox/__tests__/VCheckbox.spec.browser.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { ref } from 'vue'
+
+describe('VCheckbox exposed focus', () => {
+  it('should focus and blur from a ref', async () => {
+    const cmp = ref<any>()
+
+    render(() => <VCheckbox ref={ cmp } />)
+    await waitIdle()
+
+    // eslint-disable-next-line @vitest/prefer-expect-type-of
```
KaelWD, line 15: ?

packages/vuetify/src/components/VRadio/__tests__/VRadio.spec.browser.tsx
```diff
@@ -0,0 +1,26 @@
+// Components
+import { VRadio } from '@/components/VRadio'
+
+// Utilities
+import { render, screen, waitIdle } from '@test'
+import { ref } from 'vue'
+
+describe('VRadio', () => {
+  it('should work standalone, outside a group and expose VSelectionControl API', async () => {
```
KaelWD: This isn't actually testing that it works standalone, only that focus does.

packages/vuetify/src/components/VCheckbox/__tests__/VCheckbox.spec.browser.tsx
userquin: eslint replacing this with expectTypeOf

packages/vuetify/src/components/VRadio/__tests__/VRadio.spec.browser.tsx
userquin: render shouldn't fail 🤔 ? I'll check the RenderResult

packages/vuetify/src/components/VSelectionControlGroup/VSelectionControlGroup.tsx
userquin: I completely forgot about the Shadow DOM boundary where `document.activeElement` just returns the host element. I've updated it to use `getActiveElement()` to properly pierce the shadow root

packages/vuetify/src/components/VSwitch/VSwitch.tsx
userquin: Arrrgg, you are completely right. I was overthinking the scope shadowing, forgetting that it doesn't affect anything unless explicitly passed into `forwardRefs`. I'll revert those aliasing changes in VCheckbox and VSwitch right away to keep the diff clean as John requested.

packages/vuetify/src/components/VRadio/__tests__/VRadio.spec.browser.tsx
userquin: fixing test

userquin: done, split existing one into two:
- check it works standalone and the v-model and
- api check

userquin:
@KaelWD @johnleider 

I've reverted the `useFocus` aliasing back to the original destructuring, you were totally right that it was purely cosmetic since we don't pass focus to `forwardRefs`.

I kept `{ isFocused }` in the target object: `forwardRefs({ isFocused }, checkboxBtnRef, inputRef)`. This one is actually not cosmetic and strictly necessary (check the thread at discord).
Without it, if a user passes `<VCheckbox focused/>`, the proxy resolves `isFocused` through the `checkboxBtnRef` chain (which defaults to `VSelectionControl`'s `shallowRef(false)`), exposing the wrong state. Shadowing it at the target level ensures we expose `VCheckbox`'s actual `useProxiedModel` state.

I'm going to add some changes for `VSelectionControl` and `VRadioGroup` at api docs and the PR should be ready for review.

userquin:
I need to revert a change...

EDIT: VRadioGroup logic is fine at [125f1a2](https://github.com/vuetifyjs/vuetify/pull/23148/commits/125f1a285f6c19c9eba8f5510d6cab3a7b6f0bf7), so no changes are necessary

johnleider:
## Rebase complete (overnight burn)

Rebased onto current master. Conflicts resolved in VColorInput / VRadio (+ specs) keeping both:
- this PR’s `forwardRefs` focus/blur/validate exposure
- master’s openOnFocus / standalone v-model work

Was CONFLICTING → now MERGEABLE. CI kicked.

packages/vuetify/src/components/VCheckbox/__tests__/VCheckbox.spec.browser.tsx
KaelWD: https://github.com/vitest-dev/eslint-plugin-vitest/pull/896

KaelWD:
Can you tell your clanker to merge instead of rebase next time so I can tell what was actually changed.

KaelWD (approved):

PR #23199, opened 2026-09-18
fix(VSelect): continue typeahead from the current item

johnleider:
## Description

VSelect type-ahead diverged from native `<select>` in two ways, measured in Chromium against a native select with the same options (Discord thread, userquin):

- Every new lookup session (first press, or after the 1s timeout) restarted from the top of the list, so with `Arizona` selected, pressing `a` jumped to `Alabama` instead of `Arkansas`.
- A repeated character tried the extended prefix first, so on `9, 90, 99, Nine` pressing `9` gave `9 → 99 → 99 → 99` and stuck; native cycles `90 → 99 → 9 → 90`.

Now a session continues from the focused option (via `aria-posinset`, so it works under virtual scroll) or from the selection when the list has no focus, and a buffer of one repeated character cycles through the items starting with it. Multi-character prefixes (`a`,`m` → `American Samoa`) are unchanged. `keyboardLookupIndex` state is gone; the anchor is derived per press.

Behaviour change asserted by an existing test: in "should follow typeahead with list focus while the menu is open", `9`,`9` now lands on `90` instead of `99`.

Tests: two new browser specs (repeated-key cycling past a longer-prefix match; resuming after the current selection with a timeout in between). Full `VSelect.spec.browser.tsx` 82/82, tsgo + eslint clean.

## Markup:

```vue
<template>
  <v-app>
    <v-container>
      <v-select :items="['9', '90', '99', 'Nine']" label="Focus, press 9 repeatedly" />
      <v-select v-model="state" :items="['Alabama', 'Alaska', 'Arizona', 'Arkansas', 'California']" label="Focus, press a" />
    </v-container>
  </v-app>
</template>

<script setup>
  import { ref } from 'vue'

  const state = ref('Arizona')
</script>
```


J-Sek:
If the items list contains a value with duplicated characters it is now harder to reach. Native behavior for "9" skipping to "90" instead of "9" is both a downgrade and a breaking change. VSelect was never meant to replace native `<select>` with all quirks. I agree there is a gap and there are things (like using up/down arrows to cycle without opening dropdown menu) that make sense to port.. this is not one of them.

